PR #17, opened 2025-04-27
docs: fix example todo app path in README

RMNCLDYO:
Updated the incorrect path to the example todo app in the README file to point to the correct location at examples/react/todo


KyleAMathews (approved):
Thanks!

PR #30, opened 2025-04-30
feat: subscribeChanges

samwillis:
This adds a few public APIs:

- `collection.derivedChanges` is a tanstack/store `Derived` state, it is replaced with the latest changes (insert/update/delete) each time the derived state changes. This can be used inside an `Effect` along with other collections to essentially subscribe to a set of charges across the collections. This will be the foundation of the IVM queries.
- `collection.currentStateAsChanges` returns the current state as a set of inserts. It's needed to bootstrap the initial state of a query.
- `collection.subscribeChanges` is a thin wrapper of the above as a higher level public api


packages/optimistic/tests/collection-subscribe-changes.test.ts
```diff
@@ -0,0 +1,565 @@
...
+
+    // Verify that callback was called with initial state
+    expect(callback).toHaveBeenCalledTimes(1)
+    const changes = callback.mock.calls[0]![0] as ChangesPayload<{
+      value: string
+    }>
+    expect(changes).toHaveLength(2)
+
+    // Check that both items were emitted as inserts
+    if (changes.length > 0) {
```
KyleAMathews: this if statement isn't necessary as it should fail before getting here

packages/optimistic/tests/collection-subscribe-changes.test.ts
```diff
@@ -0,0 +1,565 @@
...
+    emitter.emit(`testEvent`, [
+      {
+        type: `delete`,
+        key: `syncItem1`,
+        value: { value: `updated sync value` },
+      },
+    ])
+
+    // Wait for changes to propagate
+    await waitForChanges()
```
KyleAMathews: I'm surprised this is necessary? I thought all of this code would be sync?

packages/optimistic/src/collection.ts
```diff
@@ -279,6 +280,59 @@ export class Collection<T extends object = Record<string, unknown>> {
...
+          if (prevDerivedState.has(key) && !derivedState.has(key)) {
+            changes.push({
+              type: `delete`,
+              key,
+              value: prevDerivedState.get(key)!,
+            })
+          } else if (!prevDerivedState.has(key) && derivedState.has(key)) {
+            changes.push({ type: `insert`, key, value: derivedState.get(key)! })
+          } else if (prevDerivedState.has(key) && derivedState.has(key)) {
+            // Check if the value has changed and only emit a change if it has
```
KyleAMathews: I'd say it's the collections job to figure out if something has changed. Deep equality is pretty fast but not *that* fast and it should be far more efficient for someone else to do it. So I'd remove all this checking code and just push out the updates.

packages/optimistic/src/collection.ts
```diff
@@ -279,6 +280,59 @@ export class Collection<T extends object = Record<string, unknown>> {
...
+          }
+        }
+
+        this.syncedKeys.clear()
+        // Initially I had this as a Store and that I had to mutate here in the effect.
+        // Mutating a dependency in an effect seems like a bad idea as we could
+        // end in in strange and unnecessary loops.
+        // So I went with the alternative that `syncedKeys` is not a Store and we then
+        // rely on the `derivedState` to trigger this effect.
+        // I think that this works?
```
KyleAMathews: 👍 yeah seems fine.

packages/optimistic/tests/collection-subscribe-changes.test.ts
samwillis: also blaming Claude for this one 😂

packages/optimistic/tests/collection-subscribe-changes.test.ts
samwillis: Blaming Claude for that one... 😂

packages/optimistic/src/collection.ts
samwillis: If we remove this check the tests fail, we get an insert and then two updates, when it should just be an insert.

When doing an inserting transaction we are somewhere triggering `optimisticOperations` to be derived three times. That causes `derivedState` and `derivedChanges` to then be run three times.

I would only expect this to be triggered once.

KyleAMathews: Oh I was reacting to the code below:

```ts
 // TODO: Better way to do this? deep equality?
            const prevValue = JSON.stringify(prevDerivedState.get(key))
            const newValue = JSON.stringify(derivedState.get(key))
            if (prevValue !== newValue) {
```

Is this check necessary?

KyleAMathews: But sounds like this is a bug somewhere else?

samwillis: Yep, removing the `JSON.stringify` check that the two values have changed results in the test failing.

I'm suspicious something else is triggering excessive recomputations.

KyleAMathews (approved):
:shipit:

PR #37, opened 2025-05-02
Don't wait for transaction to finish before applying synced transactions

### LINKED ISSUE #34: Remove unnecessary queuing/locking behavior
There were some things I added in attempt to make the library more predictable/less surprising. But now with explicit transactions on their way #29 — they make less sense as they could be solved in user land.

E.g. transactions right now queue. Imagine a slider or someone typing into a text field. You'd want the *optimistic* state to update immediately on every change but you don't want every single change to trigger a write to the backend. As that's both more load than needed on the backend but there's also then the problem of ordering the updates somehow to ensure a slow earlier change doesn't overwrite a faster later change.

To avoid that, I made transactions queue. But that's perhaps also really not what people want. E.g. for a users collection, if there's an admin view where the user is making edits to a bunch of users, each user modification should be saved immediately, not queued behind others.

There's also the danger in queuing that perhaps a transaction gets stuck and a later change doesn't get saved fast enough before the user closes the app and they lose the change.

The assumption most devs/users have is that a change is saved immediately. So that should probably be the default and any queuing is done manually.

E.g. with explicit transactions — for the typing thing, you could immediately persist the first transaction & then in the component, see if there's an persisting transaction and if there is, create a new one to collect changes and subscribe to the state of the first and then as soon as that one finishes, commit the new one. And so on.

Another helpful but perhaps too helpful locking behavior is that apply synced transactions to a collection is paused while a transaction is active in the collection. That's probably too much and not needed most of the time. Instead we can just overwrite any synced objects w/ the optimistic object to maintain consistency there & then perhaps in the future allow devs to put a sync-lock on more objects or even the whole collection if they desire.

KyleAMathews:
Partially fixes #34

### DIFF
--- a/packages/optimistic/src/collection.ts
+++ b/packages/optimistic/src/collection.ts
@@ -227,8 +227,11 @@ export class Collection<T extends object = Record<string, unknown>> {
     this.derivedState = new Derived({
       fn: ({ currDepVals: [syncedData, operations] }) => {
         const combined = new Map<string, T>(syncedData)
+        const optimisticKeys = new Set<string>()
+
         // Apply the optimistic operations on top of the synced state.
         for (const operation of operations) {
+          optimisticKeys.add(operation.key)
           switch (operation.type) {
             case `insert`:
               combined.set(operation.key, operation.value)
@@ -243,11 +246,6 @@ export class Collection<T extends object = Record<string, unknown>> {
         }
 
         // Update object => key mappings
-        const optimisticKeys = new Set<string>()
-        for (const operation of operations) {
-          optimisticKeys.add(operation.key)
-        }
-
         optimisticKeys.forEach((key) => {
           if (combined.has(key)) {
             this.objectKeyMap.set(combined.get(key)!, key)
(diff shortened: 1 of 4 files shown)


packages/optimistic/src/collection.ts
```diff
@@ -235,17 +235,12 @@ export class Collection<T extends object = Record<string, unknown>> {
...
             case `insert`:
               combined.set(operation.key, operation.value)
               break
             case `update`:
-              existingValue = syncedData.get(operation.key)
-              combined.set(operation.key, {
-                ...(existingValue || {}),
-                ...operation.value,
-              })
+              combined.set(operation.key, operation.value)
```
KyleAMathews: don't merge optimistic objects into synced objects — that could be odd

packages/optimistic/src/collection.ts
```diff
@@ -200,11 +199,6 @@ export class Collection<T extends object = Record<string, unknown>> {
     this.optimisticOperations = new Derived({
       fn: ({ currDepVals: [transactions] }) => {
         const result = Array.from(transactions.values())
-          .filter(
-            (transaction) =>
-              transaction.state !== `completed` &&
```
KyleAMathews, line 205: This PR exposed a potential bug where we'd lose the object -> key mapping if a transaction was removed but the synced data wasn't changed as then the object would revert to the synced version w/o updating the mapping.

So to prevent this from happening, we do need to loop over all transaction operations to get their keys (just applying the not completed/failed ones).

PR #40, opened 2025-05-04
feat: query collections

samwillis:
Working query compiler and execution for Collections:

```ts
// Start with a collection
const collection = new Collection<Person>({
  // config...
})

// Build a query
const query = queryBuilder()
  .from({ collection })
  .where(`@age`, `>`, 30)
  .keyBy(`@id`) // <- required at the moment
  .select(`@id`, `@name`)

// Compile it
const compiledQuery = compileQuery(query)

// Start the query
compiledQuery.start()

// Get the "ResultCollection"
const result = compiledQuery.results
```

Outstanding things to do - probably in a followup PR:
- Make it work with un-keyed queries
- More tests! (we only have a small basic test at the moment...
- Refine what a `ResultCollection` is
- React hooks

### DIFF
--- a/packages/optimistic/src/query/compiled-query.ts
+++ b/packages/optimistic/src/query/compiled-query.ts
@@ -1,29 +1,30 @@
-import {
-  D2,
-  type IStreamBuilder,
-  MessageType,
-  MultiSet,
-  type MultiSetArray,
-  output,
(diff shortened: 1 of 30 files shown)


KyleAMathews (approved):
:shipit:

PR #46, opened 2025-05-05
feat: useLiveQuery react hook

samwillis:
Draft - opening for visibility


samwillis:
@KyleAMathews this is ready for review. The main question is the return signature of the hook - currently is returns just the map in the collection, but we may want to add other return objects in futures. What do you think?

KyleAMathews:
I don't think the map is that useful — so I'd return it & the array like in useCollection https://github.com/TanStack/optimistic/blob/9a389df32bdd50dba2882d2acb6c57a6abda9a4e/packages/react-optimistic/src/useCollection.ts#L142-L146

packages/react-optimistic/src/useLiveQuery.ts
```diff
@@ -0,0 +1,34 @@
...
+  Schema,
+} from "@tanstack/optimistic"
+
+export function useLiveQuery<
+  TResultContext extends Context<Schema> = Context<Schema>,
+>(
+  queryFn: (
+    q: InitialQueryBuilder<Context<Schema>>
+  ) => QueryBuilder<TResultContext>,
+  deps: Array<unknown> = []
```
KyleAMathews, line 23: what is the `deps` here?

packages/react-optimistic/src/useLiveQuery.ts
```diff
@@ -0,0 +1,34 @@
...
+): Map<string, ResultsFromContext<TResultContext>> {
+  const compiledQuery = useMemo(() => {
```
KyleAMathews, line 25: really love how slim the React integration is. This will be super fast to port to other frameworks.

KyleAMathews (changes requested):
Looks great! Could you clean up the lint errors before merging? I thought that was enforced but clearly not.

packages/react-optimistic/src/useLiveQuery.ts
samwillis: If you use any state in building your query, these trigger a rebuild. Stuff like user provided filtering in the UI.

KyleAMathews: Ah ok so standard react stuff

PR #63, opened 2025-05-10
docs: update README.

thruflo:
Update the readme to better describe / reflect what the library is. Adds an FAQ section and aligns better with the README structure and furniture of other TanStack libraries.

Currently a draft because I haven't yet updated the more detailed usage docs (extracted temporarily to `./docs/index.md`).


README.md
```diff
@@ -1,339 +1,147 @@
...
-})
+  return <List items={todos} />
+}
 
-// Update multiple items
-update([todo1, todo2], (drafts) => {
-  drafts.forEach((draft) => {
-    draft.completed = true
+// Make writes using transactional mutations
+const tx = createTransaction({ mutationFn: /* ... */ })
```
KyleAMathews: Let's use useOptimisticMutation

docs/index.md
```diff
@@ -0,0 +1,292 @@
...
+> [!WARNING]
+> This file contains temporary usage docs (extracted and extended from the initial project README)
+>
+> It is intended to be superceeded by published project documentation.
+
+## API and usage docs
+
+### React Hooks
+
+#### `useCollection`
```
KyleAMathews: This hook will be deleted soon

KyleAMathews (commented):
Looking great!

README.md
thruflo: I just fleshed out a bit and pushed some tweaks.

What's the best way to approach the usage example here and the longer usage docs -- could you do a pass with your preferred APIs? Or want to give me some pointers and I'll try and dig out of the tests and react/todo/example?

README.md
```diff
@@ -1,339 +1,196 @@
...
+  }
 })
 
-// Update multiple items
-update([todo1, todo2], (drafts) => {
-  drafts.forEach((draft) => {
-    draft.completed = true
+// Triggers the mutationFn
+tx.mutate(() =>
+  // Applies the local optimistic state
```
thruflo: I'm eyeballing these comments wondering if they're wrong. I.e.: when does the local optimistic state get applied?

KyleAMathews: Yeah this is correct. They're applied immediately

README.md
KyleAMathews: The tests could be enough to get you going but I'm happy to pick it up Monday too

thruflo: I updated the README usage example.

docs/index.md
```diff
@@ -0,0 +1,303 @@
...
+The collection will be populated with the query results.
+
+### base Collection
+
+There is a base `Collection` class in [`../packages/optimistic/src/collection.ts`](../packages/optimistic/src/collection.ts). You can use this directly or as a base class for implementing your own collection types.
+
+See the existing implementations in [`../packages/db-collections`](../packages/db-collections) for reference.
+
+
+## Live Queries
```
thruflo, line 109: @samwillis perhaps there's more you'd like to add here?

For example, the non-hook query interface? Also I think the "it's collections all the way down" concept is powerful, i.e.: that queries can be composed. Perhaps we can surface that in the usage docs here?

KyleAMathews: Yeah, it's very powerful and unique. 

Two examples I've liked of the creating materialized views for easier querying in the app and cloning a collection to scope optimistic updates to part of the app.

thruflo:
How about we merge this and improve in subsequent PRs?

samwillis (approved):
Yep, let's do that.

I need to add a bunch of docs for the query builder - I'll try and get on to that today.

PR #73, opened 2025-05-12
docs: update the documentation page.

thruflo:
Add a conceptual introduction to describe how to use the library. Reorganise the API reference. Add e2e examples for TanStack Query and Electric usage that illustrates typical collection type + mutationFn pattern.

https://github.com/TanStack/db/blob/thruflo/docs/docs/index.md


docs/index.md
```diff
@@ -1,23 +1,164 @@
...
+
+- loading data into your app; from
+- binding data to your components
+
+Collections can be populated in many ways, including:
+
+- fetching data, for example [from API endpoints using TanStack Query](#)
+- syncing data, for example [using a sync engine like ElectricSQL](#)
+- storing local data, for example [in-memory client data or UI state](#)
+- from live queries, creating [derived collections as materialised views](#)
```
KyleAMathews:
```suggestion
- from live collection queries, creating [derived collections as materialised views](#)
```

docs/index.md
```diff
@@ -1,23 +1,164 @@
...
+Once you have your data in collections, you can query across them to bind results to your components. This allows you to:
+
+1. load normalised data into collections and then de-normalise it through queries; simplifying your backend by avoiding the need for bespoke API endpoints that match your client
+2. join data from multiple sources; for example, syncing some data out of a database, fetching some other data from an external API and then joining these into a unified data model for your front-end code
+
+Your components don't need to know where the data in the collections they're working with came from.
+
+### Using live queries
+
+Live queries are used to query data out of collections and bind the results to state variables in your components. Live queries are reactive: when the underlying data changes in a way that would affect the query result, the new result is automatically assigned to the state variable, triggering a re-render.
```
KyleAMathews: not really sure what "bind the results to state variables"? Seems like you could drop the first sentence?

docs/index.md
```diff
@@ -1,23 +1,164 @@
...
+
+TanStack DB live queries are implemented using [d2ts](https://github.com/electric-sql/d2ts), a Typescript implementation of differential dataflow. This allows the query results to update *incrementally* (rather than by re-running the whole query). This makes them blazing fast, usually sub-millisecond, even for highly complex queries.
+
+Live queries support joins across collections. In fact, live queries return new collections. It's collections all the way down.
+
+### Making local writes
```
KyleAMathews:
```suggestion
### Making optimistic mutations
```

docs/index.md
```diff
@@ -1,23 +1,164 @@
...
+
+Collections support `insert`, `update` and `delete` operations. These operations must be made within the context of a transactional mutator. Rather than mutating the collection data directly, the collection internally treats its synced/loaded data as immutable and maintains a seperate set of local mutations as optimistic state.
```
KyleAMathews: very nice

docs/index.md
```diff
@@ -1,23 +1,164 @@
...
+
+In addition, the local mutations are passed to the async `mutationFn` that's passed in when creating the mutator.
+
+```ts
+const addTodo = useOptimisticMutation({ mutationFn })
+```
+
+This mutationFn is responsible for handling the writes, usually by sending them to a server or a database. TanStack DB waits for the function to resolve before then removing the optimistic state that was applied when the local writes was made.
+
+For example, in the following code, the mutationFn first sends the write to the server using `await api.todos.create(newTodo)` and then calls `await collection.invalidate()` to trigger a re-fetch of the collection contents using TanStack Query. When this second await resolves, the collection is up-to-date with the latest changes and the optimistic state is safely discarded.
```
KyleAMathews: I'd add something about how the collection blocks applying updates while this function runs so that the optimistic state is dropped at the exact same time as the backend state is applied so there's never any flicker in the UI.

docs/index.md
```diff
@@ -1,23 +1,164 @@
...
+
+This combines to support a model of uni-directional data flow, extending the redux/flux style state management pattern beyond the client, to take in the server as well:
+
+<figure>
+  <a href="./unidirectional-data-flow.png" target="_blank">
+    <img src="./unidirectional-data-flow.png" />
+  </a>
+</figure>
+
+With a kind-of instant inner loop of optimistic state, that becomes superseded in time by the slower outer loop of persisting to the server and syncing the updated server state back into the collection.
```
KyleAMathews: lovely!

You could drop the "kind-of" as this is exactly what it's doing.

KyleAMathews: lovely!

You could drop the "kind-of" as this is exactly what it's doing.

KyleAMathews (commented):
Looking great!

We need to fit in somewhere that about how transaction lifecycles can be manually controlled.

E.g. 

```ts
const addTodo = useOptimisticMutation({ mutationFn })

const addTodoTx = addTodo.createTransaction()

addTodoTx.mutate(() => {}

// user reviews change

// Another mutation
addTodoTx.mutate(() => {}

// Mutation is approved
addTodTx.commit()
```

docs/index.md
```diff
@@ -1,23 +1,170 @@
...
+const Todos = () => {
+  // Bind data using live queries
+  const { data: todos } = useLiveQuery((query) =>
+    query
+      .from({ todoCollection })
+      .where('@completed', '=', false)
+  )
+
+  // Make local writes using optimistic mutations
+  const updateTodo = useOptimisticMutation({ mutationFn })
```
KyleAMathews, line 38: I'd add an extra new line here

docs/index.md
```diff
@@ -1,23 +1,170 @@
...
+- fetching data, for example [from API endpoints using TanStack Query](#)
+- syncing data, for example [using a sync engine like ElectricSQL](#)
+- storing local data, for example [in-memory client data or UI state](#)
+- from live collection queries, creating [derived collections as materialised views](#)
+
+Once you have your data in collections, you can query across them to bind results to your components.
+
+### Using live queries
+
+Live queries are used to query data out of collections and bind the query results to a state variable in your components. Live queries are reactive: when the underlying data changes in a way that would affect the query result, the new result is automatically assigned to the state variable, triggering a re-render.
```
KyleAMathews: "bind the query results" suggests something like knockout.js or whatever which had formal data bindings. With newer js frameworks, I don't hear "binding" used really anymore with the re-render everything model.

```suggestion
Live queries are used to query data out of collections. Live queries are reactive: when the underlying data changes in a way that would affect the query result, the result is incrementally updated and returned from the query, triggering a re-render.
```

docs/index.md
```diff
@@ -1,23 +1,170 @@
...
+
+#### `QueryCollection`
+
+[TanStack Query](https://tanstack.com/query) fetches data using managed queries. Use `createQueryCollection` to fetch data into a collection using TanStack Query:
+
+```ts
+const todoCollection = createQueryCollection({
+  queryKey: ['todoItems'],
+  queryFn: async () => fetch('/api/todos'),
+  getPrimaryKey: (item) => item.id,
```
KyleAMathews:
```suggestion
  getId: (item) => item.id,
```

docs/index.md
```diff
@@ -237,24 +388,53 @@ const AddTodo = () => {
...
+// Another mutation
+tx.mutate(() => {}
+
+// Mutation is approved
+tx.commit()
+```
+
 Transactions progress through the following states:
 
 1. `pending`: Initial state when a transaction is created
```
KyleAMathews:
```suggestion
1. `pending`: Initial state when a transaction is created and optimistic mutations can be applied
```

docs/index.md
thruflo: Have adjusted. I guess for me "bind" communicates the live-ness. I wonder how else to say it that still captures the difference between a static query result and a live query?

PR #78, opened 2025-05-12
fix optimistic state removal from queries

samwillis:


KyleAMathews (approved):
:shipit: 

Though actually — could you add a changeset to trigger a new release w/ this and other changes?

PR #83, opened 2025-05-14
feat: add Vue support

wobsoriano:
This PR adds Vue support for TanStack DB! It should be a very good starting point 🫡 

Screenshot 2025-05-14 at 8 18 25 AM


KyleAMathews (approved):
Amazing! I'm not a vue dev but it looks good to me! I have to do a little manual fiddling locally to get the first release out but then I'll get it going.

PR #85, opened 2025-05-15
refactor: vue implementation

teleskop150750:
refactor: vue implementation

### DIFF
--- a/packages/vue-db/src/useLiveQuery.ts
+++ b/packages/vue-db/src/useLiveQuery.ts
@@ -1,6 +1,6 @@
-import { computed, toValue, watch } from "vue"
-import { useStore } from "@tanstack/vue-store"
+import { onWatcherCleanup, shallowRef, toValue, watchEffect } from "vue"
 import { compileQuery, queryBuilder } from "@tanstack/db"
+import { shallow } from "./useStore"
 import type {
   Collection,
   Context,
@@ -9,12 +9,12 @@ import type {
   ResultsFromContext,
   Schema,
 } from "@tanstack/db"
-import type { ComputedRef, MaybeRefOrGetter } from "vue"
+import type { MaybeRefOrGetter, Ref } from "vue"
 
 export interface UseLiveQueryReturn<T extends object> {
-  state: ComputedRef<Map<string, T>>
-  data: ComputedRef<Array<T>>
-  collection: ComputedRef<Collection<T>>
+  state: Ref<Map<string, T>>
+  data: Ref<Array<T>>
+  collection: () => Collection<T>
 }
 
 export function useLiveQuery<
@@ -23,38 +23,63 @@ export function useLiveQuery<
   queryFn: (
     q: InitialQueryBuilder<Context<Schema>>
   ) => QueryBuilder<TResultContext>,
-  deps: Array<MaybeRefOrGetter<unknown>> = []
+  deps: () => Array<MaybeRefOrGetter<unknown>> = () => []
 ): UseLiveQueryReturn<ResultsFromContext<TResultContext>> {
-  const compiledQuery = computed(() => {
-    // Just reference deps to make computed reactive to them
-    deps.forEach((dep) => toValue(dep))
+  const compiledQuery = shallowRef() as Ref<
+    ReturnType<typeof compileQuery<TResultContext>>
+  >
+
+  watchEffect(() => {
+    toValue(deps)
 
     const query = queryFn(queryBuilder())
     const compiled = compileQuery(query)
     compiled.start()
-    return compiled
+
+    compiledQuery.value = compiled
+
+    onWatcherCleanup(compiled.stop)
   })
 
-  const state = computed(() => {
-    return useStore(compiledQuery.value.results.derivedState).value
-  })
-  const data = computed(() => {
-    return useStore(compiledQuery.value.results.derivedArray).value
-  })
+  let stateRef: Map<string, ResultsFromContext<TResultContext>>
+  let dataRef: Array<ResultsFromContext<TResultContext>>
+  const state = shallowRef(stateRef!)
+  const data = shallowRef(dataRef!)
 
-  watch(compiledQuery, (newQuery, oldQuery, onInvalidate) => {
-    if (newQuery.state === `stopped`) {
-      newQuery.start()
-    }
+  watchEffect(() => {
+    const results = compiledQuery.value.results
+    const derivedState = results.derivedState
+    const derivedArray = results.derivedArray
+    stateRef = derivedState.state
+    dataRef = derivedArray.state
+    state.value = stateRef
+    data.value = dataRef
 
-    onInvalidate(() => {
-      oldQuery.stop()
+    const unsubDerivedState = derivedState.subscribe(() => {
+      const newValue = derivedState.state
+      if (shallow(stateRef, newValue)) return
+
+      stateRef = newValue
+      state.value = newValue
+    })
+
+    const unsubDerivedArray = derivedArray.subscribe(() => {
+      const newValue = derivedArray.state
+      if (shallow(dataRef, newValue)) return
+
+      dataRef = newValue
+      data.value = newValue
+    })
+
+    onWatcherCleanup(() => {
+      unsubDerivedState()
+      unsubDerivedArray()
     })
   })
 
   return {
     state,
     data,
-    collection: computed(() => compiledQuery.value.results),
+    collection: () => compiledQuery.value.results,
   }
 }
--- a/packages/vue-db/src/useLiveQuery.ts
+++ b/packages/vue-db/src/useLiveQuery.ts
@@ -1,4 +1,4 @@
-import { onWatcherCleanup, shallowRef, watchEffect } from "vue"
+import { computed, onScopeDispose, shallowRef } from "vue"
 import { compileQuery, queryBuilder } from "@tanstack/db"
 import { shallow } from "./useStore"
 import type {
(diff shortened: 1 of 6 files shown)


wobsoriano:
Hi @teleskop150750! Curious as to what's wrong with the current one? I tried to keep it 1:1 with the React implementation (including the use of `@tanstack/store`) and make it idiomatic by using `computed` instead of setting refs inside watchers. 

Happy to know if there's a specific issue or limitation with the current implementation 🫡

packages/vue-db/src/useLiveQuery.ts
```diff
@@ -23,38 +23,60 @@ export function useLiveQuery<
...
+      state.value = newValue
+    })
+
+    const unsubDerivedArray = derivedArray.subscribe(() => {
+      const newValue = derivedArray.state
+      if (shallow(dataRef, newValue)) return
+
+      dataRef = newValue
+      data.value = newValue
+    })
```
wobsoriano: I think `@tanstack/store` handles this already?

packages/vue-db/src/useLiveQuery.ts
```diff
@@ -23,38 +23,60 @@ export function useLiveQuery<
...
-    // Just reference deps to make computed reactive to them
-    deps.forEach((dep) => toValue(dep))
+  const results = shallowRef() as Ref<
+    ReturnType<typeof compileQuery<TResultContext>>[`results`]
+  >
+
+  const state = shallowRef() as Ref<
+    Map<string, ResultsFromContext<TResultContext>>
+  >
+  const data = shallowRef() as Ref<Array<ResultsFromContext<TResultContext>>>
```
wobsoriano: Any reason on why we are not passing the types to the primitive instead?

like

```ts
const results = shallowRef<ReturnType<typeof compileQuery<TResultContext>>[`results`]>()
```

packages/vue-db/src/useLiveQuery.ts
teleskop150750: Vue and tanstack libraries unfortunately have a very difficult fate (((

https://github.com/TanStack/store/pull/179

1. the first reason is the reactors of the library @tanstack/store (drop `vue2`, refactor `unsub`)
2. the current implementation of @tanstack/store uses a deep link, which makes no sense since the entire object gets replaced. Unfortunately, vue primitives are woefully behind solidjs.

https://github.com/TanStack/store/blob/cf37b85ddecdcb6f52ad930dcd53e294fb4b03a7/packages/solid-store/src/index.tsx#L31

Solid js uses reconcile for fine-grained updating. If desired, you can try writing an analogue for vue

packages/vue-db/src/useLiveQuery.ts
teleskop150750: For
```ts
const foo = shallowRef<T>()
// foo: Ref<T | undefined>
```
The output will actually be `Ref <T | undefined>` because we didn't pass a value

teleskop150750:
> Hi @teleskop150750! Curious as to what's wrong with the current one? I tried to keep it 1:1 with the React implementation (including the use of `@tanstack/store`) and make it idiomatic by using `computed` instead of setting refs inside watchers.
> 
> Happy to know if there's a specific issue or limitation with the current implementation 🫡

My implementation still follows the react code. The only difference is that I don't use vue-store

For optimization purposes, 1 effect is used instead of 5 (1-watch and 4 computed)

Because of some stereotypes and limitations of Api ref, people are afraid to use getter functions.
[Link](https://github.com/TanStack/db/blob/ac560ec5d3313cee4c09eba83f9c5428cf5794d5/packages/vue-db/src/useLiveQuery.ts#L58)
It seems to me that in this line the extra computed does not optimize anything, but is used only to be in the style of vue api.
An easier way would be to simply return the function
```ts
{
  collection: () => compiledQuery.value.results
}
```

Changed input options interface to.
```ts
deps: () => Array<unknown> = () => []
```

I wrote about it in great detail here

https://github.com/vuejs/rfcs/discussions/747

It seems to me that deps are not needed at all, because when calling `queryFn`, all reactive variables will automatically fall into dependencies. For more detailed configuration, we need support for the `untrack` method in vue or use `watch`

wobsoriano:
> > Hi @teleskop150750! Curious as to what's wrong with the current one? I tried to keep it 1:1 with the React implementation (including the use of `@tanstack/store`) and make it idiomatic by using `computed` instead of setting refs inside watchers.
> > Happy to know if there's a specific issue or limitation with the current implementation 🫡
> 
> My implementation still follows the react code. The only difference is that I don't use vue-store
> 
> For optimization purposes, 1 effect is used instead of 5 (1-watch and 4 computed)
> 
> Because of some stereotypes and limitations of Api ref, people are afraid to use getter functions. [Link](https://github.com/TanStack/db/blob/ac560ec5d3313cee4c09eba83f9c5428cf5794d5/packages/vue-db/src/useLiveQuery.ts#L58) It seems to me that in this line the extra computed does not optimize anything, but is used only to be in the style of vue api. An easier way would be to simply return the function
> 
> ```ts
> {
>   collection: () => compiledQuery.value.results
> }
> ```
> 
> Changed input options interface to.
> 
> ```ts
> deps: () => Array<unknown> = () => []
> ```
> 
> I wrote about it in great detail here
> 
> [vuejs/rfcs#747](https://github.com/vuejs/rfcs/discussions/747)
> 
> It seems to me that deps are not needed at all, because when calling `queryFn`, all reactive variables will automatically fall into dependencies. For more detailed configuration, we need support for the `untrack` method in vue or use `watch`

Thanks for the detailed explanation and the references!

From what I understand, your refactor aims to optimize for certain edge cases and performance, but it seems to come at the cost of clarity and idiomatic Vue code. Is that correct?

re: deps array - I think this feature is similar to [Nuxt's useAsyncData watch option](https://nuxt.com/docs/api/composables/use-async-data#params) for cases where you want to refresh the query based on other reactive variables that aren’t directly referenced in `queryFn`

teleskop150750:
> deps array

There is no point in this because the result of the query will be the same
I think the reason is that react requires explicit transmission of all dependencies

teleskop150750:
@wobsoriano   Unfortunately I couldn't figure out where it came from.

https://github.com/TanStack/db/blob/ac560ec5d3313cee4c09eba83f9c5428cf5794d5/packages/vue-db/src/useLiveQuery.ts#L47

https://github.com/TanStack/db/blob/ac560ec5d3313cee4c09eba83f9c5428cf5794d5/packages/db/src/query/compiled-query.ts#L153

teleskop150750:
> idiomatic Vue code

Example [vueuse](https://github.com/vueuse/vueuse/blob/2e0d9a3e4c3e73a1e75c086baaf90f4ddd8abca7/packages/shared/computedEager/index.ts#L22-L25)

teleskop150750:
I can agree with you that computed can be useful for lazy invocation.

teleskop150750:
@wobsoriano  

look at the experimental implementation with computed, I return the entire `ref` so that the subscription is not created every time the value is updated

teleskop150750:
https://github.com/vuejs/core/blob/163b3651d174321911648a164052effa9249a2aa/packages/reactivity/__tests__/computed.spec.ts#L766-L768

KyleAMathews:
Sorry for the delay getting to this — I've been out on vacation. Our goal here is to just use tanstack/store for framework integration. So I'm very keen on making those as fast as possible but let's concentrate on making those contributions upstream in the tanstack/store repo instead of bespoke framework integrations here.

Thanks for all your efforts @teleskop150750 and @wobsoriano and let's move this conversation over to tanstack/store.

teleskop150750:
> Sorry for the delay getting to this — I've been out on vacation. Our goal here is to just use tanstack/store for framework integration. So I'm very keen on making those as fast as possible but let's concentrate on making those contributions upstream in the tanstack/store repo instead of bespoke framework integrations here.
> 
> Thanks for all your efforts @teleskop150750 and @wobsoriano and let's move this conversation over to tanstack/store.

Why did you close the pull request? Is it not just about using tanstack/store, but also about implementation fixes.

KyleAMathews:
Implementation fixes should be landed in the store repo. Here we just want to do essentially `return useStore(store)`

teleskop150750:
> Implementation fixes should be landed in the store repo. Here we just want to do essentially `return useStore(store)`

These changes are aimed at refactoring TanStack/db

teleskop150750:
@KyleAMathews 
besides using tanstack store. This pr makes a number of other changes

KyleAMathews:
You removed useStore? It looks like a reimplemention to me? Am I missing something? Why can't these changes be upsteamed?

teleskop150750:
@KyleAMathews 

The current implementation of tanstack store and integration into tanstack db does not work quite correctly.

[PR store vue](https://github.com/TanStack/store/pull/179#issuecomment-2911171219)
[PR store svelte](https://github.com/TanStack/store/pull/193)
[Solid db error](https://github.com/TanStack/db/pull/92#issuecomment-2888547242)
[Svelete db error](https://github.com/TanStack/db/pull/92#issuecomment-2888547242)

Possible solutions for corrective integration
[Use `effect`](https://github.com/TanStack/db/blob/efe925017f550635004e15b8673cdb99d76ccde9/packages/vue-db/src/useLiveQuery.ts#L45-L59)
[Use `computed`](https://github.com/TanStack/db/blob/492596da0ae0b4b945adb0d8ffcc205debd1e330/packages/vue-db/src/useLiveQuery.ts#L40-L70)

Deps props have been removed due to being no longer needed.
If I understand correctly, in React they are only needed to register variables that are used in the request. In signal-based frameworks, dependencies can be automatically registered when used

Removed unreachable code
https://github.com/TanStack/db/pull/85#issuecomment-2884775055

teleskop150750:
@KyleAMathews Hi, can you open this PR?

KyleAMathews:
hey @teleskop150750 we're planning now on not using store (because of the same problems you identified) which will make the whole update cycle really efficient even for million+ collections. See this PR https://github.com/TanStack/db/pull/155 we'd love your help working on the new framework integrations.

PR #87, opened 2025-05-16
rollback batching of transaction commit

### LINKED ISSUE #86: React Example - Uncaught Error: Object not found in collection
Using the react todo example provided. 

When you add or update a todo, you cannot perform any more actions on that todo until the next refetch occurs. Despite the todos being invalidated by the action. 

It seems the reference in `collection.state.values()` does not match the one stored in objectKeyMap anymore?

Same issue occurs with both Electric & Query Collections.

https://github.com/user-attachments/assets/2126013c-364f-4b12-b5e7-805c166392b3

samwillis:
This fixes #86, but a larger refactor of all this is needed.

@KyleAMathews I think this change didn't fix the flicker you experienced in the Electric dashboard, so there is no negative impact on rolling it back.

### DIFF
--- a/packages/db/src/transactions.ts
+++ b/packages/db/src/transactions.ts
@@ -1,4 +1,3 @@
-import { batch } from "@tanstack/store"
 import { createDeferred } from "./deferred"
 import type { Deferred } from "./deferred"
 import type {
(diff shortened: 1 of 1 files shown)


KyleAMathews (approved):
Needs a changeset too to release

PR #91, opened 2025-05-17
feat: add Svelte support

wobsoriano:
Adds Svelte support!

Screenshot 2025-07-21 at 5 29 35 PM


packages/svelte-db/tests/useLiveQuery.svelte.test.ts
wobsoriano, line 1: We have to wrap the tests here in a `$effect.root` since `useLiveQuery` has usage of `$effect`. More info here - https://svelte.dev/docs/svelte/testing#Unit-and-integration-testing-using-Vitest

packages/svelte-db/tests/useLiveQuery.svelte.test.ts
```diff
@@ -0,0 +1,873 @@
...
+      expect(result.state.size).toBe(4)
+      expect(result.state.get(`temp-key`)).toEqual({
+        id: `temp-key`,
+        name: `John Doe`,
+        title: `New Issue`,
+      })
+    })
+
+    // Wait for the transaction to be committed
+    await tx.isPersisted.promise
```
wobsoriano: Since `$effect.root` can't have an async callback, we have to move this promise out of it and create different `$effect.root`s

packages/svelte-db/src/useLiveQuery.svelte.ts
```diff
@@ -0,0 +1,58 @@
...
+    deps.forEach((dep) => dep())
+
+    const query = queryFn(queryBuilder())
+    const compiled = compileQuery(query)
+    compiled.start()
+    return compiled
+  })
+
+  const state = () => useStore(compiledQuery.results.derivedState).current
+  const data = () => useStore(compiledQuery.results.derivedArray).current
```
wobsoriano: Originally tried to use a `$derived` function here, but `useStore` for some reason is not reactive. This is still valid usage.

MAST1999: Same problem in Solid, I think if we want it to work `useStore` needs to receive a function otherwise the signal/rune for `compiledQuery` is not registered in `useStore`.

wobsoriano: Yeah. Should we open a PR to tanstack store?

wobsoriano: Yeah I updated it to not use `useStore` for now. I think as long as the tests pass then we're good 👍🏼

teleskop150750: The tests may pass but the value will not be reactive, which may cause the component not to update.

wobsoriano: Have you seen the updated code? What do you think about it?

So this means all tanstack store packages needs to be updated then?

KyleAMathews:
FYI — we've removed the tanstack/store to do a direct integration w/ the frameworks (more control means more speed). The vue code is a good example of how to directly integrate with framework reactive primitives. Would love to get a Svelte package out! https://github.com/TanStack/db/blob/main/packages/vue-db/src/useLiveQuery.ts

wobsoriano:
> FYI — we've removed the tanstack/store to do a direct integration w/ the frameworks (more control means more speed). The vue code is a good example of how to directly integrate with framework reactive primitives. Would love to get a Svelte package out! https://github.com/TanStack/db/blob/main/packages/vue-db/src/useLiveQuery.ts

Awesome. Ill take a look 👀

teleskop150750:
> The vue code is a good example

Needs some refactoring

KyleAMathews:
@teleskop150750 jump in!

wobsoriano:
@KyleAMathews good to review!

packages/svelte-db/vite.config.ts
wobsoriano, line 1: We are not using the shared `tanstackViteConfig` here as we want to use [`svelte-package`](https://svelte.dev/docs/kit/packaging) to correctly process `.svelte.ts` and `.svelte` files.

This is also what the [TanStack Svelte 5 adapter](https://github.com/TanStack/query/tree/svelte-5-adapter/packages/svelte-query) does

packages/svelte-db/src/useLiveQuery.svelte.ts
```diff
@@ -0,0 +1,383 @@
...
+}
+
+type MaybeGetter<T> = T | (() => T)
+
+function toValue<T>(value: MaybeGetter<T>): T {
+  if (typeof value === `function`) {
+    return (value as () => T)()
+  }
+  return value
+}
```
wobsoriano, line 62: In Svelte 5 with runes, the standard is to use getters when passing reactive state

packages/svelte-db/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@tanstack/svelte-db",
+  "description": "Svelte integration for @tanstack/db",
+  "version": "0.0.0",
+  "scripts": {
+    "build": "svelte-package --input ./src --output ./dist --tsconfig ./tsconfig.build.json",
```
wobsoriano, line 6: We are not using the shared `tanstackViteConfig` here as we want to use [`svelte-package`](https://svelte.dev/docs/kit/packaging) to correctly process `.svelte.ts` and `.svelte` files.

This is also what the [TanStack Svelte 5 adapter](https://github.com/TanStack/query/tree/svelte-5-adapter/packages/svelte-query) does

packages/svelte-db/tsconfig.build.json
wobsoriano, line 1: prevents `svelte-package` from generating declarations for `@tanstack/db` while still allowing proper type resolution during dev and testing

KyleAMathews (approved):
Great stuff! I'm no Svelte expert but Claude & Gemini gave their 👍 😆 & it looks good to me.

PR #109, opened 2025-05-23
collections: add `QueryCollection.refetch()`.

thruflo:
This adds a new `refetch()` method as an alternative to `invalidate()` and updates the docs and examples to use it.

### DIFF
--- a/.changeset/bumpy-wombats-yell.md
+++ b/.changeset/bumpy-wombats-yell.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db-collections": patch
+---
+
+Provide `queryCollection.refetch()` as an alternative to `queryCollection.invalidate()` that actually waits for thr collection to be updated.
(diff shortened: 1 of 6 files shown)


.changeset/bumpy-wombats-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tanstack/db-collections": patch
+---
+
+Provide `queryCollection.refetch()` as an alternative to `queryCollection.invalidate()` that actually waits for thr collection to be updated.
```
KyleAMathews:
```suggestion
Provide `queryCollection.refetch()` as an alternative to `queryCollection.invalidate()` that actually waits for the collection to be updated.
```

KyleAMathews (changes requested):
I think we just want to fully remove invalidate (for now) as looking at it more, the semantics for invalidate are refetch if the query is currently being used. But since we don't have a way of tracking that atm, just supporting an explicit refetch seems like the simple direct API to support. Once a collection can know if it's currently being used or not, then we could restore the invalidate method.

Also could you update the example.

KyleAMathews (approved):

PR #113, opened 2025-05-24
fix(db): correctly filter optimistic ops by collection

samwillis:
if you mutated two collections in a single transaction, the optimistic state would bleed into the other collections.

### DIFF
--- a/packages/db/src/collection.ts
+++ b/packages/db/src/collection.ts
@@ -202,7 +202,6 @@ export class Collection<T extends object = Record<string, unknown>> {
             return transaction.mutations
               .filter((mutation) => mutation.collection === this)
               .map((mutation) => {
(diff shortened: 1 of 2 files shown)


packages/db/src/collection.ts
```diff
@@ -199,21 +199,26 @@ export class Collection<T extends object = Record<string, unknown>> {
...
-              if (
-                mutation.metadata !== undefined &&
-                mutation.metadata !== null
-              ) {
-                message.metadata = mutation.metadata as Record<string, unknown>
-              }
-              return message
-            })
+            return transaction.mutations
+              .filter((mutation) => mutation.collection === this)
```
samwillis, line 203: this filter was missing

KyleAMathews (approved):
Oof, good catch!

Issue #42, opened 2025-05-05
Devtools


KyleAMathews:
Thinking through the devtools a bit:
- our objects are collections and transactions
- collections can be derived from other collections (#41) so showing the dependency graph visually would be nice
- click on a collection & browse it's data
- every query result is also a collection — but perhaps also showing which queries are active or recently active & on which component/url
- list of transactions ordered by creation time w/ the list of mutations linking to their collections, etc.

PR #122, opened 2025-05-30
Fix: Prevent memory leak in transaction management

jiji-hoon96:
## Problem
The global transactions array in src/transactions.ts was accumulating all created transactions indefinitely, including completed and failed ones. This caused

- Memory leaks in long-running applications
- Test isolation issues due to shared state between tests
- Potential OOM crashes in high-transaction-volume scenarios

## Reproduction Case

```ts
Before fix: All transactions stay in memory forever
for(let i = 0; i < 10000; i++) {
  const tx = createTransaction({...})
  await tx.commit() // Stays in global array
}
console.log(transactions.length) // 10000 - Memory leak!
```

## Change

```ts
setState(newState: TransactionState) {
  this.state = newState
  
  // Remove completed transactions from global tracking
  if (newState === `completed` || newState === `failed`) {
    removeFromPendingList(this)
  }
}
```

+ for Each => for of

### DIFF
--- a/.changeset/small-bushes-hide.md
+++ b/.changeset/small-bushes-hide.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+update d2ts to to latest version that improves hashing performance
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -263,9 +263,9 @@ This also works with joins to derive collections from multiple source collection
 
 #### base Collection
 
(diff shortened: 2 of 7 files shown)


KyleAMathews:
We can actually just remove it — the idea was/is in the future we'll add Devtools (#42) but that'll be an opt-in thing when it happens.

I'll just push some changes.

KyleAMathews:
Oh actually we do use the `transactions` array — so nvm, this is good 👍

KyleAMathews (approved):
Thanks!

PR #127, opened 2025-05-30
Add initial Data preparing for SSR support

andr-ec:
Hey guys I had a stab at implementing initial data as a first step for SSR support. I can do hydration provider etc similar to tanstack query in a separate pull request.

### DIFF
--- a/packages/db/src/types.ts
+++ b/packages/db/src/types.ts
@@ -118,6 +118,11 @@ export interface CollectionConfig<T extends object = Record<string, unknown>> {
   id: string
   sync: SyncConfig<T>
   schema?: StandardSchema<T>
+  /**
+   * Initial data for server-side rendering
+   * Allows hydration from server-loaded data
+   */
+  initialData?: Array<T>
 }
 
 export type ChangesPayload<T extends object = Record<string, unknown>> = Array<
--- a/packages/db-collections/src/electric.ts
+++ b/packages/db-collections/src/electric.ts
@@ -68,11 +68,41 @@ export class ElectricCollection<
     super({
       ...config,
       sync,
-      initialData:
-        config.initialData?.data.map((item) => item.value) || undefined,
+      initialData: undefined,
     })
 
     this.seenTxids = seenTxids
+
+    if (config.initialData?.data && config.initialData.data.length > 0) {
+      this.seedFromElectricInitialData(config.initialData.data)
+    }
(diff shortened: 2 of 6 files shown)


packages/db/src/types.ts
```diff
@@ -118,6 +118,11 @@ export interface CollectionConfig<T extends object = Record<string, unknown>> {
   id: string
   sync: SyncConfig<T>
   schema?: StandardSchema<T>
+  /**
+   * Initial data for server-side rendering
+   * Allows hydration from server-loaded data
+   */
+  initialData?: Array<T>
```
KyleAMathews, line 125: I'll have to think about this more but I don't think collections in general will want an `initialData` option. TanStack Query is around conventions for you directly managing a client cache. TanStack DB aims to be a bit higher-level where these collection types take over managing the cache for you.

So in this world, the QueryCollection might have an initialData option but not necessarily other collections (e.g. Electric wouldn't).

KyleAMathews: I think we'll probably just want an API which awaits running a query (i.e. all the collections are loaded and the query runs). So combination of https://github.com/TanStack/query/discussions/9135 & https://github.com/TanStack/db/issues/129

andr-ec: yes that makes sense for electric.

If there's an api which awaits running a query, would that just be a similar api to the the react query ssr hydration and rehydration? 

https://tanstack.com/query/latest/docs/framework/react/guides/advanced-ssr

KyleAMathews: > If there's an api which awaits running a query, would that just be a similar api to the the react query ssr hydration and rehydration?

Yeah exactly. Our general API goal is to copy TanStack Query directly whenever possible.

KyleAMathews:
The refactors are done if you'd like to pick things up again!

packages/db/src/types.ts
andr-ec: looks like TanStack Query uses initial data as well. still remove initial data? 

https://tanstack.com/query/latest/docs/framework/react/guides/ssr

KyleAMathews: Yeah, we can add it back later if people ask.

KyleAMathews:
This has gotten stale so closing — thanks for your effort and feel free to re-engage if you want to pick up the work again!

PR #130, opened 2025-06-01
Update overview.md

melkir:
Fix relative links on the [doc](https://tanstack.com/db/latest/docs/overview#collections).

For example https://tanstack.com/db/latest/docs/packages/db-collections instead of the source file on github.


KyleAMathews (approved):
Thanks;

PR #131, opened 2025-06-01
Update index.md

melkir:
Same as https://github.com/TanStack/db/commit/08bc4e39fa5720b17b1ae09d67b8ac174f56e7d7
Maybe `overview.md` can be removed since it is the exact same content and doesn't seems to be used?

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -259,9 +259,9 @@ This also works with joins to derive collections from multiple source collection
 
 #### base Collection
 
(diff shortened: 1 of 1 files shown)


KyleAMathews:
Oh right — actually index.md is the redundant one — I'm removing it in this PR https://github.com/TanStack/db/pull/151

KyleAMathews:
so your work is done already as it turns out 😆

PR #134, opened 2025-06-03
Collections must have a getId function & use an id for update/delete operators

### LINKED ISSUE #54: For update/delete operators, reference items in the collection by their id (instead of an object ref)
This is necessary so it's easy to update or delete those objects e.g.

```tsx
const { data: todos } = useLiveQuery(q => q.from({ todos: todosCollection })

// later
updateTodo(() => {
  todosCollection.update(todos[0], (draft) => draft.completed = !draft.completed)
})
```

Internally we'd need some way of traversing the chain of queries back to the object the collection doing the mutation knows.

KyleAMathews:
fixes #54 

Enforce globally unique ids for each item by making the internal key be a combination of the collection id & item id.

Then just reference objects for delete/update by the object id.

Todo
- [x] throw error if you try to update an id
- [x] throw error if you try to insert new object w/ same id (from collection or sync interface)
- [x] update example
- [x] add changeset


samwillis (approved):
LGTM 👍

Issue #112, opened 2025-05-23
Date field formatting and orderBy.


thruflo:
I have an ElectricCollection which loads tasks that have an `inserted_at` date field.

I'm parsing this when I sync the data:

```ts
const parser = {
  timestamp: (date: string) => new Date(date)
}
```

This is working, I see the date object in the collection data.

When I query it:

```tsx
const { data } = useLiveQuery(query => (
    query
      .from({ taskCollection })
      .select('@id', '@title', '@inserted_at')
      .orderBy({'@inserted_at': 'desc'})
      .keyBy('@id')
  ))
```

1. I get the string version of the field back from the query
2. the orderBy doesn't work

The latter is confusing -- because I believe it was working and then stopped working. OrderBy works if I sort on another field. But not for inserted_at.

gregermendle:
Care if I take a look at this?

KyleAMathews:
Feel free!

gregermendle:
Dug into this a bit and was curious if the `timestamp` parser above should in fact be `timestamptz`? Locally when testing on the todo app I can reproduce this issue when switching from `timestamptz` to `timestamp`. But I believe that is due to the fact that the todo schema is using a timestamp with a time zone.

kevin-dp:
**EDIT: Ignore this comment.** While the explanation of `orderBy` in this comment is correct for D2TS, `tanstack/db`'s `orderBy` works differently as it always passes an index column when using D2TS' `orderBy` operator. This is explained in my next comment.

----------------------

## TL;DR

The naming of the `orderBy` operator is confusing as it is not actually ordering the elements but rather re-arranging the stream to perform a `topK` operator.

## Full answer

When using `orderBy` without an index column (like in your example) the query is compiled into a pipeline that uses the `orderBy` operator (cf. https://github.com/TanStack/db/blob/main/packages/db/src/query/order-by.ts#L236-L241). Looking into D2TS, the `orderBy` is rearranging the stream such that it can be passed to the `topK` operator (cf. https://github.com/electric-sql/d2ts/blob/main/packages/d2ts/src/operators/orderBy.ts#L56). So `orderBy` in fact does not order the elements but returns a stream of elements that are in the top K. Since the example snippet did not provide a limit nor an offset, we're in fact not really doing anything since `K = infinity` and `offset = 0`, hence, all elements from the input stream are in the output stream (without order).

If you're interested in the order of the elements then you need to add an index column in your `select` clause, like:
```ts
[...]

PR #139, opened 2025-06-06
fix: comparator calls toString on undefined values in order-by

gregermendle:
When testing for [#112](https://github.com/TanStack/db/issues/112) locally, I noticed that when adding items in the todo example, you get `Cannot call toString on undefined.` during mutation. I believe this is just caused by the last bit of logic in the comparator function in `order-by`.

### DIFF
--- a/packages/db/src/query/order-by.ts
+++ b/packages/db/src/query/order-by.ts
@@ -179,8 +179,8 @@ export function processOrderBy(
       // All elements are equal up to the minimum length
       return a.length - b.length
     }
-    // if a and b are both null/undefined, return 0
-    if (a == null && b == null) {
+    // if a or b are null/undefined, return 0
+    if (a == null || b == null) {
       return 0
     }
     // Fallback to string comparison for all other cases


KyleAMathews:
How are you getting `cannot call toString on undefined`? I can add/remove as many todos as I like to the example app.

kevin-dp:
> I believe this is just caused by the last bit of logic in the comparator function in order-by.

I think you're right here. We're currently checking for `a == null && b == null` which means that if only one of them is `null` or undefined` (note: `undefined == null` is `true`) then the check fails and we will try to call `toString()` on it on the next line.

However, i don't think we should rewrite it as:
```ts
if (a == null || b == null) {
  return 0
}
```

Because now when comparing a nullish value with a non-nullish value we will say that they are equal (since we return 0).
I'd rather resort to JS semantics by using the built-in comparison operators. Something like:
```ts
if (a == null || b == null) {
  if (a == b) return 0
  if (a < b) return -1
  return 1
}
```
This ensures that our comparison is consistent with JS' comparison in cases like: `comparator(null, 5)`.

kevin-dp (changes requested):
> I believe this is just caused by the last bit of logic in the comparator function in order-by.

I think you're right here. We're currently checking for `a == null && b == null` which means that if only one of them is `null` or undefined` (note: `undefined == null` is `true`) then the check fails and we will try to call `toString()` on it on the next line.

However, i don't think we should rewrite it as:
```ts
if (a == null || b == null) {
  return 0
}
```

Because now when comparing a nullish value with a non-nullish value we will say that they are equal (since we return 0).
I'd rather resort to JS semantics by using the built-in comparison operators. Something like:
```ts
if (a == null || b == null) {
  if (a == b) return 0
  if (a < b) return -1
  return 1
}
```
This ensures that our comparison is consistent with JS' comparison in cases like: `comparator(null, 5)`.

gregermendle:
@KyleAMathews Sorry i've been away for a bit on travel. I'll try to reproduce an example of the issue and post here.

kevin-dp:
> @KyleAMathews Sorry i've been away for a bit on travel. I'll try to reproduce an example of the issue and post here.

@gregermendle i added some tests that reproduce this issue: https://github.com/TanStack/db/pull/167/commits/e0d103f8d2d0a583217d2e473e42132154f44c58
You can see the failing CI test that reproduces your issue here: https://github.com/TanStack/db/actions/runs/15613750774/job/43981203037?pr=167

samwillis:
This was fixed as part of a refactor in #167

PR #144, opened 2025-06-06
key all streams and remove keyBy

samwillis:
This removes `keyBy` by ensuring all streams a keyed throughout a query, using a composite key for joins and keying on a composite of the groupBy values for when thats used.

Joined keys are formed as `[${mainInputKey},${joinedInputKey}]` - so a key looks like: `[123,234]`, or `[123,[321,234]]` for multiple joins. This is implemented in the compiled pipeline by tanstack db query compiler.

GroupBy uses a JSON front for the key with the key+values extracted for the groupBy - this is how the key works in the D2TS groupBy operator and is just exposed on the resulting pipeline.


packages/db/src/query/compiled-query.ts
```diff
@@ -116,14 +116,20 @@ export class CompiledQuery<TResults extends object = Record<string, unknown>> {
     const input = this.inputs[inputKey]!
     const multiSetArray: MultiSetArray<unknown> = []
     for (const change of changes) {
+      // We remove the KEY::{UUID} prefix from the key used as the key for the D2
+      // pipeline otherwise it will accumulate on subsequent queries and we end up with
+      // long and complex keys
+      const key = change.key.startsWith(`KEY::`)
+        ? change.key.slice(42)
+        : change.key
```
samwillis: @KyleAMathews I am having second thoughts about this, particularly hard coding it to the length of the prefix for a UUID key - that won't work with Electric collections...

I know we need a globally unique key for transaction/mutations, but in future it would be nice if we can use the keying on the main Map in a collection as an index when querying - having the globally unique key as the map key makes that a little difficult (we would have to prefix the collection ID first).

Could we just prefix the collection id when we need the key to be globally unique rather than always when using the key?

KyleAMathews: We do have the id as well. Which the key is based on. So we can basically already do this?

We could also make the key shorter e.g. hash it if that helps.

samwillis: Yep, I've swapped this to use the `getId` callback from the collection config - much cleaner.

I still want to consider if the internal map in the collection needs to use the composite key, but this more related to when we add indexes and want to be able to select an ordered range of items based on their PK. Something for later.

I think this PR is ready if you are happy with it.

KyleAMathews (approved):
:shipit: 

great refactor!

I think you need to update the PR body again?

PR #145, opened 2025-06-06
Switch to Collection options factories instead of extending the Collection class

### LINKED ISSUE #132: Refactor core collection API
@tannerlinsley mocked this up last week and it makes a ton of sense:

```tsx
// Define collections to load data into
  queryCollectionOptions({
    queryKey: ['todos'],
    queryFn: async () =>
      fetch('/api/todos').then((res) => res.json() as Promise<Todo[]>),
    getId: (item) => item.id,
    schema: todoSchema, // any standard schema
    operations: {
      onInsert: async (mutation) => {
        await fetch('/api/todos', {
          method: 'POST',
          body: JSON.stringify(mutation.changes),
        })
      },
      onUpdate: async (mutation) => {
        await fetch(`/api/todos/${mutation.original.id}`, {
          method: 'PUT',
          body: JSON.stringify(mutation.modified),
        })
      },
      onDelete: async (mutation) => {
        await fetch(`/api/todos/${mutation.original.id}`, {
          method: 'DELETE',
        })
      },
    },
  }),
)

const Todos = () => {
    query.from({ todoCollection }).where('@completed', '=', false),
  )

  const complete = (todo) =>
    // Instantly apply your mutation
      draft.completed = true
    })

  return (
    <ul>
      {todos.map((todo) => (
        <li key={todo.id} onClick={() => complete(todo)}>
          {todo.text}
        </li>
      ))}
    </ul>
  )
}
```

This was inspired by feedback he got that `useOptimisticMutation` seems repetative/verbose. "Can't we just handle mutations in the collection?"

The original design had a `mutationFn` in the collection — which worked nicely but was somewhat verbose with if statements around handling inserts vs. updates vs. delete. The switch to `onInsert`, etc. helps a lot with this.

It also had the weakness that you persist cross-collection changes together e.g. adding a new user to a team means modifying at least 2 tables and the backend wants to do it in the same transaction vs. splitting up changes across multiple API calls. See the RFC for the full rational behind the change — https://github.com/TanStack/db/issues/29

What Tanner and I agreed on though is that while we need `useOptimistcMutation` for a lot of edge cases, ~80% of the time, simple collection mutation handlers are plenty. Tanner also suggested that `useOptimisticMutation` could not require a `mutationFn` in which case, each mutation would be handled by its collection.

The other notable thing about this proposal is we switch from our current `createQueryCollection` and `createElectricCollection` to a `createCollection(createQueryOptions)` API. This has two advantages — it makes it clear that there's nothing special about collection types to encourage more people to make them. Also it makes it a lot easier to ensure types work. Other TanStack libraries have adopted this pattern with great success.

KyleAMathews:
This refactors `ElectricCollection` and `QueryCollection` into factory functions (`electricCollectionOptions` and `queryCollectionOptions`) that return standard `CollectionConfig` objects and utility functions. Also adds a `createCollection` function to standardize collection instantiation.

Partially fixes #132

### DIFF
--- a/.changeset/smart-points-chew.md
+++ b/.changeset/smart-points-chew.md
@@ -1,0 +1,9 @@
+---
+"@tanstack/db-collections": patch
+"@tanstack/db-example-react-todo": patch
+"@tanstack/db": patch
+---
+
+Switch to Collection options factories instead of extending the Collection class
+
+This refactors `ElectricCollection` and `QueryCollection` into factory functions (`electricCollectionOptions` and `queryCollectionOptions`) that return standard `CollectionConfig` objects and utility functions. Also adds a `createCollection` function to standardize collection instantiation.


packages/db/src/collection.ts
```diff
@@ -180,16 +193,27 @@ export class Collection<T extends object = Record<string, unknown>> {
...
+    if (config.id) {
+      this.id = config.id
+    } else {
+      this.id = crypto.randomUUID()
+    }
+
+    // eslint-disable-next-line
+    if (!config.sync) {
       throw new Error(`Collection requires a sync config`)
     }
```
samwillis, line 218: Not really related to this PR, but in the thoughts about a "local collection", could we just make `sync` optional? then to create a local collection it's just a call to `createCollection()`?

samwillis (approved):
LGTM 👍

packages/db/src/collection.ts
KyleAMathews: hmm yeah possibly! We'd still have to handle "sync" internally to immediately take commited transactions and write it to the store. Perhaps this is a good base case though.

PR #149, opened 2025-06-09
Allow arrays in type of RHS in where clause when using set membership operators

kevin-dp:
Fixes https://github.com/orgs/TanStack/projects/5/views/1?pane=issue&itemId=113406395&issue=TanStack%7Cdb%7C100

Modifies the type of `BaseQueryBuilder.where` to type the RHS based on the comparator being used and if it is a set membership operator (i.e. `in` or `not in`) then the RHS **must** be an array.


KyleAMathews:
for the query example you provided — wouldn't that require parsing the array in the string? What's the use case there?

kevin-dp:
> for the query example you provided — wouldn't that require parsing the array in the string? What's the use case there?

Was raising this as i'm not sure if we want to allow raw sql strings or not.
Was chatting with @samwillis which was saying we don't want that and instead only allow JS array to be passed.
In that case, this PR should be fine.

KyleAMathews:
Ok, looks great than other than the failing check

kevin-dp:
@KyleAMathews not sure what's going on with the tests but those failures don't seem related to the changes in this PR?

KyleAMathews:
Hmm yeah dunno. I'll figure it out in a bit and merge it.

KyleAMathews:
@kevin-dp I get the same errors locally — so they do seem legit. I'm not sure either why your small change is causing so many failures.

KyleAMathews:
I realized what happened here — adding `packages/db/tests/query/query-builder/where.test-d.ts` triggers vitest to enable typechecking — which we hadn't had on before. I'll open a separate PR to fix all the type errors.

PR #152, opened 2025-06-09
feat: live query `where` clause as a callback

samwillis:
this enables the use of a callback for a `where` and `having` clauses, enabling this type of query:

```ts
  query
    .where(({ todo }) => todo.completed && todo.user_id === 'xyz')
)
```

- The existing where clauses syntax still work as before
- This implementation just passes the callback thought to the D2 pipeline filter where the where filtering happens.
- Type inference fully works in the callback, you get type hints for the based in row context
- There is a route to inferring which sources a where clause callback uses by preflight testing the callback with a proxy. this will allow a query optimiser to optimise the position of the filtering in the pipeline (such as a filter only touching the main collection running before the join). This is work we will do in future
- both `where` and `having` can be called multiple times, with both the original syntax and new callback, each essentially being `AND`ed together. This is important for future work where the original syntax (or a variant of it) is needed to query indexes before then the more expressive JS callback for further refining a query.

so say in future there was an index on `completed`:

```ts
  query
    .where('@todo.completed', '=', true) // or a new syntax that has the same effect
    .where(({ todo }) => todo.user_id === 'xyz')
)
```

### DIFF
--- a/.changeset/floppy-berries-enjoy.md
+++ b/.changeset/floppy-berries-enjoy.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+live query where clauses can now be a callback function that receives each row as a context object allowing full javascript access to the row data for filtering


KyleAMathews (approved):
nice! It'll be great having the expressivity of JS for where/having clauses.

More perhaps a question for the query syntax rewrite but I'd imagine there's use cases for OR not just AND for separate clauses?

samwillis:
> I'd imagine there's use cases for OR not just AND for separate clauses?

Yes, I think this would beed to be nested inside a where clause rather than as additional ones. Something to refine as we move forward.

PR #155, opened 2025-06-09
WIP reactor collection

samwillis:
This is a **WIP** significant refactor of the `Collection` replacing the Store based derived state with a more fine grade system that will be faster and more efficient.

Rather than computing a full "derived state" on each change, we only compute a derived upserts and deletes, along with emitting insert/update/delete messages on each change. This should result in much less object churn and far less GC.

The `Collection` is now also more `Map` like with many similar methods `keys`, `values`, `entries`, `get(key)` and `has(key)`.

You can subscribe to either changes to the whole collection, or to a single key - perfect for fine grade reactivity.

Queries still "sync" into a collection as their materialised state.

Finally, as there is only loose coupling between the live query system and collections, the latter can be used without a live query if wanted.

Still to do:

- [ ] React hook
- [ ] Vue hook

There is also a note in the code somewhere about changes to the query system for ordered collection, these should have an ordered set of keys so that all the generators (`keys`, `values`, `entries`) fall out of the `keys` generator.

*why not use a D2 pipeline to implement this?** because I think this will be more efficient (far less state is needed if we hand roll it), and keep the two loosely coupled.

More to do...

### DIFF
--- a/packages/db/src/collection.ts
+++ b/packages/db/src/collection.ts
@@ -172,18 +172,23 @@ export class SchemaValidationError extends Error {
 
 export class Collection<T extends object = Record<string, unknown>> {
   public transactions: SortedMap<string, Transaction>
-  
+
   // Core state - make public for testing
   public syncedData = new Map<string, T>()
   public syncedMetadata = new Map<string, unknown>()
-  
-  // Optimistic state tracking - make public for testing  
+
+  // Optimistic state tracking - make public for testing
   public derivedUpserts = new Map<string, T>()
   public derivedDeletes = new Set<string>()
 
   // Event system
   private eventListeners = new Set<EventListener<T>>()
   private keyListeners = new Map<string, Set<KeyListener<T>>>()
+
+  // Batching for subscribeChanges
+  private changesBatchListeners = new Set<
+    (changes: Array<ChangeMessage<T>>) => void
+  >()
 
   private pendingSyncedTransactions: Array<PendingSyncedTransaction<T>> = []
   private syncedKeys = new Set<string>()
@@ -333,16 +338,21 @@ export class Collection<T extends object = Record<string, unknown>> {
       }
     }
 
-    // Emit events for changes
-    this.emitOptimisticChanges(previousState, previousDeletes)
+    // Collect events for changes
+    const events: Array<CollectionEvent<T>> = []
+    this.collectOptimisticChanges(previousState, previousDeletes, events)
+
+    // Emit all events at once
+    this.emitEvents(events)
   }
 
   /**
-   * Emit events for optimistic changes
+   * Collect events for optimistic changes
    */
-  private emitOptimisticChanges(
+  private collectOptimisticChanges(
     previousUpserts: Map<string, T>,
-    previousDeletes: Set<string>
+    previousDeletes: Set<string>,
+    events: Array<CollectionEvent<T>>
   ): void {
     const allKeys = new Set([
       ...previousUpserts.keys(),
@@ -360,15 +370,15 @@ export class Collection<T extends object = Record<string, unknown>> {
       )
 
       if (previousValue !== undefined && currentValue === undefined) {
-        this.emitEvent({ type: `delete`, key, value: previousValue })
+        events.push({ type: `delete`, key, value: previousValue })
       } else if (previousValue === undefined && currentValue !== undefined) {
-        this.emitEvent({ type: `insert`, key, value: currentValue })
+        events.push({ type: `insert`, key, value: currentValue })
       } else if (
         previousValue !== undefined &&
         currentValue !== undefined &&
         previousValue !== currentValue
       ) {
-        this.emitEvent({
+        events.push({
           type: `update`,
           key,
           value: currentValue,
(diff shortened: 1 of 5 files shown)


packages/db/src/collection.ts
```diff
@@ -430,63 +300,380 @@ export class Collection<T extends object = Record<string, unknown>> {
...
+
+        if (event.previousValue) {
+          ;(changeMessage as any).previousValue = event.previousValue
+        }
+
+        return changeMessage
+      })
+
+      for (const listener of this.changesBatchListeners) {
+        listener([...changeMessages])
```
KyleAMathews:
```suggestion
        listener(changeMessages)
```

changeMessages is already a new array

packages/db/src/collection.ts
```diff
@@ -430,63 +300,380 @@ export class Collection<T extends object = Record<string, unknown>> {
...
+    }
+
+    // Fall back to synced data
+    return this.syncedData.has(key)
+  }
+
+  /**
+   * Get the current size of the collection (virtual)
+   */
+  public get size(): number {
```
KyleAMathews, line 524: recompute this incrementally in `recomputeOptimisticState` & when data is synced

packages/db/src/collection.ts
```diff
@@ -430,63 +300,380 @@ export class Collection<T extends object = Record<string, unknown>> {
...
+    ).length
+
+    return syncedSize - deletesFromSynced + upsertsNotInSynced
+  }
+
+  /**
+   * Get all keys (virtual derived state)
+   */
+  public *keys(): IterableIterator<string> {
+    const allKeys = new Set([
```
KyleAMathews: big allocation for large collections:

Suggested code from Gemini:

```ts
// Suggested Improvement
public *keys(): IterableIterator<string> {
  // Yield keys from synced data, skipping any that are deleted.
  for (const key of this.syncedData.keys()) {
    if (!this.derivedDeletes.has(key)) {
      yield key
    }
  }
  // Yield keys from upserts that were not already in synced data.
  for (const key of this.derivedUpserts.keys()) {
    if (!this.syncedData.has(key) && !this.derivedDeletes.has(key)) {
       // The derivedDeletes check is technically redundant if inserts/updates always remove from deletes,
       // but it's safer to keep it.
      yield key
    }
  }
}
```

KyleAMathews (approved):
very nice! I love the new Map APIs (though they should probably work on the ID not the internal key?)

wobsoriano:
Nice! I think the other frameworks were limited due to TS Store and this changes it. I'll update the Svelte PR and improve existing ones once this lands 🎉

samwillis:
@KyleAMathews I've made those changes.

100% agree on the key, the composite key should be an internal detail that isn't leaked. It would make much more sense for the internal map to just use the actual id of the collection, it makes much more sense for the public api, and then we are able to in future use it as an index on the id for faster queries that use the id.

I'm not 100% sure on how you are using the composite key, I would need to look through the transaction code to understand where to make the changes. But my thinking was to just compute it on the fly when needed, rather than persist it.

The other thing I wanted to suggest is that we remove the restriction on having to have a `sync` config. A local collection should just be a collection without one. (it may need a dummy one, I've not looked in detail)

packages/db-collections/tests/electric.test.ts
```diff
@@ -107,8 +107,8 @@ describe(`Electric Integration`, () => {
 
     expect(collection.state).toEqual(
       new Map([
-        [`KEY::${collection.id}/1`, { id: 1, name: `Test User` }],
```
KyleAMathews, line 110: lovely change haha

packages/db/src/collection.ts
```diff
@@ -1049,7 +1192,19 @@ export class CollectionImpl<T extends object = Record<string, unknown>> {
    * @returns An Array containing all items in the collection
    */
   get toArray() {
-    return this.derivedArray.state
+    const array = Array.from(this.values())
+
+    // Currently a query with an orderBy will add a _orderByIndex to the items
+    // so for now we need to sort the array by _orderByIndex if it exists
+    // TODO: in the future it would be much better is the keys are sorted - this
```
KyleAMathews, line 1199: let's create a follow-up issue

packages/db/src/collection.ts
```diff
@@ -1090,16 +1245,99 @@ export class CollectionImpl<T extends object = Record<string, unknown>> {
...
+
+    return () => {
+      this.changeListeners.delete(callback)
+    }
+  }
+
+  /**
+   * Subscribe to changes for a specific key
+   */
+  public subscribeChangesKey(
```
KyleAMathews, line 1267: throw error if key doesn't exist?

KyleAMathews: Or is it useful to preemptively subscribe to a key?

packages/db/src/collection.ts
```diff
@@ -1090,16 +1245,99 @@ export class CollectionImpl<T extends object = Record<string, unknown>> {
...
+    this.recomputeOptimisticState()
+  }
+
+  private _storeMap: Store<Map<TKey, T>> | undefined
+
+  /**
+   * Returns a Tanstack Store Map that is updated when the collection changes
+   * This is a temporary solution to enable the existing framework hooks to work
+   * with the new internals of Collection until they are rewritten.
+   * TODO: Remove this once the framework hooks are rewritten.
```
KyleAMathews, line 1314: let's create follow-up issues

packages/db/src/collection.ts
KyleAMathews: I also don't see any tests for this?

KyleAMathews (approved):
Just a few questions but otherwise looks great!

packages/db/src/collection.ts
samwillis: Yep, I think preemptively subscribing to a key is useful.

I'll add tests.

PR #161, opened 2025-06-11
Expose utilities on collection instances

### LINKED ISSUE #159: Add support for setting utility functions on collections
I took a stab at it while working on #156 but didn't get it working yet.

The problem is that right now, different collection types might want to add utilities for devs to call e.g. with QueryCollection, you'd want to be able to call `collection.refetch()`.

The trick though is how to pass utility functions into `createCollection` as options and then get them correctly assigned and typed.

KyleAMathews:
Fix #159 

Implemented a utility exposure pattern for TanStack DB collections that allows utility functions to be passed as part of collection options and exposes them under a `.utils` namespace, with full TypeScript typing.

## Changes
- Refactored `createCollection` in packages/db/src/collection.ts to accept options with utilities directly
- Added `utils` property to CollectionImpl
- Added TypeScript types for utility functions and utility records
- Changed Collection from a class to a type, updating all usages to use createCollection() instead
- Updated Electric/Query implementations

## Benefits
- Utilities are now ergonomically accessible under `.utils`
- Full TypeScript typing is preserved for both collection data and utilities
- API is clean and straightforward - users can call `createCollection(optionsCreator(config))` directly
- Zero-boilerplate TypeScript pattern that infers utility types automatically

### DIFF
--- a/packages/db/src/collection.ts
+++ b/packages/db/src/collection.ts
@@ -33,14 +33,16 @@ interface PendingSyncedTransaction<T extends object = Record<string, unknown>> {
 }
 
 /**
- * Enhanced Collection type that includes both data type T and utilities U
+ * Enhanced Collection interface that includes both data type T and utilities TUtils
  * @template T - The type of items in the collection
  * @template TUtils - The utilities record type
  */
-export type Collection<
+export interface Collection<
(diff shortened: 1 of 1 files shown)


samwillis (approved):
Looks great! :shipit:

PR #165, opened 2025-06-12
fix: string comparison for sorting in descending order

kevin-dp:
Fixes https://github.com/orgs/TanStack/projects/5/views/1?pane=issue&itemId=113401791&issue=TanStack%7Cdb%7C112

We're comparing strings using `localeCompare` and we try to sort in descending order by doing a sort in ascending order over the inverted strings. However, string inversion does not work with `localeCompare`. All details in https://github.com/TanStack/db/issues/112#issuecomment-2965781244.

### DIFF
--- a/.changeset/tricky-areas-drop.md
+++ b/.changeset/tricky-areas-drop.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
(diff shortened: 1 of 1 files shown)


KyleAMathews (approved):
Looks good!

PR #166, opened 2025-06-12
Fix ordering of ts update overloads & fix a lot of type errors in tests

### LINKED ISSUE #147: update callback isn't typed

KyleAMathews:
Fix #147


eslint.config.mjs
```diff
@@ -15,6 +15,18 @@ export default [
       "prettier/prettier": `error`,
       "stylistic/quotes": [`error`, `backtick`],
       ...prettierConfig.rules,
+      "@typescript-eslint/no-unused-vars": [
+        `error`,
+        { argsIgnorePattern: `^_`, varsIgnorePattern: `^_` },
```
samwillis, line 20: So happy to se you add this rule 😂

samwillis (approved):
LGTM - I think there are a few places we will have conflicts with #155 in the tests where we fixed the same problems 😁

eslint.config.mjs
KyleAMathews: haha I'll do the merge this time then 👍

PR #167, opened 2025-06-12
fix: fix null/undefined issue in comparator and refactor it

kevin-dp:
Fixes the issue described here: https://github.com/TanStack/db/pull/139


samwillis (approved):
Great refactor! 🥳

PR #171, opened 2025-06-16
chore: Unit test for ordering by a date property

kevin-dp:
I added a unit test to ensure ordering on a date property works. It used to be broken before (cf. https://github.com/orgs/TanStack/projects/5/views/1?pane=issue&itemId=113401831&issue=TanStack%7Cdb%7C123) because we were relying on string comparison which was broken. We've previously fixed string comparison and we're now using JS' built-in `<` comparison on Dates.


samwillis (approved):
:shipit:

PR #178, opened 2025-06-17
Show in PR comments the size of the package & any changes

KyleAMathews:


KyleAMathews:
Yeah I was just testing that the size comparison worked

Issue #170, opened 2025-06-13
Changes to the live query syntax


samwillis:
We want to take some time to re-assess the query builder syntax, and refine it to be both expressive but also very easy to understand and use. This issue is to discuss and track the changes.

## Current syntax

The current declarative syntax somewhat similar to Kysely - but with an `@column` syntax for column references, this allows mixing column references with other literals in the same expression.

This declarative syntax enables the parsing of a where clause for optimisation and the potentual use of indexes in future.

```ts
useLiveQuery((q) => q.from({ issues }).where("@status", "=", "open"));
```

We also have an imperative callback syntax that passes the function to the filter directly - this doesn't enable the parsing of the where clause for optimisation, but allows very expressive queries _now_.

```ts
useLiveQuery((q) =>
  q.from({ issues }).where(({ issue }) => issue.status === "open")
);
```

for `select` we have two current syntaxes - a declarative syntax:

```ts
useLiveQuery((q) =>
  q.from({ issues }).select("@id", "@title", "@status", "@updatedAt")
);
```

and an imperative syntax:

```ts
useLiveQuery((q) =>
  q.from({ issues }).select(({ issue }) => ({
    id: issue.id,
    title: issue.title.toLowerCase(),
    status: issue.status,
    updatedAt: issue.updatedAt.toISOString(),
  }))
);
```

## What we don't like about the current syntax

I think the main thing that we don't like is the `@column` syntax for column references, while expressive, it is unusual and probably has footguns.

Using string for the comparators and collection/column/prop references has also been questioned - would using objects and functions make for a cleaner syntax?

### Kysely

https://kysely.dev/docs/examples/where/complex-where-clause

Kysely has a declarative syntax similar to the current syntax, although with explicit column references as the first argument to the `where` method.

```ts
const result = await db
  .selectFrom("issues")
  .selectAll()
  .where("status", "=", "open")
  .execute();
```

It has an expression builder callback for more expressive where clauses:

```ts
const persons = await db
  .selectFrom("person")
  .selectAll()
  // 1. Using the `or` method on the expression builder:
  .where((eb) =>
    eb.or([
      eb("first_name", "=", "Jennifer"),
      eb("first_name", "=", "Sylvester"),
    ])
  )
  // 2. Chaining expressions using the `or` method on the
  // created expressions:
  .where((eb) =>
    eb("last_name", "=", "Aniston").or("last_name", "=", "Stallone")
  )
  .execute();

const persons = await db
  .selectFrom("person")
[...]

PR #185, opened 2025-06-18
feat: new query syntax

samwillis:
Implements the new query syntax from #170

todo:

- [x] new intermedia representation
- [x] query builder that outputs the intermedia representation
- [x] compiler that compiles the ir to a D2 pipeline
- [x] test - and it works...

features:
- [x] `from` collection
- [x] `from` subquery
- [x] `where`
- [x] `join` collection
- [x] `join` subquery
- [x] `groupBy`
- [x] `having`
- [x] `orderBy` (with limit+offset)
- [x] return type of a query fully derived from query
- [x] React `useLiveQuery`
- [x] Vue `useLiveQuery`

------

The main entry point is a new `createLiveQueryCollection` function:

```ts
createLiveQueryCollection((q) =>
  q.from({ user: usersCollection }).select(({ user }) => ({
    id: user.id,
    name: user.name,
    age: user.age,
    email: user.email,
    active: user.active,
  }))
)
```

this wraps a new `liveQueryCollectionOptions` and `createCollection` call - the former of which can be used like the other `xxxCollectionOptions` function we have:

```ts
createCollection(liveQueryCollectionOptions({
  id: 'name-for-collection`,
  query: (q) =>
    q.from({ user: usersCollection }).select(({ user }) => ({
      id: user.id,
      name: user.name,
      age: user.age,
      email: user.email,
      active: user.active,
    })),
  // any other options
}))
```

`createLiveQueryCollection` can be called with either a query builder function, or a config object that takes the `query` param as a query builder function.


samwillis:
The one failing test in this at the moment is fixed by these d2mini PRs:
- https://github.com/electric-sql/d2ts/pull/69
- https://github.com/electric-sql/d2ts/pull/70

packages/db/src/query2/builder/functions.ts
```diff
@@ -0,0 +1,392 @@
...
+): Expression<boolean>
+export function eq(
+  left: RefProxy<number>,
+  right: number | RefProxy<number> | Expression<number>
+): Expression<boolean>
+export function eq(
+  left: RefProxy<boolean>,
+  right: boolean | RefProxy<boolean> | Expression<boolean>
+): Expression<boolean>
+export function eq<T>(
```
kevin-dp: I think that this signature subsumes the 3 previous ones, could you check?

packages/db/src/query2/builder/functions.ts
```diff
@@ -0,0 +1,392 @@
...
+): Expression<boolean>
+export function eq(
+  left: string,
+  right: string | Expression<string>
+): Expression<boolean>
+export function eq(
+  left: number,
+  right: number | Expression<number>
+): Expression<boolean>
+export function eq(
```
kevin-dp: This signature and the previous 2 should be equivalent to:
```ts
export function eq<T extends string | number | boolean>(
  left: T,
  right: T | Expression<T>
): Expression<boolean>
```

packages/db/src/query2/builder/functions.ts
```diff
@@ -0,0 +1,392 @@
...
+  left: Expression<string>,
+  right: string | Expression<string>
+): Expression<boolean>
+export function eq(
+  left: Expression<number>,
+  right: number | Expression<number>
+): Expression<boolean>
+export function eq(
```
kevin-dp: This signature and the previous 2 can be reduced to this equivalent signature:
```ts
export function eq<T extends string | number | boolean>(
  left: Expression<T>,
  right: T | Expression<T>
): Expression<boolean>
```

Applying all my suggestions i get:
```ts
export function eq<T>(
  left: RefProxy<T>,
  right: T | RefProxy<T> | Expression<T>
): Expression<boolean>
export function eq<T extends string | number | boolean>(
  left: T,
  right: T | Expression<T>
): Expression<boolean>
export function eq<T extends string | number | boolean>(
  left: Expression<T>,
  right: T | Expression<T>
): Expression<boolean>
```
Which can be simplified further:
```ts
export function eq<T>(
  left: RefProxy<T>,
  right: T | RefProxy<T> | Expression<T>
): Expression<boolean>
export function eq<T extends string | number | boolean>(
  left: T | Expression<T>,
  right: T | Expression<T>
): Expression<boolean>
```
Note that the first signature doesn't restrict `T` to only `string | number | boolean`, is this on purpose?

packages/db/src/query2/builder/functions.ts
```diff
@@ -0,0 +1,392 @@
...
+  left: Expression<boolean>,
+  right: boolean | Expression<boolean>
+): Expression<boolean>
+export function eq(left: any, right: any): Expression<boolean> {
+  return new Func(`eq`, [toExpression(left), toExpression(right)])
+}
+
+export function gt(
```
kevin-dp: These signatures look very similar to `eq`. They can be simplified similarly to how we simplified `eq`. However, conceptually, they are all binary operators and that's why they are so similar. So ideally we could define a generic type signature of binary operators and each binary operator is an instance of that type. That's something we can do if we use `const eq: BinaryOperator<..., ...> = (...) => { ... }` instead of function overloading like here.

kevin-dp (commented):
I only looked into the `functions.ts` file and left a bunch of comments. I think we can greatly simplify all those types.

KyleAMathews:
Here's gemini's take — I'm also curious about errors — does it just throw errors on any issues?

---

This is a substantial and well-executed Pull Request that fundamentally refactors the live query syntax and its underlying implementation. It moves away from a somewhat unconventional string-based syntax (@column) towards a more idiomatic, type-safe, and Kysely-inspired fluent builder API.

Here's a deep analysis:

**High-Level Summary:**

The PR introduces a new query builder and compiler system. The primary goal is to improve developer experience by providing a more intuitive, type-safe, and expressive way to construct queries, while also laying a robust foundation for future optimizations through a well-defined Intermediate Representation (IR). This involves replacing the old @column string references with a proxy-based system and string comparators with function calls.

**Key Architectural Changes:**

1.  **Intermediate Representation (IR - query/ir.ts):**
    
    *   This is a cornerstone of the new system. It defines a structured, declarative representation of a query (Query interface with From, Select, Join, Where, etc.).
        
    *   Expressions (Ref, Value, Func) and Aggregations (Agg) are clearly defined classes.
        
    *   This IR is what the builder constructs and what the compiler consumes. It's essential for parsing, optimization, and transforming queries.
        
2.  **Fluent Query Builder (query/builder/\*):**
    
    *   **BaseQueryBuilder (query/builder/index.ts):** Provides methods like from(), join(), where(), select(), orderBy(), groupBy(), limit(), offset().
        
    *   **RefProxy (query/builder/ref-proxy.ts):** This is the mechanism replacing @column. It uses JavaScript Proxy objects to capture property access paths (e.g., issue.status in a callback becomes \['issue', 'status'\]). This allows for type-safe, dot-notation access to columns within builder callbacks.
        
        *   It also cleverly handles table spreads in select (e.g., ...issue) using \_\_spreadSentinels.
            
    *   **Expression Functions (query/builder/functions.ts):** Functions like eq(), gt(), and(), or(), upper(), count() are provided. These replace string-based comparators and allow for a more natural, functional way of defining query logic. They return IR Func or Agg objects.
        
    *   **Type System (query/builder/types.ts):** A sophisticated and crucial part.
        
        *   Context: Tracks the schema available at each stage of query building, including joined tables and their join types. This enables strong type inference.
            
        *   RefProxyForContext, RefProxyFor: Generate type-safe proxies based on the Context, correctly handling optionality (e.g., fields from a LEFT JOIN'ed table become optional).
            
        *   SchemaFromSource, MergeContextWithJoinType, GetResult: Powerful mapped types to infer and transform schema and result types as the query is built.
            
3.  **Query Compiler (query/compiler/\*):**
    
    *   **compileQuery (query/compiler/index.ts):** The main entry point. Takes the IR Query and a map of input streams, and produces a D2 pipeline (ResultStream).
        
    *   **evaluators.ts:** Compiles IR expressions into optimized JavaScript functions. This means for a given expression, a dedicated function is generated to evaluate it against a row, avoiding runtime interpretation or branching.
        
    *   Clause-specific processors (joins.ts, group-by.ts, order-by.ts, select.ts): Each handles its respective part of the IR, transforming the D2 stream.
        
        *   The select.ts (processSelectToResults) introduces an intermediate \_\_select\_results property on the row, which is a good pattern for decoupling subsequent operations like orderBy.
            
    *   **Subquery Handling:** The compiler includes a QueryCache (WeakMap) to avoid recompiling the same subquery if it's referenced multiple times, which is important for performance.
        
4.  **Live Query Integration (query/live-query-collection.ts):**
    
    *   **createLiveQueryCollection / liveQueryCollectionOptions:** The primary user-facing API for creating live queries with the new builder.
        
    *   It takes a builder function (e.g., (q) => q.from(...).select(...)), uses buildQuery to get the IR, then compileQuery to get the D2 pipeline.
        
    *   Manages the D2 graph, input collections, and pipes the output of the compiled query to update the live collection's state.
        
    *   Handles ordering by using orderByWithFractionalIndex and storing the generated fractional indices in a WeakMap for the collection's compare function, ensuring stable ordering.
        

**Alignment with PR Goals:**

*   **Re-assess query builder syntax:** Achieved. The new syntax is a significant departure and improvement.
    
*   **Expressive:** Yes, the combination of fluent methods, proxy-based column access, and function-based operators is highly expressive.
    
*   **Easy to understand and use:** The new syntax is more aligned with common JavaScript/TypeScript patterns and ORM/query builder libraries, making it more intuitive than the previous @column and string-based system.
    
*   **@column syntax disliked:** Successfully replaced by RefProxy (e.g., ({ issue }) => issue.status).
    
*   **String comparators questioned:** Successfully replaced by functions (e.g., eq(left, right)).
    
*   **Cleaner syntax with objects/functions:** This is the core of the new API.
    

**Detailed File Changes & Impact:**

*   **package.json:** Minor bump of @electric-sql/d2mini. Likely for new features or fixes in D2 that this PR leverages (e.g., orderByWithFractionalIndex).
    
*   **packages/db/src/index.ts:** Removal of utils export indicates the old utility functions are no longer needed or have been encapsulated elsewhere.
    
*   **NEW packages/db/src/query/builder/\*:** Entirely new, well-structured query builder.
    
    *   functions.ts: Clean, type-safe functions for all operations.
        
    *   index.ts: BaseQueryBuilder with fluent API.
        
    *   ref-proxy.ts: Clever use of Proxies for column access.
        
    *   types.ts: Advanced TypeScript for strong type safety and inference.
        
*   **NEW packages/db/src/query/compiler/\*:** New compiler logic.
    
    *   Organized by clause (joins.ts, group-by.ts, etc.).
        
    *   evaluators.ts is key for performance.
        
*   **NEW packages/db/src/query/ir.ts:** Defines the abstract syntax tree for queries. Critical for the declarative nature and enabling optimizations.
    
*   **NEW packages/db/src/query/live-query-collection.ts:** Integrates the new builder/compiler with the existing live collection mechanism. Provides a user-friendly API.
    
*   **DELETED packages/db/src/query/\* (old files):** Complete removal of the old query system, indicating a full replacement.
    
*   **NEW packages/db/tests/query/\*:** Addition of new tests for the new system, which is essential for a change of this magnitude.
    

**Positive Aspects:**

1.  **Improved Developer Experience:** The new API is much more intuitive, type-safe, and less prone to errors (e.g., typos in string columns/operators).
    
2.  **Type Safety:** The extensive use of TypeScript, especially in query/builder/types.ts, provides excellent autocompletion and compile-time checks. The way join optionality propagates to types is particularly well-handled.
    
3.  **Maintainability & Extensibility:** The separation into IR, Builder, and Compiler makes the system more modular and easier to maintain and extend. New SQL features or functions can be added by extending the IR, builder, and compiler components.
    
4.  **Performance Potential:**
    
    *   The pre-compilation of expressions in evaluators.ts into JavaScript functions should be more performant at runtime than interpreting string-based expressions repeatedly.
        
    *   The subquery caching mechanism is a good optimization.
        
5.  **Clear Foundation for Optimization:** The IR provides a solid base for implementing query optimizations in the future (e.g., predicate pushdown, join reordering).
    
6.  **Consistency with Modern Libraries:** The API style is familiar to developers who have used libraries like Kysely or other modern ORMs/query builders.
    
7.  **Handling of Spreads:** The \_\_SPREAD\_SENTINEL\_\_ approach in select for table spreads is a practical solution.
    

**Potential Concerns / Questions:**

1.  **TypeScript Complexity:** The types in query/builder/types.ts are highly advanced. While powerful, they could be a barrier to understanding for contributors and might have a non-trivial impact on TypeScript compilation times for users. This is often a trade-off for rich type inference.
    
2.  **Error Reporting:** With a compiler and IR, user-facing error messages become very important. How are errors from the builder (e.g., invalid join condition) or compiler (e.g., type mismatch not caught by TS) reported to the user?
    
3.  **Imperative Fallback:** The PR description mentions the old "imperative callback syntax that passes the function to the filter directly" for "very expressive queries now." Does the new system offer an equivalent escape hatch if some complex logic cannot be expressed through the IR? Or is the new IR/function set deemed sufficiently expressive? (The new builder uses callbacks, but they construct IR; the old system allowed _any_ JS in the callback).
    
4.  **Testing:** The scale of this refactor is large. The new tests are crucial. A review of test coverage (though not code in this PR) would be important. Particularly edge cases for joins, subqueries, and type inference.
    

**Conclusion:**

This PR represents a significant and very positive evolution of the live query system. It addresses the core concerns with the previous syntax and introduces a modern, type-safe, and extensible architecture. The use of an Intermediate Representation, a fluent builder, and a dedicated compiler is a robust approach that will benefit developers and the long-term health of the library. The complexity of the type system is a notable point but seems justified by the level of type safety and developer assistance it provides. This is a high-quality contribution.

samwillis:
Thats a good overview and assessment I think.

Error reporting is something we could improve - probably in a follow-up PR. Currently there are a few explicit errors that can be thrown during the compile step, and these will just break it and show something in the console. During pipeline execution if there is something odd that throws, the errors are going to be quite obscure. Wrapping those two steps in a try/catch with some helpful error message with anything we can extract would be good. We have no less error handing than we did on the old version.

packages/vue-db/src/useLiveQuery.ts
```diff
@@ -1,60 +1,124 @@
...
 }
 
-export function useLiveQuery<
-  TResultContext extends Context<Schema> = Context<Schema>,
->(
-  queryFn: (
-    q: InitialQueryBuilder<Context<Schema>>
-  ) => QueryBuilder<TResultContext>,
+// Overload 1: Accept just the query function
+export function useLiveQuery<TContext extends Context>(
```
KyleAMathews, line 38: we were going to support passing a query collection too right?

samwillis: Yes!

samwillis:
There are few things still needed, but I'm going to add them in follow-up PRs:

- ~~passing a liveQueryCollection to a useLiveQuery~~ (done in this)
- ~~The js function callback variants of the where/having/select clauses~~
- The multiple returned collections from a live query
- defineLiveQuery
- docs

samwillis:
Also, the react and vue hooks are a MVP, they currently rebuild the state on each render, we can clearly optimise this.

packages/vue-db/src/useLiveQuery.ts
samwillis: done

KyleAMathews:
Installed this in the expo starter which has this for the query: `const { data: todos } = useLiveQuery((q) => q.from({ todoCollection }));`

There's no loader so the initial render has 0 items but it never re-renders after that. I called `todoCollection.subscribeChanges` and it does send in the inserts.

KyleAMathews:
An optimistic/sync mutation does get it to re-render

KyleAMathews:
I tried doing a select callback and got this js error for `foo`:

```ts
    q.from({ todoCollection }).select(({ todoCollection }) => {
      return {
        foo: `${todoCollection.id}`,
        id: todoCollection.id,
        text: todoCollection.text,
        completed: todoCollection.completed,
        created_at: todoCollection.created_at,
        updated_at: todoCollection.updated_at,
      };
    }),
```

`Uncaught TypeError: Cannot convert object to primitive value`

KyleAMathews:
For the main operators — it'd be great to have an inline example or two to remind people of the syntax
Screenshot 2025-06-27 at 9 27 18 AM

KyleAMathews:
I'm not sure I love the operators being individual exports (e.g. `import { eq } from "@tanstack/react-db"` as that makes them less discoverable e.g. right now I'm testing them & I need to look through the PR to get the full list. Maybe just `operators` so you can either do `import { operators as $ } from "@tanstack/db"` or `import { operators: { eq }} from "@tanstack/db"` ?

KyleAMathews:
> I tried doing a select callback and got this js error for foo:

Ohhhh that's because this isn't supported yet... just the operator selects... hmm that's going to confuse people

KyleAMathews:
Another potentially annoying part about individual operators is that the autoimport tool suggests both drizzle & react-db 😅 

Screenshot 2025-06-27 at 10 20 39 AM

KyleAMathews:
`isInFunc` is a bit odd... `inArray` & `notInArray` is better perhaps like Drizzle https://orm.drizzle.team/docs/operators

KyleAMathews:
I've changed my mind about putting operators under `operator` — there's not that many & w/ docs they'll be plenty discoverable 👍

KyleAMathews:
`concat` doesn't seem to work?

I wrote this code: `foo: concat([todoCollection.text, '-', todoCollection.completed])` in my select and got this output when rendering: `[{},"-",{}] FOO3`

Oh actually the test is misleading I guess? Removing the array made it work i.e. `concat(todoCollection.text, '-', todoCollection.completed)`

samwillis:
> > I tried doing a select callback and got this js error for foo:
> 
> Ohhhh that's because this isn't supported yet... just the operator selects... hmm that's going to confuse people

It's planned, as I noted above and on Discord, it will come in another PR asap. We need to decide how to name them - they have to be distinct from the main method as they now take callbacks too.

samwillis:
> I'm not sure I love the operators being individual exports (e.g. `import { eq } from "@tanstack/react-db"` as that makes them less discoverable e.g. right now I'm testing them & I need to look through the PR to get the full list. Maybe just `operators` so you can either do `import { operators as $ } from "@tanstack/db"` or `import { operators: { eq }} from "@tanstack/db"` ?

There was a discussion about this over on the issue here: https://github.com/TanStack/db/issues/170#issuecomment-2975309853

Happy to change this, we didn't really reach a conclusion. They could be an import, or provided for destructuring on the useLiveQuery callback.

samwillis:
> isInFunc

I thought I had it as `isIn` and aliased as `in`... clearly I messed something up.

samwillis:
> Oh actually the test is misleading I guess? Removing the array made it work i.e. `concat(todoCollection.text, '-', todoCollection.completed)`

Yes this is the correct syntax

samwillis:
> For the main operators — it'd be great to have an inline example or two to remind people of the syntax Screenshot 2025-06-27 at 9 27 18 AM

I'de love to, I need to find a way to stop the type being so verbose though as I think it will push any jsdoc down in the prompt... 🤔

samwillis:
> There's no loader so the initial render has 0 items but it never re-renders after that. I called `todoCollection.subscribeChanges` and it does send in the inserts.

Is that with a shape that is now lazy? I think thats behaving as it should as there are no rows yet?

KyleAMathews:
>  I need to find a way to stop the type being so verbose though

Explicit named types is the trick I think? 

> Is that with a shape that is now lazy? I think thats behaving as it should as there are no rows yet?

No the collection loads data just fine — the query just doesn't seem to detect the initial sync. Subsequent syncs & any optimistic changes do trigger a render.

samwillis:
> > I need to find a way to stop the type being so verbose though
> 
> Explicit named types is the trick I think?

I've tried to make this work, but failed. No patter what I do it always shows the context in the generic in the prompt. 

I have added jsdoc comments with examples, so thats an improvement. I think we can come back to trying to minimise the blob at the top later (and maybe a TS expert will come and work it out for us!)

image

samwillis:
@KyleAMathews  on `isIn`, it was mistakingly aliased in the test `isInFunc`, it is just `isIn`.

I'm open to renaming it, but not that it operates on both arrays and strings, so you can do `isIn('string', 'isIn works on both arrays and strings')`

Obv we can't call it `in` as thats a reserved word in JS. The another option is `contains` and flip the arguments around?

We could also split into two functions `inArray` and `inString`?

DawidWraga:
> > > I need to find a way to stop the type being so verbose though
> > 
> > Explicit named types is the trick I think?
> 
> I've tried to make this work, but failed. No patter what I do it always shows the context in the generic in the prompt. 
> 

I had a similar problem a while ago, this might be helpful 

https://dev.to/dawidcodes/simplifying-complex-type-display-in-typescript-and-vs-code-4f1

KyleAMathews:
Ah ok, maybe `inArray` and `includes`? It is a bit confusing to have both `like` and another string operator. It seems we could just have like and inArray?

samwillis:
I've renamed `isIn` to `inArray`

samwillis:
> I had a similar problem a while ago, this might be helpful

Thanks @DawidWraga, I had a good go at trying to use this pattern (the `& Simplify<{}>`), but sadly haven't managed to make it minimise the types. Your `Simplify` is almost identical to a `Prettify` I already had in a few places - it's a great trick.

packages/db/src/query/ir.ts
```diff
@@ -0,0 +1,125 @@
...
+    public name: string, // such as eq, gt, lt, upper, lower, etc.
+    public args: Array<Expression>
+  ) {
+    super()
+  }
+}
+
+export type Expression<T = any> = Ref<T> | Value<T> | Func<T>
+
+export class Agg<T = any> extends BaseExpression<T> {
```
kevin-dp: I'd prefer the slightly longer but more self-documenting `Aggregate` name

packages/db/src/query/ir.ts
```diff
@@ -0,0 +1,125 @@
...
+  public type = `func` as const
+  constructor(
+    public name: string, // such as eq, gt, lt, upper, lower, etc.
+    public args: Array<Expression>
+  ) {
+    super()
+  }
+}
+
+export type Expression<T = any> = Ref<T> | Value<T> | Func<T>
```
kevin-dp: Since `Agg` is a `BaseExpression` why is it not in `Expression` ? I think `Expression` might not be the correct name here because there are a bunch of things that are expressions but are not listed in this type (like `Agg` and `QueryRef`).

packages/db/src/query/builder/types.ts
```diff
@@ -0,0 +1,265 @@
+import type { CollectionImpl } from "../../collection.js"
+import type { Agg, Expression } from "../ir.js"
+import type { QueryBuilder } from "./index.js"
+
+export interface Context {
+  // The collections available in the base schema
+  baseSchema: Record<string, any>
```
kevin-dp: Why do we need to type the values as `any` in `baseSchema` and in `schema`?
Isn't there a base type for schemas that we can use instead of any?

packages/db/src/query/builder/types.ts
```diff
@@ -0,0 +1,265 @@
...
+  // The current schema available (includes joined collections)
+  schema: Record<string, any>
+  // the name of the source that was used in the from clause
+  fromSourceName: string
+  // Whether this query has joins
+  hasJoins?: boolean
```
kevin-dp, line 13: Isn't `hasJoins` equivalent to something along the lines of `Object.keys(joinTypes).nonEmpty`?
If it is, i would avoid introducing `hasJoins` explicitly as it opens the door for inconsistencies.

packages/db/src/query/builder/index.ts
```diff
@@ -0,0 +1,651 @@
...
+  RefProxyForContext,
+  ResultTypeFromSelect,
+  SchemaFromSource,
+  SelectObject,
+  Source,
+  WhereCallback,
+  WithResult,
+} from "./types.js"
+
+export function buildQuery<TContext extends Context>(
```
kevin-dp: As we discussed before, i'm not a fan of `buildQuery` and `getQuery`. They don't do much. I'd rather let devs write `new BaseQueryBuilder().from(...).select(...).getQuery()` instead of:
```ts
buildQuery((q) => q.from(...).select(...))
```
I guess what i don't like about this is using a callback to then always call it with the same argument.

packages/db/src/query/builder/index.ts
```diff
@@ -0,0 +1,651 @@
...
+    baseSchema: SchemaFromSource<TSource>
+    schema: SchemaFromSource<TSource>
+    fromSourceName: keyof TSource & string
+    hasJoins: false
+  }> {
+    if (Object.keys(source).length !== 1) {
+      throw new Error(`Only one source is allowed in the from clause`)
+    }
+
+    const alias = Object.keys(source)[0]! as keyof TSource & string
```
kevin-dp: `as keyof TSource & string` is always gonna type to just `string` so we can replace it by `as string`. I think we don't even need the typecast because `Object.keys` returns an `Array<string>`

packages/db/src/query/builder/index.ts
```diff
@@ -0,0 +1,651 @@
...
+    >,
+    type: TJoinType = `left` as TJoinType
+  ): QueryBuilder<
+    MergeContextWithJoinType<TContext, SchemaFromSource<TSource>, TJoinType>
+  > {
+    if (Object.keys(source).length !== 1) {
+      throw new Error(`Only one source is allowed in the join clause`)
+    }
+
+    const alias = Object.keys(source)[0]!
```
kevin-dp: All the code from L140 to L157 is a duplicated of the code in `from`. Let's extract it to a helper function `createRefForSource`.

packages/db/src/query/builder/index.ts
```diff
@@ -0,0 +1,651 @@
...
+    for (const [key, value] of Object.entries(selectObject)) {
+      if (isRefProxy(value)) {
+        select[key] = toExpression(value)
+      } else if (
+        typeof value === `object` &&
+        `type` in value &&
+        value.type === `agg`
+      ) {
+        select[key] = value
+      } else if (
```
kevin-dp: i would unify this branch with the previous one by modifying the last part of the conjunction to be `value.type === 'agg' || value.type === 'func'`

packages/db/src/query/builder/types.ts
```diff
@@ -0,0 +1,265 @@
...
+type IsExactlyUndefined<T> = [T] extends [undefined] ? true : false
+
+// Helper type to check if T includes undefined (is optional)
+type IsOptional<T> = undefined extends T ? true : false
+
+// Helper type to extract non-undefined type
+type NonUndefined<T> = T extends undefined ? never : T
+
+// Helper type to create RefProxy for a specific type with optionality passthrough
+export type RefProxyFor<T> = OmitRefProxy<
```
kevin-dp, line 114: This type is mind blowing. Not sure what it does. Could use a comment explaining what it does on a high level.

packages/db/src/query/compiler/joins.ts
```diff
@@ -0,0 +1,248 @@
...
+        joinOperator(joinedPipeline, `right`),
+        consolidate(),
+        processJoinResults(joinClause.type)
+      )
+    case `full`:
+      return mainPipeline.pipe(
+        joinOperator(joinedPipeline, `full`),
+        consolidate(),
+        processJoinResults(joinClause.type)
+      )
```
kevin-dp: Let's remove the code duplication by merging all these cases:
```ts
if (![`inner`, `left`, `right`, `full`].includes(joinType)) {
  throw new Error(`Unsupported join type: ${joinClause.type}`)
}

return mainPipeline.pipe(
  joinOperator(joinedPipeline, joinType),
  consolidate(),
  processJoinResults(joinClause.type)
)
```

packages/db/src/query/compiler/group-by.ts
```diff
@@ -0,0 +1,419 @@
...
+  }
+}
+
+/**
+ * Checks if two aggregate expressions are equal
+ */
+function aggregatesEqual(agg1: Agg, agg2: Agg): boolean {
+  if (agg1.name !== agg2.name) return false
+  if (agg1.args.length !== agg2.args.length) return false
+  return agg1.args.every((arg, i) => expressionsEqual(arg, agg2.args[i]))
```
kevin-dp: can be simplified into 1 conjunction:
```ts
return
  (agg1.name === agg2.name) &&
  (agg1.args.length === agg2.args.length) &&
  agg1.args.every((arg, i) => expressionsEqual(arg, agg2.args[i]))
```

docs/overview.md
```diff
@@ -33,10 +33,10 @@ const todoCollection = createCollection({
...
 const Todos = () => {
   // Bind data using live queries
-  const { data: todos } = useLiveQuery((query) =>
-    query
-      .from({ todoCollection })
-      .where('@completed', '=', false)
+  const { data: todos } = useLiveQuery((q) =>
+    q
+      .from({ todo: todoCollection })
+      .where(({ todo }) => eq(todo.completed, false))
```
kevin-dp: Why not use `not(todo.completed)` ?

docs/overview.md
```diff
@@ -258,22 +258,21 @@ Live queries return collections. This allows you to derive collections from othe
...
-const compiled = compileQuery(query)
-compiled.start()
-
-const completedTodoCollection = compiledQuery.results()
+const completedTodoCollection = createLiveQueryCollection({
+  startSync: true,
+  query: (q) =>
+    q
+      .from({ todo: todoCollection })
+      .where(({ todo }) => eq(todo.completed, true))
```
kevin-dp: Do we really need to `eq` it with `true`? Can't we just do `where(({ todo }) => todo.completed)` ?

kevin-dp (approved):
What a tremendous job!
I love how the queries look now compared to how they were before. Awesome! 🔥

packages/db/src/query/ir.ts
samwillis: I've renamed it BasicExpression, this is used in all the expression builder callbacks, but some except more (select accepts aggregates when there is a groupBy)

packages/db/src/query/builder/types.ts
samwillis: This interface is only ever used as a type to provide context through the query builder, it's never actually used to create an actual JS object. I needed a way to infer through the builder that there were now joins being used. If there is a way to count the number of props on the `context.schema` we could infer that >1 there are joins, but I was unable to work out how to do this (I suspect its not possible.

packages/db/src/query/builder/types.ts
samwillis: The actual schema is applied when the context is constructed, I suppose this is more accurately `Record<string, unknown>`, I've changed it to that and pulled dit out into a named type.

packages/db/src/query/builder/index.ts
samwillis: I think `keyof TSource` is `string | number` and so unless I'm mistaken this ensures that it is both a key of the `TSource` and only those that are strings?

packages/db/src/query/builder/index.ts
samwillis: Yep, I have changed this so we export a `Query` constructor so that the user can do:
```
const q = new Query()
  .from(...)
  .where(...)
```

`Query` is an alias of `BaseQueryBuilder` but typecast so the it only has the `from` method, all subsequent builder instances are full and have all the other methods except are type cast to hide the `from` method. (we also validate this at runtime)

Issue #76, opened 2025-05-12
Fix and document preloadCollection


KyleAMathews:
Not as essential as other stuff but something to do sooner than later as it's essential for building many apps.

thruflo:
How does it work with the createFooCollection factory functions? I saw preloadCollection took generic sync config. Can you pass it a collection?

KyleAMathews:
oh haha — yes that is broken now.

Ok, I'll rename this to "fix and document preloadCollection" 😆

KyleAMathews:
Replaced by https://github.com/TanStack/db/issues/195

PR #186, opened 2025-06-19
If a schema is passed, use that for the collection type

### LINKED ISSUE #116: Use schema type if types aren't directly passed in
You can add a type when creating a collection e.g. `createQueryCollection<{  id: string }>()`

But you can also set a schema with a https://github.com/standard-schema/standard-schema which has a standard way to pull out the types with `InferInput`.

So while we should default to the explicitly set type, we should also use the schema's type when it's added.

KyleAMathews:
You now must either pass an explicit type or schema - passing both will conflict.

I also removed the broken `preloadCollection` so I didn't have to fix its types 😆 the plan is make collections lazy by default and expose a `.start()` method which returns a promise that resolves when the collection finishes loading.

Fix #116 #76

### DIFF
--- a/packages/db/src/query/compiled-query.ts
+++ b/packages/db/src/query/compiled-query.ts
@@ -83,8 +83,9 @@ export class CompiledQuery<TResults extends object = Record<string, unknown>> {
                   type: `insert`,
                 })
               } else if (
-                inserts >= deletes &&
-                collection.has(valueWithKey._key as string | number)
+                inserts > deletes ||
+                (inserts === deletes &&
+                  collection.has(valueWithKey._key as string | number))
               ) {
                 console.log(2)
                 write({
--- a/packages/db/src/query/compiled-query.ts
+++ b/packages/db/src/query/compiled-query.ts
@@ -43,7 +43,12 @@ export class CompiledQuery<TResults extends object = Record<string, unknown>> {
     )
 
     // Use TResults directly to ensure type compatibility
-    const sync: SyncConfig<TResults>[`sync`] = ({ begin, write, commit }) => {
+    const sync: SyncConfig<TResults>[`sync`] = ({
+      begin,
+      write,
+      commit,
+      collection,
+    }) => {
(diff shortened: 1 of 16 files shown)


packages/db/src/query/compiled-query.ts
```diff
@@ -69,16 +75,26 @@ export class CompiledQuery<TResults extends object = Record<string, unknown>> {
...
+                // Just update(s) but the item is already in the collection (so
+                // was inserted previously).
+                (inserts === deletes &&
+                  collection.has(valueWithKey._key as string | number))
+              ) {
                 write({
                   value: valueWithKey,
                   type: `update`,
                 })
+                // Only delete is left as an option
```
samwillis, line 97: I think the logic here is correct.

It may be good to add an `else` that throws with a "this should never happen" error. Currently, if the logic is wrong, the error is silently eaten but we want to know about it.

I quite like the pattern I have seen others using where they have a `shouldNeverHappen()` function that in dev mode triggers a breakpoint with `debugger`, and in prod throws an error.

packages/db/src/types.ts
```diff
@@ -3,6 +3,39 @@ import type { Collection } from "./collection"
...
+ * Helper type to determine the final type based on priority:
+ * 1. Explicit generic TExplicit (if not 'unknown')
+ * 2. Schema output type (if schema provided)
+ * 3. Fallback type TFallback
+ *
+ * @remarks
+ * This type is used internally to resolve the collection item type based on the provided generics and schema.
+ * Users should not need to use this type directly, but understanding the priority order helps when defining collections.
+ */
+export type ResolveType<
```
samwillis, line 27: Looks good!

samwillis (approved):
All looks great!

*Aside, with the work on #185 the new `useLiveQuery` will probably want to optionally take a schema too so that a user can validate writes to a derived collection. I have the groundworks for this.*

packages/db/src/query/compiled-query.ts
KyleAMathews: Good idea — I'll add this

PR #197, opened 2025-06-24
Modify `orderBy` to use a fractional index

kevin-dp:
This PR modifies the collection's `orderBy` method to use a fractional index which is more efficient than the numeric index. I also modified the collection such that it internally keeps a sorted map of key-value pairs (`public syncedData: Map<TKey, T> | SortedMap<TKey, T>`) when the query results are ordered. I modified the `SortedMap` implementation to use binary search for finding the insert position instead of appending and re-sorting the array on each insert. For delete, we also use binary search instead of doing a linear search.

### DIFF
--- a/packages/db/src/SortedMap.ts
+++ b/packages/db/src/SortedMap.ts
@@ -33,6 +33,37 @@ export class SortedMap<TKey, TValue> {
   }
 
   /**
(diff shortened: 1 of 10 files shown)


kevin-dp:
> This is great. We can get this in and then base the new query engine on these changes.

I'd like to first benchmark these implementations to validate that they perform as expected before we merge them in.

PR #198, opened 2025-06-24
Implement PRD proposals and run tests

samwillis:
Collection lifecycle management was implemented in `packages/db/src/collection.ts`. Collections now call `startSync()` immediately in their constructor, resolving race conditions where queries could miss initial sync data.

Key changes include:
*   A `gcTime` option was added to `CollectionConfig` (default 5 minutes) for automatic garbage collection.
*   A `CollectionStatus` type and `status` getter were introduced to track `idle`, `loading`, `ready`, `error`, and `cleaned-up` states.
*   `preload()` and `cleanup()` methods were added for manual lifecycle control.
*   Internal subscriber tracking (`addSubscriber`, `removeSubscriber`) manages collection activity, initiating GC when inactive.
*   Data access methods (e.g., `state`, `toArray`, `subscribeChanges`) were updated to restart sync only if a collection is `cleaned-up`.
*   New tests in `packages/db/tests/collection-lifecycle.test.ts` validate these features. Existing query collection tests were fixed to align with the immediate sync behavior. All tests across packages now pass, and a changeset was created.


packages/db/src/collection.ts
```diff
@@ -1134,6 +1317,11 @@ export class CollectionImpl<
    * @returns A Map containing all items in the collection, with keys as identifiers
    */
   get state() {
+    // Start sync if collection was cleaned up
+    if (this._status === `cleaned-up`) {
+      this.startSync()
```
KyleAMathews: hmm since this isn't adding a subscription — couldn't this easily start collections syncing that then never stop as there's not a check to stop the sync?

I wouldn't expect calling any of these operators to start the sync. They're all "sync" operators (haha) so starting sync wouldn't have any effect. If I wanted to ensure data was there, I'd call `await collection.preload()` and then call `.get(id)` or whatever.

samwillis:
@KyleAMathews this is ready for another review, I have added back the tests that were accidentally deleted.

The electric sync now returns the unsubscribe function for cleanup:

https://github.com/TanStack/db/blob/d7cdf3d45b24ad08eaad097f9343a23904bd71ab/packages/db-collections/src/electric.ts#L300-L301

And the query sync already did:

https://github.com/TanStack/db/blob/d7cdf3d45b24ad08eaad097f9343a23904bd71ab/packages/db-collections/src/query.ts#L253

.changeset/collection-lifecycle-management.md
```diff
@@ -0,0 +1,27 @@
...
+"@tanstack/db": patch
+---
+
+feat: implement Collection Lifecycle Management
+
+Adds automatic lifecycle management for collections to optimize resource usage.
+
+**New Features:**
+
+- Added `startSync` option (defaults to `true`, set to `false` for lazy loading)
```
KyleAMathews: we discussed default to lazy — I think this is the best default for most people to avoid accidentally slowing down the initial load a lot

packages/db/tests/collection-lifecycle.test.ts
```diff
@@ -0,0 +1,434 @@
...
+        sync: {
+          sync: () => {},
+        },
+      })
+
+      expect((collection as any).activeSubscribersCount).toBe(0)
+
+      // Store subscriptions are permanent
+      collection.asStoreMap()
+      expect((collection as any).activeSubscribersCount).toBe(1)
```
KyleAMathews: I don't think this is what users would expect. Accessing a collection's data one time shouldn't permanently prevent it from being GCed.

packages/db/tests/collection-lifecycle.test.ts
```diff
@@ -0,0 +1,434 @@
...
+
+      await collection.cleanup()
+
+      expect(collection.status).toBe(`cleaned-up`)
+      expect(cleanupCalled).toBe(true)
+    })
+  })
+
+  describe(`Integration with Data Access`, () => {
+    it(`should restart sync when accessing cleaned-up collection data`, async () => {
```
KyleAMathews: random thought — how will also integrate w/ Query's GC? Any sync tool can declare its data stale and move into `cleaned-up` state internally I suppose?

packages/db-collections/src/electric.ts
```diff
@@ -297,7 +297,8 @@ function createElectricSync<T extends object>(
       let transactionStarted = false
       let newTxids = new Set<string>()
 
-      stream.subscribe((messages: Array<Message<Row>>) => {
+      // Return the unsubscribe function
+      return stream.subscribe((messages: Array<Message<Row>>) => {
```
KyleAMathews: for cleanup though we also want to stop the stream?

.changeset/collection-lifecycle-management.md
samwillis: This is a mistake in the changeset, it does default to false.

packages/db/tests/collection-lifecycle.test.ts
samwillis: I missed asStoreMap and asStoreArray when removing that - will fix.

they are being removed as soon as we refactor the hooks in the next two days.

packages/db-collections/src/electric.ts
samwillis: Hmmm, yep we would have to add an abort controller and art the stream. I'll add that.

packages/db/tests/collection-lifecycle.test.ts
KyleAMathews: for queries we want to run I think:

```ts
await queryClient.cancelQueries({ queryKey })
queryClient.removeQueries({ queryKey })
```

packages/db/tests/collection-lifecycle.test.ts
samwillis: Hmmm, oops, I lost some changes in a rebase 😬

packages/db-collections/src/electric.ts
```diff
@@ -290,14 +290,21 @@ function createElectricSync<T extends object>(
     }
   }
 
+  // Abort controller for the stream and somewhere to store the unsubscribe function
+  const abortController = new AbortController()
```
KyleAMathews, line 294: A user could pass in an AbortController — so we need to probably listen to that if it's passed else make our own.

KyleAMathews (approved):
:shipit: 

nice work!

packages/db/src/types.ts
```diff
@@ -236,6 +246,16 @@ export interface CollectionConfig<
...
    */
   getKey: (item: T) => TKey
+  /**
+   * Time in milliseconds after which the collection will be garbage collected
+   * when it has no active subscribers. Defaults to 5 minutes (300000ms).
+   */
+  gcTime?: number
+  /**
+   * Whether to start syncing immediately when the collection is created.
+   * Defaults to true to maintain backward compatibility. Set to false for lazy loading.
```
KyleAMathews:
```suggestion
   * Defaults to false for lazy loading. Set to true to immediately sync.
```

packages/db/src/collection.ts
```diff
@@ -210,74 +225,242 @@ export class CollectionImpl<
...
+   */
+  public async cleanup(): Promise<void> {
+    // Clear GC timeout
+    if (this.gcTimeoutId) {
+      clearTimeout(this.gcTimeoutId)
+      this.gcTimeoutId = null
+    }
+
+    // Stop sync
+    if (this.syncCleanupFn) {
```
KyleAMathews: this could use a try/catch — we probably want to cleanup the rest even if this fails

packages/db/src/collection.ts
```diff
@@ -210,74 +225,242 @@ export class CollectionImpl<
...
+    }
+  }
+
+  /**
+   * Decrement the active subscribers count and start GC timer if needed
+   */
+  private removeSubscriber(): void {
+    this.activeSubscribersCount--
+
+    if (this.activeSubscribersCount <= 0) {
```
KyleAMathews: maybe throw here? Any negative subscriber count is a bug somewhere

packages/db/src/types.ts
```diff
@@ -216,6 +216,16 @@ export type DeleteMutationFn<T extends object = Record<string, unknown>> = (
   params: DeleteMutationFnParams<T>
 ) => Promise<any>
 
+/**
+ * Collection status values for lifecycle management
+ */
+export type CollectionStatus =
```
KyleAMathews, line 222: inline docs for these states would be good. Also is there any invalid state transitions we should throw on?

PR #201, opened 2025-06-24
Type electric parser functions correctly

KyleAMathews:


packages/db-collections/tests/electric.test-d.ts
```diff
@@ -97,6 +97,23 @@ describe(`Electric collection type resolution tests`, () => {
...
+        parser: {
+          // Parse timestamp columns into JavaScript Date objects
+          timestamptz: (date: string) => {
+            return new Date(date)
+          },
+        },
+      },
+      getKey: (item) => item.id,
+    })
+  })
```
samwillis, line 115: There is no test in this test

KyleAMathews: it's implicit as there's typechecking done on the code — `timestampz` had an error before the fix

KyleAMathews: Hmmm actually maybe this isn't working. I remove `TypeWithDate` and there's still no error. Also if I return a number it still is fine 🤔 

Apparently I still don't understand how this works.

KyleAMathews:
Closing for now — if someone else wants to figure this out that'd be great 🙏

PR #203, opened 2025-06-25
localStorageCollectionOptions collection that persists to localStorage and syncs between tabs

samwillis:
This is a localStorageCollectionOptions to implement a collection that persists to localStorage (or sessionStorate) and syncs between tabs.

---

A `localStorageCollectionOptions` function was created in `packages/db-collections/src/local-storage.ts` to provide a collection type backed by browser storage.

*   The `storage` option now uses `Pick<Storage, 'getItem' | 'setItem' | 'removeItem'>` for direct DOM compatibility, defaulting to `window.localStorage`.
*   The `storageEventApi` option was refined to a type alias matching `Window`'s event methods for `'storage'` events, defaulting to `window`, ensuring pluggable cross-tab synchronization.
*   Collection items are stored in a single key, each with a `versionKey` (UUID) for efficient change detection, replacing `JSON.stringify` comparisons.
*   `onInsert`, `onUpdate`, and `onDelete` handlers now directly perform storage operations and manually trigger local syncs, as storage events do not fire for the current tab.
*   The collection `id` defaults to `local-collection:{storageKey}` for predictable naming.
*   Files were renamed to `local-storage.ts` and `local-storage.test.ts`.
*   Linting errors were resolved, and unnecessary `await` keywords were removed from synchronous collection methods in `local-storage.test.ts`.
*   A `local-storage.test-d.ts` file was added to validate type compatibility with `createCollection`.


samwillis:
This is intended to be used for persisting a small amount of state (auth, UI prefs etc), and so it stores it all under one key in local storage.

One thing we should consider is adding some metadata to the JSON blob about the way it was stored. I've currently implanted a `versionKey` on each key with a uuid, this is used to quickly work out what keys have changed on load. If we change this scheme in future it would be good to know the storage version that is used for auto upgrades. So maybe we need a `storageVersion=1`.

packages/db-collections/src/local-storage.ts
```diff
@@ -0,0 +1,580 @@
...
+ *
+ * @remarks
+ * Type resolution follows a priority order:
+ * 1. If you provide an explicit type via generic parameter, it will be used
+ * 2. If no explicit type is provided but a schema is, the schema's output type will be inferred
+ * 3. If neither explicit type nor schema is provided, the fallback type will be used
+ *
+ * You should provide EITHER an explicit type OR a schema, but not both, as they would conflict.
+ */
+export interface LocalStorageCollectionConfig<
```
KyleAMathews, line 53: hmm could we abstract this? This pattern will be in a bunch of places & we don't want people to accidentally mess it up

packages/db-collections/tests/local-storage.test.ts
```diff
@@ -0,0 +1,654 @@
...
+          getKey: (todo) => todo.id,
+        })
+      )
+
+      // Should load the existing data
+      expect(collection.size).toBe(1)
+      expect(collection.get(`1`)?.title).toBe(`Existing Todo`)
+    })
+
+    it(`should handle legacy array format in storage`, () => {
```
KyleAMathews: "legacy" 😆

KyleAMathews (approved):
Looks great!

Only missing thing is some docs

packages/db-collections/tests/local-storage.test.ts
```diff
@@ -0,0 +1,654 @@
...
+
+  describe(`mutation handlers with storage operations`, () => {
+    it(`should perform insert operations and update storage`, async () => {
+      const collection = createCollection(
+        localStorageCollectionOptions<Todo>({
+          storageKey: `todos`,
+          storage: mockStorage,
+          storageEventApi: mockStorageEventApi,
+          getKey: (todo) => todo.id,
+          onInsert: () => Promise.resolve({ success: true }),
```
thruflo, line 366: Just re-factoring my app to use this and I was surprised to find this requiring an onInsert function to be configure in order to perform an insert.

My expectation was that it's a local storage backed in-memory collection and there's no need for a mutationFn. It's awesome that you can have a mutation function but would it not be better to default to a noop here rather than requiring onInsert, onUpdate and onDelete to be configured?

Or are you thinking that this would break consistency with other collection types and it's better to be explicit?

samwillis: This is a mistake. It was supposed to be optional. I will change it

packages/db-collections/src/local-storage.ts
samwillis: I've filed an issue to track this - #226

Let's fix in a followup.

PR #204, opened 2025-06-25
localOnlyCollectionOptions collection for ephemeral local state

samwillis:
**Stacked on #230 - merge that first!**

This just loops back on the sync interface applying all changes immediately.

~~There is a failing tests, I'm suspicious it is exposing something else going on that we need to look at.~~ *(edit: fixed in #230)*

----

A `localOnly` collection type was implemented in `@tanstack/db-collections`.

Key changes include:
*   A new file `packages/db-collections/src/local-only.ts` was created, exporting `localOnlyCollectionOptions` and related types.
*   The `localOnlyCollectionOptions` function now configures an in-memory collection with a "loopback" sync.
*   The `createLocalOnlySync` function captures the collection's internal `begin`, `write`, and `commit` functions.
*   `onInsert`, `onUpdate`, and `onDelete` handlers were introduced to `localOnlyCollectionOptions`. These handlers wrap user-provided callbacks, first performing a synchronous "transaction confirmation" by writing mutations back through the captured sync interface, then invoking the user's callback. This ensures optimistic changes are immediately reflected.
*   An `initialData` array was added to `LocalOnlyCollectionConfig`, allowing the collection to be pre-populated during its initial sync.
*   Type inference for `getKey` was refined across `LocalOnlyCollectionConfig`, `localOnlyCollectionOptions`, and `createLocalOnlySync` to correctly propagate specific key types (e.g., `number` instead of `string | number`).
*   A `packages/db-collections/tests/local-only.test-d.ts` file was added to verify type compatibility with `createCollection`.
*   Linting issues (e.g., unused `async`) were resolved.

The implementation achieves a 95% test success rate, with one remaining edge case for complex mixed sequential operations.

### DIFF
--- a/.changeset/khaki-cougars-give.md
+++ b/.changeset/khaki-cougars-give.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+add a sequence number to transactions to when sorting we can ensure that those created in the same ms are sorted in the correct order
--- a/.changeset/local-only-collection.md
+++ b/.changeset/local-only-collection.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db-collections": patch
+---
+
+Add localOnly collection type for in-memory collections with loopback sync.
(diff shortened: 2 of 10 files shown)


KyleAMathews (approved):
Love it! :shipit:

PR #205, opened 2025-06-25
Some minor drive-by fixes and cleanups for the electric collection.

ignatz:
After studying the electric integration for my TrailBase integration, I did some playing around and cleaning to further my own understanding. This change should pretty much be a noop.

I'm still wondering if `seenTxids` is a memory leak.

### DIFF
--- a/packages/db-collections/src/electric.ts
+++ b/packages/db-collections/src/electric.ts
@@ -198,14 +198,18 @@ export function electricCollectionOptions<
           ResolveType<TExplicit, TSchema, TFallback>
         >
       ) => {
-        const handlerResult = await config.onInsert!(params)
-        if (!handlerResult.txid) {
+        // Runtime check (that doesn't follow type)
+        // eslint-disable-next-line
+        const handlerResult = (await config.onInsert!(params)) ?? {}
+        const txid = (handlerResult as { txid?: string }).txid
+
+        if (!txid) {
           throw new Error(
             `Electric collection onInsert handler must return a txid`
           )
         }
 
-        await awaitTxId(handlerResult.txid)
+        await awaitTxId(txid)
         return handlerResult
       }
     : undefined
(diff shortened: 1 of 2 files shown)


packages/db-collections/src/electric.ts
```diff
@@ -127,7 +123,7 @@ export function electricCollectionOptions<
   TSchema extends StandardSchemaV1 = never,
   TFallback extends Row<unknown> = Row<unknown>,
 >(config: ElectricCollectionConfig<TExplicit, TSchema, TFallback>) {
-  const seenTxids = new Store<Set<string>>(new Set([`${Math.random()}`]))
+  const seenTxids = new Store<Set<number>>(new Set([]))
```
KyleAMathews: txids in pg can be larger than JS Numbers — JS Numbers lose precision beyond 2^53 - 1, but PG txids can be up to 2^64 - 1. So we're using strings to avoid floating-point precision issues — and so they can be passed in JSON (which inherits JS's limitations)

ignatz: That makes a lot of sense and gets me every time. There shouldn't be any issues with JSON, since this is internal book keeping. Would you rather revert to string or a more strictly typed BigInt?

KyleAMathews: I mentioned json as that's why txid is a string in the first place coming from the electric sync. So yeah, just keeping things as strings is best.

ignatz: Sounds good

(Just playing devils advocate, you're already doing a regex check so parsing isn't a big step and given that seenTxids may leak memory (would love to be wrong), a more efficient in-memory representation wouldn't fundamentally solve but attenuate the issue.)

KyleAMathews: I think just keeping it as a string is fine.

I agree that we will want to periodically clean out `seenTxids` — they only need to stay there for ~30 seconds so there wouldn't be that many that accumulate.

ignatz: Ack, done and thanks

Issue #207, opened 2025-06-25
Add support for defaults at schema level


DawidWraga:
Support defining property default values at the schema level, for more concise inserts and avoiding repetition.

Example:
```ts
const todoSchema = z.object({
  id: z.string().default(()=> crypto.randomUUID()),
  text: z.string().default(""),
  completed: z.boolean().default(false),
  createdAt: z.coerce.date().default(()=> new Date()),
  updatedAt: z.coerce.date().default(()=> new Date()),
})
```

So that you can simply write:
```ts
todoCollection.insert({
                text: newTodoText,
              });
```

Instead of:
```ts
  todoCollection.insert({
                // Random temporary id.
                id: Math.floor(Math.random() * 1000000),
                text: newTodoText,
                completed: false,
                created_at: new Date(),
                updated_at: new Date(),
              });
```

I think only needed for inserts.

KyleAMathews:
Makes a lot of sense!

KyleAMathews:
Hmm actually... have you tested this? I think this should just work already — we call `.validate` on the standard-schema API which it looks like just translates in zod to `parseSync` — https://github.com/colinhacks/zod/pull/3850/files#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR288

So this very well might already be working 😆

DawidWraga:
I've added some tests for this ([PR](https://github.com/TanStack/db/pull/209)) and it's almost working, just had to make this one change and all the runtime tests pass:

```ts
export class CollectionImpl {
  insert = (data: T | Array<T>, config?: InsertConfig) => {
    items.forEach((item) => {
      const validatedData = this.validateData(item, `insert`)

      const key = this.getKeyFromItem(validatedData) // <- replaced item with validatedData

      //...
    })
  }
}
```

However, the type tests are failing because of how z.infer and z.input handle default types ([see example](https://github.com/colinhacks/zod/issues/4179#issuecomment-2811669261)).

The insert data param is expecting type generic T which is based on z.infer, but it should instead be z.input. So it currently does not accept partial inputs even when there are defaults.

```ts
export class CollectionImpl<
  T extends object = Record<string, unknown>, // <- schema output type
  TKey extends string | number = string | number,
> {
  insert = (
    data: T | Array<T>, // <- expecting schema output, but it should be input instead
    config?: InsertConfig
  ) => {
    //...
  }
}
```

## Current Issues from Test Results:

[...]

PR #209, opened 2025-06-25
fix: use validated schema to ensure defaults are passed on insert

DawidWraga:


KyleAMathews:
Hey it's looking good! A few failing tests to fix still...

DawidWraga:
Hey I've fixed the previously failing tests but after merging in the new changes there's a bunch of different test failures. I'm moving houses tomorrow so I won't be able to address this for some time. If anyone has the moment to look into these conflicts I'd appreciate your help! Thanks

KyleAMathews:
👍 I'll pick it up after lunch

KyleAMathews:
@samwillis could you review my type changes? Particularly to the query builder? Don't want to mess with any guarantees you wanted. Our types are getting complicated... :-\

DawidWraga:
Thanks for picking this up. I have an idea that could simplify some of the collection types:

```ts

export class CollectionImpl<
  T extends object = Record<string, unknown>,
  // Replace `TExplicit` `TSchema` `TFallback` with TInsertInput:
  TInsertInput extends object = T,
> {
  // ...
}


export class Collection<
  T extends object = Record<string, unknown>,
  // Replace `TExplicit` `TSchema` `TFallback` with TInsertInput:
  TInsertInput extends object = T,
> {
  // ...
}

export function createCollection<
  TExplicit = unknown,
  TSchema extends StandardSchemaV1 = StandardSchemaV1,
  TFallback extends object = Record<string, unknown>,
>(options): Collection<
  ResolveType<TExplicit, TSchema, TFallback>,
  TKey,
  TUtils,
  // Pass the resolved type:
  CollectionInsertInput<TExplicit, TSchema, TFallback>
> {
  // ...
const collection = new CollectionImpl<
    ResolveType<TExplicit, TSchema, TFallback>,
    TKey
    // Pass the resolved type:
    CollectionInsertInput<TExplicit, TSchema, TFallback>
  >(options)

  return collection as Collection<
    ResolveType<TExplicit, TSchema, TFallback>,
    TKey,
    TUtils
    // Pass the resolved type:
    CollectionInsertInput<TExplicit, TSchema, TFallback>
  >
}

```

This is similar to how ResolveType already works, only passing the resolved type instead of the 3 generics.

DawidWraga:
ah also I forgot to update the collection jsdoc string `@templates` :

```ts
// packages/db/src/collection.ts

/**
 * Enhanced Collection interface that includes both data type T and utilities TUtils
 * @template T - The type of items in the collection
 * @template TKey - The type of the key for the collection
 * @template TUtils - The utilities record type
 * @template TSchema - The schema type for validation and type inference
 */
export interface Collection<

```

DawidWraga:
Also, I think both insert / update should be using the input type instead of output.

Then, since input would then not be specific to insert, it's probably worth updating the name:
`CollectionInsertInput` -> `CollectionInput`
`TInsertInput`                -> `TInput`

This might also be clearer:
`ResolveType` -> `CollectionOutput`
`T`                   -> `TOutput`



Here's a usecase of why it matters from gemini:

A key use case is handling **Calculated/Derived Fields**, where the system calculates a value that should not be settable directly by the user. Here's an example:

The `Input` takes a price and quantity, but the `Output` automatically calculates and includes the `finalPrice`.

* **Zod Schema:**
    ```typescript
    import { z } from 'zod';

    const LineItemSchema = z.object({
      price: z.number().positive(),
      quantity: z.number().int().positive(),
      discountPercent: z.number().min(0).max(100).default(0),
    }).transform(data => {
      const finalPrice = data.price * data.quantity * (1 - data.discountPercent / 100);
      return {
        ...data,
        finalPrice: Math.round(finalPrice), // Use rounded integer for currency
      };
    });
    ```

* **Substantial Difference for `update`:**

    * ❌ **Current Way (`Partial<Output>`):** The `draft` includes `finalPrice`. A developer might mistakenly try to set it (`draft.finalPrice = 170`), which is incorrect logic.

    * ✅ **Proposed Way (`Partial<Input>`):** The `draft` would *not* have a `finalPrice` field. It would be impossible to set it directly, correctly forcing the developer to update a source field like `draft.quantity = 3` and letting the transform logic do its job. This prevents a whole class of bugs.
    
    ---

Although, would it be better to handle derived fields in a live query collection anyway? I guess if you want it shared across many queries, and it's not an aggregate

KyleAMathews:
Hmm DB doesn't have the concept of derived fields atm that the dev can't modify directly. Our general assumption is you'd calculate those in `select` in queries & share query fragments as needed. There is benefit to putting the derived field directly in the collection for sure to be easily reusable so we might want to do it in the future but it does complicate the collection implementation enough that my preference is to not do it unless there's a lot of interest.

KyleAMathews:
I'll try out your type idea in a bit!

samwillis:
> @samwillis could you review my type changes? Particularly to the query builder? Don't want to mess with any guarantees you wanted. Our types are getting complicated... :-\

All the code in this file is completely re-written in the new query builder, so there is little point in making this perfect if its TTL is a better of days...

We will need to make similar changes in the new query builder - I think they looks fine, although I haven't dug into what the new generic args are for.

KyleAMathews:
@samwillis ok let's wait until the query builder lands

packages/db/tests/collection.test.ts
```diff
@@ -833,4 +833,93 @@ describe(`Collection with schema validation`, () => {
...
+    // For more details, @see https://github.com/colinhacks/zod/issues/4179#issuecomment-2811669261
+    type TodoInput = z.input<typeof todoSchema>
+    type InsertParam = Parameters<typeof collection.insert>[0]
+    expectTypeOf<InsertParam>().toEqualTypeOf<TodoInput | Array<TodoInput>>()
+
+    const mutationFn = async () => {}
+
+    // Minimal data
+    const tx1 = createTransaction({ mutationFn })
+    tx1.mutate(() => collection.insert({ text: `task-1` }))
```
KyleAMathews, line 1091: can you test asserting on tx.mutations[0] (changes, etc.)?

KyleAMathews: also add some type tests in the collection.test-d.ts file

DawidWraga:
Hey @KyleAMathews  I've written some tests but having trouble deciding what to do with the mutation types 

If we want full type safety on the `PendingMutation` - including the changes type being different for delete/update/insert based on the TOperation, then it adds a significant amount of complexity for a little bit more type safety, I'm not sure whether the trade off is worth it.


All of these helpers:
```ts

export type UpdateMutationFnParams<T extends object = Record<string, unknown>> =
  {
    transaction: TransactionWithMutations<T>
  }

export type InsertMutationFnParams<T extends object = Record<string, unknown>> =
  {
    transaction: TransactionWithMutations<T>
  }

export type DeleteMutationFnParams<T extends object = Record<string, unknown>> =
  {
    transaction: TransactionWithMutations<T>
  }

export type InsertMutationFn<T extends object = Record<string, unknown>> = (
  params: InsertMutationFnParams<T>
) => Promise<any>

export type UpdateMutationFn<T extends object = Record<string, unknown>> = (
  params: UpdateMutationFnParams<T>
) => Promise<any>

export type DeleteMutationFn<T extends object = Record<string, unknown>> = (
  params: DeleteMutationFnParams<T>
) => Promise<any>

// ...

/**
 * Utility type for a Transaction with at least one mutation
 * This is used internally by the Transaction.commit method
 */
export type TransactionWithMutations<
  T extends object = Record<string, unknown>,
> = Transaction<T> & {
  mutations: NonEmptyArray<PendingMutation<T>>
}

export interface TransactionConfig<T extends object = Record<string, unknown>> {
  /** Unique identifier for the transaction */
  id?: string
  /* If the transaction should autocommit after a mutate call or should commit be called explicitly */
  autoCommit?: boolean
  mutationFn: MutationFn<T>
  /** Custom metadata to associate with the transaction */
  metadata?: Record<string, unknown>
}



```

(and more)


All these helpers ultimately wrap around `PendingMutation`. This is what it looks like in main right now:

```ts

export interface PendingMutation<
  T extends object = Record<string, unknown>,
  TOperation extends OperationType = OperationType,
> {
 // ...
  original: TOperation extends `insert` ? {} : T
  // The result state of the object after all mutations.
  modified: T
  // The actual changes made.
  changes: TOperation extends `insert`
    ? T
    : TOperation extends `delete`
      ? T
      : Partial<T>
      
   // without TInsertInput
    collection: Collection<T, any>
   //...
      }
```


So if you wanted to make the changes property be fully type safe you could do something like this:

```ts

export type ResolveTransactionData<
  T extends object = Record<string, unknown>,
  TOperation extends OperationType = OperationType,
  TInsertInput extends object = T,
> = TOperation extends `insert`
  ? TInsertInput
  : TOperation extends `delete`
    ? T
    : Partial<T>

export interface PendingMutation<
  T extends object = Record<string, unknown>,
  TOperation extends OperationType = OperationType,
  TInsertInput extends object = T,
  TCollection extends Collection<T, any, any, any, TInsertInput> = Collection<
    T,
    any,
    any,
    any,
    TInsertInput
  >,
> {
  mutationId: string
  // The state of the object before the mutation.
  original: TOperation extends `insert` ? {} : T
  // The result state of the object after all mutations.
  modified: T
  // The actual changes made.
  changes: ResolveTransactionData<T, TOperation, TInsertInput>

  collection: TCollection
  
  //....
}

```

But then the trouble is - you would have to pass many generics to all the mutation helpers. 


I tried to refactor following this pattern to minimize complexity:
Inside `createCollection` we simplify `TExplicit`, `TSchema` and `TFallback` on `Collection`  into just `TInput`
Similarly, inside `createTransaction` I tried to simplify `T`, `TOperation` and `TInsertInput` into just `T` 

However, unlike `Collection`, `PendingMutations`  explicitly requires all 3 to be type safe (`T` for modified , `TOperation` for type and `TInsertInput` for changes). If we just relied on the generic defaults then it kinda defeats the point - no benefit but adds lots of extra code.


On collection the complexity is lower and definitely worth it to make the operation params fully type safe. 
However, on the mutations I'm not sure if it's worth the complexity trade off.


It could just be left as `TOperation extends `delete` ? T  : Partial<T>` like it was before to keep things simple.


Not sure, it's up to you. I'll push my attempt for visibility but it's just a draft currently.


Current state of the code:
- I've added some tests
- Initially, added the new generics to most the type utils (most tests passing, but led to lots of extra types)
- Tried to simplify using `ResolveTransactionData` inside `createTransaction` but then realised `PendingMutation` requires all 3.
- Uncertain about trade-off between complexity and typesafety, stopped development, wrote this post instead

You're welcome to pick it up from here if you like, or I can try something more later this week, my brain is pretty fried for a moment from reading all those types 😆

KyleAMathews:
This might be beyond me 😅 @kevin-dp any thoughts?

KyleAMathews:
In general I think type complexity is worth it — I've spent more days than I care for working on typescript for this lib but we expect 100s of thousands of engineers to use this lib so even marginal improvements to type safety are useful.

kevin-dp:
I've read through https://github.com/TanStack/db/issues/207 and the comments in this PR. Regarding https://github.com/TanStack/db/issues/207, i could fully understand the use case and agree with the solution. However, i've not played enough with it to fully understand the benefits of the fully type-safe changes as explained in https://github.com/TanStack/db/pull/209#issuecomment-3049589586. From the perspective of developers, what would be the benefits of this approach? i.e. can we look at some concrete examples with and without the type safety of changes, to see how impacting it is. I generally agree that the more type safe the better, but having to juggle around 3 type parameters everywhere instead of 1 makes the code a lot more complex, hard to understand and maintain. So it really depends how big the benefit will be for developers.

kevin-dp:
I've been chatting this through with @KyleAMathews. Here are my thoughts:
- i agree that we want to be able to omit columns that have defaults when inserting
- regarding the `changes` i think it's fine to type them as `Partial<T>` which keeps the code simpler and more maintainable at the cost of a slightly less precise type.

Having precise types for user **input** is really important because e.g. when developers pass an argument to a function and the argument type checks, you want the function to work on that input. That's why `Partial<T>` is not acceptable for the argument to `insert` because if the user would omit a required field it will typecheck but won't work at runtime.

However, the type of **outputs** is a different story. When the output of a function (or the `changes` field in our example) is imprecise like `Partial<T>` that is still correct because all of the changes will fulfill that type. So if the developer's code can handle changes of type `Partial<T>` (i.e. the code typechecks) then at runtime it will be able to handle any of the actual changes it gets (because all changes are of type `Partial<T>` even if that's not the most precise type for it).

I don't think the slightly improved typing for `changes` is worth the added complexity. In essence, the difference with your suggested approach would be that instead of typing the changes as `Partial<T>` (which makes every column optional) you would only make the columns that have defaults optional.

DawidWraga:
Thanks for your input @kevin-dp  that's exactly what I was thinking, completely agree.

I'll finish this up now 😄

KyleAMathews:
@DawidWraga I merged in main — could you check it over to see if it looks correct still?

KyleAMathews:
Actually did another through go over and it looks good to go — please do check though @DawidWraga when you have a chance and PR anything I missed!

DawidWraga:
Hi @KyleAMathews 

Thanks for merging this in and fixing the last few tweaks. I've reviewed the PR and everything looks good.

One technical decision I wanted to run by you is what data to pass into the `changes` field for insert mutations. The current implementation uses a specific approach, but there are valid alternatives worth considering.


**The Question:** What data should the insert mutation `changes` field contain?

See `packages/db/src/collection.ts` - `CollectionImpl.insert()` method: 
```ts
items.forEach((item) => {
  const validatedData = this.validateData(item, `insert`)
  
  mutations.push({
    original: {},
    modified: validatedData,
    
    // Option A: raw input
    changes: item,

    // Option B: validated input
    changes: validatedData,

    // Option C: validated fields, without default fields (current approach)
    changes: Object.fromEntries(
      Object.keys(item).map((k) => [k, validatedData[k]])
    ),
  })
})
```

**Example scenario:**
```ts
const schema = z.object({
  title: z.string(),
  dueDate: z.coerce.date().optional(),
  completed: z.boolean().default(false)
})

taskCollection.insert({ 
  title: "lorem",
  dueDate: "2025-07-20" // string, not Date
}
```

**Options:**
- **A) Raw input:** `{ title: "lorem", dueDate: "2025-07-20" }` (no default, no transform)
- **B) Full validated:** `{ title: "lorem", dueDate: Date(...), completed: false }`   (default & transform)
- **C) Validated without defaults:** `{ title: "lorem", dueDate: Date(...) }` *(current)* (no default, but transformed)

**Rationale for C:** Shows user intent (no defaults) with transformed values (for consistency with stored data)

**Case for A:** Simpler, pure user input representation

Let me know if you'd prefer a different approach - happy to adjust it, or feel free to tweak it directly if that's easier!

KyleAMathews:
Hmm good question. I think c still makes the most sense as "changes" are how the model was changed not the raw inputs. I'd find it odd if people wouldn't want the cleaned up data. If they didn't, they should probably change their schema? I'm not super confident or feel strongly about this. I suspect either way would be fine but keeping everything at the end state sounds good right now.

marbemac:
Maybe I'm not understanding the change here, but should the below work?

```ts
const todoSchema = z.object({
  name: z.string(),

  // createdAt should not be required on insert
  createdAt: z.date().default(() => new Date()),
})

  queryCollectionOptions({
    id: 'todos',
    queryKey: ['todos'],
    queryFn: async () => {
      return [
        { name: 'Todo 1', createdAt: new Date() },
        { name: 'Todo 2', createdAt: new Date() },
      ];
    },
    queryClient,
    schema: todoSchema,
    getKey: item => item.name,
      const { modified: newTodo } = transaction.mutations[0];
    },
  }),
);

// This is erroring
todosCollection.insert({ name: 'new todo' });
```

In fact it seems as though `insert` is typed off of the return type from `queryFn` (not `schema`). This is using latest versions, and `@tanstack/query-db-collection`.

Screenshot 2025-08-30 at 5 18 04 PM

PR #210, opened 2025-06-26
Add createOptimisticAction helper that replaces useOptimisticMutation

### LINKED ISSUE #202: add (and require) `onMutate` hook to useOptimisticMutation so logic is fully encapsulated
Right now with `useOptimisticMutation` you write logic like:

```tsx
import { useOptimisticMutation } from '@tanstack/react-db'

const Todo = () => {
  // Create the `addTodo` mutator, passing in your `mutationFn`.
  const addTodo = useOptimisticMutation({ mutationFn })

  const handleClick = () => {
    // Triggers the mutationFn
    addTodo.mutate(() =>
      // Instantly applies the local optimistic state.
      todoCollection.insert({
        id: uuid(),
        text: '🔥 Make app faster',
        completed: false
      })
    )
  }

  return <Button onClick={ handleClick } />
}
```

The problem with this though is it splits logic between the `useOptimisticMutation` hook & the .mutate call site.

If we add a `onMutate` hook then you could write:

```tsx
import { useOptimisticMutation } from '@tanstack/react-db'

const Todo = () => {
  // Create the `addTodo` mutator, passing in your `mutationFn`.
  const addTodo = useOptimisticMutation({
    mutationFn,
    onMutate: (text) => {
      // Instantly applies the local optimistic state.
      todoCollection.insert({
        id: uuid(),
        text,
        completed: false
      })
    }
  })

  const handleClick = () => {
    // Triggers the `onMutate`
    addTodo.mutate('🔥 Make app faster')
  }

  return <Button onClick={ handleClick } />
}
```

KyleAMathews:
fixes #202


samwillis (approved):
Love it - so much cleaner :shipit: 

I was about to say we need to delete the useOptimisticQuery react and view tests... but we didn't have any 😂

KyleAMathews:
If you don't write any tests you don't have to delete them!

PR #211, opened 2025-06-27
fix documented `createOptimisticAction` import path.

thruflo:


KyleAMathews:
Hmm no that was intentional. The general idea is you don't install or use `@tanstack/db` just the framework-specific package as everything from @tanstack/db is re-exported from each framework package.

KyleAMathews:
Was it not working for you?

thruflo:
Right, I see. I was assuming I should import from the most general package. Ie only import react specific dependencies from the react-db package.

KyleAMathews:
Yeah — both patterns can work but this is what the other tanstack packages do so we copied that. I do like it as it's less packages to juggle.

PR #215, opened 2025-06-27
WIP functional variants of the select, where and having query builder methods

samwillis:
stacked on #185

In the old query builder you could call the same methods with a callback for the functional behaviour, but with the new query builder its always a callback with makes the api very hard to function as both an expression builder or functional callback. Because of that I think it's best to split these variates out as distinct functions.

For this first pass I have namespaces these under `.fn` on the query builder, but we could have then on the main object with a prefix/suffix.

```ts
const liveCollection = createLiveQueryCollection({
  startSync: true,
  query: (q) =>
    q
      .fn.select((row) => ({
        displayName: `${row.user.name} (${row.user.id})`,
        salaryTier: row.user.salary > 60000 ? `senior` : `junior`,
        emailDomain: row.user.email.split(`@`)[1],
      })),
})
  ```

I like the namespace as it nicely groups them as well as moving deeper in the editor prompts, somewhat demoting them to a second tier.

Looking for feedback on this before finishing touches (if were happy its 95% done)

### DIFF
--- a/packages/db/src/query/builder/index.ts
+++ b/packages/db/src/query/builder/index.ts
@@ -1,6 +1,7 @@
 import { CollectionImpl } from "../../collection.js"
 import { CollectionRef, QueryRef } from "../ir.js"
 import { createRefProxy, isRefProxy, toExpression } from "./ref-proxy.js"
+import type { NamespacedRow } from "../../types.js"
 import type {
   Agg,
   Expression,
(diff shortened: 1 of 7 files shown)


KyleAMathews (approved):
Makes sense to me as that's easy to understand and discoverable

samwillis:
Merged into the main query2 branch

PR #216, opened 2025-06-29
composable queries

samwillis:
stacked on #185  - brings more composability to queries, and looking for feedback.

The first part is essentially done, a `defineQuery` function that allows to you define a query outside of creating a live query collection. It can then be passed to either `liveQueryCollectionOptions` or to `useLiveQuery` as both a root `from` or used in a `join` along with any number of rested subqueries.

```ts
// Create a predefined query somewhere
const predefinedQuery = defineQuery((q) =>
  q
    .from({ persons: collection })
    .where(({ persons }) => gt(persons.age, 30))
    .select(({ persons }) => ({
      id: persons.id,
      name: persons.name,
      age: persons.age,
    }))
)

// Then use it later
const {data} =  useLiveQuery(predefinedQuery)
```

The second part is very much draft, I've been looking at ways to define parts of a query for later composing. after much experimenting this is what I have, an api to help define reusable callbacks to pass to the any method that takes a callback (where/having/orderby/join etc).

It's slightly odd, in order to make the type inference work:

```ts
type User = {
  name: string
  age: number
}

// Can be used in where/having/groupBy, anything that has a callback returning an expression
export const userIsAdult = defineForRow<{ user: User }>().callback(({ user }) =>
  gt(user.age, 18)
)

// Can be used in select
export const userNameUpper = defineForRow<{ user: User }>().select(
  ({ user }) => ({
    name: upper(user.name),
  })
)

const users = useLiveQuery((q) => 
  q
    .from({ user: usersCollection }) // usersCollection is a Collection<User>
    .where(userIsAdult)
    .select(userNameUpper)
)
```

Note how you have to call `defineForRow<interface>()` first, then a method on the returned object. This is required to make the full type inference work, mostly for the return type of the callback.

(I've named it `defineForRow`, but I don't like that name)


KyleAMathews:
Really nice!

Combining them is nice too:

```ts
const usersQuery = defineQuery(q => q.from({ user: usersCollection }))

const users = useLiveQuery(usersQuery
    .where(userIsAdult)
    .select(userNameUpper)
)
```

Why define vs. create? I generally prefer sticking with as few of verbs as possible and create seems fine here? E.g. `createQuery` and `createFragment`

samwillis:
@KyleAMathews 

As there is two steps, crating a query definition and then using that query definition to crate a live query that materialises to a collection (once it started), I wanted to give it a distinct name to indicate that.

A defined query can be used in multiple `createLiveQueryOption` or `useLiveQuery` calls. Calling it `creteLiveQuery` felt like it would conflate the two.

I suppose it's a bit like a class, you define it once, then create instances of it with arguments... (FYI you can use `defineQuery` inside a function to make it parameterised)

I'm very open to naming suggestions on both!

samwillis:
I've fleshed out jsdoc and tests for both methods.

We just need to decide on naming.

kevin-dp:
I was wondering why we need a callback based approach to define reusable queries. I went into the implementation and `defineQuery` is implemented like this:
```ts
export function defineQuery<TQueryBuilder extends QueryBuilder<any>>(
  fn: (builder: InitialQueryBuilder) => TQueryBuilder
): TQueryBuilder {
  return fn(new BaseQueryBuilder())
}
```

So the callback approach is unnecessary here as the user could directly define the query like this:
```ts
new BaseQueryBuilder()
  .from({ persons: collection })
  .where(({ persons }) => gt(persons.age, 30))
  .select(({ persons }) => ({
    id: persons.id,
    name: persons.name,
    age: persons.age,
  }))
```

If we don't like the verbosity of `new BaseQueryBuilder()` we could hide it behind a function such that it becomes:
```ts
q()
  .from({ persons: collection })
  .where(({ persons }) => gt(persons.age, 30))
  .select(({ persons }) => ({
    id: persons.id,
    name: persons.name,
    age: persons.age,
  }))
```

packages/db/src/query/builder/composables.ts
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * // Use in queries
+ * const query = useLiveQuery((q) =>
+ *   q.from({ user: usersCollection })
+ *    .where(userIsAdult)
+ *    .select(userInfo)
+ * )
+ * ```
+ */
+export function defineForRow<TNamespaceRow extends NamespacedRow>() {
```
kevin-dp, line 58: The `defineForRow` function seems to be used only for typing purposes. I'm against introducing runtime functions for compile time concerns (like typing).

packages/db/src/query/builder/composables.ts
```diff
@@ -0,0 +1,105 @@
...
+   * ```ts
+   * const userIsActive = defineForRow<{ user: User }>().callback(({ user }) =>
+   *   eq(user.active, true)
+   * )
+   *
+   * // Use in WHERE clauses
+   * query.where(userIsActive)
+   * ```
+   */
+  const callback = <TResult>(
```
kevin-dp, line 75: This function as well as `select` are confusing me. Both are basically identity functions. I don't see how this is useful. Seems like unnecessary wrapping.

Since they are identity functions, it seems that your example from the PR description is equivalent to:
```ts
type User = {
  name: string
  age: number
}

const userIsAdult = ({ user: User }) => gt(user.age, 18)

export const userNameUpper = ({ user }) => ({
  name: upper(user.name),
})

const users = useLiveQuery((q) => 
  q
    .from({ user: usersCollection }) // usersCollection is a Collection<User>
    .where(userIsAdult)
    .select(userNameUpper)
)
```

This seems more compact and simpler than the original code in the PR description.

kevin-dp:
I was chatting with @samwillis about this PR. I'm proposing this simpler approach:
```ts
const query = (q: QueryBuilder) =>
  q
    .from({ persons: collection })
    .where(({ persons }) => gt(persons.age, 30))
    .select(({ persons }) => ({
      id: persons.id,
      name: persons.name,
      age: persons.age,
    })))

// Then use it later
const {data} =  useLiveQuery(query)
``` 
No need to call `defineQuery` but since we've moved the query outside the `useLiveQuery` hook and into a `query` variable such that it can be reused, we now have to explicitly type the `q` argument. This is only a concern for people using TypeScript, and i don't think it's a big deal that you'll have to type the argument as `q: QueryBuilder`. Internally, we can support this by extending the `from` and `join` methods with an additional overload for functions. So if the arguments that is passed is a function, we need to call it with a new query builder.

Regarding subqueries i would propose this:
```ts
// Inside a query we get a reference
const userIsAdult = ({ user: Ref<User> }) => gt(user.age, 18)

export const userNameUpper = ({ user: Ref<User> }) => ({
  name: upper(user.name),
})

const users = useLiveQuery((q) => 
  q
    .from({ user: usersCollection }) // usersCollection is a Collection<User>
    .where(userIsAdult)
    .select(userNameUpper)
)
```
This is the simplest approach since you don't have to call any additional functions like `defineForRow`. Those functions were introduced purely for typing purposes and we can solve the typing issues by correctly typing the argument of the subqueries (in this example using `{ user: Ref<User> }`).

Essentially, my proposal boils down to using Javascript's built-in abstraction mechanisms rather than introducing our own (`defineQuery`, `defineForRow`, etc.). As for the types, we can type the (sub)queries directly.

DawidWraga:
I was thinking the exact same thing as Kevin.

samwillis:
Yep, we chatted it over and are going to go with the simpler option. It's then just docs and renaming a few of the internal types 👍 

Closing this

PR #225, opened 2025-07-03
use useSyncExternalStore for useLiveQuery

samwillis:


KyleAMathews (approved):
Lovely! :shipit:

packages/react-db/src/useLiveQuery.ts
```diff
@@ -55,58 +55,121 @@ export function useLiveQuery(
...
-  type KeyType =
-    typeof collection extends Collection<any, infer K, any>
-      ? K
-      : string | number
+  // Use refs to track version and memoized snapshot
+  const versionRef = useRef(0)
+  const snapshotRef = useRef<{
+    state: Map<any, any>
+    data: Array<any>
+    collection: Collection<any, any, any>
```
KyleAMathews, line 103: can we still infer the collection types?

samwillis: Yes, they are inferred in the function overloads, this is just internal and the type declaration was redundant.

PR #228, opened 2025-07-04
Add TrailBase integration and example

ignatz:
I've also started working on an example, as you suggested, but wanted to send out an RFC early.

Don't hold back. Let me know if this is going roughly the expected direction. If so, also let me know how you'd like to handle a potential example, e.g. should it be part of the same PR, should it be a full replica of the todo example?

Thanks!


packages/db-collections/src/trailbase.ts
```diff
@@ -0,0 +1,243 @@
...
+            },
+          })
+          cursor = response.cursor
+        }
+
+        commit()
+      }
+
+      // Afterwards subscribe.
+      async function subscribe() {
```
KyleAMathews: You could potentially miss events by waiting to subscribe after doing the fetch. I'd suggest subscribing and then fetching.

ignatz: Great catch, thanks!

ignatz:
> BTW, thinking about this PR made me realize we really should have collection types be in their own packages so e.g. if you want to do a major release you can #235

Sounds good. I was thinking the same, let alone when adding the backend-specific dependency.
 
> We'll be migrating the other packages in the next day or two so you can just move this PR to create a new package.

If I understood correctly, we'll table this review until the split has happened for me to rebase? Sounds good, just ping this PR and I'll get the train rolling.

Did you think about the example, should we handle it in a follow-up?

Thanks!

KyleAMathews:
The refactor is merged so you can now move your code to a `trailbase-db-collection` package https://github.com/TanStack/db/pull/252

ignatz:
Awesome! Moved to separate package.

KyleAMathews:
Great! I also refactored the example so you can easily add a trail base version. Could you look at doing that? The docker backend perhaps could have a trail base setup too?

ignatz:
> Great! I also refactored the example so you can easily add a trail base version. Could you look at doing that? The docker backend perhaps could have a trail base setup too?

Sure can do. Quick questions:

* Would you like me to add to the existing example?
* If so, just add to the existing docker-compose or have a separate compose setup?
* Would you like to make it part of this initial PR?

KyleAMathews:
Yeah same example. It's meant to be an omnibus example so easily see every collection type. And yes, one command to start all of them. And yes in this pr as that'll make it easier to test.

ignatz:
I added TrailBase to the example. I tried to fumble my way through...

* Additions and deletions work, however update both for the config and todo fail due to zod validation. I have differnt data types. I already made `created_at` and `updated_at` strings, which worked for insertions but not for updates :woman_shrugging:

KyleAMathews:
hmm interesting — you're using the zod schemas from drizzle it looks like? So you should use zod schemas consistent with the trailblaze schema? I suppose you can't have Date types w/ sqlite?

ignatz:
> hmm interesting — you're using the zod schemas from drizzle it looks like?

That my understanding, at first I tried to make mine compatible because I didn't necessarily want to change the downstream code.

 > So you should use zod schemas consistent with the trailblaze schema?

 I guess so.

> I suppose you can't have Date types w/ sqlite?

SQLite has a DATETIME type, which internally is just an INT64. That said, for type-safety reasons TrailBase requires STRICT mode (see migrations), which doesn't allow DATETIME. W/o STRICT, types are merely affinities and SQLite let's you stick any value into any column :)

KyleAMathews:
Ok — yeah the other collections serialize & deserialize to/from Date objects — I imagine most people might want that for Trailblaze? Perhaps you could add a `parser` & `serialize` config option to the Trailblaze collection?

ignatz:
> Ok — yeah the other collections serialize & deserialize to/from Date objects — I imagine most people might want that for Trailblaze? Perhaps you could add a `parser` & `serialize` config option to the Trailblaze collection?

That's an interesting proposal. I would certainly love not to to diverge the downstream code of the example. Could you elaborate a bit how this could look like, is there any precedent for me to look at? Thanks

EDIT: Would this just be two functions:

```ts
parser: (record: Record) -> TItem,
serialize: (item: TItem) -. Record,
```

? If so, yeah, yeah I'll do that. It will certainly be a good starting point.

ignatz:
Ok, I found https://github.com/TanStack/db/blob/b2ce3edeee1c64e42b50365bf0d0cf870f5d7586/examples/react/todo/src/lib/collections.ts#L21, which is probably what you meant. I'm not sure the same approach would work for me, since my base type is just `INTEGER` as opposed to PG `TIMESTAMPZ`.

Anyway, it works now :tada: . Do you think this is a workable starting point?

KyleAMathews:
Nice!

Yeah, parsing/serializing the whole row works — perhaps doing it per column would be better though as for big tables, it'd be tedious to have to pass through all the untouched stuff. Electric has a parser by pg type as you found.

KyleAMathews:
I tested it briefly — update works if I refresh the page after adding a todo. But not if I add it and then try to check it off. Changing the color of the background doesn't work either.

Screenshot 2025-07-15 at 7 05 22 PM

ignatz:
> perhaps doing it per column would be better though as for big tables, it'd be tedious to have to pass through all the untouched stuff. 

I like that, I'm just not sure my TS foo is strong enough. I'm aware that it can render Doom, but how would I define/infer a type `Parse` as follows:

```ts
type TItem = { 
 id: string;
 created: Date;
}
type TRecord = {
 id: string;
 created: number;
}

type Parse = {
 id?: (id: string) => string;
 created: (id: number) => Date;
}

function parse(parse: Parse, record: Partial<TRecord> | Record) : Partial<TItem> | TItem;
```

Instead I rewrote the serialize/parse implementations to only explicit set the relevant fields. Do you think this could be a workable starting point?

```ts
{
  parse: (record: Todo): SelectTodo => ({
    ...record,
    created_at: new Date(record.created_at * 1000),
    updated_at: new Date(record.updated_at * 1000),
  }),
}
```

> I tested it briefly — update works if I refresh the page after adding a todo. But not if I add it and then try to check it off. Changing the color of the background doesn't work either.

My bad. Fixed

https://github.com/user-attachments/assets/77a878ed-97f1-49ac-ab1d-5c9a8b5a7637

KyleAMathews:
No you have to test it in the same browser. 

https://github.com/user-attachments/assets/e3c03a85-02e0-4955-9d78-55aaf0dd287d

KyleAMathews:
I'm not very good at types either which is why I ask Claude 😆 This is what it came up with:

```ts
// Parser type: takes input values and converts them to the target type
type Parser<T, InputType = unknown> = {
  [K in keyof T]?: (value: InputType) => T[K];
};

// Serializer type: takes the typed values and converts them to output format
type Serializer<T, OutputType = string> = {
  [K in keyof T]?: (value: T[K]) => OutputType;
};
```

ignatz:
> No you have to test it in the same browser.

Did you update and rebuild? I force-pushed an update that should fix your issue. In theory at least :hide:

KyleAMathews:
ah oops! I ran `git pull` but I hadn't setup my checkout correctly and didn't notice I didn't get the latest changes.

It's all working now :-D

Let's get the parser/serializer slimmed down so you only specify what you actually want changed & then I think we're ready to go!

ignatz:
> I'm not very good at types either which is why I ask Claude 😆 This is what it came up with:
> 
> ```ts
> // Parser type: takes input values and converts them to the target type
> type Parser<T, InputType = unknown> = {
>   [K in keyof T]?: (value: InputType) => T[K];
> };
> 
> // Serializer type: takes the typed values and converts them to output format
> type Serializer<T, OutputType = string> = {
>   [K in keyof T]?: (value: T[K]) => OutputType;
> };
> ```

I should have been more clear about knowing types well enough that I suspect that it's not trivial. No offense to Claude, would still love to meet them over a beverage and talk about their outlook on humanity :upside_down_face: 

(There might be a simpler way but that's the easiest I could come up with)

I did do the gymnastics now. It's not pretty. They should be doing what you were asking for but now there's always a parse/serialize parameter needed. Otherwise the gymnastics get even worse. The more positive way to look at it is that previously it wasn't properly type-safe and now it is :woman_shrugging:

KyleAMathews:
hmm why can't we make parse/serialize optional?

ignatz:
Because they can only be optional when TItem and TRecord are fully
compatible.

KyleAMathews:
👍 works for me! Excited to get this out! Finishing up some other stuff but I'll get this merged and released a bit later today 🚀

ignatz:
> 👍 works for me! Excited to get this out! Finishing up some other stuff but I'll get this merged and released a bit later today 🚀

Yay :tada: - It was very fun working with you - looking forward to more.

PS: We can make `parse`/`serialize` conditionally optional in the future. It's just just more gymnastics but the change would be api backward compatible. As someone who mostly doesn't read docs, I'm just always happy when I can jump to definition and reasonably reason about the code.

KyleAMathews:
> It's just just more gymnastics

TypeScript always is haha.

It was fun working with you too! Hopefully TanStack/DB is a super nice addition for your community. I'll tweet it out too — are you / trailbase on the various socials?

ignatz:
I can lookup my Twitter handle in a bit, I don't think I've ever used it 😇.

Quick question, would you mind if I also update the example readme to
clarify Migrations and point out the admin dash?

Thanks again

KyleAMathews:
Yeah sure push away. Just lemme know when it's ready

ignatz:
That turned out to be pretty minor :sweat_smile: . Feel free to change it as you see fit. All good from my side.

My pristine twitter handle is @trailbase_io (in case that's what you were asking for otherwise no need to mention :woman_shrugging: ).

Last question, will you take care of publishing the npm package or would you rather have me do that? (not sure if there's some namespacing issue for the package)

Thanks again. This was fun, I'm very excited :pray:  :partying_face:

KyleAMathews:
https://github.com/user-attachments/assets/126eee08-c902-4622-b615-2879f5a9c5f2

you need to rebuild your admin w/ db to get realtime updates 😆

KyleAMathews:
https://x.com/kylemathews/status/1945577670550720583

ignatz:
> you need to rebuild your admin w/ db to get realtime updates 😆

You may be joking but the admin refresh does feel incredibly jarring :eyes:  :sweat_smile:

samwillis:
Hey @ignatz 

I spotted that the CI test that checks building the example site on PRs had started failing with a hang at the end of the build. I dug in a little and it was this PR that introduced the problem.

It looks like with TanStack Start as it runs the font end code it will initiate the collections. The `setInterval` inside `trailBaseCollectionOptions` then causes the Node process to never exit:

https://github.com/TanStack/db/blob/d08c46d699e06bba4b362760c115812c2f16831b/packages/trailbase-db-collection/src/trailbase.ts#L147-L169

I don't suppose you could take a look if there is a way around this?

You can see an example the CI hang here - it aborted after 6hrs: https://github.com/TanStack/db/actions/runs/16402266480/job/46343380659

ignatz:
@samwillis , will take a look. Thanks for bringing this up

ignatz:
@samwillis https://github.com/TanStack/db/pull/293 should fix the issue. Sorry for the inconvenience.

PR #230, opened 2025-07-06
fix transaction sorting order and enforce them having an id

samwillis:
While working on #204 I found two bugs:

1. direct insert/update/delete mutations where not settings an id on the transaction, this resulted in them being overwritten with further synchronous mutations as they had an id or `undefined` when being looked up. I have fixed that by ensuring that a transaction id will always fallback to a new uuid if not set.
  @KyleAMathews you used `new Transaction()` rather than `createTransaction()` in these mutations, this means they are not added to the global list of transactions. Is this intentional or should then use `createTransaction()`? If so I will fix that here as it's a very related bug, the id was enforced in `createTransaction`, but not in the Transaction constructor.

2. transactions were sorted by `createdAt` when rebasing over the latest synced data, however if two transactions happened in the same ms they would be ordered arbitrarily. This could result in the optimistic mutations happing in the wrong order - in #204 this manifested as a very strange situation where half the tests would fail, but if you only ran one of the failing tests it would work... I think this is also *highly* likely the cause of a flaky test I have seen when running the tests locally but never seen in CI.
  To fix this I have added an additional `sequenceNumber` to transaction, a monotonic counter in the current session - when we sort we compare by `createdAt` first, then fall back to `sequenceNumber` when they are created at the same time. I have left `createdAt` as this will be useful in future when we have some level of persistence for optimistic mutations.

### DIFF
--- a/packages/db/src/transactions.ts
+++ b/packages/db/src/transactions.ts
@@ -15,21 +15,8 @@ let transactionStack: Array<Transaction<any>> = []
 export function createTransaction<
   TData extends object = Record<string, unknown>,
 >(config: TransactionConfig<TData>): Transaction<TData> {
-  if (typeof config.mutationFn === `undefined`) {
-    throw `mutationFn is required when creating a transaction`
-  }
-
-  let transactionId = config.id
-  if (!transactionId) {
-    transactionId = crypto.randomUUID()
-  }
-  const newTransaction = new Transaction<TData>({
-    ...config,
-    id: transactionId,
-  })
-
+  const newTransaction = new Transaction<TData>(config)
   transactions.push(newTransaction)
-
   return newTransaction
 }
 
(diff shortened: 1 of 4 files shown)


packages/db/src/transactions.ts
```diff
@@ -67,20 +56,25 @@ export class Transaction<
   public isPersisted: Deferred<Transaction<T, TOperation>>
   public autoCommit: boolean
   public createdAt: Date
+  public sequenceNumber: number
```
KyleAMathews, line 59: nice!

On `new Transaction` — we could just only export the type so people can't use it. It was definitely my mistake using `new Transaction`.

KyleAMathews (approved):

packages/db/src/transactions.ts
samwillis: agreed, I changed it to only export the type, and changed the direct mutations to use createTransaction.

PR #232, opened 2025-07-07
add support for composable queries

samwillis:
replaces #216

This allows you to predefine a query with this syntax:

```ts
import { Query } from '@tanstack/db'

const myQuery = new Query()
  .from(...)
  .where(...)
```

then use it with:

```ts
const liveQueryCollection = createCollection(liveQueryCollectionOptions({
  query: myQuery,
}))
// or
const { data } = useLiveQuery({
  query: myQuery,
})
```

It also enables creating reusable callback functions for use in `where`, `groupBy`, `select` etc. with a `Ref<MyType>` to correctly cast to the type for the expression builder functions.

```ts
import type { Ref } from '@tanstack/db'

const userIsAdult = ({ user }: { user: Ref<User> }) => gt(user.age, 18)

const userIsActive = ({ user }: { user: Ref<User> }) =>
  eq(user.active, true)

const userIsYoung = ({ user }: { user: Ref<User> }) => lt(user.age, 25)

// Use the predicates in a query
const myQuery = new Query()
  .where(userIsAdult)
  .where(userIsActive)
  .where(userIsYoung)
```

Or in a select:

```ts
// Define reusable select objects using defineForRow
const userBasicInfo = ({ user }: { user: Ref<User> }) => ({
  id: user.id,
  name: user.name,
  email: user.email,
})

const userNameTransforms = ({ user }: { user: Ref<User> }) => ({
  nameUpper: upper(user.name),
  nameLower: lower(user.name),
})

// Use the select objects in a query
const myQuery = new Query()
  .where(({ user }) => eq(user.active, true))
  .select(({ user }) => ({
    ...userBasicInfo({ user }),
    ...userNameTransforms({ user }),
    age: user.age,
  }))
```

### DIFF
--- a/packages/db/src/query/builder/ref-proxy.ts
+++ b/packages/db/src/query/builder/ref-proxy.ts
@@ -1,4 +1,4 @@
-import { Ref, Value } from "../ir.js"
+import { PropRef, Value } from "../ir.js"
 import type { BasicExpression } from "../ir.js"
 
 export interface RefProxy<T = any> {
@@ -124,7 +124,7 @@ export function toExpression<T = any>(value: T): BasicExpression<T>
 export function toExpression(value: RefProxy<any>): BasicExpression<any>
 export function toExpression(value: any): BasicExpression<any> {
   if (isRefProxy(value)) {
-    return new Ref(value.__path)
+    return new PropRef(value.__path)
   }
   // If it's already an Expression (Func, Ref, Value) or Agg, return it directly
   if (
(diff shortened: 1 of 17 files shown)


kevin-dp:
@samwillis LGTM. One question i have though is why we need to wrap the query in an object with a `query` key?

```ts
const { data } = useLiveQuery({
  query: myQuery,
})
```

Couldn't we just pass the query as an argument directly without wrapping?

samwillis:
> Couldn't we just pass the query as an argument directly without wrapping?

The wrapping object is currently the liveQueryCollectionOptions, so there is more we can pass in there. I left it as either the callback or this options object to simplify the overloads and not add another. We can always add it later if there is desire to.

PR #242, opened 2025-07-08
DRAFT live query docs

samwillis:


docs/live-queries.md
```diff
@@ -0,0 +1,1272 @@
...
+        name: user.name,
+      })),
+  getKey: (user) => user.id, // Optional: uses stream key if not provided
+  startSync: true, // Optional: starts sync immediately
+}))
+```
+
+### Convenience Function
+
+For simpler cases, you can use `createLiveQueryCollection` as a shortcut:
```
KyleAMathews, line 99: do we want this? We explicitly went down the two function calls as a way of being explicit about how collections work & I'd rather keep things consistent — people can certainly add their own convenience functions if they'd like

docs/live-queries.md
```diff
@@ -0,0 +1,1272 @@
...
+  id: 'active-users', // Optional: auto-generated if not provided
+  query: (q) =>
+    q
+      .from({ user: usersCollection })
+      .where(({ user }) => eq(user.active, true))
+      .select(({ user }) => ({
+        id: user.id,
+        name: user.name,
+      })),
+  getKey: (user) => user.id, // Optional: uses stream key if not provided
```
KyleAMathews, line 84: Is there an actual use case for explicitly defining the key? We should probably then say what it is or if not, just disallow it altogether.

docs/live-queries.md
```diff
@@ -0,0 +1,1272 @@
...
+| `id` | `string` | Unique identifier (auto-generated if not provided) |
+| `query` | `QueryBuilder` or function | The query definition |
+| `getKey` | `(item) => string \| number` | Function to extract keys from results |
+| `schema` | `Schema` | Optional schema for validation |
+| `startSync` | `boolean` | Whether to start syncing immediately |
+| `gcTime` | `number` | Garbage collection time in milliseconds |
+
+## Basic Queries
+
+The foundation of every query is the `from` method, which specifies the source collection or subquery. You can alias the source using object syntax.
```
KyleAMathews: you can probably remove this section

docs/live-queries.md
```diff
@@ -0,0 +1,1272 @@
...
+    q
+      .from({ u: usersCollection })
+      .select(({ u }) => ({
+        name: u.name,
+        email: u.email,
+      }))
+}))
+```
+
+### Query Builder Pattern
```
KyleAMathews: I'd expect this to show something new from the previous ones e.g. an `if (filter === 'foo') { /* add a where clause */ }`

docs/live-queries.md
```diff
@@ -0,0 +1,1272 @@
...
+    .select(({ user }) => ({
+      id: user.id,
+      name: user.name,
+      isAdult: gt(user.age, 18),
+      nameLength: length(user.name),
+    }))
+)
+```
+
+### Using Functions
```
KyleAMathews: this could be combined with the section above

docs/live-queries.md
```diff
@@ -0,0 +1,1272 @@
...
+      ({ user, post }) => eq(user.id, post.userId),
+      'inner'
+    )
+)
+
+// Right join - all posts, even without users
+const allPosts = createLiveQueryCollection((q) =>
+  q
+    .from({ user: usersCollection })
+    .join(
```
KyleAMathews: hmm I kinda liked `.rightJoin` more — that seems clearer than a lonely third argument

docs/live-queries.md
```diff
@@ -0,0 +1,1272 @@
...
+// Final query with selection
+const result = createLiveQueryCollection(activeUsersWithPosts
+  .select(({ user, post }) => ({
+    userName: user.name,
+    postTitle: post.title,
+  }))
+)
+```
+
+### Conditional Query Building
```
KyleAMathews, line 966: redundant?

docs/live-queries.md
```diff
@@ -0,0 +1,1272 @@
...
+  return query.select(({ user }) => ({
+    id: user.id,
+    name: user.name,
+  }))
+}
+
+const activeUsers = createLiveQueryCollection(buildUserQuery({ activeOnly: true, limit: 10 }))
+```
+
+### Caching Intermediate Results
```
KyleAMathews, line 993: redundant?

docs/live-queries.md
```diff
@@ -0,0 +1,1272 @@
...
+
+Build complex queries by composing smaller, reusable parts. This approach makes your queries more maintainable and allows for better performance through caching.
+
+### Building Queries Incrementally
+
+Create query parts and combine them:
+
+```ts
+// Base query for active users
+const activeUsers = q
```
KyleAMathews: redundant w/ subqueries section?

docs/live-queries.md
```diff
@@ -0,0 +1,1272 @@
...
+      .where(({ user }) => and(isActiveUser(user), isAdultUser(user)))
+      .select(({ user }) => ({
+        id: user.id,
+        name: user.name,
+        age: user.age,
+      }))
+}))
+```
+
+You can also create callbacks that take the whole row and pass them directly to `where`:
```
KyleAMathews, line 1074: 🔥

KyleAMathews (commented):
Love it!

docs/live-queries.md
samwillis: My thinking here is that this could be the exception to the rule. Users will be *very* regularly be creating liveQueryCollections outside of components, in loaders and other places, and if we don't provide this wrapper they will after a while get frustrated and make it themselves. The intention is that it exposes the same api as the useLiveQuery, you just use the callback to build your query, so moving between the two doesn't require any different structures in your code.

This is a strong opinion (I would expect it from the lib), but but weakly held - if you felt equally strongly that we should remove it then let's do that.

docs/live-queries.md
samwillis: Yes, with joins. We generate a composite key for the join using the key from each side, but often there would only need to be one side of the key. 

For example if you join an aggregate query bringing in stats for "issues" - comment counts, velocity, ect. - you are joining the PK on the same PK. Issue 1 would end up with a key of `[1,1]`, 2 with `[2,2]` ect.

In that cases using `getKey: (issue) => issue.id` would ensure that the resulting collection used the same PK as the original collection.

There is also a similar thing with groupBy, it generates a json string as the key, but often there is a better natural key that could have been used.

docs/live-queries.md
samwillis: Agreed, I meant to add this but it got forgotten in the mix. @thruflo was keen on a simple `join()` for a left join, and we then discussed having `leftJoin`, `rightJoin`, `innerJoin` ect. that wrapped it and set that final param.

How does that sound?

docs/live-queries.md
KyleAMathews: yeah I don't feel that strongly against it 👍 so let's keep it

docs/live-queries.md
KyleAMathews: ok, I do remember you saying this. Then yeah, adding the above as context in the docs would be great

docs/live-queries.md
KyleAMathews: 👍

docs/live-queries.md
samwillis: @KyleAMathews on this, we are likely to want a `createLiveQueryCollections` (plural) that can return multiple multiple collection from a single query pipeline. Thats where the `liveQueryCollectionOptions` falls down, it imposes a 1:1 query-to-collection relationship. We certainly keep it, but it's worth considering what this looks like when we move on to that.

KyleAMathews (approved):
:shipit: great stuff!

docs/config.json
```diff
@@ -20,6 +20,10 @@
         {
           "label": "Installation",
           "to": "installation"
+        },
```
thruflo, line 23: I would imagine that in addition to adding to the nav, there must be some parts of the existing documentation that would also benefit from referring to the new live query docs.

docs/live-queries.md
```diff
@@ -0,0 +1,1373 @@
+---
+title: Live Queries
+id: live-queries
+---
+
+# TanStack DB Live Queries
+
+TanStack DB provides a powerful, type-safe query system that allows you to fetch, filter, transform, and aggregate data from collections using a SQL-like fluent API. All queries are **live** by default, meaning they automatically update when the underlying data changes.
+
+The query system is built around a fluent API similar to SQL query builders like Kysely or Drizzle. The query builder doesn't perform operations in the order of method calls - instead, it composes your query into an optimal incremental pipeline that gets compiled and executed efficiently. Each method returns a new query builder, allowing you to chain operations together.
```
thruflo: Perhaps a hyperlink for fluent API -- I don't really know what a fluent api is. Or deprioritise the concept and say "built around a query builder API similar to ...".

docs/live-queries.md
```diff
@@ -0,0 +1,1373 @@
...
+
+```ts
+import { createCollection, liveQueryCollectionOptions, eq } from '@tanstack/db'
+
+const activeUsers = createCollection(liveQueryCollectionOptions({
```
thruflo, line 17: I was surprised to see the use of `createCollection(liveQueryCollectionOptions({` in most of the code examples.

So far, most of my use of live queries has been in a component context with `useLiveQuery`. Below later on you switch to the more convenient `createLiveQueryCollection`. And then also introduce the `Query` class.

Using `createCollection(liveQueryCollectionOptions({` is a lot of "collection" for creating a query.

docs/live-queries.md
```diff
@@ -0,0 +1,1373 @@
...
+        id: user.id,
+        name: user.name,
+        email: user.email,
+      }))
+}))
+```
+
+The result types are automatically inferred from your query structure, providing full TypeScript support. When you use a `select` clause, the result type matches your projection. Without `select`, you get the full schema with proper join optionality.
+
+Live queries are just collections that automatically update when their underlying data changes. You can subscribe to changes, iterate over results, and use all the standard collection methods.
```
thruflo: > Live queries are just collections that automatically update when their underlying data changes.

This text is duplicated in the next section lead in. It might also be worth pulling up above the first code example, to make it clearer why the live query is being created using two collection function calls.

docs/live-queries.md
```diff
@@ -0,0 +1,1373 @@
...
+- [Expression Functions Reference](#expression-functions-reference)
+- [Functional Variants](#functional-variants)
+
+## Creating Live Query Collections
+
+Live queries are collections that automatically update when their underlying data changes. You can create them using `liveQueryCollectionOptions` with `createCollection`, or use the convenience function `createLiveQueryCollection`.
+
+### Using liveQueryCollectionOptions
+
+The fundamental way to create a live query is using `liveQueryCollectionOptions` with `createCollection`:
```
thruflo, line 52: I see the existing PR discussion below. To share my viewpoint (which is perhaps slightly fresh / less deep in the weeds):

1. if we have createLiveQueryCollection then that is simpler for examples (that don't need the extra flexibility) than the `createCollection(liveQueryCollectionOptions({` 
2. the most common starting points to use live queries is in component context, with useLiveQuery

I think this section `Creating Live Query Collections` all really emphasises that live queries are collections. I'm not sure this is necessary. It's like "yes a live query is a collection but it's also a live query". Like the page is called "live queries". Could this section be "Creating Live Queries" and start with a "live queries can be created and used in a few different ways. The most common is in a component using useLiveQuery but they can also be created standalone using createLiveQuery and composable using the lower level Query builder".

Note my dropping of the Collection suffix from createLiveQuery. Then still explain that a live query is a collection and show that the longer createCollection(liveQueryCollectionOptions( is equivalent with more flexibility. But then the examples can more clearly generally use `createLiveQuery`?

docs/live-queries.md
```diff
@@ -0,0 +1,1373 @@
...
+
+// Logical operators
+and(condition1, condition2)
+or(condition1, condition2)
+not(condition)
+```
+
+For a complete reference of all available functions, see the [Expression Functions Reference](#expression-functions-reference) section.
+
+## Select Projections
```
thruflo: "Projections" reads a bit odd here. Could this just be `Select`?

docs/live-queries.md
```diff
@@ -0,0 +1,1373 @@
...
+Result type:
+{
+  user: User,
+  post?: Post, // post is optional because it is a left join
+}
+*/
+```
+
+### Right Join
+```ts
```
thruflo, line 509: Nit: line space.

docs/live-queries.md
```diff
@@ -0,0 +1,1373 @@
...
+      userName: user.name,
+      postTitle: post.title,
+      commentText: comment.text,
+    }))
+)
+```
+
+## Subqueries
+
+Subqueries allow you to use the result of one query as input to another. They're different from using a live query collection directly - subqueries are embedded within the query itself.
```
thruflo: Could you perhaps expand a bit more here on the difference?

docs/live-queries.md
```diff
@@ -0,0 +1,1373 @@
...
+        email: u.email,
+      }))
+}))
+```
+
+## Where Clauses
+
+Use `where` clauses to filter your data based on conditions. You can chain multiple `where` calls - they are combined with `and` logic.
+
+The `where` method takes a callback function that receives an object containing your table aliases and returns a boolean expression. You build these expressions using comparison functions like `eq()`, `gt()`, and logical operators like `and()` and `or()`. This declarative approach allows the query system to optimize your filters efficiently. These are described in more detail in the [Expression Functions Reference](#expression-functions-reference) section.
```
thruflo: I think it may be worth having a section that explains what you can and can't do in the where clause (and other) functions. I.e.: there's using the imported operation functions and then there's just doing arbitrary JS logic.

This is partly addressed in the last "advanced warning `.fn.` function expressions" section. Which kind of shows you to use `.fn.` if you want to do arbitrary stuff and warns about lack of optimisation. But these where and select etc callbacks are normal functions. So should we be explicit about this up front and be clear what you can and can't do in them? And flag up the existence of the .fn. escape hatch?

I'm not quite sure where this goes in the doc flow but I think it is one of the important hidden assumptions/concerns that devs will have to tease out themselves unless we state it clearly.

thruflo (commented):
Awesome stuff. Fantastic reference, lots of information, very clear and well organised.

docs/live-queries.md
thruflo: Another perspective on this is that if live queries are just collections then perhaps we should be talking about querying collections, not about live queries. I.e.: either a live query is a thing and a collection is a thing. Or they are one thing and collections are just queryable.

docs/live-queries.md
```diff
@@ -0,0 +1,1373 @@
...
+        email: user.email,
+      }))
+}))
+```
+
+The result types are automatically inferred from your query structure, providing full TypeScript support. When you use a `select` clause, the result type matches your projection. Without `select`, you get the full schema with proper join optionality.
+
+Live queries are just collections that automatically update when their underlying data changes. You can subscribe to changes, iterate over results, and use all the standard collection methods.
+
+## Table of Contents
```
thruflo, line 32: I think another section or aspect that's maybe not currently covered in this guide is query dependencies.

When queries are defined/derived from other queries/collections then they automatically update when their source queries/collections change.

When queries are defined using JS variables then their definitions do not change. In a component context, this is addressed with useLiveQuery where you pass a dependencies array.

The mechanism for handling dependency changes in the latter case is much more blunt than in the former.

I think it's worth outlining how this stuff works. Perhaps the meat of the usage with dependencies belongs to the specific framework docs. However I think the principle should be established here.

docs/live-queries.md
samwillis: Yes, having leand into the concept that a "live query is a collection", I fear we are going to break that distinction more as we expand the live query capabilities.

We currently have this 1:1 relationship of live query to collection but thats not going to be the case soon.

This is a common pattern that I think we will want to support:

```ts
const { totalCount, pagedResults } = useLiveQuery((q) => {
  const baseQuery = q
    .from({ issues: issueCollection })
    .where(({ issue }) => eq(issue.projectId, 123))
    .where(/* more complex expressions from UI filters */)
  
  const totalCount = baseQuery
    .select(({ issue} ) => ({
      count: count(issue.id)
    }))
  
  const pagedResults = baseQuery
    .orderBy(({ issue }) => issue.data)
    .limit(30)
    
  return { totalCount, pagedResults }
})

// Render the total results count in a header, and then the paged results below.
// (note the I have some ideas now to reuse the same query for paging through 
// results, but thats another thing later on)
```

With that I feel we need to more away from the 1:1 relationship.

docs/config.json
samwillis: I've added a few links, but we may want more. I think we should go back and have a new pass on the overview now that we are getting more of these pages.

docs/live-queries.md
samwillis: I'm not sure where best to place info on the deps array, but suspect in the React docs. Vue for example should not need the deps array and its not part of the apis discussed here. I do agree is complex and maybe needs highlighting more.

samwillis:
I'm merging this now as there is useful information, we can continue to iterate in further PRs

docs/live-queries.md
KyleAMathews: Wouldn't totalCount and pagedResults both be collections? We're not breaking the connection if useLiveQuery gets more complex.

samwillis: Yes, they are still collections. However rather than a live query collection being constructed by using liveQueryCollectionOptions and passing that to a collection, we *have* to have a factory that takes a query and returns 1+ collections. 

It's no longer Collection passed Options, which are in turn passed a Query, it's a Query processed to output multiple Collections.

There are lots of nuances that will need solving, such as the lifecycle of the Collection no longer being 1:1 with its query as that query may be being used by multiple collections.

PR #244, opened 2025-07-09
feat: Distinct operator for queries

kevin-dp:
This PR adds support for a SQL-like `distinct` operator on queries.
It works on top of https://github.com/electric-sql/d2ts/pull/80 so we need to wait until that PR is merged and a new version of d2mini is released. Until then CI will be red.

### DIFF
--- a/.changeset/slick-ghosts-shine.md
+++ b/.changeset/slick-ghosts-shine.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+New distinct operator for queries.
--- a/packages/db/package.json
+++ b/packages/db/package.json
@@ -3,7 +3,7 @@
   "description": "A reactive client store for building super fast apps on sync",
   "version": "0.0.21",
   "dependencies": {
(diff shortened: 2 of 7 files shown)


samwillis:
@kevin-dp note that I bumped d2mini on your branch after releasing it. CI is green 😁

kevin-dp:
> This is great!
> 
> We should add a test to validate the behaviour when there is no `select`. We either support that, or don't. If it doesn't worth with no select it should throw an error when you try to use without a select. (I think it should work)

Yes distinct without select works. There's a test for that here: https://github.com/TanStack/db/pull/244/files#diff-92b5216ddf806ff91a92eeda5701f1093a170697b0328f3297f31ae30d2cde7bR180-R189
But it's not very useful because every row has a unique ID so there's nothing to de-duplicate.

KyleAMathews:
If not useful should we just disallow it?

samwillis:
Yep, let's require a select with a distinct. It make it much more explicit what it is doing - fewer foot guns.

kevin-dp:
@KyleAMathews @samwillis i added an explicit check to disallow DISTINCT without a SELECT and modified the corresponding unit test accordingly.

samwillis (approved):
🚢

PR #245, opened 2025-07-09
[Electric]: Use numbers for txid

KyleAMathews:
Fix https://github.com/TanStack/db/issues/246

After some more research, this is the best way to get the 32bit internal xid (tx id) that is also sent out over logical replication and then through Electric i.e. what we'll match against inside Electric collections.

```ts
async function generateTxId(tx: any): Promise<Txid> {
  // The ::xid cast strips off the epoch, giving you the raw 32-bit value
  // that matches what PostgreSQL sends in logical replication streams
  // (and then exposed through Electric which we'll match against
  // in the client).
  const result = await tx`SELECT pg_current_xact_id()::xid::text as txid`
  const txid = result[0]?.txid

  if (txid === undefined) {
    throw new Error(`Failed to get transaction ID`)
  }

  return parseInt(txid, 10)
}
```


examples/react/todo/src/api/server.ts
```diff
@@ -23,15 +23,15 @@ app.get(`/api/health`, (req, res) => {
...
+async function generateTxId(tx: any): Promise<number> {
   const result = await tx`SELECT txid_current() as txid`
   const txid = result[0]?.txid
 
   if (txid === undefined) {
     throw new Error(`Failed to get transaction ID`)
   }
 
-  return String(txid)
+  return parseInt(txid, 10)
```
kevin-dp, line 39: Perhaps good to check that the parsed int is not `NaN`? Just to safeguard against that even though it doesn't look very plausible.

KyleAMathews: Yeah not sure pg could return an unparsable number? Though I did some more research and `txid_current()` is actually deprecated. But with the new function, we can cast it to xid then to text and then parse it in js to get the 32bit internal xid that's also on the logical replication:

```ts
async function generateTxId(tx: any): Promise<Txid> {
  // The ::xid cast strips off the epoch, giving you the raw 32-bit value
  // that matches what PostgreSQL sends in logical replication streams
  // (and then exposed through Electric which we'll match against
  // in the client).
  const result = await tx`SELECT pg_current_xact_id()::xid::text as txid`
  const txid = result[0]?.txid

  if (txid === undefined) {
    throw new Error(`Failed to get transaction ID`)
  }

  return parseInt(txid, 10)
}

```

PR #254, opened 2025-07-10
Refactor example app to use TanStack Start and separate routes / collection type

KyleAMathews:


KyleAMathews:
As a follow-up we can add the two local collection types

examples/react/todo/src/App.tsx
```diff
@@ -12,7 +12,7 @@ import type { SelectConfig, SelectTodo } from "./db/validation"
 import type { FormEvent } from "react"
 
 // API helper for todos and config
-const API_BASE_URL = `http://localhost:3001/api`
```
samwillis, line 15: I don't think this App.tsx file is used any more

KyleAMathews: Yes, I'll remove that.

PR #256, opened 2025-07-13
query optimiser that pushes where clauses down to subqueries closer to the source

samwillis:
This leads on to pushing those where clauses down to the collection subscription which can use the index...

There is a detailed comment explains the optimiser at the top of `packages/db/src/query/optimizer.ts`, but in essence it rewrites the query IR moving things around.


packages/db/src/query/optimizer.ts
```diff
@@ -0,0 +1,750 @@
...
+
+  // Create an AND function with all expressions as arguments
+  return new Func(`and`, expressions)
+}
+
+/**
+ * Deep equality check for QueryIR objects
+ */
+function areQueriesEqual(query1: QueryIR, query2: QueryIR | null): boolean {
+  if (query2 === null) return false
```
kevin-dp: Why is this needed? `deepEquals` can handle null values.

packages/db/src/query/optimizer.ts
```diff
@@ -0,0 +1,750 @@
...
+ */
+function splitAndClauses(
+  whereClauses: Array<BasicExpression<boolean>>
+): Array<BasicExpression<boolean>> {
+  const result: Array<BasicExpression<boolean>> = []
+
+  for (const clause of whereClauses) {
+    if (clause.type === `func` && clause.name === `and`) {
+      // Recursively split nested AND clauses to handle complex expressions
+      const splitArgs = splitAndClauses(
```
kevin-dp, line 345: The docstring on this function says:
> Step 1: Split all AND clauses at the root level into separate WHERE clauses.

But since we call `splitAndClauses` recursively here this is not only at the root level but also nested. So it turns any nested conjunctions (i.e. AND clauses) into a flatened array of all conjuncts.

packages/db/src/query/optimizer.ts
```diff
@@ -0,0 +1,750 @@
...
+): GroupedWhereClauses {
+  const singleSource = new Map<string, Array<BasicExpression<boolean>>>()
+  const multiSource: Array<BasicExpression<boolean>> = []
+
+  // Categorize each clause based on how many sources it touches
+  for (const clause of analyzedClauses) {
+    if (clause.touchedSources.size === 1) {
+      // Single source clause - can be optimized
+      const source = Array.from(clause.touchedSources)[0]
+      if (source) {
```
kevin-dp: Do we need this if test? We know that the size of the array is 1 and we took the only element from that array and assigned it to `source`. Or could it be that the source itself is a falsy value? In which case would that happen?

packages/db/src/query/optimizer.ts
```diff
@@ -0,0 +1,750 @@
...
+  const optimizedFrom = optimizeFromWithTracking(
+    query.from,
+    groupedClauses.singleSource,
+    actuallyOptimized
+  )
+
+  // Optimize JOIN clauses and track what was optimized
+  const optimizedJoins = query.join
+    ? query.join.map((joinClause) => ({
+        type: joinClause.type,
```
kevin-dp: Perhaps more robust to spread the joinClause in and override `from` in case we add/remove fields to join clauses later.

packages/db/src/query/compiler/index.ts
```diff
@@ -10,30 +11,35 @@ import type {
...
   if (cachedResult) {
     return cachedResult
   }
 
+  // Optimize the query before compilation
+  const query = optimizeQuery(rawQuery)
+
+  // Create mapping from optimized query to original for caching
+  queryMapping.set(query, rawQuery)
+  mapNestedQueries(query, rawQuery, queryMapping)
```
kevin-dp, line 41: I don't really like that we're passing around mutable state and relying on `mapNestedQueries` to mutate that state. It feels like this should be a class since it keeps some state and provides methods over that state.

packages/db/src/query/compiler/index.ts
```diff
@@ -274,3 +291,55 @@ function processFrom(
...
+  if (optimizedQuery.join && originalQuery.join) {
+    for (
+      let i = 0;
+      i < optimizedQuery.join.length && i < originalQuery.join.length;
+      i++
+    ) {
+      const optimizedJoin = optimizedQuery.join[i]
+      const originalJoin = originalQuery.join[i]
+
+      if (optimizedJoin && originalJoin) {
```
kevin-dp: This condition seems to always be true since the loop iteration condition is: `i < optimizedQuery.join.length && i < originalQuery.join.length` which is equivalent to `i < Math.min(optimizedQuery.join.length, originalQuery.join.length)` so inside the loop both arrays should have an entry at position `i`.

I'm slightly wondering what this does though. If there are N elements in `originalQuery.join` are there always going to be N elements in `optimizedQuery.join` or less than N or more than N?
If:
- there will always be N elements, then we can simplify the for loop condition and remove the `if (optimizedJoin && originalJoin)`
- there will by <= N or >= N elements then how do we know which optimized query maps to which original query? (since there's no one-on-one mapping this seems impossible to determine without additional information)

packages/db/src/utils.ts
```diff
@@ -0,0 +1,86 @@
...
+  return deepEqualsInternal(a, b, new WeakMap())
+}
+
+/**
+ * Internal implementation with cycle detection to prevent infinite recursion
+ */
+function deepEqualsInternal(
+  a: any,
+  b: any,
+  visited: WeakMap<object, object>
```
kevin-dp: Why does this need to be a WeakMap? It will be GCed after the call to `deepEquals` anyway.

packages/db/src/query/optimizer.ts
samwillis: yep, removed

packages/db/src/query/compiler/index.ts
samwillis: Agreed, it's getting quite complex. I would like to refactor the compiler into a class, but let's do it in a follow-up PR or this will blow up.

PR #257, opened 2025-07-13
enable adding indexes to collections to speed up initial data lookups

samwillis:
**stacked on #256**

Indexes are created like so:

```ts
collection.createIndex(
  (row) => row.age
  {
    name: `ageIndex`, // optional name in an optional options object
  }
)
```

the callback uses the same expression builder to extract a value to index as the query builder does for where classes.

You can add as many indexes to a collection as you like, and they can be created after the collection was created initially.

This pr does not link this up to live queries...


packages/db/src/query/compiler/evaluators.ts
```diff
@@ -75,91 +101,121 @@ function compileRef(ref: PropRef): CompiledExpression {
...
+  if (propertyPath.length === 0) {
+    // Empty path - return the whole item
+    return (item) => item
+  } else if (propertyPath.length === 1) {
+    // Single property access - most common case
+    const prop = propertyPath[0]!
+    return (item) => item[prop]
+  } else {
+    // Multiple property navigation
+    return (item) => {
```
kevin-dp: Why do we need the `if` and `else if` branches above? It seems that the function returned in this `else` branch works for all paths, thus also for the empty path and the path containing a single element.

packages/db/src/query/compiler/order-by.ts
```diff
@@ -52,46 +53,6 @@ export function processOrderBy(
...
-    if (bothObjects && notNull) {
-      return a.toString().localeCompare(b.toString())
-    }
-
-    if (a < b) return -1
-    if (a > b) return 1
-    return 0
-  }
-
   const descComparator = (a: unknown, b: unknown): number => {
```
kevin-dp: If we move `ascComparator` to a dedicated `comparison.ts` file, we should also move `descComparator`. ALso note that their signatures differ sligthly, `ascComparator` takes `any` arguments and `descComparator` takes `unknown` args. We should be consistent and if possible use `unknown` for both.

packages/db/src/types.ts
```diff
@@ -499,6 +501,59 @@ export type KeyedNamespacedRow = [unknown, NamespacedRow]
...
+  TKey extends string | number = string | number,
+> {
+  /** Unique identifier for this index */
+  id: string
+  /** Optional name for the index */
+  name?: string
+  /** The expression that defines what to index (a Ref expression) */
+  expression: BasicExpression
+  /** Ordered array of [indexedValue, Set<TKey>] pairs, sorted by indexedValue */
+  orderedEntries: Array<[any, Set<TKey>]>
```
kevin-dp: We have the `valueMap` for fast equality lookups. I assume we have the `orderedEntries` map for range queries. The ordered array of indexed values does provide efficient (logarithmic-time) lookups but inserting newly indexed values is slow since insertion in an array is in linear time. So this can imply a significant overhead on writes. If we want to support efficient range queries, we should move to a B+ tree for the index such that lookups, insertions, and deletions are in logarithmic time.

packages/db/src/types.ts
```diff
@@ -499,6 +501,59 @@ export type KeyedNamespacedRow = [unknown, NamespacedRow]
...
+export interface IndexOptions {
+  /** Optional name for the index, used for referencing and debugging */
+  name?: string
+}
+
+/**
+ * Represents an index for fast lookups on a collection
+ * All indexes are ordered to support range queries
+ */
+export interface CollectionIndex<
```
kevin-dp: I think it would be more appropriate for `CollectionIndex` to be a class instead of a plain JS object. Right now, it's an object leaking internal data structures to whoever wants to use it. As a result, the `CollectionImpl` class now contains a lot of additional complexity for building and maintaining indexes. It would be better to abstract all of that away behind a `CollectionIndex` class. That way, the changes to `CollectionImpl` will be minimal.

So, I'd rather have a `CollectionIndex` that implements an index and is a black box (i.e. all implementation details are contained within the class and the caller doesn't need to know about them). That also simplifies code evolution since we could swap out internal data structures without requiring changes to the calling code, as long as the public interface remains the same.

Such a `CollectionIndex` class would expose an `add(value, key)` method which will then update its internal data structures such as `valueMap`, `orderedEntries`, and `indexedKeys`. It would also expose methods for equality lookups and range scans.

packages/db/src/collection.ts
```diff
@@ -1197,6 +1227,335 @@ export class CollectionImpl<
...
+   * const fullNameIndex = collection.createIndex((row) => `${row.firstName} ${row.lastName}`)
+   */
+  public createIndex(
+    indexCallback: (row: SingleRowRefProxy<T>) => any,
+    options: IndexOptions = {}
+  ): CollectionIndex<TKey> {
+    this.validateCollectionUsable(`createIndex`)
+
+    // Generate unique ID for this index
+    const indexId = `index_${++this.indexCounter}`
```
kevin-dp: Why not just the `indexCounter` number?

packages/db/src/collection.ts
```diff
@@ -1197,6 +1227,335 @@ export class CollectionImpl<
...
+    // Create a comparison function for ordering
+    const compareFn = this.createComparisonFunction()
+
+    // Create the index
+    const index: CollectionIndex<TKey> = {
+      id: indexId,
+      name: options.name,
+      expression,
+      orderedEntries: [],
+      valueMap: new Map<any, Set<TKey>>(),
```
kevin-dp: No need to pass the type arguments explicitly, they can probably be inferred.

packages/db/src/collection.ts
```diff
@@ -1197,6 +1227,335 @@ export class CollectionImpl<
...
+      indexedKeys: new Set<TKey>(),
+      compareFn,
+    }
+
+    // Build the index with current data
+    this.buildIndex(index)
```
kevin-dp: This should be a method of the index, where we do `index.build(data)`

packages/db/src/collection.ts
```diff
@@ -1197,6 +1227,335 @@ export class CollectionImpl<
...
+      } else {
+        definedValues.push(value)
+      }
+    }
+
+    // Sort defined values normally
+    definedValues.sort(index.compareFn)
+
+    // Sort undefined values (they should all be equal, but just in case)
+    undefinedValues.sort(index.compareFn)
```
kevin-dp: They are not all equal because you push `null` and `undefined` to this array. So `sort` will return an array that first contains all the `null` values and then the `undefined` values. Note that, our comparison function considers `null` to be smaller than any other value so we can keep `null` values in the `definedValues` and rely on sorting that one and only push `undefined` values to `undefinedValues`. Then no need to sort that array.

packages/db/src/collection.ts
```diff
@@ -1197,6 +1227,335 @@ export class CollectionImpl<
...
+
+    // Combine with undefined/null first (to match orderBy compiler behavior)
+    const sortedValues = [...undefinedValues, ...definedValues]
```
kevin-dp: Spreading these arrays is more expensive than mutating them. Ideally we would:
- group `definedValues` into 2 groups: undefined and the rest
- sort the array of non-undefined values
- push the sorted array to the end of `undefinedValues` array

packages/db/src/collection.ts
```diff
@@ -1197,6 +1227,335 @@ export class CollectionImpl<
...
+        const keySet = new Set<TKey>([key])
+        index.valueMap.set(indexedValue, keySet)
+
+        // Find correct position in ordered entries using binary search
+        const insertIndex = this.findInsertPosition(
+          index.orderedEntries,
+          indexedValue,
+          index.compareFn
+        )
+        index.orderedEntries.splice(insertIndex, 0, [indexedValue, keySet])
```
kevin-dp: This `splice` operation is in linear time and defeats all the performance benefits of doing a logarithmic lookup to find the insert position. We should use a B+ tree instead of a sorted array for the `orderedEntries` such that lookup, insertions, and deletions are in logarithmic time.

packages/db/src/collection.ts
```diff
@@ -1197,6 +1227,335 @@ export class CollectionImpl<
...
+
+      const keysForValue = index.valueMap.get(indexedValue)
+      if (keysForValue) {
+        keysForValue.delete(key)
+        if (keysForValue.size === 0) {
+          // Remove from valueMap
+          index.valueMap.delete(indexedValue)
+
+          // Remove from orderedEntries
+          const entryIndex = index.orderedEntries.findIndex(
```
kevin-dp: Don't do a linear `findIndex` when you can leverage the sorted order and do a logarithmic binary search. The `findInsertPosition` will return you the index of the value if it is in the array. So you could use `findInsertPosition` and then do a check to see whether or not the element at that index is the one you're looking for. If it is, then you can delete it.

packages/db/src/collection.ts
```diff
@@ -1197,6 +1227,335 @@ export class CollectionImpl<
...
+  ): void {
+    this.removeFromIndex(index, key, oldItem)
+    this.addToIndex(index, key, newItem)
+  }
+
+  /**
+   * Finds the correct insert position for a value in the ordered entries array
+   * @private
+   */
+  private findInsertPosition<V>(
```
kevin-dp: This should definitely not be a method of this collection class but should be a function exported by a utility file.

packages/db/src/collection.ts
```diff
@@ -1197,6 +1227,335 @@ export class CollectionImpl<
...
+      // Fast equality lookup
+      const keys = index.valueMap.get(value)
+      if (keys) {
+        keys.forEach((key) => result.add(key))
+      }
+      return result
+    }
+
+    // Range operations - iterate through ordered entries
+    for (const [indexValue, keys] of index.orderedEntries) {
```
kevin-dp: [B+ tree library](https://www.npmjs.com/package/sorted-btree) gives you range queries out of the box :-)

kevin-dp: Also, this is scanning linearly over all entries which is naive. The array is ordered so we can do a binary search to find the index of the value. Then, depending on the operation we would want to take a slice of the array:
- `[0, idx[` for `lt`
- `[0, idx]` for `lte`
- `[idx, end]` for `gte`
- `]idx, end]` for `gt`

That's more efficient because we find the index in logarithmic time and the slice only takes K steps where K is the size of the slice.

packages/db/src/collection.ts
```diff
@@ -1815,20 +2174,487 @@ export class CollectionImpl<
...
+    }
+
+    return result
+  }
+
+  /**
+   * Optimizes a query expression using available indexes
+   * @private
+   */
+  private optimizeQuery(expression: BasicExpression): {
```
kevin-dp: I would move all the query optimization stuff into a separate file. It could be an abstract class that is extended by the collection for instance. But i don't like how big this file is becoming and mixing concerns.

packages/db/src/collection.ts
```diff
@@ -1815,20 +2174,487 @@ export class CollectionImpl<
...
+        case `and`:
+          return this.canOptimizeAndExpression(expression)
+
+        case `or`:
+          return this.canOptimizeOrExpression(expression)
+
+        case `in`:
+          return this.canOptimizeInArrayExpression(expression)
+
+        default:
```
kevin-dp: No need for this default. We can just let evaluation continue and it will hit the return on L2343.

packages/db/src/collection.ts
```diff
@@ -1815,20 +2174,487 @@ export class CollectionImpl<
...
+    if (expression.type !== `func` || expression.args.length !== 2) {
+      return false
+    }
+
+    const leftArg = expression.args[0]!
+    const rightArg = expression.args[1]!
+
+    // Check if this is a simple field comparison: field op value
+
+    if (leftArg.type === `ref` && rightArg.type === `val`) {
```
kevin-dp: What about the other way around?
e.g. imagine someone wrote `lte(18, user.age)`.

packages/db/src/collection.ts
```diff
@@ -1815,20 +2174,487 @@ export class CollectionImpl<
...
+   * Checks if an AND expression can be optimized
+   * @private
+   */
+  private canOptimizeAndExpression(expression: BasicExpression): boolean {
+    if (expression.type !== `func` || expression.args.length < 2) {
+      return false
+    }
+
+    // All arguments must be optimizable
+    return expression.args.every((arg) => this.canOptimizeExpression(arg))
```
kevin-dp: Can't we optimize the parts that are optimizable to already gain some speedup there?

packages/db/src/collection.ts
```diff
@@ -1815,20 +2174,487 @@ export class CollectionImpl<
...
+
+    if (
+      fieldArg.type === `ref` &&
+      arrayArg.type === `val` &&
+      Array.isArray(arrayArg.value)
+    ) {
+      const fieldPath = fieldArg.path
+
+      // Find an index that matches this field
+      for (const index of this.indexes.values()) {
```
kevin-dp: I've seen this exact code also in `canOptimizeSimpleComparison`. Let's extract it to a function `findIndex`.

packages/db/src/collection.ts
```diff
@@ -1815,20 +2174,487 @@ export class CollectionImpl<
...
+    const results: Array<{ canOptimize: boolean; matchingKeys: Set<TKey> }> = []
+
+    for (const arg of expression.args) {
+      const result = this.optimizeQueryRecursive(arg)
+      results.push(result)
+    }
+
+    if (results.length > 0) {
+      // Intersect all matching keys (AND logic)
+      let intersectedKeys = results[0]!.matchingKeys
```
kevin-dp: Let's move all this set intersection logic to a utility function

packages/db/src/collection.ts
```diff
@@ -1815,20 +2174,487 @@ export class CollectionImpl<
...
+  ): (item: T) => boolean {
+    return (item: T): boolean => {
+      try {
+        // First try the RefProxy approach for query builder functions
+        const singleRowRefProxy = createSingleRowRefProxy<T>()
+        const whereExpression = whereCallback(singleRowRefProxy)
+        const expression = toExpression(whereExpression)
+        const evaluator = compileSingleRowExpression(expression)
+        const result = evaluator(item as Record<string, unknown>)
+        return Boolean(result)
```
kevin-dp: Why do we need to convert it to a boolean? Should't a WHERE clause always be a predicate (i.e. evaluate to a boolean value) ?

samwillis:
@kevin-dp I have significantly refactored this.

It now also supports lazy loading of an Index implementation as suggested by @KyleAMathews 

I also looked into alternatives to the `.splice()`, particularly using a chunked array, but an array is still faster.

I would prefer to get this in as it should be a big speedup, and we can optimise the index structure later.

packages/db/src/collection.ts
```diff
@@ -1263,6 +1303,157 @@ export class CollectionImpl<
...
+   */
+  public createIndex<
+    TResolver extends IndexResolver<TKey> = typeof OrderedIndex,
+  >(
+    indexCallback: (row: SingleRowRefProxy<T>) => any,
+    config: IndexOptions<TResolver> = {}
+  ): IndexProxy<TKey> {
+    this.validateCollectionUsable(`createIndex`)
+
+    const indexId = `${++this.indexCounter}`
```
kevin-dp: Why not keeping the index ID as a number?

packages/db/src/collection.ts
```diff
@@ -1903,20 +2094,142 @@ export class CollectionImpl<
...
+        }
+      }
+    } catch (error) {
+      // If anything goes wrong with the where clause, fall back to full scan
+      console.warn(
+        `Error processing where clause, falling back to full scan:`,
+        error
+      )
+
+      const filterFn = this.createFilterFunction(options.where)
```
kevin-dp, line 2177: This code with the for loop is duplicated from the else branch above. Let's created a nested function or an arrow function assigned to a variable that we call in both places.

kevin-dp: Note that it's only very similar to the code at the start of this method when there is no where clause. If we make an inner function that receives the filter function as argument, then all 3 places can call that function. The first one for missing where clause would simply call it with a function that always returns true in order not to filter anything out.

packages/db/src/collection.ts
```diff
@@ -1903,20 +2094,142 @@ export class CollectionImpl<
...
+  ): (item: T) => boolean {
+    return (item: T): boolean => {
+      try {
+        // First try the RefProxy approach for query builder functions
+        const singleRowRefProxy = createSingleRowRefProxy<T>()
+        const whereExpression = whereCallback(singleRowRefProxy)
+        const expression = toExpression(whereExpression)
+        const evaluator = compileSingleRowExpression(expression)
+        const result = evaluator(item as Record<string, unknown>)
+        // WHERE clauses should always evaluate to boolean predicates (Kevin's feedback)
```
kevin-dp: No need to refer to my feedback ;D

packages/db/src/collection.ts
```diff
@@ -1933,28 +2246,104 @@ export class CollectionImpl<
...
+      for (const change of changes) {
+        if (change.type === `insert`) {
+          // For inserts, check if the new value matches the filter
+          if (filterFn(change.value)) {
+            filteredChanges.push(change)
+          }
+        } else if (change.type === `update`) {
+          // For updates, we need to check both old and new values
+          const newValueMatches = filterFn(change.value)
+          const oldValueMatches = change.previousValue
```
kevin-dp: No need for ternary here, can simplify to `change.previousValue && filterFn(change.previousValue)`

packages/db/src/indexes/lazy-index.ts
```diff
@@ -0,0 +1,251 @@
...
+   * Get index statistics (throws if not ready)
+   */
+  getStats() {
+    return this.index.getStats()
+  }
+
+  /**
+   * Check if index matches a field path (available immediately)
+   */
+  matchesField(fieldPath: Array<string>): boolean {
```
kevin-dp, line 199: I think we can delegete this method to the index that is being proxied? Since there is already a `matchesField` method on the base index. So we could do: `return this.index.matchesField(fieldPath)`

packages/db/src/indexes/ordered-index.ts
```diff
@@ -0,0 +1,268 @@
...
+    `in`,
+  ])
+
+  // Internal data structures - private to hide implementation details
+  private orderedEntries: Array<[any, Set<TKey>]> = []
+  private valueMap = new Map<any, Set<TKey>>()
+  private indexedKeys = new Set<TKey>()
+  private compareFn: (a: any, b: any) => number = ascComparator
+
+  protected initialize(options?: OrderedIndexOptions): void {
```
kevin-dp, line 45: Why do we have an `initialize` method? This is something we could do it in the constructor. Right now the flow is: constructor -> super constructor -> initialize on subclass.

packages/db/src/indexes/ordered-index.ts
```diff
@@ -0,0 +1,268 @@
...
+  /**
+   * Adds a value to the index
+   */
+  add(key: TKey, item: any): void {
+    let indexedValue: any
+    try {
+      indexedValue = this.evaluateIndexExpression(item)
+    } catch (error) {
+      console.warn(`Failed to evaluate index expression for key ${key}:`, error)
+      return
```
kevin-dp: Shouldn't we throw an error instead of returning? Right now we return and the caller has no way of knowing something went wrong so it looks like the item was added.

packages/db/src/indexes/ordered-index.ts
```diff
@@ -0,0 +1,268 @@
...
+      const keySet = new Set<TKey>([key])
+      this.valueMap.set(indexedValue, keySet)
+
+      // Find correct position in ordered entries using binary search
+      const insertIndex = findInsertPosition(
+        this.orderedEntries,
+        indexedValue,
+        this.compareFn
+      )
+      this.orderedEntries.splice(insertIndex, 0, [indexedValue, keySet])
```
kevin-dp, line 77: Really think this is going to be problematic on large collections when JS is no longer able to optimize large arrays and every `add` will become very slow.

packages/db/src/indexes/ordered-index.ts
```diff
@@ -0,0 +1,268 @@
...
+   * Performs an equality lookup
+   */
+  equalityLookup(value: any): Set<TKey> {
+    return new Set(this.valueMap.get(value) ?? [])
+  }
+
+  /**
+   * Performs a range query
+   */
+  rangeQuery(operation: `gt` | `gte` | `lt` | `lte`, value: any): Set<TKey> {
```
kevin-dp: A more general approach to range queries would not take the `operation` but two (optional) values `from`, `to` and an option to indicate whether they should be inclusive or exclusive. That would also be easier to optimize queries later. Imagine the user writes a query: `WHERE a > 5 AND a < 10`. With the current `rangeQuery` you would need to do it in two steps: `set_intersect(rangeQuery('gt', 5), rangeQuery('lt', 10))` which is suboptimal since you would scan from 5 till the end for the first one and from the beginning till 10 for the 2nd one. But if you would support `from` and `to` you would only scan from 5 till 10.

packages/db/src/collection.ts
samwillis: (I may have pasted your comment into Cursor... ;-))

packages/db/src/indexes/lazy-index.ts
samwillis: The intention is that we can validate the matchesField before the index is loaded. However, this may want to change when we get to actually creating the other index types. I will leave as is for now to preserve the current intent.

packages/db/src/indexes/ordered-index.ts
samwillis: The intention is that in subclasses a user who creates a custom index implementation doesn't need to call `super()` with the constructor arguments. They only need to handle the options in their own `initialize` method. lets leave like this for now, and reassess when we create the other index types.

packages/db/src/indexes/ordered-index.ts
samwillis: Yep, changed!

packages/db/src/indexes/ordered-index.ts
samwillis: As discussed we will explore moving to a b-tree in a followup PR.

packages/db/src/indexes/ordered-index.ts
samwillis: Yep, good idea. I've fully optimised it in https://github.com/TanStack/db/pull/257/commits/4b5320fb44d8876bbd5b68faaffd26900ce150f3

PR #258, opened 2025-07-13
enable live queries to use indexes on collections

samwillis:
**stacked on #257**

This connects the optimised queries with the collections indexes, pushing where clauses that only touch a single collection to the subscription. The subscription then handles using the indexes if they are available.

Note that the design of this means that the query optimiser and compiler does not need to have knowledge of indexes, any where clause that only touches a single collection is pushed to the subscription rather than handled as a `filter` int he D2mini pipeline.


samwillis:
We need to add tests that validate what happens when a row that was filtered out/in based on an index value during inital run, then changes to be filtered in/out. I have a feeling there may be a bug where we need to do a little book keeping on the subscribe to know when rows were sent, and then know if it's an update/insert or deleted/update on the pipeline.

packages/db/src/collection.ts
```diff
@@ -2459,7 +2478,7 @@ export class CollectionImpl<
       const operation = expression.name as `eq` | `gt` | `gte` | `lt` | `lte`
 
       // Find an index that matches this field
-      for (const index of this.indexes.values()) {
+      for (const [, index] of this.indexes) {
```
kevin-dp: i much preferred how it was before

packages/db/src/query/compiler/expressions.ts
```diff
@@ -0,0 +1,103 @@
...
+        return expr.args.every((arg) =>
+          checkExpression(arg as BasicExpression<boolean>)
+        )
+      }
+      default:
+        return false
+    }
+  }
+
+  return checkExpression(whereClause)
```
kevin-dp: Why would we define an inner function for this?
I'm not sure i'm a find of the `switch` statement, this is equivalent:
```ts
const tpe = expr.type
if (tpe === `func`) {
  // Check if this function is supported
  if (!SUPPORTED_COLLECTION_FUNCS.has(expr.name)) {
    return false
  }
  // Recursively check all arguments
  return expr.args.every((arg) =>
    checkExpression(arg as BasicExpression<boolean>)
  )
}
return [`val`, `ref`].includes(tpe)
```

packages/db/src/query/compiler/expressions.ts
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * Converts a WHERE clause to BasicExpression format compatible with collection indexes.
+ * This function creates proper BasicExpression class instances that the collection
+ * index system can understand.
+ *
+ * @param whereClause - The WHERE clause to convert
+ * @param collectionAlias - The alias of the collection being filtered
+ * @returns The converted BasicExpression or null if conversion fails
+ */
+export function convertToBasicExpression(
```
kevin-dp, line 53: Same comments on this function. No need for an inner function + i'd prefer an if statement instead of a switch.

packages/db/src/query/live-query-collection.ts
```diff
@@ -179,10 +184,15 @@ export function liveQueryCollectionOptions<
...
     )
-    pipelineCache = compileQuery(
+
+    // Compile the query and get both pipeline and collection WHERE clauses
+    const compilationResult = compileQuery(
       query,
       inputsCache as Record<string, KeyedStream>
     )
+
+    pipelineCache = compilationResult.pipeline
```
kevin-dp: Why not destructuring it? :-)

packages/db/src/utils/comparison.ts
```diff
@@ -31,14 +47,42 @@ export const ascComparator = (a: any, b: any): number => {
...
-  // If at least one of the values is an object, convert to strings
-  const bothObjects = typeof a === `object` && typeof b === `object`
-  const notNull = a !== null && b !== null
-  if (bothObjects && notNull) {
-    return a.toString().localeCompare(b.toString())
+  // If at least one of the values is an object, use stable IDs for comparison
+  const aIsObject = typeof a === `object` && a !== null
+  const bIsObject = typeof b === `object` && b !== null
+
+  if (aIsObject || bIsObject) {
```
kevin-dp, line 54: Checking for `null` is actually not needed because at the very beginning of this function we already check for `null` and if `a` or `b` is null we immediately return.

packages/db/src/utils/comparison.ts
```diff
@@ -31,14 +47,42 @@ export const ascComparator = (a: any, b: any): number => {
...
+  }
+
+  // For primitive values, use direct comparison
+  try {
+    if (a < b) return -1
+    if (a > b) return 1
+    return 0
+  } catch {
+    // If comparison fails, fall back to string comparison
+    return safeStringify(a).localeCompare(safeStringify(b))
```
kevin-dp: I don't think comparison will ever throw on primitive values?

samwillis: I think there are a few edge cases were it can arround Symbols - and it could throw for objects but they never reach here now. I cant see a reason this would ever be used with a Symbol, so removing.

kevin-dp (approved):

PR #259, opened 2025-07-14
WIP Devtools

samwillis:
This is a WIP devtools implementation based on the Tanstack Router dev tools.

It uses a WeakMap and polling to allow collections to register themselves but still be GCed by the browser. When a collection is opened in the right hand panel and hard ref is made, but that is disposed of then it's closed.

More is needed:
- inspection of transactions
- view synced data
- filtering of root collections from live queries
- live query timing
- the actual query of a live query 

Note that this branch currently has a copy of the router and query dev tools under `/devtools-reference/`, this is to enable AI agents to use them as a reference while implementing it. Buch of the groundwork was done with a Curser background agent, with Claude-4-sonnet Max.

https://github.com/user-attachments/assets/3994e52e-8fa1-4b4e-9d37-195e4b6ac9c4


KyleAMathews:
Closing this as we're not planning on shipping devtools for a bit

PR #265, opened 2025-07-15
[Electric] Immediately do commit if the shape stream errors (e.g. auth, server down, etc.) so preload will finish

KyleAMathews:


samwillis (approved):
LGTM

PR #271, opened 2025-07-16
Fix iterator-based change tracking in proxy system

KyleAMathews:
## Summary

This fixes several critical issues with iterator-based change tracking for Maps and Sets in the proxy system:

- **Map.entries()** now correctly updates actual Map entries instead of creating duplicate keys
- **Map.values()** now tracks back to original Map keys using value-to-key mapping instead of using symbol placeholders  
- **Set iterators** now properly replace objects in Set when modified instead of creating symbol-keyed entries
- **forEach()** methods continue to work correctly

## Technical Details

The implementation now uses a sophisticated parent-child tracking system with specialized `updateMap` and `updateSet` functions to ensure that changes made to objects accessed through iterators are properly attributed to the correct collection entries.

This brings the proxy system in line with how mature libraries like Immer handle iterator-based change tracking, using method interception rather than trying to proxy all property access.

## Tests Added

Added 5 comprehensive new tests covering:
- Multiple modifications via different iterators
- Nested object modifications  
- Set with duplicate objects
- Reverting iterator changes
- Mixed Map and Set operations

## Test Results

- ✅ All 83 proxy tests passing
- ✅ All 9 iterator tests passing (4 fixed + 5 new)
- ✅ No lint errors or warnings
- ✅ 88.75% code coverage on proxy.ts

### DIFF
--- a/.changeset/shiny-grapes-count.md
+++ b/.changeset/shiny-grapes-count.md
@@ -1,0 +1,16 @@
+---
+"@tanstack/db": patch
+---
+
+Fix iterator-based change tracking in proxy system
+
+This fixes several issues with iterator-based change tracking for Maps and Sets:
+
+- **Map.entries()** now correctly updates actual Map entries instead of creating duplicate keys
+- **Map.values()** now tracks back to original Map keys using value-to-key mapping instead of using symbol placeholders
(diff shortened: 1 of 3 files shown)


samwillis (approved):
Tests look good and it passes.
:shipit:

PR #283, opened 2025-07-18
refactor: vue

teleskop150750:


teleskop150750:
It makes sense to reduce the possible props options.

KyleAMathews:
Yeah using vue's Auto dependency tracking makes sense to me. @samwillis thoughts?

samwillis:
Absolutely, this was very much a first pass based on the react one. With Vues reactivity model the deps are not needed.

I think it's also worth exploring if there is a better model with Vue where we mutate a `reactive` state object rather than returning the internal derived state. It should enable more fine grade reactivity.

KyleAMathews:
@samwillis would that be a non-breaking change? Given this is a breaking change, I'd like this in sooner than later. Fine-grained reactivity would be great though.

KyleAMathews:
@teleskop150750 how are you feeling about this PR? Anything left you want to do?

targetlucked69:
why did yall go with getters here instead of refs for return properties? Most of Vue composables/hooks excepts Refs as return properties... while this works it feels really weird

teleskop150750:
@KyleAMathews

Look at this example for different frameworks.

https://github.com/fabian-hiller/formisch/blob/main/frameworks/solid/src/primitives/useFieldArray/useFieldArray.ts

We can also decide which API to use. The downside of the example is that destructuring does not work.
When passed to other hooks later, this again requires wrapping in a function.

@targetlucked69 Vue users often use Ref for returning and passing, but this does not mean that this way is always good.

In this PR I swapped ref, reactive, computed. Using getter helps to apply such changes without changing the API

teleskop150750:
@targetlucked69

Watch lecture 26:00 - 32:00

https://www.youtube.com/live/J0O69dGljx8

@KyleAMathews discussion api getter

1:59:00 I don't remember the exact time now

https://www.youtube.com/live/J0O69dGljx8

targetlucked69:
I feel like this is entering premature optimization area

KyleAMathews:
@targetlucked69 how so? This isn't a hugely complex integration so it seems like we can anticipate necessary optimizations upfront? What would we need to learn before doing more optimizations?

teleskop150750:
> I feel like this is entering premature optimization area

Why? this PR does nothing except: 
1) remove deps because Vue doesn't need them 
2) simplify the creation of collection 
3) unify the output format, which can be replaced with return functions if desired

PR #285, opened 2025-07-18
Add comprehensive error handling documentation and improve collection auto-restart behavior

KyleAMathews:
## Summary

- Added comprehensive error handling documentation covering all TanStack DB error scenarios
- Fixed collection operations to automatically restart from cleaned-up state (improves DX)
- Updated navigation to include Error Handling guide in the docs

### Documentation
- **New `docs/error-handling.md`** - Comprehensive guide covering:
  - `SchemaValidationError` with validation issues and paths
  - Collection status states and error handling
  - Transaction error handling and automatic rollbacks
  - Collection operation errors (missing handlers, duplicate keys, etc.)
  - Sync error handling for different collection types (Query, Electric, etc.)
  - Error recovery patterns and best practices
  - Complete examples with real-world error handling

### Code Improvements
- **Fixed collection auto-restart behavior** - Collections in `cleaned-up` state now automatically restart when operations are called, matching the behavior of other collection access patterns
- **Updated navigation** - Added Error Handling to the Guides section

## Test plan

- [x] Documentation follows existing conventions and style
- [x] All error handling examples use real APIs from the codebase
- [x] Collection auto-restart behavior tested and working
- [x] Navigation properly includes the new error handling page


packages/db/src/collection.ts
```diff
@@ -328,10 +328,9 @@ export class CollectionImpl<
             `Try calling cleanup() and restarting the collection.`
         )
       case `cleaned-up`:
-        throw new Error(
```
KyleAMathews, line 331: cc @samwillis this didn't seem right to error when mutating a cleaned-up collection as that seems like it'd often happen e.g. you pull in collection to mutate it but you're not necessarily viewing it. 

Though... now as I'm writing this, it also does make sense as mutating a cleaned-up collection (i.e. one not syncing) dramatically ups the odds that you'll mutate an old version of data thereby overwriting newer changes 🤔 

So perhaps inserts are always safe but updates can't be done to a cleaned-up collection?

KyleAMathews (commented):
ss

docs/error-handling.md
```diff
@@ -0,0 +1,494 @@
...
+
+### Invalid Collection State
+
+Collections in an `error` state cannot perform operations and must be manually recovered:
+
+```ts
+try {
+  todoCollection.insert(newTodo)
+} catch (error) {
+  if (error.message.includes("collection is in error state")) {
```
samwillis, line 174: This patten feels a little flaky if the message changes. A check for a prop or error type may be better?

I wander is an `error.reason` which is a fixed set of strings `'collection-error-sate' | 'key-conflict' | 'no-insert-handler' | 'validation-error'`?

Or a distinct Error subclass for each type?

packages/db/src/collection.ts
samwillis: Mutations are always rebased over synced state, so it's probably fine in a cleaned up state, we are leaning in to server side validation - the user must do that. I think this change is correct.

samwillis (approved):
This looks good, the one thing I would consider doing (which could be in a followup) is to make it easer to determine the reason for an error on a mutation. See the other comment.

packages/db/src/collection.ts
KyleAMathews: 👍

docs/error-handling.md
KyleAMathews: yeah — agree that named errors would be good — easy follow-up for AI to flip through

PR #290, opened 2025-07-20
minor: simplify example app to draw more focus to the data operations

ignatz:
This is a collection of minor tweaks to have less clutter and make the data manipulations more obvious on skim:

* move color arithmetic out
* move navigation to separate component
* reduce style clutter and minor adjustments
* make the select/unselect all button and footer always visible
* streamline dom hierarchy.


KyleAMathews (approved):
nice, thanks!

PR #291, opened 2025-07-20
fix join expression order bug and add validation

### LINKED ISSUE #277: Add support for SQL-style self-joins on the same table
Self-joins currently crash the process.

**Minimal reproduction:**
```ts
useLiveQuery(
  q => q.from({ users })
    .join(
      { parentUsers: users },
      ({ users, parentUsers }) => eq(users.parentId, parentUsers.id),
      'inner',
    )
)
```

**Expected:**
returns joined rows.

**Actual:**
memory exhaustion / crash (tab becomes unresponsive without showing error messages in the console and with memory usage climbing).

samwillis:
fixes #277

Currently the order of the references in the join `eq(left.prop, right.prop)` was required but not validated. This changes it to work either way (you can specify them in reverse) and validates and rejects other mistakes.

### DIFF
--- a/packages/db/src/query/compiler/joins.ts
+++ b/packages/db/src/query/compiler/joins.ts
@@ -163,7 +163,7 @@ function analyzeJoinExpressions(
     return { mainExpr: right, joinedExpr: left }
   }
 
-  // If both expressions refer to the same table, this is an invalid join
+  // If both expressions refer to the same alias, this is an invalid join
   if (leftTableAlias === rightTableAlias) {
     throw new Error(
       `Invalid join condition: both expressions refer to the same table "${leftTableAlias}"`
@@ -203,10 +203,8 @@ function getTableAliasFromExpression(expr: BasicExpression): string | null {
       // If all arguments refer to the same table, return that table alias
       return tableAliases.size === 1 ? Array.from(tableAliases)[0]! : null
     }
-    case `val`:
-      // Values don't reference any table
-      return null
     default:
+      // Values (type='val') don't reference any table
       return null
   }
 }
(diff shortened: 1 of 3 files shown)


KyleAMathews (approved):
Oh nice! I didn't like having to remember which order to put things in haha.

packages/db/src/query/compiler/joins.ts
```diff
@@ -121,6 +134,83 @@ function processJoin(
...
+      for (const arg of expr.args) {
+        const alias = getTableAliasFromExpression(arg)
+        if (alias) {
+          tableAliases.add(alias)
+        }
+      }
+      // If all arguments refer to the same table, return that table alias
+      return tableAliases.size === 1 ? Array.from(tableAliases)[0]! : null
+    }
+    case `val`:
```
kevin-dp: Could remove this case in favor of the default case.

packages/db/src/query/compiler/joins.ts
```diff
@@ -121,6 +134,83 @@ function processJoin(
...
+
+  // If left expression refers to joined table and right refers to main table, swap them
+  if (
+    leftTableAlias === joinedTableAlias &&
+    rightTableAlias === mainTableAlias
+  ) {
+    return { mainExpr: right, joinedExpr: left }
+  }
+
+  // If both expressions refer to the same table, this is an invalid join
```
kevin-dp: This comment got me wondering if it was not possible to self-join. But in the tests i see that it is possible but the aliases need to be different. Perhaps we can improve this comment by saying:
> if both expressions refer to the same **alias**, this is an invalid join

kevin-dp (approved):

PR #292, opened 2025-07-20
automatically create indexes when a collection is queried 

samwillis:
**stacked on #258**

This adds a new config option `autoIndex: 'eager' | 'off'` to the collection options, and sets its default to `eager`. We may in future add additional modes where are lazy, or asynchronous that perform a level of analysis first.

When a collection if queried, if there is not an index on a column with a basic `eq`, `gt`, `gte`, `lt`, `lte`, or `in` expression, including multiple `and`ed together, then it will synchronously add the index before executing the query.


packages/db/src/collection.ts
```diff
@@ -1360,12 +1365,18 @@ export class CollectionImpl<
...
     this.lazyIndexes.set(indexId, lazyIndex)
 
-    // For synchronous constructors (classes), resolve immediately
-    // For async loaders, wait for collection to be ready
-    if (typeof resolver === `function` && resolver.prototype) {
-      // This is a constructor - resolve immediately and synchronously
+    // For OrderedIndex, resolve immediately and synchronously
+    if ((resolver as unknown) === OrderedIndex) {
+      try {
+        const resolvedIndex = lazyIndex.getResolved()
```
kevin-dp, line 1370: The code to resolve the index and set it in `this.resolvedIndexes` is the same as the one in the `else if` below. Let's unify both branches.

packages/db/src/indexes/auto-index.ts
```diff
@@ -0,0 +1,108 @@
...
+  ) {
+    return
+  }
+
+  // Extract all indexable expressions and create indexes for them
+  const indexableExpressions = extractIndexableExpressions(expression)
+
+  for (const { fieldName, fieldPath } of indexableExpressions) {
+    // Check if we already have an index for this field
+    const existingIndex = Array.from(collection.indexes.values()).find(
```
kevin-dp, line 37: Potentially expensive operation here as we have to loop over all indexes to find the one that matches `fieldPath`. Could the index not keep a map from `fieldPath` to the index such that this becomes an O(1) lookup?

kevin-dp (approved):
LGTM. Left some minor comments.

PR #293, opened 2025-07-21
Delay initializing periodic cleanup until sync has actually started.

ignatz:
Also avoid use of WeakRef by explicitly tying the cleanup lifetime to the livelihood of the reader.


KyleAMathews (approved):
Thanks!

PR #302, opened 2025-07-22
Index that uses B+ tree

kevin-dp:
Currently, we have an `OrderedIndex` that internally keeps an ordered array of indexed values. While this allows for efficient range queries, it introduces significant overhead on inserts and deletes because inserting/deleting in the array is in linear time.

This PR introduces a `BTreeIndex` which internally keeps a B+ tree of indexed values. This still allows for efficient range queries but also makes insertions/deletions efficient (logarithmic time). I think it makes the `OrderedIndex` completely obsolete and we should remove it in favor of `BTreeIndex`.

Note that the B+ tree is only needed for range queries but some collections might never use range queries. So it would be good to introduce also a `HashIndex` in a follow up PR which only keeps a map of indexed values to PKs. That will reduce bookkeeping overhead on inserts/deletes further from logarithmic to constant time.

### DIFF
--- a/.changeset/better-owls-judge.md
+++ b/.changeset/better-owls-judge.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+Remove OrderedIndex in favor of more efficient BTree index.


kevin-dp:
Some quick microbenchmarks confirm that the B+ tree index is faster for inserts/updates/deletes and similar for range queries:
```
==== INDEX BUILDING ====
OrderedIndex: 2004.6950840000002ms for 100000 users
BTreeIndex: 63.54179199999999ms for 100000 users

==== INDEX INSERTION ====
OrderedIndex 0.056916999999884865ms per insert
BTreeIndex 0.004040999999688211ms per insert

==== INDEX DELETION ====
OrderedIndex 0.07920799999919836ms per delete
BTreeIndex 0.01920799999970768ms per delete

==== INDEX UPDATE ====
OrderedIndex 0.09737500000028376ms per update
BTreeIndex 0.010333000000173342ms per update

==== INDEX RANGE QUERY ====
OrderedIndex range query: 10.354792000000089ms per query
BTreeIndex range query: 11.601208000000042ms per query
```
All operations were performed on an index containing 100K rows.

PR #308, opened 2025-07-23
fix: prevent user actions from being blocked by event filtering optimizations

KyleAMathews:
## Summary
- Fixes issue where clicking multiple todo checkboxes causes them to stop responding when optimistic updates back up with slow Materialize responses
- Modifies event filtering logic in `collection.ts` to distinguish between user-triggered and sync-triggered recomputations
- Adds `triggeredByUserAction` parameter to `recomputeOptimisticState()` method
- User actions now bypass both sync status filtering and complex filtering to prevent UI blocking
- Includes test case that reproduces and verifies the fix

## Test plan
- [x] Added comprehensive test that reproduces the original bug scenario
- [x] Test verifies that rapid user actions on recently synced keys are not blocked
- [x] Existing tests continue to pass
- [x] Manual testing in browser confirms fix resolves the issue

### DIFF
--- a/packages/db/src/collection.ts
+++ b/packages/db/src/collection.ts
@@ -687,7 +687,9 @@ export class CollectionImpl<
   /**
    * Recompute optimistic state from active transactions
    */
-  private recomputeOptimisticState(): void {
+  private recomputeOptimisticState(
+    triggeredByUserAction: boolean = false
+  ): void {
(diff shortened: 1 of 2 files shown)


samwillis (approved):
LGTM, let's get this out but keep an eye out for side effects.

KyleAMathews:
Side effects like excessive renders?

PR #309, opened 2025-07-23
fix: ensure LiveQueryCollection markReady is called when source collections have no data

### LINKED ISSUE #306: LiveQueryCollection preload doesn't resolve when source collection is empty
I have been trying to narrow down an issue relating to `preload` calls never resolving and managed to narrow it down to the following test case:

https://stackblitz.com/edit/tanstack-db-km8idwxc?file=src%2FApp.tsx

Any advice on how to try to troubleshoot this?

KyleAMathews:
## Summary

Fixed an issue where `LiveQueryCollection.preload()` would hang indefinitely when source collections call `markReady()` without data changes (e.g., when queryFn returns empty array from query-db-collection).

## Root Cause

The bug occurred when source collections became ready but had no data to emit:
1. LiveQueryCollection waits for source collections to reach `ready` or `initialCommit` status
2. Collections that call `markReady()` without data changes don't emit change events  
3. LiveQueryCollection never receives notification that these collections are ready
4. This was especially problematic with WHERE clause optimized subscriptions

## Solution

Implemented a proper event-based solution with two parts:

1. **Collection Level**: Modified `markReady()` in `collection.ts` to emit empty change events when a collection becomes ready with no data
2. **Filtering Level**: Updated `createFilteredCallback` in `change-events.ts` to distinguish between empty filtered results (normal) and empty original events (ready signals), ensuring WHERE clause subscriptions pass through ready signals

## Test Coverage

Added comprehensive tests that reproduce the bug:
- Source collections that only call `markReady()` without `begin()`/`commit()`
- Both regular and WHERE clause optimized LiveQueryCollections
- All tests now pass correctly

## Files Changed

- `packages/db/src/collection.ts` - Added empty ready event emission
- `packages/db/src/change-events.ts` - Fixed filtered callback logic for ready signals
- `packages/db/tests/query/live-query-collection.test.ts` - Added reproduction tests

Fixes #306

### DIFF
--- a/.changeset/ninety-parts-check.md
+++ b/.changeset/ninety-parts-check.md
@@ -5,4 +5,9 @@
 
 Fix LiveQueryCollection hanging when source collections have no data
 
-Fixed an issue where `LiveQueryCollection.preload()` would hang indefinitely when source collections call `markReady()` without data changes (e.g., when queryFn returns empty array). The LiveQueryCollection now properly detects when all source collections are ready, even without data changes, by polling collection status until all dependencies are ready.
+Fixed an issue where `LiveQueryCollection.preload()` would hang indefinitely when source collections call `markReady()` without data changes (e.g., when queryFn returns empty array). 
+
+The fix implements a proper event-based solution:
+- Collections now emit empty change events when becoming ready with no data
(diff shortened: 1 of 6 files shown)


samwillis (approved):
LGTM!

Issue #229, opened 2025-07-05
orderBy on dates works incorrectly when mixing formats


ruudandriessen:
When using dates - order by does not work consistently when formats are mixed. For example, the following does not sort correctly:

```ts
 ['2025-07-01 16:47:13', '2025-07-02 22:00:00', null, null, '2025-07-12 09:06:56']
```

Where the expectation would be that it is sorted like:
```ts
[
    "2025-07-01 16:47:13",
    "2025-07-02 22:00:00",
    "2025-07-12 07:06:56",
    null,
    null
]
```

On versions
```ts
    "@tanstack/db-collections": "^0.0.18",
    "@tanstack/react-db": "^0.0.14",
```

Related issue: https://github.com/TanStack/db/issues/123

KyleAMathews:
Are these dates or strings?

This seems related

https://github.com/TanStack/db/issues/212

ruudandriessen:
> Are these dates or strings?
> 
> This seems related
> 
> [#212](https://github.com/TanStack/db/issues/212)

I tried both. Neither of them work.  Looking at that related thread. Does that imply we don't deal with nulls at all? We just leave them wherever? Because the dates *assuming I ignore the nulls* are sorted correctly.

LE: The workaround there - splitting it into two queries where one filters by date != null and the other by date == null works. But it's not ideal.

Issue #212, opened 2025-06-27
Feature request: nulls first and custom comparator functions.


thruflo:
I have a live query like this to get all the user's threads:

```ts
const { data: threads } = useLiveQuery(
    (query) =>
      query
        .from({ t: threadCollection })
        .join({
          type: `inner`,
          from: { m: membershipCollection },
          on: [`@t.id`, `=`, `@m.thread_id`],
        })
        .select('@t.id', '@t.name')
        .where('@m.user_id', '=', currentUserId)
        .orderBy({ '@inserted_at': 'desc' }),
    [currentUserId]
  )
```

Trouble is, when I insert optimistic records without an inserted_at, they sort last. Now, I can fix this by jigging my schema and I can do different queries for inserted_at / no-inserted_at. But ideally I'd like to be able to control the sorting more expressively.

So one idea would be:

```ts
.orderBy({ 
  '@t.inserted_at': { direction: 'desc', nulls: 'first' }
})
```

Another would be to allow me to provide the comparator function.

thruflo:
Example workaround:

```ts
const { collection: baseQuery } = useLiveQuery(query =>
  query
    .from({ t: threadCollection })
    .join({
      type: `inner`,
      from: { m: membershipCollection },
      on: [`@t.id`, `=`, `@m.thread_id`],
    })
    .select('@t.id', '@t.name', '@t.inserted_at')
    .where('@m.user_id', '=', currentUserId),
  [currentUserId]
)

const { data: nonSyncedThreads } = useLiveQuery(query =>
  query
    .from({ t: baseQuery })
    .where(({ t }) => t.inserted_at === undefined)
    .select('@t.id', '@t.name')
    .orderBy({ '@t.name': 'desc' })
)

const { data: syncedThreads } = useLiveQuery(query =>
  query
    .from({ t: baseQuery })
    .where(({ t }) => t.inserted_at !== undefined)
    .select('@t.id', '@t.name')
    .orderBy({ '@t.inserted_at': 'desc' })
)

const threads = nonSyncedThreads.concat(syncedThreads)
```

KyleAMathews:
Curious why you're not adding `inserted_at` to the optimistic record? BTW, this is a very nice PR to improve the experience there as you can add defaults to your schema to auto-insert stuff like that https://github.com/TanStack/db/pull/209

thruflo:
Yup, that’s another solution.

[...]

PR #314, opened 2025-07-24
Configuration options for sortBy

kevin-dp:
Fixes https://github.com/TanStack/db/issues/212 and fixes https://github.com/TanStack/db/issues/229 by adding a configuration argument to `orderBy`:
```ts
orderBy(
  callback: OrderByCallback<TContext>,
  options: OrderByDirection | OrderByOptions = `asc`
)

type OrderByOptions = {
  direction?: OrderByDirection
  nulls?: `first` | `last`
} & StringSortOpts

type StringSortOpts =
  | {
      stringSort?: `lexical`
    }
  | {
      stringSort?: `locale`
      locale?: string
      localeOptions?: object
    }
```

### DIFF
--- a/packages/db/src/query/builder/index.ts
+++ b/packages/db/src/query/builder/index.ts
@@ -478,7 +478,8 @@ export class BaseQueryBuilder<TContext extends Context = Context> {
    */
   orderBy(
     callback: OrderByCallback<TContext>,
-    direction: OrderByDirection = `asc`
+    direction: OrderByDirection = `asc`,
+    nulls: `first` | `last` = `first`
   ): QueryBuilder<TContext> {
     const aliases = this._getCurrentAliases()
     const refProxy = createRefProxy(aliases) as RefProxyForContext<TContext>
@@ -488,6 +489,7 @@ export class BaseQueryBuilder<TContext extends Context = Context> {
     const orderByClause: OrderByClause = {
       expression: toExpression(result),
       direction,
+      nulls,
     }
 
     const existingOrderBy: OrderBy = this.query.orderBy || []
--- a/packages/db/src/query/builder/index.ts
+++ b/packages/db/src/query/builder/index.ts
@@ -19,6 +19,7 @@ import type {
   QueryIR,
 } from "../ir.js"
 import type {
+  CompareOptions,
   Context,
   GroupByCallback,
   JoinOnCallback,
@@ -485,22 +486,25 @@ export class BaseQueryBuilder<TContext extends Context = Context> {
     const refProxy = createRefProxy(aliases) as RefProxyForContext<TContext>
     const result = callback(refProxy)
 
-    const opts: Required<OrderByOptions> =
+    const opts: CompareOptions =
       typeof options === `string`
-        ? { direction: options, nulls: `first` }
+        ? { direction: options, nulls: `first`, stringSort: `locale` }
         : {
-            direction:
-              typeof options === `string`
-                ? options
-                : (options.direction ?? `asc`),
+            direction: options.direction ?? `asc`,
             nulls: options.nulls ?? `first`,
+            stringSort: options.stringSort ?? `locale`,
+            locale:
+              options.stringSort === `locale` ? options.locale : undefined,
+            localeOptions:
+              options.stringSort === `locale`
+                ? options.localeOptions
+                : undefined,
           }
 
     // Create the new OrderBy structure with expression and direction
     const orderByClause: OrderByClause = {
       expression: toExpression(result),
-      direction: opts.direction,
-      nulls: opts.nulls,
+      compareOptions: opts,
     }
 
     const existingOrderBy: OrderBy = this.query.orderBy || []
--- a/packages/db/src/query/builder/index.ts
+++ b/packages/db/src/query/builder/index.ts
@@ -25,6 +25,7 @@ import type {
   MergeContext,
   MergeContextWithJoinType,
   OrderByCallback,
+  OrderByOptions,
   RefProxyForContext,
   ResultTypeFromSelect,
   SchemaFromSource,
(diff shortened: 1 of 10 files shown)


asabil:
Would it make sense to mimic PostgreSQL's default ordering?

> By default, null values sort as if larger than any non-null value; that is, `NULLS FIRST` is the default for `DESC` order, and `NULLS LAST` otherwise.

kevin-dp:
> Would it make sense to mimic PostgreSQL's default ordering?
> 
> > By default, null values sort as if larger than any non-null value; that is, `NULLS FIRST` is the default for `DESC` order, and `NULLS LAST` otherwise.

I have been thinking about this. I didn't do that because that would be a breaking change, but we could certainly do that if we want. @KyleAMathews @samwillis any opinions on this?

samwillis:
On the api design, I'm tempted to not add it as a third arg, but make the second arg either a string for direction, or an object for extended options:

```ts

interface OrderByOptions {
  direction: OrderByDirection = `asc`
  nulls: `first` | `last` = `first`
}

orderBy(
  callback: OrderByCallback<TContext>,
  directionOrOptions: OrderByDirection | OrderByOptions  = `asc`
)
```

This makes it much more clear what the options are, and makes it mush easer to extend to option in future if we want to.

@kevin-dp what do you think?

kevin-dp:
@samwillis agreed. Cleaner with a configuration object. I'm not sure about the union type `OrderByDirection | OrderByOptions` though. It's probably fine because most of the time one would only specify asc/desc order without specifying for nulls?

KyleAMathews:
Yeah I like avoiding a third arg too. Union seems fine to me as yeah, 99% of time people will just specify the direction.

packages/db/src/query/builder/index.ts
```diff
@@ -478,18 +479,28 @@ export class BaseQueryBuilder<TContext extends Context = Context> {
...
     const aliases = this._getCurrentAliases()
     const refProxy = createRefProxy(aliases) as RefProxyForContext<TContext>
     const result = callback(refProxy)
 
+    const opts: Required<OrderByOptions> =
+      typeof options === `string`
+        ? { direction: options, nulls: `first` }
+        : {
+            direction:
+              typeof options === `string`
```
thruflo: Presumably this can never be true at this point in the ternary clause.

thruflo:
N.b.: docs https://tanstack.com/db/latest/docs/live-queries#order-by-limit-and-offset

thruflo:
Also, whilst we're considering the orderBy signature https://github.com/TanStack/db/issues/316

thruflo (changes requested):
Just actually testing this with my app. I get an error because my schema is an `optional()`, not `nullable()`.

Using this branch, I updated my code to:

```ts
const { data: events } = useLiveQuery(
     (query) => (
      query
        .from({ event: eventCollection })
        .innerJoin(
          { user: userCollection },
          ({ event, user }) => eq(user.id, event.user_id)
        )
        .orderBy(
          ({ result }) => result.inserted_at
          { direction: 'asc', nulls: 'last' }
        )
        .select(({ event, user }) => ({
          data: event.data,
          id: event.id,
          inserted_at: event.inserted_at,
          type: event.type,
          user_id: user.id,
          user_name: user.name,
          user_type: user.type
        }))
        .where(({ event }) => eq(event.thread_id, threadId))
    ),
    [threadId]
  )
```

I get this error:

```
react-dom-client.production.js:5788 TypeError: Cannot read properties of undefined (reading 'inserted_at')
    at query.from.innerJoin.orderBy.direction (ChatArea.tsx:82:34)
    at BaseQueryBuilder.orderBy (index.js:349:20)
    at ChatArea.tsx:81:10
    at buildQuery (index.js:582:18)
    at liveQueryCollectionOptions (live-query-collection.js:9:54)
    at createLiveQueryCollection (live-query-collection.js:200:21)
    at useLiveQuery (useLiveQuery.js:16:33)
    at ChatArea (ChatArea.tsx:73:28)
```

I'm not sure if we should:

1. handle undefined fields / zod `foo.optional()`s
2. not handle them and require the property but allow `foo.nullable()`s

I think that the ability to support both would be ideal as people may well use optionals in their schema definitions. If not then we need to document the limitation.

packages/db/src/query/builder/index.ts
kevin-dp: oops, indeed. Good catch!

kevin-dp:
@thruflo i'll open an issue for ordering on an optional column. It's a separate issue that's best addressed with a separate PR.

kevin-dp:
> Would it make sense to mimic PostgreSQL's default ordering?
> 
> > By default, null values sort as if larger than any non-null value; that is, `NULLS FIRST` is the default for `DESC` order, and `NULLS LAST` otherwise.

@asabil Since TanStack DB is backend-agnostic we decided to stick to JS semantics wrt how null values sort.

kevin-dp:
@thruflo I think the problem is in your callback itself:
```ts
({ event }) => event.inserted_at
```
You're trying to access `inserted_at` but `event` is undefined. That's throwing the error you're seeing. Can you try replacing it by this:
```ts
({ event }) => event?.inserted_at
```

thruflo:
Just trying to wrap my head around how event could be `undefined`?

kevin-dp:
> Just trying to wrap my head around how event could be `undefined`?

@thruflo could you console.log it just to confirm that `event` is undefined. Could be that we try to access `inserted_at` somewhere internally and that the error comes from there, but from the stack trace i got the impression it is the `event` that is undefined.

thruflo:
Ok, I'm running the same code against this branch and it's now working, i.e.: I can use the inserted_at in the orderBy fine.

thruflo (approved):

PR #323, opened 2025-07-28
[WIP]: Add Firebase integration

KyleAMathews:


stevecastaneda:
https://github.com/TanStack/db/blob/8296f8ec0668ccf49e801394226ed1570b696b00/packages/firebase-db-collection/src/firestore.ts#L108

@KyleAMathews  What are your thoughts about changing this parameter to `collectionPath`?  I think collection name implies the name of the collection when it also accepts a path.  eg. sub-collections

devstojko:
Hello there,
  
  I took this in progress code and added locally to my project because I need it for firestore and I notice issue that is described below.
  
  The firebaseCollectionOptions function was causing type inference issues when used with
  createCollection from @tanstack/react-db. Specifically, when using the collection in queries with
  useLiveQuery, TypeScript was unable to infer the correct types for the collection items, resulting in
  RefProxyFor<never> always.
  
  The issue stemmed from a type signature mismatch between firebaseCollectionOptions and what
  createCollection expects:

  1. firebaseCollectionOptions was returning: CollectionConfig<TItem, TKey> & { utils: 
  FirebaseCollectionUtils }
  2. createCollection expects: CollectionConfig<TItem> & { utils: ... }

  The key difference is that CollectionConfig<TItem> uses a default key type of string | number, while
  CollectionConfig<TItem, TKey> uses a generic TKey. This mismatch prevented TypeScript from properly
  inferring the item types when the collection was used in queries.
  
  firebaseCollectionOptions with the pattern used by queryCollectionOptions (which works
  correctly). The changes include:

  1. Updated return type:
  // Before
`  ): CollectionConfig<TItem, TKey> & { utils: FirebaseCollectionUtils } {`

  // After
`  ): CollectionConfig<TItem> & { utils: FirebaseCollectionUtils } {`
  2. Updated sync configuration:
  // Before
`  const sync: SyncConfig<TItem, TKey> = {`

  // After
`  const sync: SyncConfig<TItem> = {`
  3. Updated mutation function signatures:
  // Before
`  onInsert: async (params: InsertMutationFnParams<TItem, TKey>): Promise<Array<TKey>> => {`

  // After
`  onInsert: async (params: InsertMutationFnParams<TItem>): Promise<Array<string | number>> => {`
  4. Updated getKey function:
  // Before
`  const getKey = config.getKey || ((item: TItem) => (item as any).id as TKey)`

  // After
`  const getKey = config.getKey || ((item: TItem) => (item as any).id as string | number)`

With these changes when I use firebaseCollectionOptions inside useLiveQuery combining with leftJoin from another query-db-collection types are inferred properly .

KyleAMathews:
@devstojko yeah it's pretty rough atm. I'm about to leave on parental leave so if you want to pick up taking it to production-ready, feel free to take it over!

codercatdev:
oh this is exciting!

BrianGrug:
> @devstojko yeah it's pretty rough atm. I'm about to leave on parental leave so if you want to pick up taking it to production-ready, feel free to take it over!

Congratulations!! This is actually the last thing I'm looking for for my project! What needs to be finished here, I may be able to pick it up.

KyleAMathews:
@BrianGrug that'd be amazing! Tbh I don't entirely remember haha. I think I stopped around when I was getting the tests running but you'd need to look around

PR #329, opened 2025-07-29
docs: add type-doc and update docs config

KevinVandy:


samwillis (approved):
This is awesome, thanks @KevinVandy 

I made a couple of small tweaks and will merge now.

Will follow up with a PR with links from the guides through to the generated docs.

PR #335, opened 2025-07-30
Optimize joins to use index when possible

kevin-dp:
This PR optimizes joins based on available indexes.

For a left join we always have to iterate over the left collection, but we don't need to iterate over the entire right collection. Based on the rows in the left collection we can lookup the rows that match in the right collection (based on the key we're joining on). That lookup is efficient if there's an index on the join key. We can do the same for right joins. For inner joins, we can loop over the smallest collection and lookup matching rows in the bigger collection such that we don't have to loop over the bigger collection.

Here's a concrete example, imagine we're left joining a `Comments` collection with a `Users` collection on `Comments.user_id = Users.id`. And imagine we have an index on `Users.id`. Now, we can loop over all `Comments` and for each comment we can lookup it's `user_id` in the index for `Users.id`. This will give us the corresponding user that we need to join with the comment, without having to loop over the entire `Users` collection.

### Implementation Overview

We don't actually have to modify the existing D2 `join` operator to do this. The `join` operator takes two streams, a stream for the left collection and a stream for the right collection. For left/right/inner joins we only need to loop over one of the streams and we don't need to loop over the entire other stream. Therefore, the idea is to modify the streams such that there is an active stream and a lazy stream. We process the entire active stream and use it to dynamically populate the lazy stream. This is depicted in the following diagram:

shapes at 25-08-05 09 34 45

The diagram above depicts a left-join for comments with users (the example from before). The comments are filtered (e.g. to only get the comments for a certain issue). Then, we want to left-join it with users. To do this, we add a `tap` operator after the `filter` and before the `join`. This operator doesn't modify the stream, but will for every row, look up the join key in the index of the `Users` collection and dynamically load the matching user into the lazy users stream. In other words, we're populating the lazy stream with users that are matching the comments as we process them. Note that the lazy users stream can apply additional operators before being joined in. In the diagram, we're doing an additional filter over the lazy stream before joining it in.

### Implementation Challenges

The D2 pipeline from the diagram above is created at compile time but the indexes are created at runtime. Hence, when creating this special `tap` operator we don't know if the collection that we want to lazily load via an index on the join key, will actually have the index that is required on the join key. We will only know this at runtime, when the `map` operator runs for the first time. At that point, if we notice that the index does not exist, we cannot apply the optimization so we then turn the Users collection back into a regular collection (instead of a lazy collection).

Currently, inner joins loop over the smallest collection and lookup matching rows in the index on the bigger collection. But that index may not exist, in which case we will need to loop over the entire bigger collection. In that case, it may be more efficient to loop over the bigger collection and try to find matching rows in the smaller collection (because that one might have an index on the join key). However, flipping these collections around is going to complicate the code quite a lot (because again this would need to happen at runtime) so we decided not to do it yet.

### TODOs

- [x] Add unit tests to check that indexes are correctly used for left/right/inner joins
- [x] Automatically create indexes for join keys (in `eager` mode)
- [ ] ~~Always create an index on the PK of a collection (in `eager` and `pk` mode)~~ Better as a follow-up PR.

### DIFF
--- a/packages/db/src/query/compiler/joins.ts
+++ b/packages/db/src/query/compiler/joins.ts
@@ -10,6 +10,7 @@ import {
   InvalidJoinConditionSameTableError,
   InvalidJoinConditionTableMismatchError,
   InvalidJoinConditionWrongTablesError,
+  JoinCollectionNotFoundError,
   UnsupportedJoinSourceTypeError,
   UnsupportedJoinTypeError,
 } from "../../errors.js"
@@ -119,15 +120,11 @@ function processJoin(
   const joinedCollection = collections[joinedCollectionId]
 
   if (!mainCollection) {
-    throw new Error(
-      `Internal error in join. Collection not found for ${mainTableId}`
-    )
+    throw new JoinCollectionNotFoundError(mainTableId)
   }
 
   if (!joinedCollection) {
-    throw new Error(
-      `Internal error in join. Collection not found for ${joinedCollectionId}`
-    )
+    throw new JoinCollectionNotFoundError(joinedCollectionId)
   }
 
   const { activeCollection, lazyCollection } = getActiveAndLazyCollections(
--- a/packages/db/src/errors.ts
+++ b/packages/db/src/errors.ts
@@ -377,6 +377,12 @@ export class UnknownFunctionError extends QueryCompilationError {
   }
 }
 
+export class JoinCollectionNotFoundError extends QueryCompilationError {
+  constructor(collectionId: string) {
+    super(`Collection "${collectionId}" not found during compilation of join`)
+  }
+}
+
 // JOIN Operation Errors
 export class JoinError extends TanStackDBError {
   constructor(message: string) {


samwillis (commented):
~~We have central error classes in `src/errros.ts` we should use one form there or create a new one.~~

Ignore, clicked wrong btutton

packages/db/src/query/compiler/joins.ts
```diff
@@ -64,29 +87,54 @@ function processJoin(
...
+    throw new Error(
+      `Internal error in join. Collection not found for ${mainTableId}`
+    )
+  }
+
+  if (!joinedCollection) {
+    throw new Error(
+      `Internal error in join. Collection not found for ${joinedCollectionId}`
+    )
+  }
```
samwillis: We have central error classes in `src/errros.ts` we should use one form there or create a new one.

packages/db-ivm/src/operators/tap.ts
```diff
@@ -0,0 +1,53 @@
...
+    output: DifferenceStreamWriter<T>,
+    f: (data: T) => void
+  ) {
+    super(id, inputA, output)
+    this.#f = f
+  }
+
+  inner(collection: MultiSet<T>): MultiSet<T> {
+    return collection.map((data) => {
+      this.#f(data)
```
samwillis, line 25: It could be useful in future to pass the multiplicity to the callback so that it's aware if it's an insert/delete. Not important for now.

packages/db/src/indexes/auto-index.ts
```diff
@@ -6,6 +6,55 @@ export interface AutoIndexConfig {
...
+  }
+
+  // Create a new index for this field using the collection's createIndex method
+  try {
+    collection.createIndex((row) => (row as any)[fieldName], {
+      name: `auto_${fieldName}`,
+      indexType: BTreeIndex,
+    })
+  } catch (error) {
+    console.warn(`Failed to create auto-index for field "${fieldName}":`, error)
```
samwillis, line 54: we've used the `debug` package elsewhere for logging. We should maybe use it here.

kevin-dp: I didn't actually change this. I took this piece of code from `ensureIndexForExpression` and moved it here. We could use `debugLog` but that would only log it in debug mode. I think this warning is useful also in non-debug mode to warn you that for some reason the index could not be created and thus the queries might be less efficient.

packages/db-ivm/src/operators/tap.ts
kevin-dp: We can add this later when we need it :-)

packages/db/src/query/compiler/joins.ts
```diff
@@ -132,11 +177,96 @@ function processJoin(
...
+
+    const fieldName = followRefResult.path[0]
+    if (fieldName) {
+      ensureIndexForField(fieldName, followRefResult.path, followRefCollection)
+    }
+
+    const activePipelineWithLoading: IStreamBuilder<
+      [key: unknown, [originalKey: string, namespacedRow: NamespacedRow]]
+    > = activePipeline.pipe(
+      tap(([joinKey, _]) => {
```
samwillis, line 224: Not something to change right now, but with the current version of tap we processes and ask for each joined key one at a time. If in future we then want to batch load via sync we need to try and reassembly a batch to ask for.

the tap operator iterates over the items in the multiset, calling this function here, and it then for each row asked for the joined items to be injected. The alternative would to do so tot as the multiset level, we have a batch of items from the left, so ask for a batch from the right to be injected all at once. These batches then naturally can be pushed back down to the sync layer and asked for from there. 

But for later!

PR #349, opened 2025-07-31
failing test for stable ref from useLiveQuery

mharris717:


falkfrentzen:
This issue (probably) causes infinite loops when useLiveQuery (+ElectricSQL) is used in the same component, that tanstack-table is implemented in. I have not confirmed, if this is an issue with other tanstack-db collection types. 

I prepared a project showcasing the infinite loop issue, as well as a workaround, using a parent wrapper for data aggregation and a child table component : https://github.com/falkfrentzen/tanstack-db-table-issue-showcase

samwillis:
I've incorporated the tests from this PR into #388 - we should get this out on Monday after it's reviewed.

PR #365, opened 2025-08-02
Doc: documents that RN needs a crypto.randomUUID polyfill

### LINKED ISSUE #362: cypto module doesn't exist in expo app
`ReferenceError: Property 'crypto' doesn't exist, js engine: hermes [Component Stack]`

```tsx
import { Stack } from "expo-router";
import {
  QueryClient,
  QueryClientProvider,
} from '@tanstack/react-query'
import { createCollection, useLiveQuery } from '@tanstack/react-db'
import { queryCollectionOptions } from '@tanstack/query-db-collection'

const queryClient = new QueryClient()

  queryCollectionOptions({
    queryClient,
    queryKey: ['todos'],
    queryFn: async () => {
      const response = await fetch('/api/todos')
      return response.json()
    },
    getKey: (item) => item.id,
      const { original, modified } = transaction.mutations[0]
      await fetch(`/api/todos/${original.id}`, {
        method: 'PUT',
        body: JSON.stringify(modified),
      })
    },
  })
)

export default function RootLayout() {
  return <QueryClientProvider client={queryClient}>
    <Stack />
  </QueryClientProvider>
}

```

KyleAMathews:
Fix #362

### DIFF
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -885,6 +885,24 @@ const AddTodo = () => {
 }
 ```
 
+## React Native
+
+When using TanStack DB with React Native, you need to install and configure a UUID generation library since React Native doesn't include crypto.randomUUID() by default.
+
+Install the `react-native-random-uuid` package:
+
+```bash
+npm install react-native-random-uuid
+```
+
+Then import it at the entry point of your React Native app (e.g., in your `App.js` or `index.js`):
+
+```javascript
+import 'react-native-random-uuid'
+```
+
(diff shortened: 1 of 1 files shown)


samwillis:
@KyleAMathews back in the day of the old electric, I think we ended on vendoring a uuid implementation into the client.l to solve this. Reduces the friction a little. What do you think?

(obv ship this as is!)

KyleAMathews:
@samwillis like lazy load it if in React Native?

PR #377, opened 2025-08-04
fix issue with using optional props in select and join clauses

### LINKED ISSUE #299: Typescript error joining on an optional field with `eq`.
I have event and user collections defined with the following (abbreviated) zod schemas:

```ts
const userSchema = z.object({
  id: z.uuid()
})

const eventSchema = z.object({
  user_id: z.uuid().optional()
})
```

If I try to do an inner join between them like this:

```ts
query
  .from({ event: eventCollection })
  .innerJoin(
    { user: userCollection },
    ({ event, user }) => eq(event.user_id, user.id)
  )
```

I get a typescript error:

```
No overload matches this call.
  Overload 1 of 3, '(left: RefProxy<string | undefined>, right: string | RefProxy<string | undefined> | BasicExpression<string | undefined> | undefined): BasicExpression<...>', gave the following error.
    Argument of type 'RefProxy<string | undefined> | undefined' is not assignable to parameter of type 'RefProxy<string | undefined>'.
      Type 'undefined' is not assignable to type 'RefProxy<string | undefined>'.
  Overload 2 of 3, '(left: string | number | boolean | BasicExpression<string | number | boolean>, right: string | number | boolean | BasicExpression<string | number | boolean>): BasicExpression<...>', gave the following error.
    Argument of type 'RefProxy<string | undefined> | undefined' is not assignable to parameter of type 'string | number | boolean | BasicExpression<string | number | boolean>'.
      Type 'undefined' is not assignable to type 'string | number | boolean | BasicExpression<string | number | boolean>'.
  Overload 3 of 3, '(left: Aggregate<unknown>, right: any): BasicExpression<boolean>', gave the following error.
    Argument of type 'RefProxy<string | undefined> | undefined' is not assignable to parameter of type 'Aggregate<unknown>'.
      Type 'undefined' is not assignable to type 'Aggregate<unknown>'.

         .innerJoin({ user: userCollection }, ({ event, user }) => eq(event.user_id, user.id))
```

If I swap the order of the arguments to `eq` then I don't get the error. Is this correct / expected? Or should `eq` handle optional fields as the first argument in this situation?

### LINKED ISSUE #300: Typescript error selecting an optional field.
I have a collection defined with a zod schema with `inserted_at: z.date().optional()`.

If I try to select this field in a live query:

```ts
query
  .from({ thread: threadCollection })
  .select(({ thread }) => ({
    inserted_at: thread.inserted_at
  }))
```

I get a Typescript error:

```
Type 'RefProxy<Date | undefined> | undefined' is not assignable to type 'RefProxy<any> | RefProxyFor<any> | BasicExpression | Aggregate<any>'.
  Type 'undefined' is not assignable to type 'RefProxy<any> | RefProxyFor<any> | BasicExpression | Aggregate<any>'.

50           inserted_at: thread.inserted_at
             ~~~~~~~~~~~
```

This may relate to #299.

samwillis:
This also fixes using `createCollection` with a schema, and then using a live query on it - the types were not inferred in the query. This did not affect query or electric collections, but was apparent in the new tests.

fixes #299 
fixes #300


KyleAMathews:
Also.... what if you could just write type logic in typescript or something. This is pretty simple logic but sooo complicated in Typescript.

Issue #344, opened 2025-07-31
Add Query Invalidation Support to query-db-collection


KyleAMathews:
### Background

TanStack Query provides `invalidateQueries` to refresh multiple queries by pattern matching. Currently, `query-db-collection` only supports manual refetch per collection. This requires developers to import and call refetch on each related collection manually.

### Problem

When using multiple query collections, invalidating related queries requires boilerplate:

```typescript
// Current approach - manual coordination
await Promise.all([
  todoCollection.utils.refetch(),
  userStatsCollection.utils.refetch(),
  projectSummaryCollection.utils.refetch()
])
```

### Proposed Solution

Add invalidation support specifically to `query-db-collection` by:

1. Allowing collections to subscribe to invalidation events
2. Pattern matching against the collection's queryKey
3. Automatically calling refetch when matched

### Implementation

```typescript
// Add to query-db-collection
export interface QueryCollectionConfig<TItem, TError, TQueryKey> {
  // ... existing options ...
}

// In queryCollectionOptions implementation
const observerOptions = {
  // ... existing options ...
  // Subscribe to invalidation events
  onInvalidate: () => {
    // This is called by TanStack Query when invalidateQueries matches
    refetch()
  }
}

// The queryClient.invalidateQueries will work automatically
queryClient.invalidateQueries({ queryKey: ['todos'] })
// All collections with queryKey starting with ['todos'] will refetch
```

### Usage Example

```typescript
// Define collections
const todoCollection = createCollection(
  queryCollectionOptions({
    queryKey: ['todos', userId],
    queryFn: fetchTodos,
    getKey: (item) => item.id,
    queryClient,
  })
)

const todoStatsCollection = createCollection(
  queryCollectionOptions({
    queryKey: ['todos', 'stats', userId],
    queryFn: fetchTodoStats,
    getKey: (item) => item.id,
    queryClient,
  })
)

// Later, invalidate all todo-related queries
await queryClient.invalidateQueries({ 
  queryKey: ['todos'] // Matches both collections
})
```

### Benefits

1. Works with existing TanStack Query invalidation patterns
2. No new APIs to learn
3. Leverages QueryClient's pattern matching
4. Each collection manages its own invalidation response

kels-c:
Hi @KyleAMathews, can I pick this up?

KyleAMathews:
Yes please!

kels-c:
[...]

PR #381, opened 2025-08-06
feat(query-db-collection): support automatic refetch on query invalidation

kels-c:
From #344:

Adds support for automatic refetching of collections when queryClient.invalidateQueries is called with a matching queryKey. This removes the need for manually coordinating utils.refetch() calls across related collections.

Tests are passing, but I’d appreciate a second look to confirm they cover the right scenarios — especially around observed vs. passive (unobserved) invalidation.

### DIFF
--- a/packages/query-db-collection/tests/query.test.ts
+++ b/packages/query-db-collection/tests/query.test.ts
@@ -834,7 +834,7 @@ describe(`QueryCollection`, () => {
 
       // Wait for refetch to complete
       await vi.waitFor(() => {
-        expect(queryFn).toHaveBeenCalledTimes(2)
+        expect(queryFn).toHaveBeenCalledTimes(3)
         expect(collection.size).toBe(2)
       })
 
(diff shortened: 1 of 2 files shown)


packages/query-db-collection/tests/query.test.ts
```diff
@@ -834,7 +834,7 @@ describe(`QueryCollection`, () => {
 
       // Wait for refetch to complete
       await vi.waitFor(() => {
-        expect(queryFn).toHaveBeenCalledTimes(2)
+        expect(queryFn).toHaveBeenCalledTimes(3)
```
kels-c, line 837: This test might need to be changed, open to feedback!

KyleAMathews:
The build is failing — also let's update docs/jsdocs where we recommend calling refetch on collections to also mention invalidate as an option.

packages/query-db-collection/tests/query.test.ts
KyleAMathews: huh wait... did we already have support for invalidating queries?? 😅 It looks like we just added another implementation... it seems Query was already handling this internally... if this is the case, then we have a lot less to do, just update the docs/jsdocs a bit.

kels-c: Definitely appears so, and to be honest, that's what I've been confused about. Still not completely following the need for refetch in general, if under the hood we're reading from the cache from query? But I don't have full context as well. 😄

I'll go ahead and remove the changes.

If this behavior is handled by Query already, do we want to mention that directly in docs/query-collection.md? What did you have in mind?

KyleAMathews: Yeah when building this originally and writing this issue it didn't occur to me that the query client would just keep calling the queryFn like normal. 

Refetch is nice as not everyone is migrating from a full blown normal Query setup so they might just have a one-off collection that they want to be able to manually refetch. It depends I suppose on your mental model of your state. A web of data or standalone table-like collections.

There's a few places where we mention calling other collections to refetch (search for "refetch". So we should mention first that you can invalidate a tag or just directly tell those collections to refetch.

Issue #383, opened 2025-08-06
liveQueryCollections are set to ready even if one of it's source collections is not ready yet


tomkuehl:
```ts
const collection = createCollection<{ id: string }>({
  sync: {
    sync({ begin, commit }) {
      begin()
      commit()
    },
  },
  getKey: (item) => item.id,
  startSync: true,
})

const liveQuery = createLiveQueryCollection({
  query: (q) => q.from({ collection }),
  startSync: true,
})
```

`liveQuery.isReady()` is `true` even though `collection` is not ready yet

PR #384, opened 2025-08-06
Ensure liveQueryCollections are not ready until all it's source collections are ready

tomkuehl:
fixes https://github.com/TanStack/db/issues/383


samwillis:
Hey @tomkuehl thanks for this! I've made some additions to it in a new PR here: https://github.com/TanStack/db/pull/390
We should get this out tomorrow if all goes to plan.

PR #386, opened 2025-08-07
Refactor Ref System with Nested Optional Properties

samwillis:
## 🎯 Overview

This PR introduces a comprehensive refactor of the ref system to properly support nested structures and optionality, bringing the type system in line with JavaScript's optional chaining behavior.

## ✨ Key Improvements

- **🔗 Nested Optional Properties**: Full support for deeply nested optional objects (`employees.profile?.bio`, `orders.customer?.address?.street`)
- **🛡️ Enhanced Type Safety**: Optional types now correctly typed as `RefProxy<T> | undefined` with optionality outside the ref, enabling proper optional chaining
- **🔧 New Query Functions**: Added `isUndefined`, `isNotUndefined`, `isNull`, `isNotNull` for proper null/undefined checks in queries  
- **⚙️ Improved JOIN Handling**: Fixed optionality in JOIN operations and multiple GROUP BY clause support

## 🚨 Breaking Changes

**IMPORTANT**: Code that previously worked by ignoring optionality now requires proper optional chaining syntax.

```typescript
// Before (worked but was type-unsafe)
employees.profile.bio         // ❌ Now throws type error

// After (correct and type-safe)  
employees.profile?.bio        // ✅ Required syntax
```

### Type System Refactor
- **`RefProxyForContext`**: Now handles optionality outside refs (`RefProxy<T> | undefined`)
- **`PrecomputeRefStructure`**: Only wraps leaf values, leaving objects as plain TypeScript
- **Enhanced `SelectValue`**: Support for nested projections and JavaScript literals
- **New Helper Types**: Comprehensive null/undefined handling

### Runtime Improvements  
- **Compiler Functions**: Implemented `isNotUndefined`, `isUndefined`, `isNull`, `isNotNull`
- **GROUP BY Fixes**: Multiple `.groupBy()` calls now accumulate instead of replace
- **JOIN Optionality**: Correct handling of optional tables in different join types

## 📊 Testing

- ✅ **1,050 tests passing** (60 test files)
- ✅ **0 type errors** 
- ✅ **0 linting errors**
- ✅ **Enhanced type safety** with comprehensive nested property coverage

### Core Type System
- `packages/db/src/query/builder/types.ts` - Major refactor of ref types
- `packages/db/src/query/builder/index.ts` - Enhanced query builder methods
- `packages/db/src/query/builder/functions.ts` - New null/undefined functions

### Compiler & Runtime
- `packages/db/src/query/compiler/evaluators.ts` - Added missing function implementations
- `packages/db/src/query/compiler/group-by.ts` - Fixed multiple GROUP BY validation

### Tests & Documentation
- Updated 100+ test files to use proper optional chaining syntax
- Fixed linting issues from improved type inference
- Added comprehensive changeset documentation

## 🚀 Migration Guide

**For existing codebases**: Add `?.` when accessing potentially undefined nested properties. The TypeScript compiler will guide you to the correct syntax.

## 🎉 Benefits

- **Eliminates runtime errors** from accessing undefined nested properties
- **Better IDE experience** with cleaner type displays and proper IntelliSense  
- **JavaScript alignment** - type system now mirrors optional chaining behavior
- **Improved SQL compliance** with better validation and null handling

This enhancement significantly improves developer experience while maintaining the robustness and performance of the existing query system.


kevin-dp:
@samwillis  Do we really need `isNotUndefined` and `isNotNull`? I would expect to write that as `not(isUndefined(...))` and `not(isNull(...))`.

samwillis:
> @samwillis Do we really need `isNotUndefined` and `isNotNull`? I would expect to write that as `not(isUndefined(...))` and `not(isNull(...))`.

Agreed, lets drop them

packages/db/src/query/builder/types.ts
```diff
@@ -296,20 +837,85 @@ export type HasJoinType<
...
+ * The key difference from `MergeContextWithJoinType` is that NO optionality is applied here.
+ *
+ * **Why No Optionality?**
+ * In SQL, join conditions are evaluated BEFORE optionality is determined. Both tables
+ * must be treated as available (non-optional) within the join condition itself.
+ * Optionality is only applied to the result AFTER the join logic executes.
+ *
+ * **Example**:
+ * ```typescript
+ * .leftJoin({ orders }, (refs) => {
```
kevin-dp: In the callback we use `refs.users` but `users` come out of the blue. It seems like we are missing the `q.from(...)` before the `.leftJoin(...)` in this example snippet.

packages/db/src/query/builder/types.ts
```diff
@@ -43,49 +102,208 @@ export type SchemaFromSource<T extends Source> = Prettify<{
...
-      : TSelectObject[K] extends Aggregate<infer T>
-        ? T
-        : TSelectObject[K] extends RefProxyFor<infer T>
-          ? T
-          : TSelectObject[K] extends undefined
-            ? undefined
-            : TSelectObject[K] extends { __type: infer U }
-              ? U
-              : never
+      : TSelectObject[K] extends Ref<infer T> | undefined
```
kevin-dp, line 237: Since we're here we know it does not extend `Ref<infer T>` so should we just check whether it extends `undefined` ?

kevin-dp: I believe this is equivalent:
```ts
TSelectObject[K] extends Ref<infer T>
  ? T
  : TSelectObject[K] extends undefined
      ? undefined
      : TSelectObject[K] extends null
          ? null
          : ...
```
The same applies a bit further down when checking for extensions of `RefProxy`.

packages/db/src/query/builder/types.ts
```diff
@@ -43,49 +102,208 @@ export type SchemaFromSource<T extends Source> = Prettify<{
...
+              ? T | null
+              : TSelectObject[K] extends BasicExpression<infer T>
+                ? T
+                : TSelectObject[K] extends Aggregate<infer T>
+                  ? T
+                  : TSelectObject[K] extends RefProxyFor<infer T>
+                    ? T
+                    : TSelectObject[K] extends SpreadableRefProxy<infer T>
+                      ? ResultTypeFromSelect<SpreadableRefProxy<T>>
+                      : TSelectObject[K] extends string
```
kevin-dp: Do we really need to make all these separate checks?
I would simplify this to:
```ts
TSelectObject[K] extends string | number | boolean | null | undefined
? TSelectObject[K]
: ...
```
There's a difference compared to your original type though which is that your original type widens `TSelectObject[K]` to a `string`, `number`, ... whereas the type i propose preserves the original type which might be narrower, e.g. `'a'` instead of `string`.

packages/db/src/query/builder/types.ts
```diff
@@ -104,91 +329,315 @@ export type CompareOptions = {
...
+ * does allow it because the undefined is "outside" the proxy.
+ */
 export type RefProxyForContext<TContext extends Context> = {
-  [K in keyof TContext[`schema`]]: RefProxyFor<TContext[`schema`][K]>
+  [K in keyof TContext[`schema`]]: IsExactlyUndefined<
+    TContext[`schema`][K]
+  > extends true
+    ? // T is exactly undefined - wrap in RefProxy as-is
+      RefProxy<TContext[`schema`][K]>
+    : IsExactlyNull<TContext[`schema`][K]> extends true
```
kevin-dp: Looks like we can merge the `IsExactlyUndefined` and `IsExactlyNull` cases:
```ts
IsExactlyUndefined<TContext[`schema`][K]> extends true | IsExactlyNull<TContext[`schema`][K]> extends true
? RefProxy<TContext[`schema`][K]>
: ...
```

packages/db/src/query/builder/types.ts
```diff
@@ -104,91 +329,315 @@ export type CompareOptions = {
...
+      ? // T is exactly null - wrap in RefProxy as-is
+        RefProxy<TContext[`schema`][K]>
+      : IsOptional<TContext[`schema`][K]> extends true
+        ? // T is optional (T | undefined) but not exactly undefined
+          // Extract the non-undefined part and place undefined outside
+          RefProxy<NonUndefined<TContext[`schema`][K]>> | undefined
+        : IsNullable<TContext[`schema`][K]> extends true
```
kevin-dp: It's wrong to split the `isOptional` and `isNullable` checks because if a type is both, e.g. `string | null | undefined` then it will match the first case and we will only remove the `undefined` but not the `null` from the type.

packages/db/src/query/builder/types.ts
```diff
@@ -104,91 +329,315 @@ export type CompareOptions = {
...
+          ? // T is nullable (T | null) but not exactly null
+            // Extract the non-null part and place null outside
+            RefProxy<NonNull<TContext[`schema`][K]>> | null
+          : // T is not optional or nullable - always wrap in RefProxy for top-level schema types
+            RefProxy<TContext[`schema`][K]>
```
kevin-dp: Since this is the default case we could simplify the type to check for non-exact optionality/nullability and return the right type in that case, otherwise default to `RefProxy<TContext['schema'][K]>`.
Something like:
```ts
type IsNonExactOptional<T> = IsOptional<T> extends true
  ? IsExactlyUndefined<T> extends false
     ? true
     : false
  : false

type IsNonExactNullable<T> = ...

export type RefProxyForContext<TContext extends Context> = {
  [K in keyof TContext[`schema`]]:
    IsNonExactOptional<TContext[`schema`][K]> extends true
    ? RefProxy<NonUndefined<TContext[`schema`][K]>> | undefined
    : IsNonExactNullable<TContext[`schema`][K]> extends true
       ? RefProxy<NonNull<TContext[`schema`][K]>> | null
       : RefProxy<TContext[`schema`][K]>
}
```

Keeping into account my earlier comment that we should also handle the case where the type is both non exact optional and non exact nullable, it would be something like this:
```ts
export type RefProxyForContext<TContext extends Context> = {
  [K in keyof TContext[`schema`]]:
    (IsNonExactOptional<TContext[`schema`][K]> & IsNonExactNullable<TContext[`schema`][K]>) extends true
    ? RefProxy<NonUndefined<NonNull<TContext[`schema`][K]>>> | undefined
    : IsNonExactOptional<TContext[`schema`][K]> extends true
       ? RefProxy<NonUndefined<TContext[`schema`][K]>> | undefined
       : IsNonExactNullable<TContext[`schema`][K]> extends true
          ? RefProxy<NonNull<TContext[`schema`][K]>> | null
          : RefProxy<TContext[`schema`][K]>
}
```

packages/db/src/query/builder/types.ts
```diff
@@ -104,91 +329,315 @@ export type CompareOptions = {
...
+ * Input:  { bio: string, contact: { email: string, phone?: string } }
+ * Output: { bio: Ref<string>, contact: { email: Ref<string>, phone: Ref<string> | undefined } }
+ *
+ * The recursion handles nested objects while preserving optionality/nullability:
+ * - For optional objects: The object structure is preserved, undefined goes outside
+ * - For optional leaves: Ref<T> | undefined (undefined outside the Ref)
+ * - For nullable objects: The object structure is preserved, null goes outside
+ * - For nullable leaves: Ref<T> | null (null outside the Ref)
+ */
+export type PrecomputeRefStructure<T extends Record<string, any>> = {
```
kevin-dp, line 486: Same comments here as for the `RefProxyForContext` type:
- need to explicitly handle the case where a type is both optional and nullable
- can simplify the type by handling `Ref<T[K]>` as the default case by using `IsNonExact...` helper types

packages/db/src/query/builder/types.ts
```diff
@@ -104,91 +329,315 @@ export type CompareOptions = {
...
+ * - Standalone composable functions
+ * - Reusable query fragments
+ * - Backward compatibility with existing code
+ *
+ * It follows the same principles as PrecomputeRefStructure and RefProxyForContext:
+ * - Place undefined/null outside refs for optional chaining support
+ * - Only wrap leaf values in RefProxy/Ref
+ * - Preserve object structures for spread operations
+ */
+export type RefProxyFor<T> =
```
kevin-dp, line 531: Same comments as for the other 2 types + this type is not handling the case where `T` is exactly null. Is this an oversight or is this on purpose?

packages/db/src/query/builder/types.ts
```diff
@@ -104,91 +329,315 @@ export type CompareOptions = {
...
+  /** @internal - Runtime marker to identify ref proxy objects */
   readonly __refProxy: true
-  /** @internal */
+  /** @internal - Path segments for building query expressions */
   readonly __path: Array<string>
-  /** @internal */
+  /** @internal - Type information for TypeScript inference */
   readonly __type: T
-}
+} & (T extends undefined
```
kevin-dp, line 578: Why don't we need to check `T` for `IsExactlyUndefined<T>` here?

packages/db/src/query/builder/types.ts
```diff
@@ -104,91 +329,315 @@ export type CompareOptions = {
...
+  ? {} // undefined types get no additional properties
+  : T extends Record<string, any>
+    ? {
+        // Object types get recursive property transformation
+        [K in keyof T]: IsExactlyUndefined<T[K]> extends true
```
kevin-dp: Same comment here as before:
- need to handle the case where the type is both optional and nullable
- can simplify it by handling `Ref<T[K]>` as the default case, use the `isNonExact...` helper types for that

packages/db/src/query/builder/types.ts
```diff
@@ -104,91 +329,315 @@ export type CompareOptions = {
...
+ * Structurally compatible with `RefProxy<T>` so runtime proxies can be used
+ * wherever a `Ref<T>` is expected, while hiding internal proxy properties.
+ *
+ * - For object types: exposes the same nested property structure, with leaves
+ *   typed as `Ref<Leaf>` and optionality/nullability preserved outside the Ref.
+ * - For primitives: an opaque branded wrapper so hovers show `Ref<T>`.
+ */
+declare const RefBrand: unique symbol
+type RefLeaf<T> = { readonly [RefBrand]?: T }
+export type Ref<T = any> = T extends undefined
```
kevin-dp, line 638: All same comments as before

packages/db/src/query/builder/functions.ts
```diff
@@ -2,71 +2,189 @@ import { Aggregate, Func } from "../ir"
...
+
+// Helper type to extract the underlying type from various expression types
+type ExtractType<T> =
+  T extends RefProxy<infer U>
+    ? U
+    : T extends Ref<infer U>
+      ? U
+      : T extends BasicExpression<infer U>
+        ? U
+        : T extends undefined
```
kevin-dp: Can't we replace `T extends undefined ? ... : ...` by just `T` since in all cases we return T.

packages/db/src/query/builder/functions.ts
```diff
@@ -2,71 +2,189 @@ import { Aggregate, Func } from "../ir"
...
+          : T extends null
+            ? null
+            : T
+
+// Helper type to determine aggregate return type based on input nullability
+type AggregateReturnType<T> =
+  ExtractType<T> extends infer U
+    ? U extends number | undefined | null
+      ? Aggregate<U>
+      : U extends number
```
kevin-dp: How can this condition ever be true? If we reach here then `U extends number | undefined | null` is false so `U extends number` can't be true.

packages/db/src/query/builder/functions.ts
```diff
@@ -2,71 +2,189 @@ import { Aggregate, Func } from "../ir"
...
+        ? Aggregate<number>
+        : Aggregate<number | undefined | null>
+    : Aggregate<number | undefined | null>
+
+// Helper type to determine string function return type based on input nullability
+type StringFunctionReturnType<T> =
+  ExtractType<T> extends infer U
+    ? U extends string | undefined | null
+      ? BasicExpression<U>
+      : U extends string
```
kevin-dp: Can't ever be true

packages/db/src/query/builder/functions.ts
```diff
@@ -2,71 +2,189 @@ import { Aggregate, Func } from "../ir"
...
+        : BasicExpression<string | undefined | null>
+    : BasicExpression<string | undefined | null>
+
+// Helper type to determine numeric function return type based on input nullability
+// This handles string, array, and number inputs for functions like length()
+type NumericFunctionReturnType<T> =
+  ExtractType<T> extends infer U
+    ? U extends string
+      ? BasicExpression<number>
+      : U extends string | undefined
```
kevin-dp: Why are we checking for `string | ...` here and in the next cases? We already know that it does not extend `string`.

packages/db/src/query/builder/functions.ts
```diff
@@ -2,71 +2,189 @@ import { Aggregate, Func } from "../ir"
...
+              ? BasicExpression<number>
+              : U extends Array<any> | undefined
+                ? BasicExpression<number | undefined>
+                : U extends Array<any> | null
+                  ? BasicExpression<number | null>
+                  : U extends Array<any> | undefined | null
+                    ? BasicExpression<number | undefined | null>
+                    : U extends number | undefined | null
+                      ? BasicExpression<U>
+                      : U extends number
```
kevin-dp: We won't ever reach here if  `U extends number` because it would match the previous check `U extends number | undefined | null`.

packages/db/src/query/builder/types.ts
```diff
@@ -104,91 +329,315 @@ export type CompareOptions = {
...
-// source being `SourceType | undefined`.
-// We then follow this through the ref proxy system so that accessing a property on
-// and optional source will itsself be optional.
-// If for example we join in `joinedTable` with a left join, then
-// `where(({ joinedTable }) => joinedTable.name === `John`)`
-// we want the the type of `name` to be `RefProxy<string | undefined>` to indicate that
-// the `name` property is optional, as the joinedTable is also optional.
-export type RefProxyFor<T> = OmitRefProxy<
+// Helper type to extract non-null type
+type NonNull<T> = T extends null ? never : T
```
kevin-dp, line 463: Do we really need to define our own `NonUndefined` and `NonNull` helper types?
Can't we use Typescript's built in `NonNullable<T>` type? This does the same as `NonUndefined<NonNull<T>>` but it seems that any place where we use `NonUndefined` or `NonNull` it wouldn't matter if we would just remove both (i.e. use `NonNullable`).

packages/db/tests/query/basic.test.ts
```diff
@@ -715,6 +779,324 @@ function createBasicTests(autoIndex: `off` | `eager`) {
...
+      expect(profileCollection.size).toBe(4) // All users still there, Bob will have undefined values
+      expect(profileCollection.get(2)).toMatchObject({
+        id: 2,
+        name: `Bob`,
+        theme: undefined,
+        notifications: undefined,
+      })
+    })
+
+    test(`should work with spread operator on nested objects`, () => {
```
kevin-dp: The title of this test mentions the spread operator but i don't see it being used in the test

packages/db/src/query/builder/types.ts
samwillis: I think we have to leave these as they are, these union are `extends (Ref<infer T> | undefined)` not `(extends Ref<infer T>)  OR  (extends undefined)` (fake syntax for clarity). They preserver the optionality when extracting the wrapped type.

kevin-dp (approved):

PR #387, opened 2025-08-07
Refactor devtools to be event-based and add new features

samwillis:
This pull request contains changes generated by Cursor background composer.


samwillis:
replaced with #411

Issue #385, opened 2025-08-06
Types tracking


jherr:
Types aren't coming through in the `useLiveQuery` hook. `data` here needs to be cast to `Message[]`. 

```ts
import {
  createCollection,
  localOnlyCollectionOptions,
  useLiveQuery
} from "@tanstack/react-db";
import { z } from "zod";

const MessageSchema = z.object({
  id: z.number(),
  text: z.string(),
  user: z.string(),
});

export type Message = z.infer<typeof MessageSchema>;

messagesCollection = createCollection(
  localOnlyCollectionOptions({
    getKey: (message) => message.id,
    schema: MessageSchema,
  })
);

export function useMessages() {
  const { data: messages } = useLiveQuery((q) =>
    q.from({ message: messagesCollection }).select(({ message }) => ({
      ...message,
    }))
  );

  return messages as Message[];
}
```

samwillis:
Hey @jherr 
I think this is related to the spread in the `select`, I'm going to try and get a fix in as part of #386 as it's very related to the work there.

PR #389, opened 2025-08-10
Refactor select improving spread (`...obj`) and enabling nested projection.

samwillis:
This branch started as a fix for the broken types when using a spread in a select (#385), but identified a couple of problems:

- we didn't have any way to track the order of spreads within an object, so multiple spreads could result in incorrect overwriting of properties. This is now fixed by tracking the order of operations (ref proxy access and spreads) within the select:

```ts
.select({ user }) => ({
  ...user, // <-- spread first
  name: upper(user.name) // <-- override name
})
```

- select didn't have proper support for nested projection, you can now do this:

```ts
select({ user }) => ({
  ...user,
  meta: {  // <= nested object within the select
    email: user.email
  }
})
```

### DIFF
--- a/packages/db/src/query/builder/index.ts
+++ b/packages/db/src/query/builder/index.ts
@@ -1,5 +1,13 @@
 import { CollectionImpl } from "../../collection.js"
-import { CollectionRef, QueryRef, Aggregate as AggregateExpr, Func as FuncExpr, PropRef, Value as ValueExpr } from "../ir.js"
+import {
+  Aggregate as AggregateExpr,
+  CollectionRef,
+  Func as FuncExpr,
+  PropRef,
+  QueryRef,
+  Value as ValueExpr,
+  isExpressionLike,
+} from "../ir.js"
 import {
   InvalidSourceError,
   JoinConditionMustBeEqualityError,
@@ -438,8 +446,8 @@ export class BaseQueryBuilder<TContext extends Context = Context> {
       return (
         value !== null &&
         typeof value === `object` &&
-        !(`type` in value) &&
-        !(value as any).__refProxy
+        !isExpressionLike(value) &&
+        !value.__refProxy
       )
     }
 
@@ -461,7 +469,7 @@ export class BaseQueryBuilder<TContext extends Context = Context> {
 
     return new BaseQueryBuilder({
       ...this.query,
-      select: select as any,
+      select: select,
       fnSelect: undefined, // remove the fnSelect clause if it exists
     }) as any
   }
--- a/packages/db/src/query/builder/index.ts
+++ b/packages/db/src/query/builder/index.ts
@@ -26,7 +26,6 @@ import type {
   OrderByDirection,
   QueryIR,
 } from "../ir.js"
-// import { PropRef } from "../ir.js"
 import type {
   CompareOptions,
   Context,
--- a/packages/db/src/query/builder/index.ts
+++ b/packages/db/src/query/builder/index.ts
@@ -1,5 +1,5 @@
 import { CollectionImpl } from "../../collection.js"
-import { CollectionRef, QueryRef } from "../ir.js"
+import { CollectionRef, QueryRef, Aggregate as AggregateExpr, Func as FuncExpr, PropRef, Value as ValueExpr } from "../ir.js"
 import {
   InvalidSourceError,
   JoinConditionMustBeEqualityError,
@@ -18,7 +18,7 @@ import type {
   OrderByDirection,
   QueryIR,
 } from "../ir.js"
-import { PropRef } from "../ir.js"
+// import { PropRef } from "../ir.js"
 import type {
   CompareOptions,
   Context,
@@ -424,44 +424,15 @@ export class BaseQueryBuilder<TContext extends Context = Context> {
       if (value === undefined) return toExpression(null)
       if (isRefProxy(value)) return toExpression(value)
       if (
-        value &&
-        typeof value === `object` &&
-        `type` in value &&
-        (value.type === `agg` ||
-          value.type === `func` ||
-          value.type === `ref` ||
-          value.type === `val`)
+        value instanceof AggregateExpr ||
+        value instanceof FuncExpr ||
+        value instanceof PropRef ||
+        value instanceof ValueExpr
       ) {
         return value as BasicExpression | Aggregate
       }
       return toExpression(value)
     }
-
-    // Recursively compute the max order id from an expression tree
-    function computeOrderId(expr: BasicExpression | Aggregate): number {
-      if (!expr || typeof expr !== `object` || !(`type` in expr)) return 0
-      if ((expr as any).type === `ref`) return (expr as any).__orderId ?? 0
-      if ((expr as any).type === `val`) return 0
-      const args = (expr as any).args as Array<BasicExpression | Aggregate>
-      if (Array.isArray(args) && args.length) {
-        let maxId = 0
-        for (const a of args) {
-          const id = computeOrderId(a)
-          if (id > maxId) maxId = id
-        }
-        return maxId
-      }
-      return 0
-    }
-
-    // Convert the select object to ordered operations
-    type FieldOp = {
-      type: `field`
-      key: string
-      expr: BasicExpression | Aggregate
-      id: number
-    }
-    const fieldOps: Array<FieldOp> = []
 
     function isPlainObject(value: any): value is Record<string, any> {
       return (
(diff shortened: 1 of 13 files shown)


packages/db/src/query/builder/index.ts
```diff
@@ -418,40 +426,49 @@ export class BaseQueryBuilder<TContext extends Context = Context> {
...
-
-    // Convert the select object to use expressions, including spread sentinels
-    const select: Record<string, BasicExpression | Aggregate> = {}
-
-    // First, add spread sentinels for any tables that were spread
-    for (const spreadAlias of spreadSentinels) {
-      const sentinelKey = `__SPREAD_SENTINEL__${spreadAlias}`
-      select[sentinelKey] = toExpression(spreadAlias) // Use alias as a simple reference
+    // Helper to ensure we have a BasicExpression/Aggregate for a value
+    function toExpr(value: any): BasicExpression | Aggregate {
```
kevin-dp: Let's not introduce nested functions here but extract them outside this function into the scope of this module. Same for the `isPlainObject` and`buildNestedSelect` functions.

packages/db/src/query/builder/types.ts
```diff
@@ -230,47 +231,62 @@ export type SelectObject<
...
-  [K in keyof TSelectObject]: TSelectObject[K] extends RefProxy<infer T>
+  [K in keyof StripInternalKeys<TSelectObject>]: StripInternalKeys<TSelectObject>[K] extends RefProxy<
+    infer T
+  >
     ? T
-    : TSelectObject[K] extends Ref<infer T>
+    : StripInternalKeys<TSelectObject>[K] extends Ref<infer T>
       ? T
-      : TSelectObject[K] extends Ref<infer T> | undefined
+      : StripInternalKeys<TSelectObject>[K] extends Ref<infer T> | undefined
```
kevin-dp: Similar to my comments on #386, where are we checking for `extends Ref<infer T> | undefined` if we know it doesn't extend `Ref<infer T>` ? I believe we can just check for `extends undefined` here.

packages/db/src/query/builder/types.ts
```diff
@@ -230,47 +231,62 @@ export type SelectObject<
...
-                                : TSelectObject[K] extends { __type: infer U }
-                                  ? U
-                                  : TSelectObject[K] extends Record<string, any>
-                                    ? TSelectObject[K] extends {
-                                        __refProxy: true
-                                      }
-                                      ? never // This is a RefProxy, handled above
-                                      : ResultTypeFromSelect<TSelectObject[K]> // Recursive for nested objects
-                                    : never
+                  : StripInternalKeys<TSelectObject>[K] extends string
```
kevin-dp: The cases for string, number, boolean, null, and undefined can be combined:
```ts
StripInternalKeys<TSelectObject>[K] extends string | number | boolean | null | undefined
  ? StripInternalKeys<TSelectObject>[K]
  : ...
```

packages/db/src/query/compiler/index.ts
```diff
@@ -241,7 +242,16 @@ export function compileQuery(
     const resultPipeline = orderedPipeline.pipe(
       map(([key, [row, orderByIndex]]) => {
         // Extract the final results from __select_results and include orderBy index
-        const finalResults = (row as any).__select_results
+        const raw = (row as any).__select_results
+        const finalResults =
+          raw instanceof ValClass
```
kevin-dp: We could simplify the code here by extracting these checks into a predicate `isValue` that checks whether its an instance of `ValClass` or whether it's an objects with `type === 'val'`.

packages/db/src/query/compiler/index.ts
```diff
@@ -264,7 +274,16 @@ export function compileQuery(
   const resultPipeline: ResultStream = pipeline.pipe(
     map(([key, row]) => {
       // Extract the final results from __select_results and return [key, [results, undefined]]
-      const finalResults = (row as any).__select_results
+      const raw = (row as any).__select_results
+      const finalResults =
```
kevin-dp: Don't duplicate it here. Let's use the helper function as i proposed.

packages/db/src/query/compiler/index.ts
```diff
@@ -320,7 +339,15 @@ function processFrom(
       const extractedInput = subQueryInput.pipe(
         map((data: any) => {
           const [key, [value, _orderByIndex]] = data
-          return [key, value] as [unknown, any]
+          // Unwrap Value expressions that might have leaked through as the entire row
+          const unwrapped =
```
kevin-dp: Same

packages/db/src/query/compiler/index.ts
kevin-dp: If we want we could even introduce an additional helper function `unwrap` on top of that which will return `raw.value` if it's a value class/object, and `raw` otherwise.

packages/db/src/query/compiler/select.ts
```diff
@@ -16,67 +17,180 @@ export function processSelectToResults(
   select: Select,
   _allInputs: Record<string, KeyedStream>
 ): NamespacedAndKeyedStream {
```
kevin-dp: This function has grown too big. Let's extract the `addFromObject` function outside of this function.

packages/db/src/query/compiler/select.ts
```diff
@@ -16,67 +17,180 @@ export function processSelectToResults(
...
+            targetPath,
+            source: (row) => (row as any)[tableAlias],
+          })
+        }
+        continue
+      }
+
+      const expression = value as any
+      if (
+        expression &&
```
kevin-dp: Let's introduce a helper function for this predicate. Give it a good descriptive name because right now i'm not sure what this condition is checking.

packages/db/src/query/compiler/select.ts
```diff
@@ -16,67 +17,180 @@ export function processSelectToResults(
...
+          })
+        }
       }
     }
   }
 
+  addFromObject([], select)
+
   return pipeline.pipe(
     map(([key, namespacedRow]) => {
```
kevin-dp: Let's not do a huge map like this. Let's extract the arrow function we pass to `map` into a dedicated module-level function.
In the body of this arrow function we loop over `ops`. Let's make a helper function also for handling one such `op` such that the code becomes:
```ts
ops.forEach(op => meaningfulName(op))
```
Inside the body of the for loop we handle 2 cases, `op.kind === 'merge'` and `op.kind !== 'merge'`. Extract each part into its own helper function: `processMerge` and `processNonMergeOp`.
Also extract the `unwrapVals` helper function to its own module-level function instead of a nested function.
If we apply all these improvements this `processSelectToResults` function should become smaller, more understandable, and easier to maintain.

packages/db/src/query/compiler/select.ts
```diff
@@ -90,54 +204,110 @@ export function processSelect(
...
-  }> = []
-  const spreadAliases: Array<string> = []
-
-  for (const [alias, expression] of Object.entries(select)) {
-    if (alias.startsWith(`__SPREAD_SENTINEL__`)) {
-      // Extract the table alias from the sentinel key
-      const tableAlias = alias.replace(`__SPREAD_SENTINEL__`, ``)
-      spreadAliases.push(tableAlias)
-    } else {
+  type Op =
```
kevin-dp: This is not acceptable. This type has already been defined inside `processSelectToResults` and is defined again here. That's why we should not define types/functions inside functions but define them at the module level. Same for `addFromObject`. Huge amount of code duplication between this function and `processSelectToResults` that must be solved.

packages/db/src/query/live-query-collection.ts
```diff
@@ -249,11 +249,11 @@ export function liveQueryCollectionOptions<
 
               // Store the key of the result so that we can retrieve it in the
               // getKey function
-              resultKeys.set(value, rawKey)
+              resultKeys.set(value as unknown as object, rawKey)
```
kevin-dp: My IDE is saying `value` is of type `any`. Why do an unsafe typecast if the `value` is typed as `any` anyway. Could as well keep it as `any`.

packages/db/src/query/live-query-collection.ts
```diff
@@ -374,7 +374,8 @@ export function liveQueryCollectionOptions<
   return {
     id,
     getKey:
-      config.getKey || ((item) => resultKeys.get(item) as string | number),
+      config.getKey ||
+      ((item) => resultKeys.get(item as unknown as object) as string | number),
```
kevin-dp: Item is of type `TResult extends object` so why do we need to typecast it to an object?

packages/db/src/query/optimizer.ts
```diff
@@ -760,15 +761,19 @@ function optimizeFromWithTracking(
  * { from: users, select: { id, name } }
  * ```
  */
-function isSafeToPushIntoExistingSubquery(query: QueryIR): boolean {
-  // Check for aggregates in SELECT clause
+function isSafeToPushIntoExistingSubquery(
```
kevin-dp: I'm not a fan of functions that look like this:
```ts
if (pred)
  return false

return true
```
We can just compose one big predicate and return that:
```ts
return !pred
```

Now, i understand that for this function a big negated predicate might hamper readability. Therefore, i would rewrite it as such:
```ts
function isSafeToPushIntoExistingSubquery(
  query: QueryIR,
  whereClause: BasicExpression<boolean>,
  outerAlias: string
): boolean {
  if (
    // If the subquery has a SELECT clause, block pushdown when the WHERE references
    // fields that are computed by the subquery's SELECT (non pass-through projections).
    (
      query.select &&
      (selectHasAggregates(query.select) ||
       whereReferencesComputedSelectFields(query.select, whereClause, outerAlias))
    ) ||
    // Check for GROUP BY clause
    (query.groupBy && query.groupBy.length > 0) ||
    // Check for HAVING clause
    (query.having && query.having.length > 0) ||
    // Check for ORDER BY with LIMIT or OFFSET (dangerous combination)
    (
      query.orderBy && query.orderBy.length > 0 &&
      (query.limit !== undefined || query.offset !== undefined)
    ) ||
    // Check for functional variants that might have side effects
    query.fnSelect ||
    (query.fnWhere && query.fnWhere.length > 0) ||
    (query.fnHaving && query.fnHaving.length > 0) ||

  ) {
    return false
  }

  // If none of the unsafe conditions are present, it's safe to optimize
  return true
}
```

Even better would be to introduce helper functions for each unsafe case and use those.
Then it would become even more readable:
```ts
return !(unsafeSelect(query) || unsafeGroupBy(query) || unsafeHaving(query) || unsafeOrderBy(query) || unsafeFnSelect(query))
```

packages/db/src/query/optimizer.ts
```diff
@@ -801,6 +806,83 @@ function isSafeToPushIntoExistingSubquery(query: QueryIR): boolean {
...
+  const computed = new Set<string>()
+  for (const [key, value] of Object.entries(select)) {
+    if (key.startsWith(`__SPREAD_SENTINEL__`)) continue
+    if (value instanceof PropRef) continue
+    // Nested object or non-PropRef expression counts as computed
+    computed.add(key)
+  }
+
+  const refs: Array<PropRef> = []
+  function collectRefs(expr: any) {
```
kevin-dp: Extract this function to the module level.
Shall we also make helper functions for the 2 for loops?

packages/db/src/query/builder/ref-proxy.ts
```diff
@@ -97,6 +97,7 @@ export function createRefProxy<T extends Record<string, any>>(
         if (prop === `__refProxy`) return true
         if (prop === `__path`) return path
         if (prop === `__type`) return undefined // Type is only for TypeScript inference
+        if (prop === `__orderId`) return target.__orderId
```
kevin-dp: What is `__orderId` used for and where? I couldn't find any usage of this property in this PR.

kevin-dp (approved):
I reviewed this PR in detail. The code quality should be improved, i left comments on how to do this.

kevin-dp (changes requested):
I meant to request changes instead of approving in my previous review.

packages/db/src/query/builder/types.ts
samwillis: Although these look like they can be simplified, doing so breaks all the optionality tests.

packages/db/src/query/builder/index.ts
```diff
@@ -760,6 +741,44 @@ export class BaseQueryBuilder<TContext extends Context = Context> {
   }
 }
 
+// Helper to ensure we have a BasicExpression/Aggregate for a value
+function toExpr(value: any): BasicExpression | Aggregate {
+  if (value === undefined) return toExpression(null)
+  if (isRefProxy(value)) return toExpression(value)
```
kevin-dp: `toExpression(value)` is the default case when none of the if tests match so we can remove this if test altogether.

packages/db/src/query/builder/ref-proxy.ts
```diff
@@ -190,7 +183,8 @@ export function toExpression<T = any>(value: T): BasicExpression<T>
 export function toExpression(value: RefProxy<any>): BasicExpression<any>
 export function toExpression(value: any): BasicExpression<any> {
   if (isRefProxy(value)) {
-    return new PropRef(value.__path)
+    const expr = new PropRef(value.__path)
```
kevin-dp: nit: let's write this on one line: `return new PropRef(value.__path)`

packages/db/src/query/optimizer.ts
```diff
@@ -780,73 +781,145 @@ function optimizeFromWithTracking(
...
+  if (!query.select) return false
+
+  return (
+    selectHasAggregates(query.select) ||
+    whereReferencesComputedSelectFields(query.select, whereClause, outerAlias)
+  )
+}
+
+function unsafeGroupBy(query: QueryIR): boolean {
+  return !!(query.groupBy && query.groupBy.length > 0)
```
kevin-dp: Why are we using `!!` in this function and the ones below? This is usually used to check that something is not undefined but these are all predicates so doesn't seem to make sense.

packages/db/src/query/optimizer.ts
```diff
@@ -780,73 +781,145 @@ function optimizeFromWithTracking(
...
+    case `ref`:
+      refs.push(expr as PropRef)
+      break
+    case `func`:
+    case `agg`:
+      for (const arg of expr.args ?? []) {
+        refs.push(...collectRefs(arg))
+      }
+      break
+    case `val`:
```
kevin-dp: this case can be removed since it's the same as the default case

PR #395, opened 2025-08-10
Fix: Live query collections stuck in `initialCommit` status when source collections are preloaded after creation

### LINKED ISSUE #391: `useLiveQuery` isn't working properly with React Compiler
Hello!

I'm playing with TanStack DB - awesome job, really excellent experience. Happy to test the early release and report some issues I stumbled upon. 

Currently `useLiveQuery` isn't working properly with React Compiler enabled - it's not always updating correctly, especially when updating queries, mounting or optimistic mutations. When looked into the source code it seems like parts of `useLiveQuery` code breaks rules of React. 

Is this something that's on your radar, or React Compiler compatibility isn't guaranteed? Thanks a lot in advance!

samwillis:
stacked on #390, fairly sure that it fixes #391 

## Problem

Live query collections were getting stuck in `initialCommit` status instead of transitioning to `ready` status when their source collections were preloaded after the live query collection was created. This caused issues where:

- Live queries would have the correct data but remain in `initialCommit` status
- Components relying on `collection.status === 'ready'` would not update properly
- The live query would only transition to `ready` when mutations/remote updates occurred

This matches the user report: *"if I don't preload the collection before rendering, It won't update after being loaded. It'll only update, when some mutation/remote update happens"*

## Root Cause

The issue was in the `markReady()` method in `packages/db/src/collection.ts`. When a collection became ready for the first time, it would only emit an empty ready event if the collection was empty AND had change listeners:

```typescript
// Before (problematic)
if (this.size === 0 && this.changeListeners.size > 0) {
  this.emitEmptyReadyEvent()
}
```

This condition was too restrictive. Live query collections that depend on source collections often have data but still need to emit the ready event to properly transition their status.

## Solution

Changed the condition to emit the ready event whenever there are change listeners, regardless of collection size:

```typescript
// After (fixed)
if (this.changeListeners.size > 0) {
  this.emitEmptyReadyEvent()
}
```

This ensures that any collection with active subscribers will properly emit the ready event when it transitions to ready status, allowing dependent live query collections to also transition to ready status.

## Testing

Added a comprehensive test case that reproduces the exact scenario:

1. Creates a source collection with `startSync: false`
2. Creates a live query collection that depends on the source collection
3. Starts the live query's `preload()` promise before the source collection is ready
4. Preloads the source collection and triggers initial data load
5. Verifies that the live query properly transitions to `ready` status

The test uses explicit promise control instead of flaky timeouts to ensure reliable reproduction of the issue.

## Impact

This fix ensures that live query collections properly transition to `ready` status when their source collections are preloaded after the live query is created, resolving the issue where components would not update properly until mutations occurred.

### DIFF
--- a/packages/db/tests/query/live-query-collection.test.ts
+++ b/packages/db/tests/query/live-query-collection.test.ts
@@ -5,18 +5,6 @@ import { Query } from "../../src/query/builder/index.js"
 import { mockSyncCollectionOptions } from "../utls.js"
 import type { ChangeMessage } from "../../src/types.js"
 
-// This test file includes tests that reproduce a bug where live query collections
-// don't properly transition to 'ready' status when source collections are preloaded
-// after the live query collection is created.
-//
-// The issue: When a live query collection is created before its source collections
-// are preloaded, the live query gets stuck in 'initialCommit' status instead of
-// transitioning to 'ready' status, even though it has the correct data.
-//
-// This matches the original error report: "if I don't preload the collection before
-// rendering, It won't update after being loaded. It'll only update, when some
-// mutation/remote update happens"
-
 // Sample user type for tests
 type User = {
   id: number
--- a/packages/db/tests/query/live-query-collection.test.ts
+++ b/packages/db/tests/query/live-query-collection.test.ts
@@ -3,6 +3,19 @@ import { createCollection } from "../../src/collection.js"
 import { createLiveQueryCollection, eq } from "../../src/query/index.js"
 import { Query } from "../../src/query/builder/index.js"
 import { mockSyncCollectionOptions } from "../utls.js"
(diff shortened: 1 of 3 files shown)


packages/db/tests/query/live-query-collection.test.ts
```diff
@@ -189,4 +190,102 @@ describe(`createLiveQueryCollection`, () => {
     })
     expect(liveQuery.isReady()).toBe(false)
   })
+
+  it(`should update after source collection is loaded even when not preloaded before rendering - REPRODUCES BUG`, async () => {
```
kevin-dp: Let's remove the " - REPRODUCES BUG" in the test title. Dunno why LLMs always do this 😅

kevin-dp (approved):

PR #405, opened 2025-08-13
Modify topK and orderBy to track by key instead of by object reference

kevin-dp:
This PR modifies the topK and orderBy operators. Previoiusly, the topK was computing a topK per key and it tracked multiplicities for each value associated to that key. We were always keying by `null` in ts/db and the values were entire rows, i.e. objects. However, the objects are always different objects (even for the same row) so they were always treated as a new value.

As reported by community members, this led to a bug where the results of an `orderBy` query with e.g. `limit: 10` would end up with just a single item in the result. This would happen in the following case:
- initial query run, results in 10 rows (because of the limit)
- update to a row in that top 10: ts/db splits it into a delete of the old row and an insert of the new row
- topK operator sees the delete for the old row, but the object reference is different from the object reference of the old row it has seen before, so instead of finding the multiplicity for the old row (which is 1), it doesn't find it and thinks it hasn't seen it before so multiplicity is 0. It decrements it by 1, yielding -1. It concludes that since it was 0 before, this row was not visible and is still not visible. Nothing happens. The row is in fact never deleted. 
- topK operator sees the insert for the new (updated) row. It stores this new row with multiplicity 1.
- Now the old row is in the top 10 and the new row is in the top 10. But ts/db materializes the query results by key so you will only see it once (with the new value) in the results. However, since internally it's twice in the top K it kicked out another row, so there are only 9 distinct rows in the query result. If you repeat this update 8 more times, you end up with a single row in the query results

To solve this problem i modified the `topK` operator such that it takes a keyed stream, i.e. a stream of tuples: `[Key, Value]`. It computes the topK across all keys (i.e. not per key as it did before) and uses the key to identify values. As such, when a row is received we lookup its multiplicity based on the key instead of the object reference.

### DIFF
--- a/packages/db-ivm/tests/operators/orderByWithFractionalIndex.test.ts
+++ b/packages/db-ivm/tests/operators/orderByWithFractionalIndex.test.ts
@@ -7,7 +7,7 @@ import {
 } from "../../src/operators/index.js"
 import { orderByWithFractionalIndexBTree } from "../../src/operators/orderByBTree.js"
 import { loadBTree } from "../../src/operators/topKWithFractionalIndexBTree.js"
-import { MessageTracker, assertOnlyKeysAffected } from "../test-utils.js"
+import { MessageTracker } from "../test-utils.js"
 import type { KeyValue } from "../../src/types.js"
 
 const stripFractionalIndex = ([[key, [value, _index]], multiplicity]: any) => [
@@ -15,6 +15,10 @@ const stripFractionalIndex = ([[key, [value, _index]], multiplicity]: any) => [
   value,
   multiplicity,
 ]
+
+const stripFractionalIndexWithoutMultiplicity = (
+  r: [string, [{ id: number; value: string }, string]]
+) => [r[0], r[1][0]]
 
 beforeAll(async () => {
   await loadBTree()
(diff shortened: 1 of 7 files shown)


packages/db-ivm/tests/operators/orderByWithFractionalIndex.test.ts
```diff
@@ -353,41 +366,53 @@ describe(`Operators`, () => {
 
       graph.finalize()
 
+      const value1 = { id: 1, value: `a` }
+
       // Initial data
       input.sendData(
         new MultiSet([
-          [[`key1`, { id: 1, value: `a` }], 1],
+          [[`key1`, value1], 1],
```
samwillis: Test here is changed to use the same object ref, do we have a test where we explicitly check that it works with new ref?

samwillis (commented):
There is a failed test in the react-db package - need to look at that.

packages/db-ivm/tests/operators/orderByWithFractionalIndex.test.ts
kevin-dp: Yep, this is a remnant of when we tracked object refs. I will change it to use different objects.

kevin-dp:
> There is a failed test in the react-db package - need to look at that.

Seems to be a flake. CI is green now.

samwillis (approved):
:shipit:

Issue #407, opened 2025-08-13
`draft` in `update` collection method should be writable even if collection type is readonly


pawelblaszczyk5:
Hello 👋

As I understand the intention behind `update` collection method is to get mutable, proxied object, like immer. I think it that case the item passed to the update callback should be marked as mutable:

<img width="1380" height="122" alt="Image" src="https://github.com/user-attachments/assets/eec2b1b1-ac26-450d-8c2f-9e7de6e3cff8" />

<img width="966" height="119" alt="Image" src="https://github.com/user-attachments/assets/fbd872bb-1e0e-4b4d-89ee-09fac319e8e1" />

Currently my collection items are readonly (because that's reality, they're readonly by default). Would it be possible to mark them here as writable using some generic `DeepWritable` type? I'd be open to making a PR if that's something you're open to!

Thanks in advance!

KyleAMathews:
Yeah sure, makes sense to make sure the type is writable!

PR #408, opened 2025-08-13
fix: make item drafts mutable

pawelblaszczyk5:
Closes https://github.com/TanStack/db/issues/407

I wasn't sure whether it's fine to install dependency for it, or just copy paste it (but the type is a little bit bulky). As this type unfortunately has a lot of weird edge cases 😅 Lmk if you'd prefer to avoid the dependency!

### DIFF
--- a/.changeset/shiny-phones-stick.md
+++ b/.changeset/shiny-phones-stick.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+fix: make item drafts mutable
--- a/.changeset/shiny-phones-stick.md
+++ b/.changeset/shiny-phones-stick.md
@@ -2,4 +2,4 @@
 "@tanstack/db": patch
 ---
 
-fix: make item drafts mutable
+mark item drafts as a `mutable` type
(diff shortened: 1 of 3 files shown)


kevin-dp:
Since we're only using 1 type from it, first reaction would be to vendor it in. But, it's a type so we could make it a dev dependency in which case i'm less worried about the additional dependency (cc @samwillis).

pawelblaszczyk5:
Tbh I'm not sure about npm packaging and was wondering whether I should put this in `dependencies` or `devDependencies`. On one hand it's type only, but on the other hand it'll be part of generated `d.ts` files and user will need it to work correctly? 🤔

kevin-dp:
@pawelblaszczyk5 Indeed, i might be wrong here. It may need to be a regular dependency because the type is needed by the users of our library. In that case, i think it's best to vendor in that one type.

samwillis:
Yep, I would prefer to vendor a minimal type helper for this rather than have a dependancy. We are aiming for almost no dependancies (we only depend on standard schema at the moment).

pawelblaszczyk5:
I’ll change that then, thanks :)

pawelblaszczyk5:
I changed it - venodred the type from `type-fest`

kevin-dp (approved):

PR #410, opened 2025-08-14
Optimized order by that lazily loads more data when needed

kevin-dp:
This PR optimizes the `orderBy` operator by only loading the first `limit + offset` rows into the query pipeline. If later more rows are needed (e.g. because some rows are updated or deleted), then it lazily loads more rows until the query results contain `limit` rows again (or there are no more rows available).

This optimization is only possible when a range index over the ordered property is available.

### DIFF
--- a/.changeset/common-clubs-add.md
+++ b/.changeset/common-clubs-add.md
@@ -1,0 +1,6 @@
+---
+"@tanstack/db-ivm": patch
+"@tanstack/db": patch
+---
+
+Optimize order by to lazily load ordered data if a range index is available on the field that is being ordered on.


packages/db/src/query/compiler/order-by.ts
```diff
@@ -79,12 +101,85 @@ export function processOrderBy(
     return defaultComparator(a, b)
   }
 
+  let setSizeCallback: ((getSize: () => number) => void) | undefined
+
+  // Optimize the orderBy operator to lazily load elements
+  // by using the range index of the collection.
+  // Only for orderBy clause on a single column for now (no composite ordering)
```
samwillis, line 108: Worth making a tracking issue for this - we will forget.

I wander if a first step is to use the index for the first of the columns in a composite orderBy.

packages/db/src/query/live-query-collection.ts
```diff
@@ -307,18 +312,21 @@ export function liveQueryCollectionOptions<
 
       let subscribedToAllCollections = false
 
-      const maybeRunGraph = () => {
+      const maybeRunGraph = (callback?: () => boolean) => {
```
samwillis, line 329: Could we add a comment on what the callback is for?

I think it's for when loading more items into the pipeline, we can return `false` to indicate that the query is still being run - this is returned from `loadMoreIfNeeded`

packages/db/src/query/live-query-collection.ts
```diff
@@ -437,12 +460,117 @@ export function liveQueryCollectionOptions<
...
+              loadNextItems(n)
+            }
+
+            // Indicate that we're done loading data if we didn't need to load more data
+            return n === 0
+          }
+
+          // Keep track of the keys we've sent
+          // and also the biggest value we've sent so far
+          // and the smallest value we've sent so far
```
samwillis: we're not tracking the smallest any more

packages/db/tests/query/order-by.test.ts
```diff
@@ -342,6 +342,8 @@ function createOrderByTests(autoIndex: `off` | `eager`): void {
         expect(results.map((r) => r.salary)).toEqual([65000, 60000, 55000])
       })
 
+      // TODO: also test live updates with lazy loaded orderBy
```
samwillis: Yep, we need this test

packages/db/tests/query/order-by.test.ts
```diff
@@ -1228,6 +1230,6 @@ function createOrderByTests(autoIndex: `off` | `eager`): void {
 }
 
 describe(`Query2 OrderBy Compiler`, () => {
-  createOrderByTests(`off`)
+  // createOrderByTests(`off`)
```
samwillis: I assume the tests still pass with auto indexing off?

packages/db/src/query/compiler/order-by.ts
kevin-dp: That's smart! That would already be able to reduce the rows that are loaded initially. And the rest of the filtering would then happen inside the pipeline.

packages/db/src/query/live-query-collection.ts
kevin-dp: I've added a comment explaining this callback. It's called after the pipeline run which we need such that we can check whether the query has enough results or not (i.e. when using orderBy with a limit). We need to do it like because we cannot rely on the `orderBy` operator pulling data in when needed because the pipeline execution might never reach the `orderBy` operator if the data gets filtered out.

For example, if have limit to 20 results, then we only load the first 20 rows. But there could be a filter in the pipeline that filters out the first 40 rows. In which case we would never reach the `orderBy` operator. So with this solution, after the initial load of 20 rows and the pipeline run, the callback will notice that the result set is still empty. So it would load 20 more results. It would run the pipeline again, notice the empty result set, and load 20 more items. This time they don't get filtered out. So the result set contains 20 rows. The callback checks the result set, notices it has 20 rows so it doesn't load more data and marks the collection as ready.

packages/db/tests/query/order-by.test.ts
kevin-dp: This todo is obsolete as it is implemented by these tests:
- applies incremental insert of a new row before the topK correctly
- applies incremental insert of a new row inside the topK correctly
- applies incremental insert of a new row after the topK correctly
- applies incremental update of a row inside the topK correctly
- applies incremental delete of a row in the topK correctly

packages/db/src/query/compiler/order-by.ts
kevin-dp: Thinking a bit more about this i don't think this works. Say that we want the first 20 rows based on two columns: `[A, B]`. Now, assume that all rows have the same value for column `A`. If we only take into account column `A` to lazily load rows that means that we will load 20 arbitrary rows because all rows are considered equal (since they have the same value for `A`). So we cannot ignore the remaining columns in the order because we need them to tie-break equal rows. We could however lazily load the first 20 rows based on column A and make sure to also load all rows that are equal to the 20th row. That will ensure we have all equal rows and we can then let the pipeline tie-break them.

bradleybernard:
I upgraded and ran into an issue that I think might be related to this, but not too sure....
https://github.com/TanStack/db/issues/435

PR #412, opened 2025-08-15
add support for electric must-refetch

samwillis:


KyleAMathews (approved):
Is resetting to loading to also reset the query pipeline?

balegas (approved):
The code looks good, but I have a couple of observations that might make or not make sense with the collection.
- The shape is not invalid at the point of must-refetch, could you mask the loading state by keeping the state of the shape while you refetch it?
- Why do you need a new `loading` state instead of resetting to the initial state of the shape, e.g. marking it not ready?

samwillis:
@KyleAMathews 
> Is resetting to loading to also reset the query pipeline?

Good question, I don't think it will. We need a way to handle that in the live query. I think, but need to check, that the live query will not update until it goes back to ready.

@balegas 
> The shape is not invalid at the point of must-refetch, could you mask the loading state by keeping the state of the shape while you refetch it?

Yes, we could. The intention here was that the live query would see the loading state and not commit its changes - essentially masking it at the live query level.

I think there is a wider question here of is we want to go back into "loading" when this happen, but also if there are other reasons any other sync engine wants to go back into loading.

One thought I had was for partitioned collections, when requesting a partition that requires to be sure a new query is "read" we likely need to go back into loading. But again that maybe want to be more fine grained, so only triggering loading for live queries that span a loading partition.

My thinking here is that we are going from, everything breaks at the moment, to a first stop towards making this work. we can refine it.

KyleAMathews:
Another way of doing this could be to delete every key and then reinsert everything?

samwillis:
Yep, I'm now rethinking this, we should "schedule a truncate" and then do that at the bending of the next commit. It's the pattern we used in pglite.

We can go back and think about loading state later on - we need to, but no rush now.

KyleAMathews:
Yeah, maybe even just a `truncate` operator that sync engines can use so:

```ts
begin()
truncate()
insert()
insert()
insert()
commit()
```

samwillis:
Now has a transactional `truncate()`

.changeset/electric-must-refetch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tanstack/electric-db-collection": patch
+---
+
+Add must-refetch message handling to clear synced data and reset collection to loading state.
```
KyleAMathews:
```suggestion
Add must-refetch message handling to clear synced data and re-sync collection data from server.
```

packages/electric-db-collection/tests/electric.test.ts
```diff
@@ -225,6 +225,57 @@ describe(`Electric Integration`, () => {
     expect(collection.state).toEqual(new Map())
   })
 
+  it(`should handle must-refetch by clearing synced data and resetting to loading state`, () => {
```
KyleAMathews:
```suggestion
  it(`should handle must-refetch by clearing synced data and re-syncing`, () => {
```

KyleAMathews (approved):
Looks great!

PR #428, opened 2025-08-20
feat: add support for dates to max and min functions.

MAST1999:
I tried not to change too many things.
If something doesn't look right, or we need some tests that are missing, I'm happy to try and make those changes.

BTW, I tried adding date support to `eq` as well but, I couldn't figure out a way to make it work. I made a change in `packages/db/src/query/compiler/evaluators.ts` and, after checking it, I can see that it's converting date to number and comparing it correctly.

After I added a test to confirm it, the test kept failing :(
So if it's something that needs more work I can take it out and follow up later.

Feel free to make any necessary changes :)

### DIFF
--- a/packages/db-ivm/src/operators/groupBy.ts
+++ b/packages/db-ivm/src/operators/groupBy.ts
@@ -211,19 +211,19 @@ export function avg<T>(
  * Creates a min aggregate function that computes the minimum value in a group
  * @param valueExtractor Function to extract a numeric value from each data entry
  */
-export function min<T>(
-  valueExtractor: (value: T) => number = (v) => v as unknown as number
-): AggregateFunction<T, number, number> {
+export function min<T, R, V>(
+  valueExtractor: (value: T) => V = (v) => v as unknown as V
+): AggregateFunction<T, R, V> {
   return {
     preMap: (data: T) => valueExtractor(data),
-    reduce: (values: Array<[number, number]>) => {
-      let minValue = Number.POSITIVE_INFINITY
+    reduce: (values) => {
+      let minValue = Number.POSITIVE_INFINITY as V
       for (const [value, _multiplicity] of values) {
-        if (value < minValue) {
+        if (Number(value) < Number(minValue)) {
           minValue = value
         }
       }
-      return minValue === Number.POSITIVE_INFINITY ? 0 : minValue
+      return minValue === Number.POSITIVE_INFINITY ? (0 as V) : minValue
     },
   }
 }
@@ -232,19 +232,19 @@ export function min<T>(
  * Creates a max aggregate function that computes the maximum value in a group
  * @param valueExtractor Function to extract a numeric value from each data entry
  */
-export function max<T>(
-  valueExtractor: (value: T) => number = (v) => v as unknown as number
-): AggregateFunction<T, number, number> {
+export function max<T, R, V>(
+  valueExtractor: (value: T) => V = (v) => v as unknown as V
+): AggregateFunction<T, R, V> {
   return {
     preMap: (data: T) => valueExtractor(data),
-    reduce: (values: Array<[number, number]>) => {
-      let maxValue = Number.NEGATIVE_INFINITY
+    reduce: (values) => {
+      let maxValue = Number.NEGATIVE_INFINITY as V
       for (const [value, _multiplicity] of values) {
-        if (value > maxValue) {
+        if (Number(value) > Number(maxValue)) {
           maxValue = value
         }
       }
-      return maxValue === Number.NEGATIVE_INFINITY ? 0 : maxValue
+      return maxValue === Number.NEGATIVE_INFINITY ? (0 as V) : maxValue
     },
   }
 }
--- a/packages/db/src/query/compiler/evaluators.ts
+++ b/packages/db/src/query/compiler/evaluators.ts
@@ -142,8 +142,14 @@ function compileFunction(func: Func, isSingleRow: boolean): (data: any) => any {
       const argA = compiledArgs[0]!
       const argB = compiledArgs[1]!
       return (data) => {
-        const a = argA(data)
-        const b = argB(data)
+        let a = argA(data)
+        let b = argB(data)
+        if (a instanceof Date) {
+          a = a.getTime()
+        }
+        if (b instanceof Date) {
+          b = b.getTime()
+        }
         return a === b
       }
     }
--- a/packages/db/src/query/compiler/group-by.ts
+++ b/packages/db/src/query/compiler/group-by.ts
@@ -347,6 +347,20 @@ function getAggregateFunction(aggExpr: Aggregate) {
     return typeof value === `number` ? value : value != null ? Number(value) : 0
   }
 
+  // Create a value extractor function for the expression to aggregate
+  const valueExtractorWithDate = ([, namespacedRow]: [
+    string,
+    NamespacedRow,
+  ]) => {
+    const value = compiledExpr(namespacedRow)
+    // Ensure we return a number for numeric aggregate functions
+    return typeof value === `number` || value instanceof Date
+      ? value
+      : value != null
+        ? Number(value)
+        : 0
+  }
+
   // Return the appropriate aggregate function
   switch (aggExpr.name.toLowerCase()) {
     case `sum`:
@@ -356,9 +370,9 @@ function getAggregateFunction(aggExpr: Aggregate) {
     case `avg`:
       return avg(valueExtractor)
     case `min`:
-      return min(valueExtractor)
+      return min(valueExtractorWithDate)
     case `max`:
-      return max(valueExtractor)
+      return max(valueExtractorWithDate)
     default:
       throw new UnsupportedAggregateFunctionError(aggExpr.name)
   }
--- a/packages/db-ivm/tests/operators/groupBy.test.ts
+++ b/packages/db-ivm/tests/operators/groupBy.test.ts
@@ -454,6 +454,7 @@ describe(`Operators`, () => {
       const input = graph.newInput<{
         category: string
         amount: number
+        date: Date
       }>()
       let latestMessage: any = null
 
@@ -461,6 +462,8 @@ describe(`Operators`, () => {
         groupBy((data) => ({ category: data.category }), {
           minimum: min((data) => data.amount),
           maximum: max((data) => data.amount),
+          min_date: min((data) => data.date),
+          max_date: max((data) => data.date),
         }),
         output((message) => {
           latestMessage = message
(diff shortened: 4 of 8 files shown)


MAST1999:
I also checked with `lt`, `gt` and the whole family and I think it's working. Though I recommend someone double check.

MAST1999:
OK, managed to make it work.

The PR is ready for review.

samwillis:
Hey @MAST1999 just letting you know this hasn't dropped off my radar. I've asked @kevin-dp to review it as he worked on the groupBy operator recently.

packages/db-ivm/src/operators/groupBy.ts
```diff
@@ -211,19 +211,19 @@ export function avg<T>(
  * Creates a min aggregate function that computes the minimum value in a group
  * @param valueExtractor Function to extract a numeric value from each data entry
  */
-export function min<T>(
-  valueExtractor: (value: T) => number = (v) => v as unknown as number
-): AggregateFunction<T, number, number> {
+export function min<T, R, V>(
```
kevin-dp: The type parameter `R` is never used.

packages/db-ivm/src/operators/groupBy.ts
```diff
@@ -211,19 +211,19 @@ export function avg<T>(
...
+): AggregateFunction<T, R, V> {
   return {
     preMap: (data: T) => valueExtractor(data),
-    reduce: (values: Array<[number, number]>) => {
-      let minValue = Number.POSITIVE_INFINITY
+    reduce: (values) => {
+      let minValue = Number.POSITIVE_INFINITY as V
       for (const [value, _multiplicity] of values) {
-        if (value < minValue) {
+        if (Number(value) < Number(minValue)) {
```
kevin-dp: I'm not a fan of casting arbitrary values into a `Number`. This works for `Date` objects but may not work in the general case.

kevin-dp (changes requested):
I left some comments with regard to the code changes. I believe the changes here are indeed enough for `Date`s to work with aggregation functions.

On the longer term, i'm not sure if this is the approach we want to take. Turning the value into a number with `Number` seems brittle and works for `Date`s but may not work for other data types.

We may want a bigger refactor such that `min` and `max` work on any data type that has a total ordering. This means we need a custom comparator function as well as a bottom value and a top value for that specific data type. For example, for numbers the comparator is `<`, the bottom value is `-Inf` and the top value is `+Inf`.

packages/db-ivm/src/operators/groupBy.ts
```diff
@@ -211,19 +211,19 @@ export function avg<T>(
...
           minValue = value
         }
       }
-      return minValue === Number.POSITIVE_INFINITY ? 0 : minValue
+      return minValue === Number.POSITIVE_INFINITY ? (0 as V) : minValue
```
kevin-dp: let's type `let minValue: Number | V` such that we don't need the typecast here

samwillis:
I think in future a larger refactor may be good to support a wider range of types in min/max, but in the sort term this looks like a good route to getting them working for Date. Numbers and dates are by far the most common use for these functions, so lets concentrate on getting that in soon.

packages/db-ivm/src/operators/groupBy.ts
MAST1999: The issue I have here is that I need the returned type to be the type passed in. So if they pass in a Date, I want the return type to be Date as well. That's why I ended up casting as `V`.

I couldn't figure out what's the best way to handle this `minValue` and `maxValue` situation since if the user returns a number we can return 0 as a valid response, but what should we do if they pass a date? Do we still return a 0, because then the response type would always be `number | Date` which is annoying?

packages/db-ivm/src/operators/groupBy.ts
MAST1999: I needed to pass the `V` generic to `AggregateFunction`, so I added `R` as well to pass it, since `V` was after it.

I think I can replace it with `never` in `AggregateFunction` and I don't think that would break anything if we still want to remove it.

packages/db-ivm/src/operators/groupBy.ts
MAST1999: Resolving it based on @samwillis's [message](https://github.com/TanStack/db/pull/428#issuecomment-3223967100).

samwillis:
Hi @MAST1999, I'm finally catching up on this! I hope you don't mind that I have made some tweaks and pushed to your branch. You had identified much of what needed to happen, and while looking through this it's become apparent we actually have a small bug with range lookups on Date properties.

@kevin-dp could you take a look at this PR, I think its all good now but I want you to cast your eyes over the normalisation of keys for the btree index. Even though we had a comparison function that supported dates they would still be inserted using the Date object ref, and so range queries would not work. This PR changes it to normalise dates to a number as the key in the index so that the same date is inserted under the same key. As you did this initial work I want to check that we are making the correct changes. May be worth discussing on Monday.

@MAST1999 I removed the normalisation of other objects to JSON - this is an edge case that I don't think we need to support, and stringifying to json is very slow. It would be better to just go on object ref for them (essentually random order).

packages/db/src/query/compiler/evaluators.ts
```diff
@@ -144,6 +144,9 @@ function compileFunction(func: Func, isSingleRow: boolean): (data: any) => any {
       return (data) => {
         const a = argA(data)
         const b = argB(data)
+        if (a instanceof Date && b instanceof Date) {
```
kevin-dp: I don't like special casing Dates here. As i suggested in a previous comment in `btree-index.ts`, the `normalizeMapKey` method should be a utility function (perhaps best to rename it to `normalizeValue`) and here we should then simply do `return normalizeValue(a) == normalizeValue(b)`.

packages/db/src/indexes/btree-index.ts
```diff
@@ -289,4 +299,15 @@ export class BTreeIndex<
   get valueMapData(): Map<any, Set<TKey>> {
     return this.valueMap
   }
+
+  /**
+   * Normalizes values for use as Map keys to ensure proper equality and range queries
+   * For Date objects, uses timestamp.
+   */
+  private normalizeMapKey(value: any): any {
```
kevin-dp: This should be a standalone utility function instead of a class method since it doesn't use any of the class' fields or methods.

packages/db/src/query/compiler/group-by.ts
```diff
@@ -349,6 +349,20 @@ function getAggregateFunction(aggExpr: Aggregate) {
     return typeof value === `number` ? value : value != null ? Number(value) : 0
   }
 
+  // Create a value extractor function for the expression to aggregate
+  const valueExtractorWithDate = ([, namespacedRow]: [
+    string,
+    NamespacedRow,
+  ]) => {
+    const value = compiledExpr(namespacedRow)
+    // Ensure we return a number for numeric aggregate functions
```
kevin-dp: Comment isn't entirely correct anymore since we also return Date objects here

packages/db/src/utils/comparison.ts
```diff
@@ -51,7 +51,7 @@ export const ascComparator = (a: any, b: any, opts: CompareOptions): number => {
 
   // If both are dates, compare them
   if (a instanceof Date && b instanceof Date) {
-    return a.getTime() - b.getTime()
+    return a.valueOf() - b.valueOf()
```
kevin-dp: Why this change? `getTime` and `valueOf` both return the epoch time so this change doesn't make any difference. imo `getTime()` is more descriptive than `valueOf`.

packages/db-ivm/src/operators/groupBy.ts
```diff
@@ -214,19 +214,21 @@ export function avg<T>(
  * Creates a min aggregate function that computes the minimum value in a group
  * @param valueExtractor Function to extract a numeric value from each data entry
  */
-export function min<T>(
-  valueExtractor: (value: T) => number = (v) => v as unknown as number
-): AggregateFunction<T, number, number> {
+export function min<T, V extends number | Date | bigint>(
+  valueExtractor: (value: T) => V = (v) => v as unknown as V
```
kevin-dp: The type of this default extractor is not correct. The function is an identity function where T can be any type and we cast it to V that can only be number, Date, or bigint. This comment also applies to the `max` function.

packages/db-ivm/src/operators/groupBy.ts
```diff
@@ -214,19 +214,21 @@ export function avg<T>(
...
+): AggregateFunction<T, V, V> {
   return {
     preMap: (data: T) => valueExtractor(data),
-    reduce: (values: Array<[number, number]>) => {
-      let minValue = Number.POSITIVE_INFINITY
+    reduce: (values) => {
+      let minValue: V | undefined
       for (const [value, _multiplicity] of values) {
-        if (value < minValue) {
+        if (minValue && value < minValue) {
```
kevin-dp: Both branches have the same body. We can rewrite this to:
```ts
if (!minValue || value < minValue) {
  minValue = value
}
```

packages/db-ivm/src/operators/groupBy.ts
```diff
@@ -214,19 +214,21 @@ export function avg<T>(
...
+          minValue = value
+        } else if (!minValue) {
           minValue = value
         }
       }
-      return minValue === Number.POSITIVE_INFINITY ? 0 : minValue
+      return minValue ?? (0 as V)
```
kevin-dp: Why is 0 the minValue if there's no data? Shouldn't it be -Inf ?

packages/db-ivm/src/operators/groupBy.ts
kevin-dp: This comment also applies to max

packages/db-ivm/src/operators/groupBy.ts
kevin-dp: This comment also applies to max

kevin-dp (changes requested):
Overall the changes look good to me. Left some comments for small improvements.

packages/db-ivm/src/operators/groupBy.ts
samwillis: I decided to address this in this PR, it now returns `undefined` when there is no rows to aggregate

packages/db-ivm/src/operators/groupBy.ts
```diff
@@ -210,48 +210,64 @@ export function avg<T>(
...
-    preMap: (data: T) => valueExtractor(data),
+    preMap: (data: T) => extractor(data),
     reduce: (values) => {
       let minValue: V | undefined
       for (const [value, _multiplicity] of values) {
-        if (minValue && value < minValue) {
-          minValue = value
-        } else if (!minValue) {
+        if (!minValue || !value || value < minValue) {
           minValue = value
```
kevin-dp, line 237: Is it correct to add `!value` here? It wasn't in the original if test and this means that if we have an existing `minValue` but `value = undefined` (or `value = 0` because `!0 === true`) we will overwrite the `minValue` with `undefined`/`0`.

kevin-dp: Should it be `if (!minValue || (value && value < minValue))` ?

kevin-dp: These comments also apply to `max`

kevin-dp (commented):
Thanks for addressing my feedback. I left 1 comment that is important to double check whether it is correct or not before we merge.

PR #432, opened 2025-08-21
Refactor live query collection

kevin-dp:
This PR refactors the live query collection. A lot of the main logic was burried in a single big function that was closing over a lot of state. I have refactored this to use classes such that the state becomes class fields and the nested functions become class methods. This results in code being logically split into dedicated small files based on their purpose.


packages/db/src/query/live/LiveQueryCollectionConfig.ts
```diff
@@ -0,0 +1,73 @@
...
+ *       id: comment.id,
+ *       content: comment.content,
+ *       authorName: user.name,
+ *     })),
+ *   // getKey is optional - defaults to using stream key
+ *   getKey: (item) => item.id,
+ * }
+ * ```
+ */
+export interface LiveQueryCollectionConfig<
```
samwillis: Can this be in the types.ts file?

packages/db/src/query/live/CollectionConfigBuilder.ts
samwillis, line 1: we use cabal-case for file names, `collection-config-builder.ts`

packages/db/src/query/live/CollectionSubscriber.ts
samwillis, line 1: rename file `collection-subscriber.ts`

Issue #422, opened 2025-08-19
SQL STRING_AGG function in groupBy


Uziniii:
```ts
new Query().from({ brand: brandsCollection })
      .innerJoin({ productBrand: productBrandsCollection }, ({ productBrand, brand }) => eq(productBrand.brandId, brand.id))
      .groupBy(({ productBrand }) => productBrand.productId)
      .select(({ brand, productBrand }) => ({
        id: productBrand.productId,
        brands: stringAgg(brand.name, ", ", { orderBy: { val: productBrand.order, order: "asc" | "desc" }}),
      }))
```
result
```ts
[
  { id: 1, brands: ["coca", "fanta"] },
]
```

Uziniii:
Or an arrayAgg that transform brands in ["coca", "fanta"] and that can supports number too ([1,2,3])

PR #433, opened 2025-08-21
WIP adding a list aggregate for puting all of the fields or table into an array with groupBy

Uziniii:
#422 

not all tests are correct and theres change to be done with AggregateFunction type cause it doesnt return initial type value

the pull request is only for review i dont know how to implement the type in the AggregateFunction type and for knowing if a features like this can and would be added to tanstack db because i really need it and i thinks its necessary

### DIFF
--- a/.changeset/spotty-dogs-fry.md
+++ b/.changeset/spotty-dogs-fry.md
@@ -1,0 +1,47 @@
+---
+"@tanstack/react-db": patch
+---
+
+Expand `useLiveQuery` callback to support conditional queries and additional return types, enabling the ability to temporarily disable the query.
+
+**New Features:**
+
+- Callback can now return `undefined` or `null` to temporarily disable the query
+- Callback can return a pre-created `Collection` instance to use it directly
+- Callback can return a `LiveQueryCollectionConfig` object for advanced configuration
+- When disabled (returning `undefined`/`null`), the hook returns a specific idle state
+
+**Usage Examples:**
+
+```ts
+// Conditional queries - disable when not ready
+const enabled = useState(false)
+const { data, state, isIdle } = useLiveQuery((q) => {
+  if (!enabled) return undefined  // Disables the query
+  return q.from({ users }).where(...)
+}, [enabled])
+
+/**
+ * When disabled, returns:
+ * {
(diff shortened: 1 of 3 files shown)


MAST1999:
Hey, I'm in dire need of this operator :smile: 

Do you need any help or testing? I have sometime this weekend.

packages/db/src/query/compiler/group-by.ts
```diff
@@ -344,7 +344,7 @@ function getAggregateFunction(aggExpr: Aggregate) {
   const valueExtractor = ([, namespacedRow]: [string, NamespacedRow]) => {
     const value = compiledExpr(namespacedRow)
     // Ensure we return a number for numeric aggregate functions
-    return typeof value === `number` ? value : value != null ? Number(value) : 0
```
MAST1999: I'm not sure if this is a good change, since the other function rely on the value being a number.

Might be a good idea to add another value extractor specific to your need. At least that's what I've done in my PR for `min` and `max` to support dates.

Uziniii: Yes that a workaround for now I didn't have the time to do it and I needed it to work fast for my project

samwillis: I agree this should be a separate extractor

samwillis (changes requested):
Hey @Uziniii 
Thanks for picking this up - I'm supportive of adding this, although it needs a little more work first.
This will work well for small groups, but will not be ideal for large hierarchical aggregations, for that we have a plan to make them fully incidental all each level.
The key thing is tidying up the value extractor code.
We should also maybe add some type tests ( in `group-by.test-d.ts`) to ensure that the value extracted in the `list` results in the current type on the resulting collection, let me know when you are further along and I will take another look.

Uziniii:
How should the value extractor be named ?

MAST1999:
> How should the value extractor be named ?

I think the older one should be named `numberExtractor` since it always extracts numbers. This new one can be `valueExtractor`, since it just returns the value.

MAST1999:
Hey @Uziniii did you close this on purpose?

PR #434, opened 2025-08-22
fix: handle Temporal objects correctly in proxy deepClone and deepEqual

KyleAMathews:
## Summary
- Fixes an issue where Temporal objects (like `Temporal.ZonedDateTime`) were being converted to empty objects during proxy tracking operations
- Adds proper support for all Temporal API objects in the proxy's `deepClone` and `deepEqual` functions
- Prevents unnecessary proxy creation for immutable Temporal objects

## Problem
When using Temporal objects with TanStack DB's proxy system, the `deepClone` function was attempting to clone Temporal objects as regular objects by copying their enumerable properties. Since Temporal objects store their data internally (not as enumerable properties), this resulted in empty objects being produced.

## Solution
This PR adds special handling for Temporal objects:

1. **In `deepClone`**: Detects Temporal objects by their `Symbol.toStringTag` and returns them directly (since they're immutable)
2. **In `deepEqual`**: Adds proper equality checking using Temporal's built-in `equals()` method where available
3. **In proxy get handler**: Prevents creating proxies for Temporal objects since they're immutable

## Test plan
- [x] Added comprehensive test coverage for `Temporal.ZonedDateTime`, `Temporal.PlainDate`, and `Temporal.Duration`
- [x] All existing proxy tests pass

### DIFF
--- a/packages/db/src/proxy.ts
+++ b/packages/db/src/proxy.ts
@@ -133,6 +133,30 @@ function deepClone<T extends unknown>(
     return clone as unknown as T
   }
 
+  // Handle Temporal objects
+  // Check if it's a Temporal object by checking for the Temporal brand
+  if (typeof (obj as any)[Symbol.toStringTag] === `string`) {
+    const tag = (obj as any)[Symbol.toStringTag]
+    const temporalTypes = [
+      `Temporal.PlainDate`,
+      `Temporal.PlainTime`,
+      `Temporal.PlainDateTime`,
+      `Temporal.ZonedDateTime`,
+      `Temporal.Instant`,
+      `Temporal.PlainYearMonth`,
+      `Temporal.PlainMonthDay`,
+      `Temporal.Duration`,
+      `Temporal.TimeZone`,
+      `Temporal.Calendar`,
+    ]
+
+    if (temporalTypes.includes(tag)) {
+      // Temporal objects are immutable, so we can return them directly
+      // This preserves all their internal state correctly
+      return obj
+    }
+  }
+
   const clone = {} as Record<string | symbol, unknown>
   visited.set(obj as object, clone)
 
@@ -242,6 +266,47 @@ function deepEqual<T>(a: T, b: T): boolean {
     }
 
     return true
+  }
+
+  // Handle Temporal objects
+  // Check if both are Temporal objects of the same type
+  const aTag = (a as any)[Symbol.toStringTag]
+  const bTag = (b as any)[Symbol.toStringTag]
+
+  if (typeof aTag === `string` && typeof bTag === `string`) {
+    const temporalTypes = [
+      `Temporal.PlainDate`,
+      `Temporal.PlainTime`,
+      `Temporal.PlainDateTime`,
+      `Temporal.ZonedDateTime`,
+      `Temporal.Instant`,
+      `Temporal.PlainYearMonth`,
+      `Temporal.PlainMonthDay`,
+      `Temporal.Duration`,
+      `Temporal.TimeZone`,
+      `Temporal.Calendar`,
+    ]
+
+    const aIsTemporal = temporalTypes.includes(aTag)
+    const bIsTemporal = temporalTypes.includes(bTag)
+
+    if (aIsTemporal && bIsTemporal) {
+      // If they're different Temporal types, they're not equal
+      if (aTag !== bTag) return false
+
+      // Use Temporal's built-in equals method if available
+      if (typeof (a as any).equals === `function`) {
+        return (a as any).equals(b)
+      }
+
+      // For Duration, use toString comparison as it doesn't have equals
+      if (aTag === `Temporal.Duration`) {
+        return (a as any).toString() === (b as any).toString()
+      }
+
+      // Fallback to toString comparison for other types
+      return (a as any).toString() === (b as any).toString()
+    }
   }
 
   // Handle plain objects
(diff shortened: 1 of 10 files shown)


obeattie:
This works beautifully. Thanks v much for jumping on this so quickly 🏎️

obeattie:
Actually, I spoke too soon 😓 Unless I'm doing something wrong, I'm seeing some different odd behaviour now, and I put together [this little reproduction](https://stackblitz.com/edit/tanstack-db-temporal-gz42ra3g?file=src%2FApp.tsx).

If you click the button after the page loads to mutate the data in the store, you will see that the data that's returned from the live query gets out of sync with the data retrieved directly from the store, even though a refetch is happening.

KyleAMathews:
Ok interesting — thanks for the further reproduction! Me or someone else will look into this next week.

obeattie:
Thanks, I appreciate it. Looks like the code in `collection.ts` to check deep equality is the culprit – it will need some logic similar to what you added in `proxy.ts`. I just copy-pasted it across and the problem goes away.

https://github.com/TanStack/db/blob/25c43cdae02c586fcad2a1b661ab620643093bcf/packages/db/src/collection.ts#L1712-L1733

KyleAMathews:
Ah nice!

obeattie:
I took a stab at fixing this, and here's a commit you can cherry-pick onto this branch (if you agree with it!) that makes the issues I was seeing go away: https://github.com/obeattie/tanstack-db/commit/c7cf39cced6316c802f70619be3dd9391db31aeb

I noticed there are a bunch of different functions for checking deep equality, so I combined them in that commit.

packages/db/src/proxy.ts
```diff
@@ -133,6 +133,30 @@ function deepClone<T extends unknown>(
...
+    const temporalTypes = [
+      `Temporal.PlainDate`,
+      `Temporal.PlainTime`,
+      `Temporal.PlainDateTime`,
+      `Temporal.ZonedDateTime`,
+      `Temporal.Instant`,
+      `Temporal.PlainYearMonth`,
+      `Temporal.PlainMonthDay`,
+      `Temporal.Duration`,
+      `Temporal.TimeZone`,
```
kevin-dp: These 2 temporals (`TimeZone` and `Calendar`) don't seem to exist according to the docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal

packages/db/src/proxy.ts
```diff
@@ -133,6 +133,30 @@ function deepClone<T extends unknown>(
     return clone as unknown as T
   }
 
+  // Handle Temporal objects
+  // Check if it's a Temporal object by checking for the Temporal brand
+  if (typeof (obj as any)[Symbol.toStringTag] === `string`) {
```
kevin-dp: We should extract the check to a helper function `isTemporalObject` that does this logic of checking the `toStringTag`. We could also use a plain `instanceof` check unless these temporals may come from a different realm (not sure?).

packages/db/src/proxy.ts
```diff
@@ -244,6 +268,47 @@ function deepEqual<T>(a: T, b: T): boolean {
     return true
   }
 
+  // Handle Temporal objects
+  // Check if both are Temporal objects of the same type
+  const aTag = (a as any)[Symbol.toStringTag]
+  const bTag = (b as any)[Symbol.toStringTag]
+
+  if (typeof aTag === `string` && typeof bTag === `string`) {
```
kevin-dp: Let's not duplicate this temporal checking logic. We should use a helper function.

packages/db/src/proxy.ts
```diff
@@ -244,6 +268,47 @@ function deepEqual<T>(a: T, b: T): boolean {
...
+      `Temporal.TimeZone`,
+      `Temporal.Calendar`,
+    ]
+
+    const aIsTemporal = temporalTypes.includes(aTag)
+    const bIsTemporal = temporalTypes.includes(bTag)
+
+    if (aIsTemporal && bIsTemporal) {
+      // If they're different Temporal types, they're not equal
+      if (aTag !== bTag) return false
```
kevin-dp: Let's also make a helper function to get the tag of a temporal

packages/db/src/proxy.ts
```diff
@@ -244,6 +268,47 @@ function deepEqual<T>(a: T, b: T): boolean {
...
+
+      // Use Temporal's built-in equals method if available
+      if (typeof (a as any).equals === `function`) {
+        return (a as any).equals(b)
+      }
+
+      // For Duration, use toString comparison as it doesn't have equals
+      if (aTag === `Temporal.Duration`) {
```
kevin-dp: Since the default is to fallback to `toString` comparison (L308) there's no need for this if test.

kevin-dp: Actually, i wouldn't fallback to string comparison here but rather use the static `compare` method: `Temporal.Duration.compare(a, b) === 0`

packages/db/src/proxy.ts
```diff
@@ -741,12 +806,31 @@ export function createChangeProxy<
           return value.bind(ptarget)
         }
 
-        // If the value is an object, create a proxy for it
+        // Check if it's a Temporal object - don't proxy them as they're immutable
+        const isTemporalObject =
```
kevin-dp: Use a helper function instead of duplicating this code.

kevin-dp:
@obeattie Your commit is spot on! As far as i can tell, you replaced the different versions of `deepEquals` spread across the code base by a single `deepEquals` version that handles `Temporals` properly and that fixes your issue, right?

Thanks for jumping in and fixing this 👍

kevin-dp:
@obeattie I cherry-picked your commit and i further cleaned this PR. Your stackblitz example now works correctly. Before i merge this I'd like to add a unit test though that reproduces the original bug to avoid future regressions.

obeattie:
Thanks for picking this up @kevin-dp!

> As far as i can tell, you replaced the different versions of `deepEquals` spread across the code base by a single `deepEquals` version that handles `Temporals` properly and that fixes your issue, right?

Yes that's exactly right. In the longer term, and as I think is touched upon in #428's discussion, I think it might be good to allow users to provide their own comparators for different types. If that functionality existed I would probably have just used that, though it's also easy to make a case that `Date`s and `Temporal` objects ought to work out of the box.

Appreciate you both for fixing this so quickly!

kevin-dp:
I added a unit test that reproduces the bug when ran before @obeattie's commit and passes after that commit. All good on my side to merge it. I'll wait untill this afternoon to give @KyleAMathews a chance to have a look at it.

kevin-dp (approved):

KyleAMathews:
Looks great! I found another bug around live queries not updating when I modified a (regular) date which this I tested this PR on and it also fixed so glad you cleaned up our sprawling deep equals mess @obeattie :-D Let's get this out there

Issue #182, opened 2025-06-17
add a `.findOne()` to query build to return the first item


KyleAMathews:
e.g. instead of:

```tsx
const { data: users } = useLiveQuery(
  (q) =>
    q
      .from({ usersCollection })
      .where(`@usersCollection.id`, `=`, userId),
  [userId]
)
const user = users[0]
```

We could write:

```tsx
const { data: user } = useLiveQuery(
  (q) =>
    q
      .from({ usersCollection })
      .where(`@usersCollection.id`, `=`, userId)
      .findOne(),
  [userId]
)
```

samwillis:
The query is materialised into a collection, we would need to find a way to not have a collection to make this work. Maybe it's a feature of the hook rather than the query?

We will also need a stable order, so it may require an orderBy.

Alternatively we do a one off query first to get the first item, then just subscribe to that.

Note that with a `orderBy` and a `limit(1)` you don't need `users[0]`, but it's still a collection.

KyleAMathews:
Oh `limit(1)` already turns it into an object instead of an array?

KyleAMathews:
no it doesn't :-D

Yeah, the hook could just see that `findOne` is on the query & return the first object.

KyleAMathews:
> We will also need a stable order, so it may require an orderBy.

Maybe default to ordering by the key if an explicit `orderBy` isn't added? I'd assume that the item that's found would change if the order changes so either people are explicit about the order, etc. or they're accepting things potentially moving around a bit.

samwillis:
> Maybe default to ordering by the key if an explicit orderBy isn't added

orderBy adds overhead... maybe we can just document that the user should provide a stable order if they expect more than one result. I can see `fineOne` being mostly used for UI screens where there is only ever one result - a filter by PK.

internally the query returns a `CompiledQuery` object, that has a `.result` - maybe when there is a `fineOne` we keep the internal collection, but `.result` is `resultCollection[0]` rather than `resultCollection`.

It will need some jiggery pokery on the types of `useLiveQuery` to ensure that when there is a `fineOne` it not an array. Very doable.

kevin-dp:
What's the status of this?
This issue https://github.com/TanStack/db/issues/192 was related and got fixed by this PR: https://github.com/TanStack/db/pull/155. Does that simplify the implementation of this feature?

KyleAMathews:
> What's the status of this?

[...]

PR #440, opened 2025-08-23
add findOne() to query builder

fcannizzaro:
Hi! I've implemented the `findOne()` in the query builder (#182) for **db** and **react-db**.

When `findOne()` is chained in the query builder, I store a `single` variable in the context (extending also the context type for the inference) that is then inherited from the collection config to allow the `data` getter to return a single row or an array.

I also tested both `db/test` and `react-db/test` (by adding also a dedicated test for this feature).

LMK if it's ok to do like this or if you have some better ideas.


obeattie:
I'd definitely find an API like this to retrieve just a single row helpful.

samwillis:
@fcannizzaro just letting you know this hasn't dropped off my radar, we discussed it in a meeting today and are going to push through with adding this findOne api. I'll get back to you with a proper review and some guidance tomorrow.

samwillis:
Hi @fcannizzaro, I'm finally catching up on this. I've rebased your branch on main, and fixed a couple of things. One addition is that it now supports specifying `findOne()` outside the context of a `useLiveQuery` call, so constructing the live query collection elsewhere then passing that to the `useLiveQuery`. This needed feeding `single: true` through a few more layers...

@kevin-dp could you take a look at this, you may know of a cleaner way to do this.

packages/db/src/collection/index.ts
```diff
@@ -50,6 +50,7 @@ export interface Collection<
   TInsertInput extends object = T,
 > extends CollectionImpl<T, TKey, TUtils, TSchema, TInsertInput> {
   readonly utils: TUtils
+  readonly single?: true
```
kevin-dp: nit: the name is a bit cryptic. `singleResult` would be more descriptive

packages/db/src/query/builder/index.ts
```diff
@@ -615,6 +615,27 @@ export class BaseQueryBuilder<TContext extends Context = Context> {
...
+   * @example
+   * ```ts
+   * // Get the user matching the query
+   * query
+   *   .from({ users: usersCollection })
+   *   .where(({users}) => eq(users.id, 1))
+   *   .findOne()
+   *```
+   */
+  findOne(): QueryBuilder<TContext & { single: true }> {
```
kevin-dp: nit: "single result enforced" reads as if it's would throw an error if there is more than 1 result. I'd say: "A query builder that returns the first result."

packages/db/src/query/compiler/index.ts
```diff
@@ -290,7 +290,10 @@ export function compileQuery(
     cache.set(rawQuery, compilationResult)
 
     return compilationResult
-  } else if (query.limit !== undefined || query.offset !== undefined) {
+  } else if (
+    !query.single &&
```
kevin-dp: It's a bit weird that we throw an error if you use a limit or offset without an orderBy, but if you use `findOne` then it is allowed.

packages/db/src/types.ts
```diff
@@ -672,3 +685,12 @@ export type WritableDeep<T> = T extends BuiltIns
           : T extends object
             ? WritableObjectDeep<T>
             : unknown
+
+/**
+ * Utility type to infer the query result size (single row or an array)
+ */
+export type WithResultSize<TContext extends Context> = TContext extends {
```
kevin-dp: nit: naming `WithResultSize` is a bit off because it's not really returning the size it's inferring the correct return type. Can we modify the existing `GetResult` type such that it does this check?

packages/db/src/query/live-query-collection.ts
```diff
@@ -35,12 +35,20 @@ export function liveQueryCollectionOptions<
   TResult extends object = GetResult<TContext>,
 >(
   config: LiveQueryCollectionConfig<TContext, TResult>
-): CollectionConfig<TResult> {
+): TContext extends {
```
kevin-dp: Let's introduce a helper type such that we can simplify this return type. It's a bit messy now and hard to read.

packages/db/src/query/live-query-collection.ts
```diff
@@ -35,12 +35,20 @@ export function liveQueryCollectionOptions<
...
+  single: true
+}
+  ? CollectionConfig<TResult> & { single: true }
+  : CollectionConfig<TResult> & { single?: never } {
   const collectionConfigBuilder = new CollectionConfigBuilder<
     TContext,
     TResult
   >(config)
-  return collectionConfigBuilder.getConfig()
+  return collectionConfigBuilder.getConfig() as TContext extends {
```
kevin-dp: Now you can also benefit from the helper type here

packages/db/src/query/live-query-collection.ts
```diff
@@ -83,7 +91,9 @@ export function createLiveQueryCollection<
   TResult extends object = GetResult<TContext>,
 >(
   query: (q: InitialQueryBuilder) => QueryBuilder<TContext>
-): Collection<TResult, string | number, {}>
+): TContext extends { single: true }
```
kevin-dp: use the helper type

packages/db/src/query/live-query-collection.ts
```diff
@@ -92,7 +102,9 @@ export function createLiveQueryCollection<
   TUtils extends UtilsRecord = {},
 >(
   config: LiveQueryCollectionConfig<TContext, TResult> & { utils?: TUtils }
-): Collection<TResult, string | number, TUtils>
+): TContext extends { single: true }
```
kevin-dp: use the helper type

packages/db/src/query/live-query-collection.ts
```diff
@@ -103,7 +115,9 @@ export function createLiveQueryCollection<
   configOrQuery:
     | (LiveQueryCollectionConfig<TContext, TResult> & { utils?: TUtils })
     | ((q: InitialQueryBuilder) => QueryBuilder<TContext>)
-): Collection<TResult, string | number, TUtils> {
+): TContext extends { single: true }
```
kevin-dp: use the helper type

packages/db/src/query/live-query-collection.ts
```diff
@@ -113,7 +127,11 @@ export function createLiveQueryCollection<
       ) => QueryBuilder<TContext>,
     }
     const options = liveQueryCollectionOptions<TContext, TResult>(config)
-    return bridgeToCreateCollection(options)
+    return bridgeToCreateCollection(options) as TContext extends {
```
kevin-dp: use the helper type

packages/db/src/query/live-query-collection.ts
```diff
@@ -124,7 +142,9 @@ export function createLiveQueryCollection<
     return bridgeToCreateCollection({
       ...options,
       utils: config.utils,
-    })
+    }) as TContext extends { single: true }
```
kevin-dp: use the helper type

packages/react-db/src/useLiveQuery.ts
```diff
@@ -280,6 +290,26 @@ export function useLiveQuery<
   isEnabled: true // Always true for pre-created live query collections
 }
 
+// Overload 8: Accept pre-created live query collection with single: true
+export function useLiveQuery<
+  TResult extends object,
+  TKey extends string | number,
+  TUtils extends Record<string, any>,
+>(
+  liveQueryCollection: Collection<TResult, TKey, TUtils> & { single: true }
```
kevin-dp: We should probably have helper types instead of writing `& { single: true }` or `& { single: never }` in a bunch of places. That will greatly simplify maintainability in case we change something to this later.

packages/db/src/query/builder/types.ts
```diff
@@ -571,6 +573,7 @@ export type MergeContextWithJoinType<
     [K in keyof TNewSchema & string]: TJoinType
   }
   result: TContext[`result`]
+  single: TContext[`single`] extends true ? true : false
```
kevin-dp: This is a bit funky since `TContext[`single`]` is already of type boolean. But i guess we need this to avoid type widening?

samwillis: Yep, without this the types don't work... I did not have fun with these types.

packages/db/src/types.ts
fcannizzaro: I changed the name to `InferResultType` and moved together with `GetResult`. I'm not sure to merge into `GetResult` cause it's used for many different things

packages/db/src/query/compiler/index.ts
fcannizzaro: Well but how can we manage this for specific filters like `id` or unique columns where `orderBy` is not useful to keep stable results ? Should be use a default orderBy if missing (ex. sort by collection id) ?

Also this is the reason cause different tests are failing.

kevin-dp: @samwillis what do you think?

samwillis: I think for this first version of `findOne` we ignore the offset+limit - either enforce a value or set one.

Currently limit doesn't work without an orderBy, and I don't think we want to enforce an orderBy when a user uses `findOne` as often they are selecting something they know there is only one of.

If the query returns more than one row, we are just taking the first key and using that.

In future we can optimise it by enforcing a `limit=1` and automaticly setting `orderBy={primary key column}` but thats going to take more work and needs a little thought first.

I think I reverted this change in my commit before handing back to @fcannizzaro - it doesn't look like it's in the current state of this PR.

fcannizzaro:
@samwillis Can you check what's the error ? Locally `pnpm test` has no failed test

samwillis:
Seems to be in the react-db package:

```
packages/react-db test:  ✓  TS   @tanstack/react-db  tests/useLiveQuery.test-d.tsx (4 tests)
packages/react-db test: ⎯⎯⎯⎯⎯⎯ Unhandled Errors ⎯⎯⎯⎯⎯⎯
packages/react-db test: 
packages/react-db test: Vitest caught 2 unhandled errors during the test run.
packages/react-db test: This might cause false positive tests. Resolve unhandled errors to make sure your tests are not affected.
packages/react-db test: ⎯⎯⎯ Unhandled Source Error ⎯⎯⎯
packages/react-db test: TypeCheckError: No overload matches this call.
packages/react-db test:   The last overload gave the following error.
packages/react-db test:     Type 'Plugin<any>[]' is not assignable to type 'PluginOption'.
packages/react-db test:       Type 'Plugin<any>[]' is not assignable to type 'PluginOption[]'.
packages/react-db test:         Type 'Plugin<any>' is not assignable to type 'PluginOption'.
packages/react-db test:           Type 'Plugin<any>' is not assignable to type 'Plugin$1<any>'.
packages/react-db test:             Types of property 'hotUpdate' are incompatible.
packages/react-db test:               Type 'ObjectHook<(this HotUpdatePluginContext, options HotUpdateOptions) => void | EnvironmentModuleNode[] | Promise<void | EnvironmentModuleNode[]>> | undefined' is not assignable to type 'ObjectHook<(this MinimalPluginContext & { environment DevEnvironment; }, options HotUpdateOptions) => void | EnvironmentModuleNode[] | Promise<void | EnvironmentModuleNode[]>> | undefined'.
packages/react-db test:                 Type '(this HotUpdatePluginContext, options HotUpdateOptions) => void | EnvironmentModuleNode[] | Promise<void | EnvironmentModuleNode[]>' is not assignable to type 'ObjectHook<(this MinimalPluginContext & { environment DevEnvironment; }, options HotUpdateOptions) => void | EnvironmentModuleNode[] | Promise<void | EnvironmentModuleNode[]>> | undefined'.
packages/react-db test:                   Type '(this HotUpdatePluginContext, options HotUpdateOptions) => void | EnvironmentModuleNode[] | Promise<void | EnvironmentModuleNode[]>' is not assignable to type '(this MinimalPluginContext & { environment DevEnvironment; }, options HotUpdateOptions) => void | EnvironmentModuleNode[] | Promise<void | EnvironmentModuleNode[]>'.
packages/react-db test:                     The 'this' types of each signature are incompatible.
packages/react-db test:                       Type 'MinimalPluginContext & { environment DevEnvironment; }' is not assignable to type 'HotUpdatePluginContext'.
packages/react-db test:                         Types of property 'environment' are incompatible.
packages/react-db test:                           Property 'options' is missing in type 'import("/home/runner/work/db/db/node_modules/.pnpm/vite@7.1.7_@types+node@22.18.1_jiti@2.6.0_lightningcss@1.30.1_sass@1.90.0_terser@5.44.0_tsx@4.20.6_yaml@2.8.1/node_modules/vite/dist/node/index").DevEnvironment' but required in type 'import("/home/runner/work/db/db/node_modules/.pnpm/vite@6.3.6_@types+node@22.18.1_jiti@2.6.0_lightningcss@1.30.1_sass@1.90.0_terser@5.44.0_tsx@4.20.6_yaml@2.8.1/node_modules/vite/dist/node/index").DevEnvironment'.
packages/react-db test:  ❯ vite.config.ts:7:13
packages/react-db test: ⎯⎯⎯ Unhandled Source Error ⎯⎯⎯
packages/react-db test: TypeCheckError: Argument of type 'UserConfig & Promise<UserConfig> & (UserConfigFnObject & UserConfigExport)' is not assignable to parameter of type 'never'.
packages/react-db test:   Type 'UserConfig & Promise<UserConfig> & UserConfigFnObject' is not assignable to type 'never'.
packages/react-db test:  ❯ vite.config.ts:19:3
packages/react-db test: ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
packages/react-db test:  Test Files  2 passed (2)
packages/react-db test:       Tests  29 passed (29)
packages/react-db test: Type Errors  no errors
packages/react-db test:      Errors  2 errors
packages/react-db test:    Start at  12:18:39
packages/react-db test:    Duration  13.46s (transform 2.68s, setup 590ms, collect 3.51s, tests 522ms, environment 2.04s, prepare 388ms, typecheck 13.05s)
```
https://github.com/TanStack/db/actions/runs/18129520557/job/51592548474?pr=440

fcannizzaro:
It seems a vite configuration error (?)

Issue #347, opened 2025-07-31
Enhanced Error Handling and Reporting for query-db-collection


KyleAMathews:
### Background

TanStack Query provides robust error handling with retry logic, error callbacks, and error boundaries. While `query-db-collection` supports basic retry configuration, it doesn't fully expose Query's error handling capabilities. Currently, errors are logged to console and the collection is marked as ready, but the error details aren't easily accessible.

### Current Behavior

From query.ts:407-415:
```typescript
} else if (result.isError) {
  console.error(
    `[QueryCollection] Error observing query ${String(queryKey)}:`,
    result.error
  )
  
  // Mark collection as ready even on error to avoid blocking apps
  markReady()
}
```

### Problems

1. **Limited error visibility**: Errors only go to console.error
2. **No error callbacks**: Can't run custom logic on errors
3. **Missing error state in utils**: No way to check last error
4. **No error boundary integration**: Can't use React error boundaries

### Proposed Solution

Enhance error handling to match TanStack Query's capabilities while respecting TanStack DB's design:

```typescript
export interface QueryCollectionConfig<TItem, TError, TQueryKey> {
  // ... existing options ...
  
  /**
   * Callback when query encounters an error (after all retries)
   * @param error - The error that occurred
   * @param query - The query instance
   */
  onError?: (error: TError, query: Query) => void
  
  /**
   * Callback when query settles (success or error)
   * @param data - The data if successful
   * @param error - The error if failed
   */
  onSettled?: (data?: Array<TItem>, error?: TError) => void
  
  /**
   * Whether errors should be thrown to nearest error boundary
   * @default false
   */
  useErrorBoundary?: boolean | ((error: TError, query: Query) => boolean)
}

export interface QueryCollectionUtils<TItem, TKey, TInsertInput> {
  // ... existing utils ...
  
  /**
   * Get the last error encountered by the query (if any)
   */
  lastError: () => TError | undefined
  
  /**
   * Check if the collection is in an error state
   */
  isError: () => boolean
  
  /**
   * Get the number of consecutive sync failures
   */
  errorCount: () => number
  
  /**
   * Clear the error state and retry
   */
  clearError: () => Promise<void>
}
```

#### Example 1: Error Notification

```typescript
const todoCollection = createCollection(
  queryCollectionOptions({
    queryKey: ['todos'],
    queryFn: fetchTodos,
    getKey: (item) => item.id,
    
    // Show toast on error
    onError: (error) => {
[...]

PR #441, opened 2025-08-24
feat: add error tracking and retry methods to query collection utils

lucasweng:
## Summary

- Add `lastError`, `isError`, `errorCount` and `clearError` methods to query collection utils per https://github.com/TanStack/db/issues/347
- Add new tests to verify error state tracking, ensure `clearError` triggers a refetch, and collection operations function as expected with cached data while in an error state.

**Note:** To keep the diff focused, this PR does not implement `onError`, `onSettled`, and `useErrorBoundary`. These will be addressed in a separate PR.

## Test Plan

- [x] Tests pass locally

### DIFF
--- a/packages/query-db-collection/src/query.ts
+++ b/packages/query-db-collection/src/query.ts
@@ -309,11 +309,13 @@ export type RefetchFn = () => Promise<void>
  * @template TItem - The type of items stored in the collection
  * @template TKey - The type of the item keys
  * @template TInsertInput - The type accepted for insert operations
+ * @template TError - The type of errors that can occur during queries
  */
 export interface QueryCollectionUtils<
   TItem extends object = Record<string, unknown>,
   TKey extends string | number = string | number,
   TInsertInput extends object = TItem,
+  TError = unknown,
 > extends UtilsRecord {
   /** Manually trigger a refetch of the query */
   refetch: RefetchFn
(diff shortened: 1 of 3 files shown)


packages/query-db-collection/src/query.ts
```diff
@@ -573,6 +593,12 @@ export function queryCollectionOptions<
         // Mark collection as ready after first successful query result
         markReady()
       } else if (result.isError) {
+        // This prevents double counting when QueryObserver calls subscribe twice
+        if (lastError !== result.error) {
+          lastError = result.error
+          errorCount++
+        }
```
lucasweng: Added the check here since I found that the `subscribe` callback gets called twice when testing `errorCount` tracking using `collection.utils.refetch`.

KyleAMathews:
ChatGPT's review https://chatgpt.com/share/68add190-8f64-800c-bc40-1218ca9d9e7f

lucasweng:
> ChatGPT's review https://chatgpt.com/share/68add190-8f64-800c-bc40-1218ca9d9e7f

Thanks @KyleAMathews. I updated the error tracking logic and added a retry test in a7cf838 per the review.

~~I didn't adopt the suggestion to have `clearError` throw on refetch failure though. The method's contract is to reset the error state, and throwing an exception could make it appear that the error state wasn't successfully cleared when it actually was, which may confuse users. If callers need to handle refetch failures, they can check the `errorCount` or `isError` state after calling `clearError`.~~ Would like to know your thoughts on this 🙏

Edited: https://github.com/TanStack/db/pull/441/commits/62d3e864d35ba1791529a2341bd7414aa2ee316c added documentation for error tracking utils. The `onError` example in https://github.com/TanStack/db/issues/347 will be included after onError is implemented.

Edited: https://github.com/TanStack/db/pull/441/commits/1947a5751f23daa6f678e51d4a06f481c0f02c0b Thinking about `clearError` again, it's simpler to just throw an error if the refetch fails. This is easier for users to handle than checking `errorCount` after it's been reset. To avoid the confusion I mentioned, I updated the JSDoc for `clearError` explaining that it can throw an error if the refetch fails."

KyleAMathews:
@lucasweng btw I made a few follow-up issues (linked here in the PR). Feel free to pick up any of them!

lucasweng:
>Sorry for the delay getting this in. I'm just starting to come back from parental leave.

No problems at all! Thanks for taking time to look into this 🙏🏻

PR #442, opened 2025-08-25
keep residual where clause in join predicate pushdown

samwillis:
This modifies the optimiser to keep a residual copy of the where clause in place when pushing a predicate down into a left/right/full join ensuring that the semantics of the query are maintained.

The residual where close is marked as such in the IR so that it is not repeatedly pushed down during the recursive optimisation.

### DIFF
--- a/packages/db/src/query/ir.ts
+++ b/packages/db/src/query/ir.ts
@@ -39,7 +39,9 @@ export interface JoinClause {
   right: BasicExpression
 }
 
-export type Where = BasicExpression<boolean>
+export type Where =
+  | BasicExpression<boolean>
+  | { expression: BasicExpression<boolean>; residual?: boolean }
 
 export type GroupBy = Array<BasicExpression>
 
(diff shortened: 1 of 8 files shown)


packages/db/tests/query/indexes.test.ts
samwillis, line 1: Tests in this file were wrong...

packages/db/tests/query/join.test.ts
samwillis, line 1: Tests added here fail before we then fix the optimiser

packages/db/tests/query/optimizer.test.ts
samwillis, line 1: Even though this is testing the internal of the ir and optimiser i think it's important to check we add the residual flag correctly.

packages/db/src/query/ir.ts
```diff
@@ -128,3 +130,48 @@ export class Aggregate<T = any> extends BaseExpression<T> {
...
+}
+
+/**
+ * Extract the expression from a HAVING clause
+ * HAVING clauses can contain aggregates, unlike regular WHERE clauses
+ */
+export function getHavingExpression(
+  having: Where
+): BasicExpression | Aggregate {
+  return typeof having === `object` && `expression` in having
```
kevin-dp, line 154: This function seems to only differ from `getWhereExpression` by the return type. But the implementation is the same. Should we unify them?

packages/db/src/query/optimizer.ts
```diff
@@ -424,14 +432,44 @@ function isRedundantSubquery(query: QueryIR): boolean {
...
  */
 function splitAndClauses(
+  whereClauses: Array<Where>
+): Array<BasicExpression<boolean>> {
+  const result: Array<BasicExpression<boolean>> = []
+
+  for (const whereClause of whereClauses) {
+    // Skip residual WHERE clauses to prevent them from being pushed down again
+    if (isResidualWhere(whereClause)) {
+      const clause = getWhereExpression(whereClause)
```
kevin-dp: We can move this constant to L440 (i.e. before the if statement) and then remove the one on L447

packages/db/src/query/optimizer.ts
```diff
@@ -424,14 +432,44 @@ function isRedundantSubquery(query: QueryIR): boolean {
...
+      // Preserve non-AND clauses as-is (including OR clauses)
+      result.push(clause)
+    }
+  }
+
+  return result
+}
+
+// Helper function for recursive splitting of BasicExpression arrays
+function splitAndClausesBasic(
```
kevin-dp: Why do we have 2 versions (`splitAndClauses` and `splitAndClausesBasic`) of this function?
Inside `splitAndClauses` we may call into `splitAndClausesBasic` which will recursively split AND expressions but won't ignore residual where clauses. Is this safe? Can't we have the case where the `splitAndClausesBasic` will bump on a residual where clause?

kevin-dp (commented):
Reviewed the changes and everything seems fine. I'm wondering why we needs two versions of the split function though.

kevin-dp (approved):
Brill, thanks!

PR #448, opened 2025-08-25
Fix fetch function to return parsed JSON response

coryhouse:
The example won't work as is. This resolves.


KyleAMathews (approved):
Thanks!

PR #451, opened 2025-08-26
fix race condition in loading joined collections into live query

### LINKED ISSUE #438: Multiple joins not follow SQL semantics
I have an issue with multiple joins, I need some joins between a main table and other 4 tables but it seems that the output on TanStack DB is different from the same query on SQL.

These are my collection and SQL query:

**TanStack DB Collection**

```ts
const collection = createLiveQueryCollection((q) => {
    return q
        .from({ p: player })
        .leftJoin({ challenge1 }, ({ p, challenge1 }) =>
          eq(p.id, challenge1.id)
        )
        .leftJoin({ challenge2 }, ({ p, challenge2 }) =>
          eq(p.id, challenge2.id)
        )
        .leftJoin({ challenge3 }, ({ p, challenge3 }) =>
          eq(p.id, challenge3.id)
        )
        .leftJoin({ challenge4 }, ({ p, challenge4 }) =>
          eq(p.id, challenge4.id)
        )
        .select(({ p, challenge1, challenge2, challenge3, challenge4 }) => ({
          id: p.id,
          category: p.category,
          challenge1,
          challenge2,
          challenge3,
          challenge4,
        }));
    });
}
```

**SQL Query**

```sql
SELECT
  player.id AS player_id,
  player.category AS player_category,
  challenge1.score as challenge1_score,
  challenge2.score as challenge2_score,
  challenge3.score as challenge3_score,
  challenge4.score as challenge4_score
FROM player
LEFT JOIN challenge1 ON player.id = challenge1.id
LEFT JOIN challenge2 ON player.id = challenge2.id
LEFT JOIN challenge3 ON player.id = challenge3.id
LEFT JOIN challenge4 ON player.id = challenge4.id
```

(or drizzle version used below)

```ts
const rows = await db
  .select({
    id: player.id,
    category: player.category,
    challenge1: challenge1.score,
    challenge2: challenge2.score,
    challenge3: challenge3.score,
    challenge4: challenge4.score,
  })
  .from(player)
  .leftJoin(challenge1, (t) => eq(t.id, challenge1.id))
  .leftJoin(challenge2, (t) => eq(t.id, challenge2.id))
  .leftJoin(challenge3, (t) => eq(t.id, challenge3.id))
  .leftJoin(challenge4, (t) => eq(t.id, challenge4.id))
  .where(eq(player.category, data))
  .orderBy(player.id);
```

I obtain from  SQL **N** rows and **M**  from the collection (*with M >>> N*), where each row is duplicated a variable number of times (with missing part of relation).

It seems that the multiple joins feature is not following the SQL semantics.

**repo to reproduce the issue**
https://github.com/fcannizzaro/ts-db-multiple-joins-issue/tree/main

samwillis:
Fixes #438

The issue was that the subscription to the live changes could start emitting results before we had first done any join processing. The join called either `loadKeys` or `loadInitialState` depending on if there was an index it could use. These now set a flag to indicate that we now actually want to start tailing the live changes, until then they are filtered out. We still have to send an empty array of changes when there are any before then so that the query is notified of any collection status change.

Failed test run with repo before it's then fixed: https://github.com/TanStack/db/actions/runs/17233860827/job/48894002331?pr=451

### DIFF
--- a/packages/db/src/query/live/collection-subscriber.ts
+++ b/packages/db/src/query/live/collection-subscriber.ts
@@ -168,9 +168,11 @@ export class CollectionSubscriber<
     const sendVisibleChanges = (
       changes: Array<ChangeMessage<any, string | number>>
     ) => {
-      // We are filtering the changes out when `sendChanges` is false, but still sending
-      // an empty array to the pipeline. This is needed to ensure that the pipeline
-      // receives the status update that the collection is now ready.
+      // We filter out changes when sendChanges is false to ensure that we don't send
+      // any changes from the live subscription until the join operator requests either
+      // the initial state or its first key. This is needed otherwise it could receive
+      // changes which are then later subsumed by the initial state (and that would
+      // lead to weird bugs due to the data being received twice).
       this.sendVisibleChangesToPipeline(
         sendChanges ? changes : [],
         loadedInitialState
--- a/packages/db/src/query/live/collection-subscriber.ts
+++ b/packages/db/src/query/live/collection-subscriber.ts
@@ -113,6 +113,7 @@ export class CollectionSubscriber<
           // filter out deletes for keys that have not been sent
           continue
         }
+        this.sentKeys.add(change.key)
       }
       newChanges.push(newChange)
     }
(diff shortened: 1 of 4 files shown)


packages/db/src/query/live/collection-subscriber.ts
```diff
@@ -113,6 +113,7 @@ export class CollectionSubscriber<
           // filter out deletes for keys that have not been sent
           continue
         }
+        this.sentKeys.add(change.key)
```
samwillis, line 116: We were only adding keys to the `sentKeys` when they were requested, not when they were send due to just watching the live changes.

packages/db/src/query/live/collection-subscriber.ts
```diff
@@ -153,12 +154,27 @@ export class CollectionSubscriber<
...
+    // Until that point we filter out all changes from subscription to the collection.
+    let sendChanges = false
+
     const sendVisibleChanges = (
       changes: Array<ChangeMessage<any, string | number>>
     ) => {
-      this.sendVisibleChangesToPipeline(changes, loadedInitialState)
+      // We are filtering the changes out when `sendChanges` is false, but still sending
+      // an empty array to the pipeline. This is needed to ensure that the pipeline
+      // receives the status update that the collection is now ready.
```
kevin-dp: I would add some more comments explaining why the filtering is needed:
> We filter out changes when `sendChanges` is false to ensure that we don't send any changes from the live subscription until the join operator requests either the initial state or its first key. This is needed otherwise it could receive changes which are then later subsumed by the initial state (and that would lead to weird bugs due to the data being received twice).

packages/db/tests/query/live-query-collection.test.ts
```diff
@@ -313,4 +313,143 @@ describe(`createLiveQueryCollection`, () => {
     // Resubscribe should not throw (would throw "Graph already finalized" without the fix)
     expect(() => liveQuery.subscribeChanges(() => {})).not.toThrow()
   })
+
+  for (const autoIndex of [`eager`, `off`] as const) {
+    it(`should not send the initial state twice on joins with autoIndex: ${autoIndex}`, async () => {
+      type Player = { id: number; name: string }
+      type Challenge = { id: number; value: number }
+
+      let playerBeginCallback: (() => void) | undefined
```
kevin-dp: The code for creating the collections and storing their callbacks is the same for Players and challenges. Could define a helper function that does this such that we don't have to duplicate the code. And then just call it once per collection.

PR #453, opened 2025-08-26
fix count aggregate function (evaluate only not null fields)

fcannizzaro:
The current `count` aggregate function also counts `NULL` values (unlike the SQL count).

Ex. 

```ts
[
	{ category: "A", amount: 10 },
	{ category: "B", amount: 10 },
	{ category: "A", amount: null },
	{ category: "B", amount: null },
]
```

This code below

```typescript
const { data } = useLiveQuery((q) =>
  q
    .from({ t: collection })
    .groupBy(({ t }) => t.category)
    .select(({ t }) => ({
      category: t.category,
      total: count(t.category),
      notNull: count(t.amount)
    }))
);
```

returned

```ts
[
	{ category: "A", total: 2, notNull: 2 },
	{ category: "B", total: 2, notNull: 2 },
]
```

now it gives

```ts
[
	{ category: "A", total: 2, notNull: 1 },
	{ category: "B", total: 2, notNull: 1 },
]
```

I have also added a new test case that keep track of this behavior to prevent a future regression.

### DIFF
--- a/.changeset/five-singers-mate.md
+++ b/.changeset/five-singers-mate.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db-ivm": patch
+---
+
+Hybrid index implementation to track values and their multiplicities
--- a/.changeset/long-ducks-arrive.md
+++ b/.changeset/long-ducks-arrive.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+fix a bug where distinct was not applied to queries using a join
--- a/.changeset/modern-trees-mate.md
+++ b/.changeset/modern-trees-mate.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+Fix bug where too much data would be loaded when the lazy collection of a join contains an offset and/or limit clause.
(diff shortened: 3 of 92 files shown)


packages/db-ivm/src/operators/groupBy.ts
```diff
@@ -165,13 +165,16 @@ export function sum<T>(
...
  * Creates a count aggregate function
  */
-export function count<T>(): AggregateFunction<T, number, number> {
+export function count<T>(
+  valueExtractor: (value: T) => any = (v) => v
+): AggregateFunction<T, number, number> {
   return {
-    preMap: () => 1,
+    // Count only not-null values (ignore null/undefined)
+    preMap: (data: T) => (valueExtractor(data) == null ? 0 : 1),
```
samwillis, line 173: Nice use of the broader `==` match. Worth adding a comment explains that it matched both null and undefined as many won't know.

fcannizzaro: I have changed the comment to

> Count only not-null values (the `== null` comparison gives true for both null and undefined)

samwillis (approved):
Sorry it's taken a little time to get to this, it's perfect. I'm merging now

PR #460, opened 2025-08-27
fix handling of electric must-refetch

samwillis:
I tried to add a test to validate that a 409 didn't reach the onError handler, but it would require very complex mocking, and time is limited...

### DIFF
--- a/packages/electric-db-collection/src/electric.ts
+++ b/packages/electric-db-collection/src/electric.ts
@@ -540,9 +540,8 @@ function createElectricSync<T extends Row<unknown>>(
 
             truncate()
 
-            // Commit the truncate transaction immediately
-            commit()
-            transactionStarted = false
+            // Reset hasUpToDate so we continue accumulating changes until next up-to-date
+            hasUpToDate = false
           }
         }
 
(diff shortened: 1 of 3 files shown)


packages/electric-db-collection/src/electric.ts
```diff
@@ -476,17 +476,27 @@ function createElectricSync<T extends Row<unknown>>(
...
 
           if (shapeOptions.onError) {
             return shapeOptions.onError(errorParams)
+          } else {
+            console.error(
+              `An error occurred while syncing collection: ${collection.id}, \n` +
+                `it has been marked as ready to avoid blocking apps waiting for '.preload()' to finish. \n` +
+                `You can provide an 'onError' handler on the shapeOptions to handle this error, and this message will not be logged.`,
+              errorParams
+            )
```
samwillis, line 499: @KyleAMathews I have added this logging here so that a user is alerted to the fact the collection has been marked ready on an error. You added the original code to do this and I wanted to check your thinking.

packages/electric-db-collection/src/electric.ts
```diff
@@ -540,9 +550,8 @@ function createElectricSync<T extends Row<unknown>>(
 
             truncate()
 
-            // Commit the truncate transaction immediately
-            commit()
-            transactionStarted = false
+            // Reset hasUpToDate so we continue accumulating changes until next up-to-date
```
kevin-dp, line 553: If we are no longer committing it here, when will it be committed and where?

samwillis: At the next up-to-date from electric, line 558 just below.

PR #461, opened 2025-08-27
Fix order by with limit and where clause

kevin-dp:
This PR fixes a problem that resulted in queries that have an `orderBy`, `limit`, and `where` clause to initially load too few data if the `where` clauses filters out some of the first `limit + offset` rows. Claude had a shot at fixing it in https://github.com/TanStack/db/pull/459 but didn't do a good job. I stole the unit tests from that PR (thanks @samwillis for writing those) as they reproduced the issue and are now green in this PR.

When the query contains an `orderBy` and `limit` clause we optimize it by only loading the `offset + limit` first items from the range index initially (introduced in https://github.com/TanStack/db/pull/410). Then at runtime, when new updates come in that may affect the results we send them to the query pipeline and check if the result has enough rows. If not, we load more rows from the range index.

The problem is that we should also check the result set after the initial load since `where` clauses may filter out some of the rows, leading to too few rows in the result. Hence, this PR introduces the necessary changes to check the size of the result after the initial load and load more rows from the range index if needed.

When introducing this fix, the problem with infinite loading loops when the collection has not enough data to fill the result reappeared. This was fixed in https://github.com/TanStack/db/pull/450 but that solution is incompatible with the changes in this PR. Instead, i found a simpler solution which consists of not passing the callback that loads more data if we know that we already exhausted the collection.

### DIFF
--- a/packages/db/src/query/live/collection-subscriber.ts
+++ b/packages/db/src/query/live/collection-subscriber.ts
@@ -229,7 +229,7 @@ export class CollectionSubscriber<
   private subscribeToOrderedChanges(
     whereExpression: BasicExpression<boolean> | undefined
   ) {
-    const { offset, limit, comparator } =
+    const { offset, limit, comparator, dataNeeded } =
       this.collectionConfigBuilder.optimizableOrderByCollections[
         this.collectionId
       ]!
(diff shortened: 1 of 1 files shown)


wrong7:
I'm having this exact issue with v0.1.9 with this code @kevin-dp 

```tsx
  (q) =>
    q
      .from({ todoCollection })
      .where(({ todoCollection }) =>
        like(todoCollection.user_id, user.id)
      )
      .orderBy(({ todoCollection}) => todoCollection.updated_at, "desc")
      .limit(5),
  [user.id]
);
console.log(data.length) // 1, but the query without limit returns 60 rows.
```

kevin-dp:
@wrong7 Could you try out the latest version to see if the problem persists. If it does, could you please open an issue such that we can track this bug and keep you posted when we fix it.

PR #462, opened 2025-08-27
Automatically subscribe/unsubscribe from TanStack Query based on collection subscriber count

### LINKED ISSUE #348: Fix staleTime behavior in query-db-collection by tracking component subscriptions
### Background

Currently, `query-db-collection` creates a single QueryObserver that stays subscribed to TanStack Query for the entire lifecycle of the collection. This breaks TanStack Query's staleTime behavior because from Query's perspective, there's always an "active" observer.

### The Problem

**Expected behavior (like useQuery):**
1. Component mounts → Query becomes active
2. Component unmounts → Query becomes inactive, staleTime timer starts  
3. Component remounts within staleTime → No refetch occurs
4. Component remounts after staleTime → Refetch occurs

**Current behavior in query-db-collection:**
1. Collection starts sync → QueryObserver subscribes and stays subscribed
2. Query is ALWAYS active from TanStack Query's perspective
3. staleTime never applies because query never goes "inactive"
4. refetchOnWindowFocus always triggers regardless of staleTime

### Evidence

From query.ts:303-332:
```typescript
const internalSync: SyncConfig<TItem>['sync'] = (params) => {
  // Creates observer when collection starts sync
  const localObserver = new QueryObserver(queryClient, observerOptions)
  
  // Subscribes immediately and stays subscribed
  const actualUnsubscribeFn = localObserver.subscribe((result) => {
    // ... sync logic
  })
  
  // Only unsubscribes when collection is cleaned up
  return async () => {
    actualUnsubscribeFn()
    // ...
  }
}
```

### Root Cause

Collections already track subscriber count (lines 663-685 in collection.ts), but query-db-collection doesn't use this information to manage its QueryObserver lifecycle.

### Proposed Solution

Modify query-db-collection to subscribe/unsubscribe its QueryObserver based on the collection's subscriber count:

```typescript
const internalSync: SyncConfig<TItem>['sync'] = (params) => {
  const { collection } = params
  let localObserver: QueryObserver | null = null
  let queryUnsubscribe: (() => void) | null = null

  // Track when collection gains/loses subscribers
  const originalAddSubscriber = collection.addSubscriber
  const originalRemoveSubscriber = collection.removeSubscriber

  // Override to manage QueryObserver lifecycle
  collection.addSubscriber = function() {
    const wasInactive = this.activeSubscribersCount === 0
    originalAddSubscriber.call(this)
    
    // Create QueryObserver when first subscriber added
    if (wasInactive && \!localObserver) {
      localObserver = new QueryObserver(queryClient, observerOptions)
      queryUnsubscribe = localObserver.subscribe((result) => {
        // ... existing sync logic
      })
    }
  }

  collection.removeSubscriber = function() {
    originalRemoveSubscriber.call(this)
    
    // Cleanup QueryObserver when last subscriber removed
    if (this.activeSubscribersCount === 0 && localObserver) {
      queryUnsubscribe?.()
      localObserver = null
      queryUnsubscribe = null
      // Now TanStack Query's staleTime timer can start
    }
  }

  return async () => {
    // Restore original methods
    collection.addSubscriber = originalAddSubscriber
    collection.removeSubscriber = originalRemoveSubscriber
    
    // Final cleanup
    queryUnsubscribe?.()
    await queryClient.cancelQueries({ queryKey })
    queryClient.removeQueries({ queryKey })
  }
}
```

### Alternative Approach

Instead of overriding methods, collections could expose subscriber count changes via events:

```typescript
// In collection.ts
private notifySubscriberCountChange(count: number) {
  this.emit('subscriberCountChanged', count)
}

// In query-db-collection
collection.on('subscriberCountChanged', (count) => {
  if (count === 0 && localObserver) {
    // Unsubscribe QueryObserver
  } else if (count === 1 && \!localObserver) {
    // Subscribe QueryObserver  
  }
})
```

### Benefits

1. **Proper staleTime behavior**: Data goes stale when no components are using it
2. **Better performance**: Avoids unnecessary refetches within staleTime window
3. **Memory efficiency**: Queries can be garbage collected after gcTime
4. **TanStack Query compatibility**: Behaves like regular useQuery hooks

### Test Cases

1. Component mounts → Query should become active, fetch if stale
2. Component unmounts → Query should become inactive after delay
3. Component remounts within staleTime → Should not refetch
4. Component remounts after staleTime → Should refetch
5. Multiple components → Query stays active until all unmount
6. Window focus within staleTime → Should not refetch
7. Window focus after staleTime → Should refetch

### Breaking Changes

This should be backwards compatible - existing behavior is preserved, but staleTime will now work correctly.

obeattie:
Previously, `query-db-collection` created a persistent `QueryObserver` that remained subscribed throughout the collection's lifecycle, breaking TanStack Query's `staleTime` behavior. From Query's perspective, there was always an "active" observer, so queries never went inactive and `staleTime` never applied.

This PR fixes the issue by subscribing/unsubscribing the `QueryObserver` based on collection subscriber count. When no components are using the collection, the query becomes inactive. When the collection is observed again, TanStack Query can then refetch automatically. So `query-db-collection` now behaves similarly to standard TanStack Query hooks for caching and staleness.

Fixes #348.


samwillis:
Thanks for this @obeattie, it's looking great.

This indicates we need an event system on the collection. I'm not so keen on adding a dep for this, I would prefer to vendor a simple event system.

@KyleAMathews these events are a new user facing API, I would like your input on this.

KyleAMathews:
Yeah, we don't want a new dependency. We already have event system(s?) in the library which we could repurpose or replaced with a vendored lib.

I'll look at the API soon.

alexander-densley:
should be closed by #555

samwillis:
I've merged main into this, and fixed the conflicts, plus changed it to use the new event system from #555 - @obeattie I hope you don't mind me pushing to your branch.

@KyleAMathews could you cast your eyes over this, I don't think I know enough to really check it.

There seems to also be a type issue during build that isn't apparent in vscode... thats odd

samwillis:
Fixed the type issue 🥳

packages/query-db-collection/tests/query.test.ts
```diff
@@ -1064,6 +1117,71 @@ describe(`QueryCollection`, () => {
       const finalItem = collection.get(`1`)
       expect(finalItem?.name).toMatch(/^Item \d+$/)
     })
+
+    it(`should manage startSync vs subscriber count priority correctly`, async () => {
```
KyleAMathews, line 1121: None of the other collections have this behavior where `startSync: true` overrides gctime.

We document startSync as:

```
  /**
   * Whether to start syncing immediately when the collection is created.
   * Defaults to false for lazy loading. Set to true to immediately sync.
   */
  startSync?: boolean
```

`startSync` should just start the collection but still respect the gctime.

KyleAMathews: oh nvm 😅

kevin-dp:
Hi @obeattie,

Thank you for working on this. I have question regarding the subscription to the `QueryObserver`. On each change of the subscribers count we call `subscribeToQuery`:
```ts
const subscribeToQuery = () => {
  if (!isSubscribed) {
    actualUnsubscribeFn = localObserver.subscribe(handleQueryResult)
    isSubscribed = true
  }
}

// Set up event listener for subscriber changes
const unsubscribeFromCollectionEvents = collection.on(
  `subscribers:change`,
  ({ subscriberCount }) => {
    if (subscriberCount > 0) {
      subscribeToQuery()
    } else if (subscriberCount === 0) {
      unsubscribeFromQuery()
    }
  }
)
```
Isn't that potentially registering the `handleQueryResult` multiple times and hence we would apply the query results several times on the collection?

obeattie:
Hi @kevin-dp, good question. With the `isSubscribed` guard, `handleQueryResult` should only ever be subscribed once at a time, unless I'm missing something?

kevin-dp:
@obeattie I totally missed that guard. Everything's fine indeed! Thanks

PR #463, opened 2025-08-27
fix disabling of gc by settings `gcTime=0`

samwillis:
The intention was that setting `gcTime=0` would disable it, this was not true.

I've also added tests to the electric collection to validate that it will restart the sync after being GC'd and resubscribed too.

### DIFF
--- a/.changeset/neat-queens-smile.md
+++ b/.changeset/neat-queens-smile.md
@@ -1,0 +1,6 @@
+---
+"@tanstack/electric-db-collection": patch
+"@tanstack/db": patch
+---
+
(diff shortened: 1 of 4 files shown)


.changeset/neat-queens-smile.md
```diff
@@ -0,0 +1,6 @@
+---
+"@tanstack/electric-db-collection": patch
+"@tanstack/db": patch
+---
+
+fix disabling of gc by settings `gcTime: 0` on the collection options
```
kevin-dp: typo:
> setting~s~

kevin-dp (approved):
Good catch!

PR #473, opened 2025-08-31
Investigate query collection stuck empty state

samwillis:
Add a test to reproduce the race condition where GC-triggered async cleanup cancels a newly started sync, leaving the collection empty.

This test simulates the scenario where a `Collection`'s `syncCleanupFn` is asynchronous. If a new subscription starts after the GC timer triggers `cleanup()` but before the async `syncCleanupFn` completes, the "old" cleanup can interfere with and cancel the "new" sync's operations, preventing the collection from populating. This test uses fake timers to deterministically demonstrate this interaction.

---
<a href="https://cursor.com/background-agent?bcId=bc-60eebbbf-55e2-46cf-a710-e2bd88be24b2">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-cursor-dark.svg">
    <source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-cursor-light.svg">
    Open in Cursor
  </picture>
</a>
<a href="https://cursor.com/agents?id=bc-60eebbbf-55e2-46cf-a710-e2bd88be24b2">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-web-dark.svg">
    <source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-web-light.svg">
    Open in Web
  </picture>
</a>

### DIFF
--- a/packages/db/tests/async-cleanup-race.test.ts
+++ b/packages/db/tests/async-cleanup-race.test.ts
@@ -1,0 +1,104 @@
+import { describe, expect, it, vi } from "vitest"
+import { createCollection } from "../src/collection.js"
+
+type Row = { id: string; name: string }
+
+// Simulate an external observer keyed by collection id (like a query observer)
+// The cleanup() will cancel whatever observer is currently active for that id
+const observers = new Map<string, { fetchTimer: ReturnType<typeof setTimeout> | null; cancelled: { value: boolean } }>()
+const startCounts = new Map<string, number>()
+
+describe(`Async sync cleanup race with GC`, () => {
+  it(`reproduces: GC cleanup cancels the restarted sync, leaving collection stuck empty`, async () => {
+    vi.useFakeTimers()
+
+    try {
+      const collection = createCollection<Row>({
+        id: `race-test`,
+        getKey: (row) => row.id,
+        gcTime: 50, // fast GC to trigger cleanup quickly
+        // Do not start immediately - start occurs on first subscription
+        startSync: false,
+        sync: {
+          sync: ({ begin, write, commit, markReady, collection }) => {
+            const id = collection.id
+
+            // Simulate an async initial fetch that would populate and mark ready
+            const count = (startCounts.get(id) ?? 0) + 1
+            startCounts.set(id, count)
+            const delay = count === 1 ? 200 : 20
+
+            const cancelledRef = { value: false }
+            const state = { fetchTimer: null as ReturnType<typeof setTimeout> | null, cancelled: cancelledRef }
+            const fetchTimer = setTimeout(() => {
+              // If this observer has been cancelled (e.g. by cleanup racing), do nothing
+              const current = observers.get(id)
+              if (cancelledRef.value || current !== state) return
+              // Simulate successful sync signaling readiness
+              markReady()
+            }, delay)
+            state.fetchTimer = fetchTimer
+            observers.set(id, state)
+
+            // Cleanup that (1) immediately unsubscribes the current fetch
+            // and (2) after a small async delay, cancels the latest fetch that
+            // might have been started by a restart (racing)
+            return async () => {
+              // Immediate unsubscribe of the fetch from this sync instance
+              if (fetchTimer) {
+                clearTimeout(fetchTimer)
+              }
+
+              // Simulate async cleanup (e.g., await cancelQueries)
+              await new Promise<void>((resolve) => setTimeout(resolve, 5))
+
+              // After awaiting, cancel whatever fetch is currently registered
+              // (this likely belongs to the restarted sync)
+              const current = observers.get(id)
+              if (current) {
+                // Mark current observer as cancelled so its pending fetch (if any) won't write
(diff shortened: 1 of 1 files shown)


packages/db/src/collection.ts
```diff
@@ -536,8 +537,21 @@ export class CollectionImpl<
       return // Already started or in progress
     }
 
+    // Enter loading state immediately so callers see progress
     this.setStatus(`loading`)
 
+    // If a previous cleanup is still in flight, wait for it to finish before actually starting
+    if (this.cleanupInFlight) {
+      this.cleanupInFlight.then(() => {
```
kevin-dp, line 545: This method is called `startSync` but is no longer really synchronous. If there is a cleanup in flight then this function will return before we actually started. As a caller of `startSync` i would expect the collection to be started which is no longer the case. Should we make this function async `private async start(): Promise<void>` ?

packages/db/src/collection.ts
```diff
@@ -335,6 +335,7 @@ export class CollectionImpl<
   private gcTimeoutId: ReturnType<typeof setTimeout> | null = null
   private preloadPromise: Promise<void> | null = null
   private syncCleanupFn: (() => void) | null = null
+  private cleanupInFlight: Promise<void> | null = null
```
kevin-dp, line 338: I don't like that this can be a promise or null. I'd rather have it always as a promise. So the start value here would be `Promise.resolve<void>()`. This would simplify the handling of `cleanupInFlight` as we always just await it. This also means we no longer have to set this to `null` in the finally blocks on L720 and L728.

packages/db/src/collection.ts
```diff
@@ -698,10 +712,22 @@ export class CollectionImpl<
     // Stop sync - wrap in try/catch since it's user-provided code
     try {
       if (this.syncCleanupFn) {
-        this.syncCleanupFn()
+        const result = this.syncCleanupFn()
+        // Track in-flight cleanup to block subsequent starts until finished
+        this.cleanupInFlight = Promise.resolve(result as unknown as void)
```
kevin-dp, line 717: I'm wondering what the purpose of this Promise really is?
We are only assigning the promise after the synchronous cleanup function has ran. And the promise we assign is a resolved promise with an empty `catch` (but the promise will never throw since we resolve it...) and a finally block that immediately assigns `this.cleanupInFlight` back to `null`. So essentially, on the next tick, the `finally` block will run and re-assign it to null. I'm failing to see what this does/how this solves our problems.

packages/db/tests/async-cleanup-race.test.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+>()
+const startCounts = new Map<string, number>()
+
+describe(`Async sync cleanup race with GC`, () => {
+  it(`reproduces: GC cleanup cancels the restarted sync, leaving collection stuck empty`, async () => {
+    vi.useFakeTimers()
+
+    try {
+      const collection = createCollection<Row>({
```
kevin-dp, line 22: The definition of the collection is identical to the one in `async-cleanup-livequery-race.test.ts`. Extract it to a constant/function in a helper file.

packages/db/src/collection.ts
kevin-dp: After analyzing the tests i think i understand it now. This works because `result` is a Promise and JS will do promise chaining when passing a promise into `Promise.resolve`. So, this is a trick because we don't actually know whether `result` is a promise or not. Why not just await the promise?
```ts
const result = await this.syncCleanupFn()
```

Also, currently this asynchronous cleanup function will return a resolved promise before the cleanup is actually finished because it never await the `this.cleanupInFlight`. That seems wrong to me. I would just assign `this.cleanupInFlight = await this.syncCleanupFn()`.

Note that the naming and types are very misleading here `private syncCleanupFn: (() => void) | null = null` this should be `private cleanupFn: (() => void) | (() => Promise<void>) | null = null`.

kevin-dp (changes requested):
Great catch and reproduction of the issue, however, i think the solution needs to be improved. Especially, naming and types seem wrong now and the collection's async cleanup function will resolve before the cleanup is actually finished which seems to potentially introduce a new race condition.

packages/db/src/collection.ts
```diff
@@ -698,10 +712,22 @@ export class CollectionImpl<
     // Stop sync - wrap in try/catch since it's user-provided code
     try {
       if (this.syncCleanupFn) {
-        this.syncCleanupFn()
+        const result = this.syncCleanupFn()
```
kevin-dp, line 715: Don't we need to `await this.cleanupInFlight` before calling `this.syncCleanupFn()` in case there's still a previous cleanup ongoing?

samwillis:
I don't believe this is a bug that a user has seen, nor is it something that really internal to tsdb at the moment. The sync cleanup function is defined as non-asynchronous.

Closing for now.

PR #477, opened 2025-09-01
Also optimize orderBy for singleton array

kevin-dp:
We optimize orderBy clauses that order by a single column. However, there was a bug that caused `orderBy(row => row.column)` to be optimized but `orderBy(row => [row.column])` not to be optimized. This PR reproduces the bug with a unit test and fixes it.


samwillis (approved):
:shipit:

PR #482, opened 2025-09-02
fix a bug where a pending sync transaction could be committed early

samwillis:
This resolves a bug where a local optimistic change that is "resolved" would trigger all pending sync transactions to be applied to the collection, even if they haven't been "committed" themselves.

This could be seen if for example an electric collection received a `must-refetch` which would trigger a `truncate`, that transaction is then left open which the data is refetched. If a local mutation was resolved or rejected, that pending sync transaction was applied.

The root cause was that `commitPendingTransactions` would apply to *all* pending sync transactions, both uncommitted and committed - this is wrong. This fixes it by filtering out only committed transactions and applying them.

Failed test run from before fix committed: https://github.com/TanStack/db/actions/runs/17399754657/job/49389905839?pr=482

### DIFF
--- a/.changeset/eighty-hoops-report.md
+++ b/.changeset/eighty-hoops-report.md
@@ -2,4 +2,4 @@
 "@tanstack/db": patch
 ---
 
-fixed a bug where a pending sync transaction could be applied early when an optimistic mutation was resolved to rolled back
+fixed a bug where a pending sync transaction could be applied early when an optimistic mutation was resolved or rolled back
--- a/.changeset/eighty-hoops-report.md
+++ b/.changeset/eighty-hoops-report.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+fixed a bug where a pending sync transaction could be applied early when an optimistic mutation was resolved to rolled back
--- a/packages/db/src/collection.ts
+++ b/packages/db/src/collection.ts
@@ -1195,7 +1195,16 @@ export class CollectionImpl<
       }
     }
 
-    const hasTruncateSync = this.pendingSyncedTransactions.some(
+    // pending synced transactions could be either `committed` or still open.
+    // we only want to process `committed` transactions here
+    const committedSyncedTransactions = this.pendingSyncedTransactions.filter(
+      (t) => t.committed
+    )
+    const uncommittedSyncedTransactions = this.pendingSyncedTransactions.filter(
+      (t) => !t.committed
+    )
+
+    const hasTruncateSync = committedSyncedTransactions.some(
       (t) => t.truncate === true
     )
 
@@ -1205,7 +1214,7 @@ export class CollectionImpl<
 
       // First collect all keys that will be affected by sync operations
       const changedKeys = new Set<TKey>()
-      for (const transaction of this.pendingSyncedTransactions) {
+      for (const transaction of committedSyncedTransactions) {
         for (const operation of transaction.operations) {
           changedKeys.add(operation.key as TKey)
         }
@@ -1228,7 +1237,7 @@ export class CollectionImpl<
       const events: Array<ChangeMessage<T, TKey>> = []
       const rowUpdateMode = this.config.sync.rowUpdateMode || `partial`
 
-      for (const transaction of this.pendingSyncedTransactions) {
+      for (const transaction of committedSyncedTransactions) {
         // Handle truncate operations first
         if (transaction.truncate) {
           // TRUNCATE PHASE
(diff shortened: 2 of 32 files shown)


packages/db/src/collection.ts
```diff
@@ -1195,7 +1195,16 @@ export class CollectionImpl<
       }
     }
 
-    const hasTruncateSync = this.pendingSyncedTransactions.some(
+    // pending synced transactions could be either `committed` or still open.
+    // we only want to process `committed` transactions here
+    const committedSyncedTransactions = this.pendingSyncedTransactions.filter(
```
kevin-dp: We are looping twice over the same array here and potentially a third time for `hasTruncateSync`. We could iterate over it once and accumulate those 3 states (using a for loop or `reduce`). Another option would be to define a helper function akin to Lodash's `groupBy` to group by the committed state. But then we would still need to iterate over it a second time for `hasTruncateSync`.

packages/db/src/collection.ts
```diff
@@ -1304,13 +1313,13 @@ export class CollectionImpl<
       // re-apply optimistic mutations on top of the fresh synced base. This ensures
       // the UI preserves local intent while respecting server rebuild semantics.
       // Ordering: deletes (above) -> server ops (just applied) -> optimistic upserts.
-      const hadTruncate = this.pendingSyncedTransactions.some(
+      const hadTruncate = committedSyncedTransactions.some(
```
kevin-dp: Why are we repeating this piece of code? We already defined a `hasTruncateSync` variable on L1207

packages/db/tests/collection.test.ts
```diff
@@ -1508,4 +1509,64 @@ describe(`Collection with schema validation`, () => {
     expect(collection.syncedData.size).toBe(0)
     expect(collection.syncedMetadata.size).toBe(0)
   })
+  it(`open sync transaction doesn't get applied when optimistic mutation is resolved`, async () => {
```
kevin-dp: nit: add a blank line between the previous test and this one

kevin-dp: nit: "doesn't get" -> "isn't"

packages/db/tests/collection.test.ts
```diff
@@ -1508,4 +1509,64 @@ describe(`Collection with schema validation`, () => {
...
+    expect(collection.state.size).toBe(2)
+
+    // we now do a local optimistic insert
+    collection.insert({ id: 3, name: `three` })
+
+    // we should immediately see the optimistic state
+    expect(collection.state.size).toBe(3)
+    expect(collection.state.get(3)?.name).toBe(`three`)
+
+    // resolve the sync, this is a bit of an anti-pattern here as we haven't
```
kevin-dp: Why do we need this antipattern in the test?

packages/db/tests/collection.test.ts
```diff
@@ -1508,4 +1509,64 @@ describe(`Collection with schema validation`, () => {
...
+    // actually seen the insert from the server yet... but this could easily happen in
+    // a misconfigured setup.
+    // you would see the same effect if the local mutation was rejected and rolled back.
+    collection.utils.resolveSync()
+    await new Promise((resolve) => setTimeout(resolve, 0))
+
+    // we should now be back to the original state
+    expect(collection.state.size).toBe(2)
```
kevin-dp, line 1562: We no longer see the optimistic update because we called `resolveSync` and that assumes that the write has already made its way back to the client?
And since that's not the case here, that optimistic update is gone. Is that right?
Still not completely sure why we can't just let the write sync back before calling `resolveSync`.

packages/db/tests/utls.ts
```diff
@@ -271,6 +271,7 @@ export function mockSyncCollectionOptionsNoInitialState<
   let write: Parameters<SyncConfig<T>[`sync`]>[0][`write`]
   let commit: () => void
   let markReady: () => void
+  let truncate: () => void
```
kevin-dp, line 274: Can we fix the name of this file please? It contains a typo. Should be `utils.ts`

.changeset/eighty-hoops-report.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+fixed a bug where a pending sync transaction could be applied early when an optimistic mutation was resolved to rolled back
```
kevin-dp: This sentence seems broken to me.
> when an optimistic mutation was resolved to rolled back

kevin-dp (approved):
Changes look good. Just left some comments wrt code quality.

packages/db/tests/collection.test.ts
samwillis: In order to trigger the error we cant wait for a sync back as that would imply that the sync transaction was committed. I tried to reject the mutation as it would have the same effect, but it results in an unhandled promise rejection thats thrown in a new microtask. Thats very hard to catch in the test and silence - would require a load of boiler plate.

samwillis: Actually, I changed this. I've made a utility to catch unhanded rejections, and changed the test to reject a optimistic mutation. This is not representative of a real thing that will happen often in user code.

PR #489, opened 2025-09-03
Hybrid index implementation

kevin-dp:
This PR modifies the index such that it is a composition of two indexes:
- a single value index
- a multi value hash index

As long as a key only has a single value associated to it, it is stored in the single value index and no hashes are computed. Once a key has two or more distinct values, it is moved to the multi value index where they are distinguished based on their hash.

This ensures that objects are distinguished by structural equality (as is needed by differential dataflow) and not by referential equality (as is the case with objects in JS). Thus guaranteeing correctness while also minimizing the amount of hashes that are computed.

### DIFF
--- a/packages/db-ivm/src/indexes.ts
+++ b/packages/db-ivm/src/indexes.ts
@@ -46,9 +46,6 @@ export class Index<K, V> {
     return this.#hashIndex.getMultiplicity(key, value)
   }
 
-  // TODO: make it back public if we notice that it's not used externally
-  //       but we made it private just to notice if anyone is using it externally
-
   /**
    * This returns an iterator that iterates over all key-value pairs.
    * @returns An iterable of all key-value pairs (and their multiplicities) in the index.


packages/db-ivm/src/indexes.ts
```diff
@@ -1,83 +1,145 @@
...
+      return this.#valueIndex.getMultiplicity(key)
+    }
+    return this.#hashIndex.getMultiplicity(key, value)
   }
 
-  entries() {
-    return this.#inner.entries()
+  // TODO: make it back public if we notice that it's not used externally
+  //       but we made it private just to notice if anyone is using it externally
+
```
samwillis: Can we lose the todo now?

kevin-dp:
@samwillis i would prefer to keep them as separate files. Regarding the naming of the files, it seems that the `db-ivm` package uses camelCase also for file names (see for instance `topKWithFractionalIndex.ts`), probably a remnant from the past. Do you want me to fix all file names to be hyphenated instead of camelCased?

samwillis:
ah, ok lets leave as is 👍

PR #491, opened 2025-09-03
Replace JSON.stringify based hashing by structural hashing

kevin-dp:
This PR changes how we hash values. We used to `JSON.stringify` the values and then hash them. Now, we walk over the value's structure and hash each part. This implementation adapts the implementation from the [composites polyfill](https://github.com/tc39/proposal-composites). The composites hashing only supports hashing composites, we adapted it too also handle arrays, maps, and sets. We also support `Buffer`, `Uint8Array`, and `File` but hash them based on their object reference rather than their value because their value is likely to be big and hence hashing it would be too costly.

### DIFF
--- a/.changeset/tame-times-create.md
+++ b/.changeset/tame-times-create.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db-ivm": patch
+---
+
+Replace JSON.stringify based hash function by structural hashing function.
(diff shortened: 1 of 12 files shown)


kevin-dp:
@samwillis I would be keen to micro benchmark this PR against the base one to see how this new hashing function performs.

kevin-dp:
We benchmarked initial load time for ts/db queries with the old hashing version and this structural hashing version. The performance figures looked similar with no noticeable differences. So we will merge this structural hashing implementation as it doesn't construct (big) intermediate strings and hence is more memory efficient.

PR #495, opened 2025-09-03
Fix: Ensure query collections initialize with existing data

obeattie:
Previously, when creating a collection from a TanStack Query that already had data, the collection would remain stuck in the `loading` state until the query refreshed. This happens because [`QueryObserver`](https://tanstack.com/query/latest/docs/reference/QueryObserver) only invokes its callback on _changes_, not when attaching to an already-loaded query.

The issue was especially visible for me with [Fast Refresh](https://nextjs.org/docs/architecture/fast-refresh): working locally, after a hot reload, all query state disappeared until the next time the queries refreshed.

This change makes the callback fire immediately with the current query state when we set up the observer, so collections correctly initialize with existing data. (As far as I could tell, `QueryObserver` doesn’t provide a built-in way to fire the callback automatically with initial state.)

### DIFF
--- a/packages/query-db-collection/src/query.ts
+++ b/packages/query-db-collection/src/query.ts
@@ -497,7 +497,8 @@ export function queryCollectionOptions<
       TQueryKey
     >(queryClient, observerOptions)
 
-    const actualUnsubscribeFn = localObserver.subscribe((result) => {
+    type UpdateHandler = Parameters<typeof localObserver.subscribe>[0]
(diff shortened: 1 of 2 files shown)


obeattie:
This is similar in symptoms to, but a distinct issue from #485 I think.

KyleAMathews (approved):
Great! It occurred to me the other day that we need to handle this so glad you saw it and fixed it!

Issue #402, opened 2025-08-12
Add flexible transaction matching strategies to electric-db-collection


KyleAMathews:
The electric-db-collection currently uses PostgreSQL transaction ID (txid) matching to synchronize client mutations with server responses. This is the ideal approach - it provides instant, precise matching and optimal performance. However, it requires backend APIs to return txids, which isn't always feasible or desirable for existing systems.

We should expand the matching options to support three strategies, allowing developers to choose based on their constraints:

**Current State (txid only):**
- Mutation handlers MUST return `{ txid: number }`
- Provides instant sync resumption when the transaction arrives
- Requires modifying APIs to expose PostgreSQL transaction IDs

**Why This Matters:**
TanStack DB blocks incoming sync data to a collection while waiting for backend mutations to complete. This prevents race conditions but means the matching strategy directly impacts user experience:
- **With txid**: Sync resumes immediately when the specific transaction arrives (~100ms)
- **Without precise matching**: Must wait for timeout (3 seconds of blocked sync)

**Proposed Enhancement:**
Support three transaction matching strategies:

### 1. Transaction ID (Current behavior, remains recommended)
- **Returns**: `{ txid: number | number[] }`
- **Use when**: Your API can return PostgreSQL transaction IDs
- **Performance**: Instant sync resumption
- **Status**: Already implemented, will remain the recommended approach

### 2. Custom Match Function (New)
- **Returns**: `{ matchStream: (stream) => Promise<Message> }`
- **Use when**: 
  - Can't get txid from backend
  - Multiple backend transactions involved
  - Need heuristic-based matching
- **Performance**: Resumes when match found (with timeout fallback)
- **Example**: Match by unique ID, timestamp, or combination of fields

We'll want to vendor in the fn from here https://github.com/electric-sql/electric/blob/main/packages/experimental/src/match.ts

### 3. Void/Timeout (New)
- **Returns**: Nothing/void
- **Use when**: Prototyping, testing, or accepting the delay
- **Performance**: Always waits 3 seconds
- **Trade-off**: Simple but blocks sync unnecessarily

**Implementation Example:**

```typescript
// Option 1: Transaction ID (current, optimal)
onInsert: async ({ transaction }) => {
  const result = await api.todos.create(transaction.mutations[0].modified)
  return { txid: result.txid } // Instant sync resumption
}

// Option 2: Custom matching (new)
onInsert: async ({ transaction }) => {
  const item = transaction.mutations[0].modified
[...]

PR #499, opened 2025-09-03
feat: Add flexible matching strategies for electric-db-collection (#402)

KyleAMathews:
## Summary

This PR adds a new `awaitMatch` utility function to electric-db-collection, enabling custom synchronization matching logic for backends that cannot provide PostgreSQL transaction IDs. It also reduces the default timeout for `awaitTxId` from 30 seconds to 5 seconds for faster feedback.

### `collection.utils.awaitMatch()` Utility

For cases where transaction IDs aren't available, use the `awaitMatch` utility function with custom matching logic:

```typescript
import { isChangeMessage } from '@tanstack/electric-db-collection'

  electricCollectionOptions({
      const newItem = transaction.mutations[0].modified
      await api.todos.create(newItem)

      // Wait for sync using custom match logic
      await collection.utils.awaitMatch(
        (message) => isChangeMessage(message) &&
                     message.headers.operation === 'insert' &&
                     message.value.text === newItem.text,
        5000 // timeout in ms (optional, defaults to 5000)
      )
    }
  })
)
```

### Helper Functions

Export utility functions for custom match logic:
- `isChangeMessage(message)` - Check if message is a data change (insert/update/delete)
- `isControlMessage(message)` - Check if message is a control message (up-to-date, must-refetch)

### Use in Custom Optimistic Actions

```typescript
const addTodoAction = createOptimisticAction({
  onMutate: ({ text }) => {
    const tempId = crypto.randomUUID()
    todosCollection.insert({ id: tempId, text, completed: false })
  },
  mutationFn: async ({ text }) => {
    await api.todos.create({ data: { text, completed: false } })
    
    // Wait for matching message
    await todosCollection.utils.awaitMatch(
      (message) => isChangeMessage(message) && 
                   message.headers.operation === 'insert' &&
                   message.value.text === text
    )
  }
})
```

## Breaking Changes

- Default timeout for `awaitTxId` reduced from 30 seconds to 5 seconds for faster feedback when txids don't match

## Benefits

- **Flexible**: Supports backends that can't provide transaction IDs
- **Heuristic Matching**: Custom logic for finding synchronized data
- **Backward Compatible**: Works alongside existing `{ txid }` approach (though default timeout changed)
- **Type-safe**: Full TypeScript support
- **Faster Feedback**: 5 second timeout provides quicker error detection

## Test plan
- [x] All existing tests pass
- [x] New tests for awaitMatch utility with custom match functions
- [x] Timeout behavior testing
- [x] Integration tests with Electric stream simulation
- [x] Linting and type checking passes

### DIFF
--- a/packages/electric-db-collection/src/electric.ts
+++ b/packages/electric-db-collection/src/electric.ts
@@ -309,12 +309,32 @@ export function electricCollectionOptions(
       })
     }
   }
+
+  /**
+   * Helper function to resolve and cleanup matched pending matches
+   */
+  const resolveMatchedPendingMatches = () => {
+    const matchesToResolve: Array<string> = []
+    pendingMatches.state.forEach((match, matchId) => {
+      if (match.matched) {
+        clearTimeout(match.timeoutId)
+        match.resolve(true)
+        matchesToResolve.push(matchId)
+        debug(
+          `${config.id ? `[${config.id}] ` : ``}awaitMatch resolved on up-to-date for match %s`,
+          matchId
+        )
+      }
+    })
+    removePendingMatches(matchesToResolve)
+  }
   const sync = createElectricSync<any>(config.shapeOptions, {
     seenTxids,
     seenSnapshots,
     pendingMatches,
     currentBatchMessages,
     removePendingMatches,
+    resolveMatchedPendingMatches,
     collectionId: config.id,
   })
 
@@ -563,6 +583,7 @@ function createElectricSync<T extends Row<unknown>>(
     >
     currentBatchMessages: Store<Array<Message<T>>>
     removePendingMatches: (matchIds: Array<string>) => void
+    resolveMatchedPendingMatches: () => void
     collectionId?: string
   }
 ): SyncConfig<T> {
@@ -572,6 +593,7 @@ function createElectricSync<T extends Row<unknown>>(
     pendingMatches,
     currentBatchMessages,
     removePendingMatches,
+    resolveMatchedPendingMatches,
     collectionId,
   } = options
   const MAX_BATCH_MESSAGES = 1000 // Safety limit for message buffer
@@ -785,18 +807,7 @@ function createElectricSync<T extends Row<unknown>>(
           })
 
           // Resolve all matched pending matches on up-to-date
-          const matchesToResolve: Array<string> = []
-          pendingMatches.state.forEach((match, matchId) => {
-            if (match.matched) {
-              clearTimeout(match.timeoutId)
-              match.resolve(true)
-              matchesToResolve.push(matchId)
-              debug(`awaitMatch resolved on up-to-date for match %s`, matchId)
-            }
-          })
-
-          // Remove resolved matches
-          removePendingMatches(matchesToResolve)
+          resolveMatchedPendingMatches()
         }
       })
 


packages/electric-db-collection/src/electric.ts
```diff
@@ -74,17 +83,148 @@ type InferSchemaOutput<T> = T extends StandardSchemaV1
...
    * Configuration options for the ElectricSQL ShapeStream
    */
   shapeOptions: ShapeStreamOptions<GetExtensions<T>>
+
+  /**
+   * All standard Collection configuration properties
+   */
+  id?: string
+  schema?: TSchema
+  getKey: CollectionConfig<T, string | number, TSchema>[`getKey`]
```
kevin-dp: Most of these properties and methods (`getKey`, `onInsert`, etc.) are defined in the `BaseCollectionConfig`. Let's not introduce them here but extend the `BaseCollectionConfig` class as it used to do and only add Electric-specific things that are not part of the base config.

samwillis: This PR originally predates the refactor when `BaseCollectionConfig` was introduced - all these props were here and modified in the first commit to this PR. I think Claude has not caught that context when addressing the merge conflict when merging in main.

docs/collections/electric-collection.md
```diff
@@ -54,15 +54,26 @@ The `electricCollectionOptions` function accepts the following options:
...
+
 - `onInsert`: Handler called before insert operations
 - `onUpdate`: Handler called before update operations  
 - `onDelete`: Handler called before delete operations
 
-## Persistence Handlers
+Each handler can return:
+- `{ txid: number | number[] }` - Txid strategy (recommended)
+- `{ matchFn: (message) => boolean, timeout?: number }` - Custom match function strategy
+- `{}` - Void strategy (3-second timeout)
```
samwillis: Could we let a user return the timeout on the void strategy?

```suggestion
- `{ timeout?: number }` - Void strategy (default 3-second timeout)
```

packages/electric-db-collection/src/electric.ts
kevin-dp: Yeah i was expecting that it had rebased correctly. Anyways, it will have to do that :)

packages/db/src/collection/mutations.ts
```diff
@@ -230,7 +230,8 @@ export class CollectionMutationsManager<
 
       // Apply mutations to the new transaction
       directOpTransaction.applyMutations(mutations)
-      directOpTransaction.commit()
+      // Errors still reject tx.isPersisted.promise; this catch only prevents global unhandled rejections
+      directOpTransaction.commit().catch(() => undefined)
 
       // Add the transaction to the collection's transactions store
       state.transactions.set(directOpTransaction.id, directOpTransaction)
```
kevin-dp, line 237: If the commit fails, do we still want to store this TX ID in the `state.transactions`?

packages/electric-db-collection/src/electric.ts
```diff
@@ -239,67 +406,157 @@ export function electricCollectionOptions(
...
+      }
 
-        // Handle both single txid and array of txids
-        if (Array.isArray(txid)) {
-          await Promise.all(txid.map((id) => awaitTxId(id)))
-        } else {
-          await awaitTxId(txid)
+      // Check against current batch messages first to handle race conditions
+      for (const message of currentBatchMessages.state) {
+        if (checkMatch(message)) {
```
kevin-dp: Why is `checkMatch` looking up the match and setting `match.matched` to `true` if we are also doing that here on L465? Isn't L465 going to override whatever we set on L449?

packages/electric-db-collection/src/electric.ts
```diff
@@ -239,67 +406,157 @@ export function electricCollectionOptions(
...
+   * Process matching strategy and wait for synchronization
+   */
+  const processMatchingStrategy = async (
+    result: MatchingStrategy<any>
+  ): Promise<void> => {
+    // Check for txid strategy (backward compatible)
+    if (`txid` in result) {
+      // Handle both single txid and array of txids
+      if (Array.isArray(result.txid)) {
+        await Promise.all(result.txid.map((id) => awaitTxId(id)))
```
kevin-dp: nit: we can shorten `result.txid.map((id) => awaitTxId(id))` to `result.txid.map(awaitTxId)`

packages/electric-db-collection/src/electric.ts
```diff
@@ -413,11 +696,51 @@ function createElectricSync<T extends Row<unknown>>(
...
+          }
+
           // Check for txids in the message and add them to our store
           if (hasTxids(message)) {
             message.headers.txids?.forEach((txid) => newTxids.add(txid))
           }
 
+          // Check pending matches against this message
+          // Note: matchFn will mark matches internally, we don't resolve here
+          const matchesToRemove: Array<string> = []
```
kevin-dp, line 706: It's worth extracting this logic to clean up matches into a helper function and call the helper function here in order to keep the code readable and concise.

packages/electric-db-collection/src/electric.ts
```diff
@@ -491,6 +817,26 @@ function createElectricSync<T extends Row<unknown>>(
             newSnapshots.length = 0
             return seen
           })
+
+          // Resolve all matched pending matches on up-to-date
+          const matchesToResolve: Array<string> = []
```
kevin-dp: It's worth also extracting this logic to clean up matches into a helper function and call the helper function here in order to keep the code readable and concise.

kevin-dp (changes requested):
Good work, this is almost ready to be merged. We just need to fix the rebase that went a bit wrong (i.e. use the `BaseCollectionConfig` to inherit common properties) and extract some of the logic to helper functions.

packages/db/src/collection/mutations.ts
KyleAMathews: yes because the transaction still exists even if it failed

Issue #392, opened 2025-08-10
Types mismatch between `electricCollectionOptions` and `createCollection`


pawelblaszczyk5:
Hello!

I'm playing with TanStack DB - awesome job, really excellent experience. Happy to test the early release and report some issues I stumbled upon.

Currently `electricCollection` options return this:

<img width="1345" height="534" alt="Image" src="https://github.com/user-attachments/assets/194aa8de-bd39-4f09-9e43-8fc027d7aef0" />

While `createCollection` accepts only:

<img width="1411" height="169" alt="Image" src="https://github.com/user-attachments/assets/a5062a4d-19e7-46e5-b64c-bd51bd2c42dd" />

That results in cryptic errors for users with `exactOptionalPropertyTypes ` in `tsconfig` enabled:

<img width="981" height="257" alt="Image" src="https://github.com/user-attachments/assets/f92d13a8-8c3b-4ce8-a1f3-f2a25bcba54d" />

I think it's related to the fact that `electricCollectionOptions` return type isn't explicit? It seems like most of the rest collection factories are explicitly typed so they get this type sync for free.

pawelblaszczyk5:
Will be fixed in next release with changes from: https://github.com/TanStack/db/pull/530

It also shouldn't happen anymore now, that every schema extends the same base interface

PR #502, opened 2025-09-05
fix: explicitly type `electricCollectionOptions` return value

pawelblaszczyk5:
Fixes https://github.com/TanStack/db/issues/392

Currently, if someone used `exactOptionalPropertyTypes` in their tsconfig return type from `electricCollectionOptions` wasn't assignable to `createCollection`. This could be observed in the example in the repository, when added the flag to the tsconfig:

image

After the fix in the PR:

image

### DIFF
--- a/.changeset/chilly-phones-rest.md
+++ b/.changeset/chilly-phones-rest.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/electric-db-collection": patch
+---
+
+fix: explicitly type `electricCollectionOptions` return value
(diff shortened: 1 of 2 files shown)


kevin-dp:
There is a major refactoring of the types under review (https://github.com/TanStack/db/pull/530). I will come back to this PR on Monday to see if this is still needed with the new types or not. Thanks!

pawelblaszczyk5:
Thanks a lot for taking time to look into it - from quick glance it seems like it’ll also fix the issue. Subscribed it, so I’ll also test whether it’s necessary when it got merged :)

pawelblaszczyk5:
I see that it was merged and from what I checked, it's fixed now, so closing it, thank you very much ;) 

image

Issue #507, opened 2025-09-08
Joining subquery with orderBy and limit ignores the limit


kevin-dp:
Consider this query:
```ts
q
  .from({ issue: issuesCollection })
  .join(
    {
      users: q
        .from({ user: usersCollection })
        .where(({ user }) => eq(user.status, `active`))
        .orderBy(({ user }) => user.name, `asc`)
        .limit(1),
    },
    ({ issue, users }) => eq(issue.userId, users.id),
    `left`
  )
```
It joins with a subquery that returns a single user. So the results of the query should only ever contain that one user. However, the optimized version of this query doesn't take into account the limit on the subquery and the resulting rows are joined with all users.

PR #508, opened 2025-09-08
Deoptimize join when lazy side has a limit/offset clause

kevin-dp:
Fixes https://github.com/TanStack/db/issues/507

### DIFF
--- a/.changeset/modern-trees-mate.md
+++ b/.changeset/modern-trees-mate.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+Fix bug where too much data would be loaded when the lazy collection of a join contains an offset and/or limit clause.


packages/db/src/query/compiler/joins.ts
```diff
@@ -187,90 +187,109 @@ function processJoin(
...
+
+      let index: BaseIndex<string | number> | undefined
+
+      const lazyCollectionJoinExpr =
+        activeCollection === `main`
+          ? (joinedExpr as PropRef)
+          : (mainExpr as PropRef)
+
+      console.log(`lazyCollectionJoinExpr`, lazyCollectionJoinExpr)
+      console.log(`joinClause`, joinClause)
```
samwillis: Stray console.logs

samwillis (approved):
LGTM - just the left over debugging code and needs a changeset

PR #520, opened 2025-09-09
fix: use collection key instead of source key when checking for updates in live query collections

### LINKED ISSUE #320: LiveQueryCollection item disappears after an update
In the case when the liveQueryCollection defines a `getKey` function that returns a different key from the parent collection, items will be "deleted" after receiving an update:

https://stackblitz.com/edit/tanstack-db-km8idwxc-1hae4956?file=src%2FApp.tsx

asabil:
Fixes #320


samwillis:
Great spot @asabil! this was perfect, just needed a changeset but I couldn't push to your branch so I had to fork it. New pr is #521 but it has your commits so you will be in the history. Going to merge it now.

Issue #418, opened 2025-08-18
Preload hanging / Collection lifecycle bug


fcannizzaro:
Hi, I'm having an issue with the `collection.preload()`:

stack: TanStack Start + TanStack DB + Electric

After keeping a page open for a while, when switching routes the next route remains in "pending" status because the preload() call never finishes.

```ts
export const Route = createFileRoute("/_main/alerts")({
  component: RouteComponent,
  loader: async () => {
    // console.log("before preload");
    await collection.preload();
    // console.log("after preload"); <-- unreachable when broken
  },
});
```

**Observations**
 • If the page is left in the background for 5–10 minutes, many collections break.
 • When this happens, `collection.preload()` hangs indefinitely.

> Also it seems to happen more frequently on 1.1.1+ electric (idk why).

**Possible correlation** (not 100% sure):
 • the latest Electric network call gets canceled
 • after that, no more network calls are made for that collection
 • If they resume, the collection is still empty (even though previously was not empty)

As @samwillis said on discord maybe it can be some problem of the lifecycle management of the collection (and the related GC)

**repo to reproduce the issue**
https://github.com/fcannizzaro/ts-db-electric-preload-issue/tree/main

PR #523, opened 2025-09-09
fix AbortController re-initialization (electric-db-collection)

fcannizzaro:
Hi! I debugged what was happening in #418 and finally I discovered the issue's cause.

The `AbortController` inside the sync config of `electric-db-collection` after being aborted (by the cleanup fn after the `gcTime` and 0 active subscribers) was never re-initialized again.

And cause the signal was already `aborted`, the electric client skips the sync request! (https://github.com/electric-sql/electric/blob/3baf67a11887fea0bd046b06e081a93f16c88173/packages/typescript-client/src/client.ts#L474)

I only moved the AbortController instantiation inside the sync() function instead of the parent scope.

### DIFF
--- a/packages/electric-db-collection/src/electric.ts
+++ b/packages/electric-db-collection/src/electric.ts
@@ -465,19 +465,25 @@ function createElectricSync<T extends Row<unknown>>(
 
   return {
     sync: (params: Parameters<SyncConfig<T>[`sync`]>[0]) => {
+      const { begin, write, commit, markReady, truncate, collection } = params
+
       // Abort controller for the stream - wraps the signal if provided
       const abortController = new AbortController()
 
       if (shapeOptions.signal) {
-        shapeOptions.signal.addEventListener(`abort`, () => {
-          abortController.abort()
-        })
+        shapeOptions.signal.addEventListener(
+          `abort`,
+          () => {
+            abortController.abort()
+          },
+          {
+            once: true,
+          }
+        )
         if (shapeOptions.signal.aborted) {
           abortController.abort()
         }
       }
-
-      const { begin, write, commit, markReady, truncate, collection } = params
 
       const stream = new ShapeStream({
         ...shapeOptions,
--- a/packages/electric-db-collection/src/electric.ts
+++ b/packages/electric-db-collection/src/electric.ts
@@ -478,7 +478,7 @@ function createElectricSync<T extends Row<unknown>>(
       }
 
       const { begin, write, commit, markReady, truncate, collection } = params
-      
+
       const stream = new ShapeStream({
         ...shapeOptions,
         signal: abortController.signal,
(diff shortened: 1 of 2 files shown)


examples/react/issue
```diff
@@ -0,0 +1 @@
+Subproject commit 5c115260ceee8f2c325f6a8ee2be96c9c7d6f337
```
KyleAMathews: delete this

packages/electric-db-collection/src/electric.ts
```diff
@@ -461,22 +461,24 @@ function createElectricSync<T extends Row<unknown>>(
...
-
   let unsubscribeStream: () => void
 
   return {
     sync: (params: Parameters<SyncConfig<T>[`sync`]>[0]) => {
+      // Abort controller for the stream - wraps the signal if provided
+      const abortController = new AbortController()
+
+      if (shapeOptions.signal) {
+        shapeOptions.signal.addEventListener(`abort`, () => {
```
KyleAMathews: We should also unsubscribe to this event listener then in the cleanup

KyleAMathews: Or add `once: true`

packages/electric-db-collection/src/electric.ts
```diff
@@ -461,22 +461,24 @@ function createElectricSync<T extends Row<unknown>>(
...
-    if (shapeOptions.signal.aborted) {
-      abortController.abort()
-    }
-  }
-
   let unsubscribeStream: () => void
 
   return {
     sync: (params: Parameters<SyncConfig<T>[`sync`]>[0]) => {
+      // Abort controller for the stream - wraps the signal if provided
```
samwillis, line 470: Could we move this after the destruct of the params below?

samwillis (commented):
This is great, really well spotted.

I don't suppose you could try and create a test that reproduces the problem so we can validate this fix and prevent a regression in future?

KyleAMathews (changes requested):
Looks great! Could you add a test which reproduces the bug? Call cleanup & then query again to make sure sync starts up again?

fcannizzaro:
Tomorrow I'll make these changes and write a test for it

KyleAMathews:
I'm pretty sure I've run into the same bug so thanks for figuring it out!

KyleAMathews:
A customer is running into this in production so I'm going to add a test to get this out. Thanks again for debugging things and putting up the PR!!

fcannizzaro:
Yeah if you can write the test since I tried multiple ways to write it but I don't know how to test this case.. cause the mock is done on a global `AbortController` and `ShapeStream` but now it changes for every `sync()` so how it should be done ?

KyleAMathews:
Yeah, I couldn't get it to work either after trying for a bit — our mocks need modified quite a bit to make a test work. Let's merge and release.

KyleAMathews (approved):

KyleAMathews:
I deployed this to my app and confirmed it fixed the bug

PR #524, opened 2025-09-10
fix types when selecting objects such as dates

samwillis:
fixes bug report https://discord.com/channels/719702312431386674/1369767025723052134/1415155288990023751
may fix https://discord.com/channels/933657521581858818/1374668288549847071/1415081127558910073 but unconfirmed

### DIFF
--- a/packages/db/src/query/builder/types.ts
+++ b/packages/db/src/query/builder/types.ts
@@ -500,20 +500,20 @@ export type Ref<T = any> = {
   [K in keyof T]: IsNonExactOptional<T[K]> extends true
     ? IsNonExactNullable<T[K]> extends true
       ? // Both optional and nullable
-        NonNullable<T[K]> extends Record<string, any>
+        IsPlainObject<NonNullable<T[K]>> extends true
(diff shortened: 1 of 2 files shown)


packages/db/src/query/builder/types.ts
```diff
@@ -825,3 +825,48 @@ export type WithResult<TContext extends Context, TResult> = Prettify<
...
+ */
+type IsPlainObject<T> = T extends unknown
+  ? T extends object
+    ? T extends ReadonlyArray<any>
+      ? false
+      : T extends JsBuiltIns
+        ? false
+        : true
+    : false
+  : never
```
kevin-dp: Why return `never` instead of `false` ?

kevin-dp (commented):
Looks like this solves it for now. What about objects that are instances of user-defined classes? For a class that doesn't extend one of the classes listed in `JsBuiltIns` we will still enumerate all its properties. And for classes that do extend a class listed in `JsBuiltIns` the `IsPlainObject` check will return `false` and hence we will not spread it at all? That behavior feels inconsistent for user-defined classes.

PR #525, opened 2025-09-10
Fix structural hashing of numbers

kevin-dp:


samwillis (approved):
I take it we had to use the array buffer?

LGTM

kevin-dp:
> I take it we had to use the array buffer?
> 
> LGTM

Yes, bitwise operators cast the number to 32 bit integer so can't use those. So instead have to explicitly use an array buffer, but we allocate them only once and then mutate them.

PR #526, opened 2025-09-10
optimise key loading into query graph

samwillis:


packages/db/tests/collection-auto-index.test.ts
```diff
@@ -528,13 +528,13 @@ describe(`Collection Auto-Indexing`, () => {
       status: `active`,
       createdAt: new Date(),
     })
-
+    console.log(tracker.stats.queriesExecuted)
```
kevin-dp: leftover console.log can be removed

kevin-dp (approved):
Awesome!

PR #528, opened 2025-09-10
Update dependencies & add Renovate

KyleAMathews:


KyleAMathews:
Yeah exactly. So that way we'll be alerted early if something breaks us

KyleAMathews:
Also it'll ensure that users get updates to package dependencies e.g. the electric client

PR #401, opened 2025-08-11
WIP improve schema support and refactor collection options factory function types

samwillis:
WIP - addresses #396 and #226
TODO - ensure that schema transforms are correctly done on insert and update

Issue #396, opened 2025-08-10
Types are incorrect for schema, where input & output are different


pawelblaszczyk5:
Hello 👋

It seems like currently intended way is that schema is used for validating, but isn't impacting parsed return values:

https://github.com/TanStack/db/blob/bbed2f250e4f01520d966b0fff34f32c7e654fda/packages/db/src/collection.ts#L1616-L1618

Is it set in stone? That limits a lot of usages for schemas, I'd love to see it work. 

If it's not possible, is it feasible to make it work on a type level? Currently it just silently works and user won't know he has invalid types.

samwillis:
Not set in stone - it would be good to understand the use case here a little more, could use explain where and how you would like it to work?

My colleague who worked on the schema implementation is on leave, so I don't have the full context. I believe the types are set up so that when you use a schema it uses the output types for the main collection types, and the input types for the insert/update functions. Is that what you are seeing?

I suspect that implies we should be storing the parsed/transformed data returned from the schema parse, is that not what is happening?

(I'm skimming the code and trying to get myself back up to speed on this bit)

pawelblaszczyk5:
My initial thought was to reuse backend schemas that I'm using for database modeling with `@effect/sql`. Effect heavily relies on bidirectional schemas so they inherently have stuff going inside. For example, converting nullable field to Effect's `Option`, or native JS dates to Effect's `DateTime`. Currently types looks like everything is fine and the validators run, but under the hood I'm getting the stuff without transformations. 

But maybe that's me approach it wrongly and frontend shape schema shouldn't be so tightly coupled to the model of db? I'm not sure here. Curious about your feelings/opinions there. Also I'm aware that most of different schema libraries aren't so heavily into transformations/bidirectional etc.

Also it seems like the behavior you're describing isn't fully correct? Or I'm missing something - currently for everything it looks like output schema is used? At least that what I'm observing. For schema like this:

<img width="751" height="161" alt="Image" src="https://github.com/user-attachments/assets/6a8dad63-b681-4142-8154-5e88f2d5eeea" />

Notice `DateTimeFromDate` usage, or `OptionFromNullOr`.

The inferred types has the information about both directions:

<img width="593" height="433" alt="Image" src="https://github.com/user-attachments/assets/bfa3ba7e-a6f1-4de2-af73-ab81b7c57c21" />

But insert/update & selecting both use the output:

[...]

PR #530, opened 2025-09-11
Refactoring of the types for collection configurations

kevin-dp:
This PR addresses https://github.com/TanStack/db/issues/396 and supersedes https://github.com/TanStack/db/pull/401. It refactors the types for collection configs.

## Problem: `ResolveType` helper doesn't work with inferred types

The main problem was that we had different type parameters `TExplicit`, `TSchema`, and `TFallback` and we would resolve the type from them based on the `ResolveType` helper. While `ResolveType` works correctly when you provide it the types, it doesn't play nicely when it comes to type inference. Essentially, the compiler sometimes had to infer the types for `TExplicit`, `TSchema`, and `TFallback` based on `ResolveType<TExplicit, TSchema, TFallback>` and that doesn't work:
```ts
type ResolveType<
  TExplicit,
  TSchema extends StandardSchemaV1 = never,
  TFallback extends object = Record<string, unknown>,
> = unknown extends TExplicit
  ? [TSchema] extends [never]
    ? TFallback
    : InferSchemaOutput<TSchema>
  : TExplicit extends object
    ? TExplicit
    : Record<string, unknown>

function foo<TExplicit, TSchema extends StandardSchemaV1, TFallback extends object>(
  a: ResolveType<TExplicit, TSchema, TFallback>
): TExplicit {
  return null as unknown as TExplicit // don't care about runtime here
}

const schema = z.object({
  id: z.string(),
  schema: z.boolean(),
})

// `res` is typed as typeof schema
// which means that TExplicit got inferred as typeof schema
// which messes up ResolveType logic because it means that ResolveType will return typeof schema
// instead of InferSchemaOutput<TSchema>
const res = foo(schema)
```

This is one of the problems, there were many other type problems related to this logic. For instance, a user could provide an explicit type to a collection config and then pass in a `getKey` function whose type doesn't correspond to the explicitly passed type, and it would type check without errors while you would expect it to complain because the argument of the `getKey` function has a different type than `TExplicit`.

### Remove `ResolveType` and infer the types from the schema

When creating a collection configuration the user can either pass in a schema or not pass in a schema. If a schema is provided the types need to be inferred from the schema. If no schema is provided, they need to be inferred from `getKey`. Note: if a schema is provided then the schema types and the `getKey` types must match.

To track whether or not a schema is provided we need to overload the factories that create collection options. This is the new implementation of `createCollection`:
```ts
// Overload for when schema is provided
export function createCollection<
  T extends StandardSchemaV1,
>(
  options: CollectionConfig<InferSchemaOutput<T>, TKey, T> & {
    schema: T
    utils?: TUtils
  }
): Collection<InferSchemaOutput<T>, TKey, TUtils, T, InferSchemaInput<T>>

// Overload for when no schema is provided
// the type T needs to be passed explicitly unless it can be inferred from the getKey function in the config
export function createCollection<
  T extends object,
>(
  options: CollectionConfig<T, TKey, never> & {
    schema?: never // prohibit schema if an explicit type is provided
    utils?: TUtils
  }
): Collection<T, TKey, TUtils, never, T>

// Implementation
export function createCollection(
  options: CollectionConfig<any, string | number, any> & {
    schema?: StandardSchemaV1
    utils?: UtilsRecord
  }
): Collection<any, string | number, UtilsRecord, any, any> {
  // actual implementation
}
```
Notice that in the first overload, Typescript can infer `T` from the schema that the user provides. We then derive the input and output types from that schema, i.e. from `T`. In the 2nd overload, we explicitly require the schema to be absent (i.e. `schema?: never`) and thus the input and output types are the same as they are just `T`. Note that `T` corresponds to the type of values that are handled by the `getKey` function from the config. So Typescript can infer it from that function, or the user can pass it explicitly (in which case it must typecheck with the actual types of the `getKey` function).

So far so good, we managed to remove the `ResolveType` logic.

### Track schemas in the output type of the create collection config factories

We managed to remove the `ResolveType` logic but the `createCollection` function now needs to know whether a schema was provided or not. However, our existing collection config factories don't provide that information. Consider for instance the localOnlyCollectionOptions factory implementation:
```ts
export function localOnlyCollectionOptions<
  TExplicit = unknown,
  TSchema extends StandardSchemaV1 = never,
  TFallback extends Record<string, unknown> = Record<string, unknown>,
>(
  config: LocalOnlyCollectionConfig<TExplicit, TSchema, TFallback, TKey>
): CollectionConfig<ResolveType<TExplicit, TSchema, TFallback>, TKey> & {
  utils: LocalOnlyCollectionUtils
} {
  // actual implementation
}
```
The return type of this factory function is:
```ts
CollectionConfig<ResolveType<TExplicit, TSchema, TFallback>, TKey> & {
  utils: LocalOnlyCollectionUtils
}
```
Since `CollectionConfig` defined a property `schema?: TSchema` that means that we can't actually know whether or not the schema was provided (since it's always typed as an optional). But we need to know whether it is provided or not in order to pick the right overload.

Hence, we need to also overload the config factories such that their return type is explicit about whether or not the schema was provided:
```ts
// Overload for when schema is provided
export function localOnlyCollectionOptions<
  T extends StandardSchemaV1,
>(
  config: LocalOnlyCollectionConfig<InferSchemaOutput<T>, T, TKey> & {
    schema: T
  }
): CollectionConfig<InferSchemaOutput<T>, TKey, T> & {
  utils: LocalOnlyCollectionUtils
  schema: T
}

// Overload for when no schema is provided
// the type T needs to be passed explicitly unless it can be inferred from the getKey function in the config
export function localOnlyCollectionOptions<
  T extends object,
>(
  config: LocalOnlyCollectionConfig<T, never, TKey> & {
    schema?: never // prohibit schema
  }
): CollectionConfig<T, TKey, never> & {
  utils: LocalOnlyCollectionUtils
  schema?: never // no schema in the result
}

export function localOnlyCollectionOptions(
  config: LocalOnlyCollectionConfig<any, any, string | number>
): CollectionConfig<any, string | number, any> & {
  utils: LocalOnlyCollectionUtils
  schema?: StandardSchemaV1
} {
  // actual implementation
}
```
Finally, now we are tracking whether or not the schema was provided throughout the factories and all types can be inferred correctly.

### DIFF
--- a/packages/query-db-collection/src/query.ts
+++ b/packages/query-db-collection/src/query.ts
@@ -425,7 +425,12 @@ export function queryCollectionOptions<
   }
 ): CollectionConfig<InferSchemaOutput<T>, TKey, T> & {
   schema: T
-  utils: QueryCollectionUtils<InferSchemaOutput<T>, TKey, InferSchemaInput<T>>
+  utils: QueryCollectionUtils<
+    InferSchemaOutput<T>,
+    TKey,
+    InferSchemaInput<T>,
+    TError
+  >
 }
 
 // Overload for when no schema is provided
@@ -446,7 +451,7 @@ export function queryCollectionOptions<
   }
 ): CollectionConfig<T, TKey> & {
   schema?: never // no schema in the result
-  utils: QueryCollectionUtils<T, TKey>
+  utils: QueryCollectionUtils<T, TKey, T, TError>
 }
 
 export function queryCollectionOptions(
--- a/packages/query-db-collection/tests/query.test.ts
+++ b/packages/query-db-collection/tests/query.test.ts
@@ -1836,7 +1836,6 @@ describe(`QueryCollection`, () => {
 
       const config: QueryCollectionConfig<
         TestItem,
-        never,
         typeof queryFn,
         CustomError
       > = {


kevin-dp:
@ignatz I had a look at the trailbase collection to see if it needs similar changes. But i wasn't sure. The changes in this PR are needed for collections that accept a user-provided schema. At the type level the trailbase collection does accept the `schema` option since the `TrailBaseCollectionConfig` interface extends `Omit<CollectionConfig<TItem, TKey>, "sync" | "onInsert" | "onUpdate" | "onDelete">`. So, since it's not omitting `schema` it will inherit it from `CollectionConfig`.

However, looking further at the `TrailBaseCollectionConfig` options i notice that it requires users to pass a `parse` function. That seems to be very similar to a user-provided `schema`. So i'd say that one of them is obsolete and we have 2 options here:
1. keep it as it is but modify the type to also omit the `schema` key
2. replace the `parse` option and use the `schema` option instead

ignatz:
@kevin-dp , thanks for improving inference.

The TB collection was very closely modeled after the electric one. It has been a few days. Let me have a quick look, to make sure I'm up-to-date with the latest changes, and come back to you :pray:

ignatz:
> 2. replace the parse option and use the schema option instead

IIUC, for this to work the trailbase client would need to emit a zod schema for tanstack/db to use, similar to how drizzle produces one that is used in the example. I think that's a fine approach but I wouldn't want to hold you up.

> 1. keep it as it is but modify the type to also omit the schema key

This seems like the straightforward approach. Removing the schema argument from the example seems to just work. Do you foresee any issues with this approach?

Naively, I'd be leaning (1). Thanks :pray:

kevin-dp:
> This seems like the straightforward approach. Removing the schema argument from the example seems to just work. Do you foresee any issues with this approach?

Yep that's what i was thinking. Since the user can provide a custom parser it seems that they can do in the parser what the schema would usually do. Although it may be nicer with a schema. Wondering now if from a user perspective it makes sense to provide a parser to parse `TRecord`s into `TItem`s and also provide a schema to further validate those `TItem`s and possibly also transform them. What do you think? @ignatz

kevin-dp:
> IIUC, for this to work the trailbase client would need to emit a zod schema for tanstack/db to use, similar to how drizzle produces one that is used in the example. I think that's a fine approach but I wouldn't want to hold you up.

Not really, the schema is usually provided by the user. It's optional though. If the user provides a schema, then tanstack/db uses that schema to validate the input and possible transform it into the output type. Tanstack/db can also infer the type of the items in the collection from the schema. If no schema is provided, then ts/db doesn't do any validation and infers the type of the collection from the `getKey` method.

ignatz:
> > IIUC, for this to work the trailbase client would need to emit a zod schema for tanstack/db to use, similar to how drizzle produces one that is used in the example. I think that's a fine approach but I wouldn't want to hold you up.
> 
> Not really, the schema is usually provided by the user. It's optional though. If the user provides a schema, then tanstack/db uses that schema to validate the input and possible transform it into the output type. Tanstack/db can also infer the type of the items in the collection from the schema. If no schema is provided, then ts/db doesn't do any validation and infers the type of the collection from the `getKey` method.

The TrailBase client is also typed. If schema, it would make sense to infer it. Otherwise, there's the risk of skew and parlor usability.

The `parse`approach was very adhoc just to make the Todo example work with a shared schema. Based on your reply that may have been misguided. Happy to revisit

kevin-dp:
@ignatz if it makes sense to replace `parser` by the `schema` option we already support that would be great!

PR #535, opened 2025-09-11
expand `useLiveQuery` callback to support conditional queries and additional return types

samwillis:


packages/react-db/src/useLiveQuery.ts
```diff
@@ -312,33 +443,48 @@ export function useLiveQuery(
...
+      returnedRef.current = {
+        state: undefined,
+        data: undefined,
+        collection: undefined,
+        status: undefined,
+        isLoading: false,
+        isReady: false,
+        isIdle: true,
+        isError: false,
+        isCleanedUp: false,
```
samwillis, line 466: @KyleAMathews This is the "disabled" state, I've used `isIdle` to indicate that nothing is happening/started, but maybe we need a new `isDisabled` or similar?

KyleAMathews: Query uses `isEnabled` so we can reuse that

https://tanstack.com/query/latest/docs/framework/react/reference/useQuery

packages/react-db/src/useLiveQuery.ts
```diff
@@ -209,32 +314,53 @@ export function useLiveQuery(
...
+          // (not the result, since the result might be from a different query builder instance)
+          collectionRef.current = createLiveQueryCollection({
+            query: configOrQueryOrCollection,
+            startSync: true,
+            gcTime: DEFAULT_GC_TIME_MS,
+          })
+        } else if (result && typeof result === `object`) {
+          // Assume it's a LiveQueryCollectionConfig
+          collectionRef.current = createLiveQueryCollection({
+            ...result,
```
KyleAMathews: we should copy result over the defaults right?

packages/react-db/src/useLiveQuery.ts
```diff
@@ -312,33 +443,48 @@ export function useLiveQuery(
...
-    const entries = Array.from(snapshot.collection.entries())
-    let stateCache: Map<string | number, unknown> | null = null
-    let dataCache: Array<unknown> | null = null
+    // Handle null collection case (when callback returns undefined/null)
+    if (!snapshot.collection) {
+      returnedRef.current = {
+        state: undefined,
+        data: undefined,
+        collection: undefined,
+        status: undefined,
```
KyleAMathews: Seems like it should have a status still?

```suggestion
        status: `disabled`,
```

packages/react-db/src/useLiveQuery.ts
```diff
@@ -209,32 +314,53 @@ export function useLiveQuery(
...
-          gcTime: 0, // Live queries created by useLiveQuery are cleaned up immediately
-        }) as unknown as Collection<object, string | number, {}>
+        // Call the function with a query builder to see what it returns
+        const queryBuilder = new BaseQueryBuilder() as InitialQueryBuilder
+        const result = configOrQueryOrCollection(queryBuilder)
+
+        if (result === undefined || result === null) {
+          // Callback returned undefined/null - disabled query
+          collectionRef.current = null
+        } else if (result instanceof CollectionImpl) {
```
KyleAMathews, line 335: it'd be nice to have a `isCollection`

packages/react-db/src/useLiveQuery.ts
```diff
@@ -209,32 +314,53 @@ export function useLiveQuery(
...
+          // Callback returned a Collection instance - use it directly
+          result.startSyncImmediate()
+          collectionRef.current = result
+        } else if (result instanceof BaseQueryBuilder) {
```
KyleAMathews, line 339: and `isQueryBuilder`

packages/react-db/src/useLiveQuery.ts
samwillis: I've left isCollection and isQueryBuilder for a followup, there is some ambiguity on when each should be set as everything becomes a collection at the end.

PR #541, opened 2025-09-12
fix: getServerSnapshot issue

### LINKED ISSUE #361: Error in renderToPipeableStream: Error: Missing getServerSnapshot
I'm not able to start the application, at first paga load I'm getting this:

```javascript
Error in renderToPipeableStream: Error: Missing getServerSnapshot, which is required for server-rendered content. Will revert to client rendering.
    at Object.useSyncExternalStore (/Users/XXX/node_modules/.pnpm/react-dom@19.1.1_react@19.1.1/node_modules/react-dom/cjs/react-dom-server.node.development.js:9123:19)
    at process.env.NODE_ENV.exports.useSyncExternalStore (/Users/XXX/node_modules/.pnpm/react@19.1.1/node_modules/react/cjs/react.development.js:1229:34)
    at useLiveQuery (/Users/XXX/node_modules/.pnpm/@tanstack+react-db@0.1.0_react@19.1.1_typescript@5.9.2/node_modules/@tanstack/react-db/src/useLiveQuery.ts:296:20)
```

Any advice?

ogzhanolguncu:
This PR fixes #361. 

I tested it on our [repo](https://github.com/unkeyed/unkey/tree/main/apps/dashboard) and error went away.

### DIFF
--- a/.changeset/crazy-crabs-study.md
+++ b/.changeset/crazy-crabs-study.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/react-db": patch
+---
+
+Adds a server snapshot function to useSyncExternalStore in useLiveQuery to fix "Missing getServerSnapshot" errors during SSR. The server snapshot returns a mock collection with empty data and allows proper hydration. Fixes: https://github.com/TanStack/db/issues/361


KyleAMathews:
Hey thanks for the PR! I'm curious if we want to do this vs a proper SSR setup? This "solves" the issue but you'd have a flash of empty SSR shape followed by the client render. We generally recommend people just disable SSR for now.

ogzhanolguncu:
My reasoning was to put this as a band-aid until you guys have a better solution. You'll probably have something decent soon. By the way, that flash was already happening in my case, this PR just prevents JS from throwing that error. And I'm not sure if we can disable SSR entirely. If you have a proper solution in mind, let me know. I'd love to tackle this issue properly.

KyleAMathews:
It'd be great if you could tackle adding SSR support! I wrote up an issue with what we're thinking https://github.com/TanStack/db/issues/545

I'll close this PR.

daveycodez:
```tsx
Error: Missing getServerSnapshot, which is required for server-rendered content. Will revert to client rendering.
dev:     at TodoList (src/app/todos/page.tsx:49:41)
dev:   47 |     const [isLoading, setIsLoading] = useState(false)
dev:   48 |     const [loadingTodoId, setLoadingTodoId] = useState<string | null>(null)
dev: > 49 |     const { data: todos } = useLiveQuery((q) =>
dev:      |                                         ^
dev:   50 |         q.from({ todo: todoCollection })
dev:   51 |     )
dev:   52 | {
dev:   digest: '3336627078'
dev: }
```

Still having issues here on Next.js client components @KyleAMathews

AdamMikacich:
@daveycodez You need to use a `dynamic` import within your client component with `ssr: false`.

For example:
```tsx
'use client';

import dynamic from 'next/dynamic';

const ComponentWithLiveQuery = dynamic(
  () =>
    import('@/components/component-with-live-query').then((mod) => ({
      default: mod.ComponentWithLiveQuery
    })),
  {
    ssr: false
  }
);

export default function Page() {
  return <ComponentWithLiveQuery />;
}
```

By default, Next.js still renders components marked with the `'use client'` directive on the server. Adding `'use client'` simply ensures the components are hydrated (i.e. for interactivity) on the client. To explicitly disable SSR, you must use the `dynamic` function with `ssr: false`.

daveycodez:
It’s not ideal, I’ll share when I make a hook that supports SSR with fallback/initial data. As of right now I’m just doing createLiveCollection -> subscribeChanges -> setData in a useEffect. 

Also shouldn’t useLiveQuery have a disabled option? Like conditional useQuery had

samwillis:
Hey @daveycodez, we do plan to enable SSR but it's a little move complex than adding a getServerSnapshot. Collections and live queries are regularly loaded in a route loader with `.preload()` - we need a solution to this. There is an issue here tracking the plan: https://github.com/TanStack/db/issues/545 

You can conditionally turn off a live query inside the callback - we added this a couple of weeks back but may not have updated the docs yet:

```ts
const { data, isDisabled } = useLiveQuery((q) => {
  if (!switch) return // return null or undefined to disable
  ...
})

**Comment by daveycodez** [comment]
@samwillis Here is the fix to suppress the warning on Next.js

```tsx
    const snapshot = useSyncExternalStore(
        subscribeRef.current,
        getSnapshotRef.current,
        () => null // Just add this line 
    )
```

daveycodez:
If you don't want snapshot to have | null on its types..

Define serverSnapshot at the root scope under the imports (cached)
```tsx
const serverSnapshot = {
    collection: null,
    version: 0
}
```

Then
```tsx
    const snapshot = useSyncExternalStore(
        subscribeRef.current,
        getSnapshotRef.current,
        () => serverSnapshot
    )
```

This would be a super simple patch that would address Next.js warning issues until your full SSR implementation is ready.

Until then I'm doing this to preserve SSR support in my project:
```tsx
const [data, setData] = useState([])
const [isLoading, setIsLoading] = useState(true)

useEffect(() => {
    const liveQueryCollection = createLiveQueryCollection({
        startSync: true,
        query: (q) => q.from({ table: tableCollection })
    })

    liveQueryCollection.onFirstReady(() => {
        setData(liveQueryCollection.toArray)
        setIsLoading(false)
    })

    liveQueryCollection.subscribeChanges(() => {
        setData(liveQueryCollection.toArray)
    })
}, [])
```

PR #550, opened 2025-09-13
Fix memory leak

samwillis:
Thanks to @sorenbs for finding and fixing this.


sorenbs:
So stoked for this! Reduced memory from ~6GB to ~200MB in one of my tests.

kevin-dp:
@samwillis could you explain what was causing the memory leak? I see that you removed the `#streams` property and the `addStream` method. Why was it causing a memory leak and why do we no longer need it?

samwillis:
@kevin-dp we were adding an output from each operator which would accumulate messages into a list in the graph, these outputs were never drained. It was a remnant of how things worked from very early in D2TS that was overlooked.
It meant we had these unused growing lists holding hard references to messages forever.

kevin-dp (approved):

Issue #345, opened 2025-07-31
Add select Option to query-db-collection


KyleAMathews:
### Background

Many APIs return data wrapped in metadata (pagination info, timestamps, etc.). Currently, `query-db-collection` expects `queryFn` to return just the array of items. This forces developers to transform data in the queryFn, losing access to metadata.

### Problem

```typescript
// API returns: { data: Todo[], total: number, page: number }
// But collection needs: Todo[]

const todoCollection = createCollection(
  queryCollectionOptions({
    queryKey: ['todos'],
    queryFn: async () => {
      const response = await api.getTodos()
      // Metadata is lost here
      return response.data
    },
    getKey: (item) => item.id,
    queryClient,
  })
)
```

### Proposed Solution

Add a `select` option that extracts the item array from the query response:

```typescript
export interface QueryCollectionConfig<
  TItem extends object,
  TError = unknown,
  TQueryKey extends QueryKey = QueryKey,
  TQueryData = unknown, // New generic for raw query data
> {
  // ... existing options ...
  
  /**
   * Transform the query response to extract the items array
   * @param data - The raw response from queryFn
   * @returns The array of items for the collection
   */
  select?: (data: TQueryData) => Array<TItem>
}
```

#### Example 1: Paginated API Response

```typescript
interface PaginatedResponse<T> {
  data: T[]
  total: number
  page: number
  pageSize: number
  hasMore: boolean
}

const todoCollection = createCollection(
  queryCollectionOptions({
    queryKey: ['todos', { page: 1 }],
    queryFn: async () => {
      // Returns PaginatedResponse<Todo>
      return api.getTodos({ page: 1, pageSize: 100 })
    },
    select: (response) => response.data, // Extract just the items
    getKey: (item) => item.id,
    queryClient,
  })
)

// The full response is still cached by TanStack Query
// Can access it via queryClient if needed:
const fullResponse = queryClient.getQueryData<PaginatedResponse<Todo>>(['todos', { page: 1 }])
console.log(`Total items: ${fullResponse?.total}`)
```

#### Example 2: GraphQL Response

```typescript
interface GraphQLResponse<T> {
  data: T
  errors?: Array<{ message: string }>
}

interface TodosQuery {
  todos: {
    edges: Array<{ node: Todo }>
    pageInfo: { hasNextPage: boolean }
  }
}

const todoCollection = createCollection(
  queryCollectionOptions({
    queryKey: ['todos-graphql'],
    queryFn: async () => {
[...]

PR #551, opened 2025-09-13
feat: Add select option to extract items while preserving metadata

0xlakshan:
## Description
Added a new select option to **QueryCollectionConfig** that lets developers extract item arrays from wrapped API responses (e.g., with pagination or metadata) while preserving the full response in the TanStack Query cache.

**Changes** 
- Added select field to **QueryCollectionConfig** interface
- In addition to updating the core interface as the issue suggests, i've added two more function overloads, since the code structure was recently updated with use of overloads. to maintain the consistency & strong type inference (https://github.com/TanStack/db/pull/530)
- Updated internal sync logic to apply select before processing query results

**Todo** 
- [x] Test that select properly extracts array from wrapped response
- [x] Test error handling when select returns non-array
- [x] Test that full response is cached in QueryClient
- [x] Test type inference with select option
- [x] Test that mutations still work correctly

Related Issue - #345


0xlakshan:
Hi @KyleAMathews @samwillis would you mind taking a quick look at my PR ? im open to any feedback you might have

packages/query-db-collection/tests/query.test.ts
```diff
@@ -1305,7 +1427,7 @@ describe(`QueryCollection`, () => {
       const updatedItem = cacheAfterUpdate.find((item) => item.id === `1`)
       expect(updatedItem?.name).toBe(`Updated Item 1`)
       expect(updatedItem?.value).toBe(10) // Original value preserved
-
+      // console.log(cacheAfterUpdate)
```
samwillis: stray console.log

packages/query-db-collection/src/query.ts
```diff
@@ -249,6 +249,63 @@ export interface QueryCollectionUtils<
  *   })
  * )
  */
+// Overload for when schema is provided and select present
```
samwillis, line 265: Could you add an example of how to use `select` in the jsdoc above?

samwillis (commented):
Thanks @0xlakshan, this is great and perfectly matches the proposed design from @KyleAMathews. A couple of other inline comments along with:

- I don't suppose you could add the option to the docs in `docs/collections/query-collection.md`?

- It also needs a changeset - run `pnpm changeset` repo root and follow the instructions. This would be a patch release at this point.

0xlakshan:
All done! @samwillis  , thanks for the feedback

KyleAMathews (approved):
Great work! Thanks for taking this on! Sorry for the delay getting it merged 🙏

PR #552, opened 2025-09-14
feat: add exact refetch targeting and refetchType option

lucasweng:
## Summary
- Implemented exact query key targeting to prevent unintended cascading refetches of related queries
- Added a `refetchType` option to query collections for more granular control with `'all'`, `'active'`, and `'inactive'` types
- The default for `refetch()` is `'all'`, which matches the underlying TankStack [queryClient.refetchQueries()](https://tanstack.com/query/v5/docs/reference/QueryClient#queryclientrefetchqueries) default. This behavior is documented.

## Test Plan
- [x] Tests pass locally

### DIFF
--- a/.changeset/soft-doodles-cover.md
+++ b/.changeset/soft-doodles-cover.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/query-db-collection": patch
+---
+
+Implement exact query key targeting to prevent unintended cascading refetches of related queries, and add refetchType option to query collections for granular refetch control with 'all', 'active', and 'inactive' modes
--- a/docs/collections/query-collection.md
+++ b/docs/collections/query-collection.md
@@ -56,6 +56,14 @@ The `queryCollectionOptions` function accepts the following options:
 ### Query Options
 
 - `enabled`: Whether the query should automatically run (default: `true`)
+- `refetchType`: The type of refetch to perform (default: `all`)
+  - `all`: Refetch this collection regardless of observer state
+  - `active`: Refetch only when there is an active observer
+  - `inactive`: Refetch only when there is no active observer
+  - Notes:
+    - Refetch only targets queries that already exist in the TanStack Query cache for the exact `queryKey`
+    - If `enabled: false`, `utils.refetch()` is a no-op for all `refetchType` values
+    - An "active observer" exists while the collection is syncing (e.g. when `startSync: true` or once started manually)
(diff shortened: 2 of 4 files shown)


KyleAMathews:
Hey @lucasweng! Sorry for not getting to this for so long! I/we somehow missed it.

Thanks for tackling this issue. I reviewed your PR and made some refinements based on a careful analysis of the requirements and TanStack Query's behavior. Here's what changed:

  Changes Made

  1. Changed utils.refetch() to use queryObserver.refetch() instead of queryClient.refetchQueries()

  This was the key insight: we need different behavior for imperative refetch vs batch refetch operations.

  - Why: According to https://tanstack.com/query/latest/docs/framework/react/guides/disabling-queries, the hook's refetch() method bypasses
  enabled: false to support manual fetch patterns
  - Benefit: Fixes clearError() which calls refetch() internally - it now works even when enabled: false
  - Benefit: Supports the common "manual fetch on button click" pattern with enabled: false

  2. Removed refetchType option

  After discussion, we realized refetchType doesn't add value with exact targeting:

  - With exact: true, we're already targeting just this specific collection (no cascade)
  - Users always want their collection data refetched, whether from explicit utils.refetch() calls or after mutations
  - The refetchType filter made sense for preventing cascades in batch operations, but with exact targeting it's unnecessary

  3. Changed return type to QueryObserverResult

  - Returns the query result instead of void so users can inspect the result
  - Better DX and more aligned with hook behavior

  4. Added more comprehensive tests

  - Test for clearError() using exact targeting
  - Test for throwOnError propagation
  - Test for enabled: false bypass behavior

  The Core Issue

  The original issue was about preventing cascade refetches. Your solution with exact: true solved that perfectly! The additional changes ensure
   that imperative refetch() calls work as users expect (bypassing enabled: false like hooks do).

  Let me know if you have any questions about these changes!

lucasweng:
Hey @KyleAMathews, appreciate you taking the time to review this! No worries at all, I know the team's been working on a lot of features.

Thanks for the clear explanation of the refinements! The `queryObserver.refetch` approach makes way more sense for better DX and bypassing `enabled: false`. And yeah, you're right about `refetchType`. With exact targeting, it's redundant.

Just pushed https://github.com/TanStack/db/pull/552/commits/d199a7abd5e0ace4693278d23157a4ccfc711da9 to fix a type error in query.test. 🙏🏻

KyleAMathews:
Thanks!

PR #555, opened 2025-09-15
New collection event system

samwillis:
This adds a new event system for generic collection events, such as status changes or notification of the change in subscribers. It's strongly typed based on the passed in event name.

This can be used by #462 as the events to trigger subscribing/unsubscribing from Query. I also intend to use it in a refactor of the live query collection and its status triggers.

We should look to replace `.onFirstReady` with `.once('status:ready', () => ...)` in a followup, and deprecate that api.

### DIFF
--- a/packages/db/src/collection-events.ts
+++ b/packages/db/src/collection-events.ts
@@ -38,6 +38,11 @@ export type AllCollectionEvents = {
   [K in CollectionStatus as `status:${K}`]: CollectionStatusEvent<K>
 }
 
+export type CollectionEvent =
+  | AllCollectionEvents[keyof AllCollectionEvents]
+  | CollectionStatusChangeEvent
+  | CollectionSubscribersChangeEvent
+
 export type CollectionEventHandler<T extends keyof AllCollectionEvents> = (
   event: AllCollectionEvents[T]
 ) => void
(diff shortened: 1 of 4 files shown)


packages/db/src/collection-events.ts
```diff
@@ -0,0 +1,127 @@
...
+    callback: CollectionEventHandler<T>
+  ) {
+    this.listeners.get(event)?.delete(callback)
+  }
+
+  emit<T extends keyof AllCollectionEvents>(
+    event: T,
+    eventPayload: AllCollectionEvents[T]
+  ) {
+    this.listeners.get(event)?.forEach((listener) => listener(eventPayload))
```
KyleAMathews: a listener exception could break the loop so let's wrap the listener call in a try/catch

packages/db/src/collection-events.ts
```diff
@@ -0,0 +1,127 @@
...
+    callback: CollectionEventHandler<T>
+  ) {
+    const unsubscribe = this.on(event, (eventPayload) => {
+      callback(eventPayload)
+      unsubscribe()
+    })
+    return unsubscribe
+  }
+
+  off<T extends keyof AllCollectionEvents>(
```
KyleAMathews, line 86: we want to cleanup all listeners when the collection is cleaned up

packages/db/src/collection-events.ts
```diff
@@ -0,0 +1,127 @@
...
+ * Event emitted when the number of subscribers to the collection changes
+ */
+export interface CollectionSubscribersChangeEvent {
+  type: `subscribers:change`
+  collection: Collection
+  previousSubscriberCount: number
+  subscriberCount: number
+}
+
+export type AllCollectionEvents = {
```
KyleAMathews, line 34: Export a union: export type CollectionEvent = AllCollectionEvents[keyof AllCollectionEvents] for switch(event.type) ergonomics.

packages/db/src/collection-events.ts
```diff
@@ -0,0 +1,127 @@
+import type { Collection } from "./collection"
+import type { CollectionStatus } from "./types"
+
```
KyleAMathews, line 3: Provide a small helper waitFor(event, {  timeout? }) => Promise<event> that internally uses once.

kevin-dp (approved):

PR #557, opened 2025-09-15
Improve merge logic for applyMutations

### LINKED ISSUE #556: Improve merge logic for applyMutations
https://github.com/TanStack/db/blob/dcfda3751988b90808c492ed53193e323b012b65/packages/db/src/transactions.ts#L212-L226

- an update that's applied after an insert should stay an insert, merge the update's changes w/ the insert's changes but otherwise take the updates values
- a delete mutation should not be applied if there's an insert (both should be deleted). If there's an update, it should overwrite the update like now.

KyleAMathews:
## Summary

Improves the merge logic for `applyMutations` method in transactions to handle mutation combinations more intelligently:

- **Update after insert**: Keeps the mutation as an `insert` type with empty `original`, merges changes from both mutations, and uses the update's final `modified` value
- **Delete after insert**: Removes both mutations since they cancel each other out  
- **Delete after update**: Maintains current behavior of replacing with delete

## Changes

- Enhanced `applyMutations` method in `src/transactions.ts` with smarter merge logic
- Added comprehensive test suite in `tests/apply-mutations.test.ts` covering all scenarios
- Fixed TypeScript errors with proper type annotations
- Added changeset for the improvements

## Test Plan

- [x] All existing transaction tests pass
- [x] New comprehensive test suite covering all mutation merge scenarios  
- [x] Build passes with no TypeScript errors
- [x] Maintains backward compatibility for existing behavior

Fixes #556

### DIFF
--- a/.changeset/dirty-trams-nail.md
+++ b/.changeset/dirty-trams-nail.md
@@ -2,8 +2,6 @@
 "@tanstack/db": patch
 ---
 
-Improve merge logic for applyMutations
+Improve mutation merging from crude replacement to sophisticated merge logic
 
-- Update after insert: keeps the insert type with empty original, merges changes from both mutations
-- Delete after insert: removes both mutations (they cancel each other out)
-- Delete after update: maintains current behavior of replacing with delete
+Previously, mutations were simply replaced when operating on the same item. Now mutations are intelligently merged based on their operation types (insert vs update vs delete), reducing network overhead and better preserving user intent.
--- a/.changeset/dirty-trams-nail.md
+++ b/.changeset/dirty-trams-nail.md
@@ -1,0 +1,9 @@
+---
+"@tanstack/db": patch
+---
+
+Improve merge logic for applyMutations
+
+- Update after insert: keeps the insert type with empty original, merges changes from both mutations
+- Delete after insert: removes both mutations (they cancel each other out)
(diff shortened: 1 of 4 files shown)


samwillis (commented):
Implementation looks nice a clean, but I would like to question if it would be better to throw errors for some of these combination, or if the current logic (which is sort of sql like in that the effect is the same) is what we want.

I wander if these should throw errors, as either the row doesn't exist, or already exists, in causal order?

delete -> delete
delete -> update
update -> insert
insert -> insert

also by dropping the delete when followed by an insert we are loosing user intent. I think there we need to send both to the server.

SQL is a little different as operations are essentially a query with a where clause, and so a delete/update with no matching rows will always go through, but thats not really how rest apis work which is what the majority of users would be doing.

KyleAMathews:
The default case is to throw.

We don't have a way to store a delete and then insert so I think we can just throw on that as well (or keep the current pattern). The dev should put a delete then insert into separate transactions.

KyleAMathews:
I guess a delete followed by an insert would need to be an update. Which is confusing. So I think we should just throw.

KyleAMathews:
Actually the collection already blocks any operations on a key that's been (optimistically) deleted so just removed those.

samwillis (approved):
Looks good to me :shipit:

PR #559, opened 2025-09-15
feat(offline-transactions): implement offline-first transaction system

KyleAMathews:
## Summary

🚀 Implements comprehensive offline-first transaction capabilities for TanStack DB that provides durable persistence of mutations with automatic retry when connectivity is restored.

Initial implementation of the RFC at https://github.com/TanStack/db/discussions/554

• **Outbox Pattern**: Persist mutations before dispatch for zero data loss during offline periods
• **Multi-tab Coordination**: Leader election via Web Locks API with BroadcastChannel fallback ensures safe storage access
• **FIFO Sequential Processing**: Simplified execution model processes transactions one at a time in creation order
• **Robust Retry Logic**: Exponential backoff with jitter and developer-controlled error classification
• **Flexible Storage**: IndexedDB primary with localStorage fallback for broad compatibility
• **Type Safety**: Full TypeScript integration preserving existing TanStack DB patterns
• **Developer Experience**: Clear APIs with leadership awareness and comprehensive error handling

### 🏗️ **Architecture**
- **8 Core Modules**: Storage, Outbox, Execution Engine, Retry System, Connectivity, Coordination, Replay, API
- **27 Source Files**: Complete implementation with proper separation of concerns
- **Production Ready**: Comprehensive error handling, quota management, and multi-environment support

### 🔧 **Key Features**
- **Zero Data Loss**: Outbox-first persistence pattern ensures mutations survive network failures
- **Multi-tab Safety**: Only one tab manages the outbox, others run in online-only mode
- **Sequential Processing**: Transactions execute one at a time in FIFO order to avoid dependency issues
- **Automatic Recovery**: Transaction replay restores optimistic state on application restart
- **Developer-Controlled Errors**: All errors retry by default unless developers throw `NonRetriableError`

### 🎯 **Developer Experience**
```typescript
// Setup offline executor with mutation functions
const offline = startOfflineExecutor({
  collections: { todos: todoCollection },
  mutationFns: {
    syncTodos: async ({ transaction, idempotencyKey }) => {
      try {
        await api.saveBatch(transaction.mutations, { idempotencyKey })
      } catch (error) {
        // Developer controls what's retriable
        if (error.status === 401 || error.status === 403) {
          throw new NonRetriableError('Authentication failed')
        }
        throw error // Everything else retries automatically
      }
    }
  },
  onLeadershipChange: (isLeader) => {
    console.log(isLeader ? 'Offline support active' : 'Online-only mode')
  }
})

// Create offline transactions that work offline
const offlineTx = offline.createOfflineTransaction({
  mutationFnName: 'syncTodos',
  autoCommit: false
})

offlineTx.mutate(() => {
  todoCollection.insert({ id: '1', text: 'Buy milk' })
})

await offlineTx.commit() // Persists to outbox if offline
```

### Storage Layer
- **IndexedDBAdapter**: Primary storage with quota exceeded handling
- **LocalStorageAdapter**: Automatic fallback for compatibility
- **TransactionSerializer**: Handles complex object serialization with Date support

### Execution Engine
- **KeyScheduler**: Manages FIFO sequential execution for transaction safety
- **TransactionExecutor**: Orchestrates execution with retry logic and error handling
- **OutboxManager**: CRUD operations for persistent transaction queue

### Multi-tab Coordination
- **WebLocksLeader**: Preferred leader election using Web Locks API (Chrome 69+, Firefox 96+)
- **BroadcastChannelLeader**: Fallback leader election for broader compatibility
- **Graceful Degradation**: Non-leaders automatically switch to online-only mode

### Connectivity & Retry
- **OnlineDetector**: Monitors network state via navigator.onLine and visibility API
- **BackoffCalculator**: Exponential backoff with configurable jitter
- **Developer Error Control**: All errors retry unless `NonRetriableError` is thrown

### 1. FIFO Sequential Processing

**RFC Proposed**: Key-based parallel execution where transactions affecting different keys could run concurrently up to a configurable `maxConcurrency` limit.

**Implemented**: FIFO sequential processing where all transactions execute one at a time in creation order.

**Reasoning**:
- **Foreign Key Safety**: Avoids dependency issues between transactions that may reference each other
- **Simpler Mental Model**: Easier to reason about transaction ordering and effects
- **Reduced Complexity**: Eliminates need for complex dependency analysis and scheduling
- **Predictable Behavior**: Transactions execute in the exact order they were created

This conservative approach ensures correctness while maintaining the core offline-first benefits. Future versions could explore more sophisticated scheduling if performance requirements demand it.

### 2. Explicit Transaction API

**RFC Proposed**: Automatic offline behavior where registered collection operations transparently gain offline capabilities:
```typescript
// RFC envisioned this working offline automatically
todoCollection.insert({ id: '1', text: 'Buy milk' })
```

**Implemented**: Explicit offline transactions where developers explicitly opt-in to offline behavior:
```typescript
const tx = offline.createOfflineTransaction({ mutationFnName: 'syncTodos' })
tx.mutate(() => todoCollection.insert({ id: '1', text: 'Buy milk' }))
await tx.commit()
```

**Reasoning**:
- **Clear Intent**: Code explicitly shows which operations are offline-capable
- **No Magic**: Developers understand exactly when offline logic is active
- **Safer Default**: No unexpected offline behavior from standard collection operations
- **Better Control**: Developers choose when to use offline vs online-only transactions
- **Future Flexibility**: Can add automatic mode later as an opt-in configuration

This approach provides better control and clearer semantics while maintaining all the offline-first capabilities. The explicit API makes it obvious which operations will persist to the outbox and retry on failure.

## Test Plan

✅ **Unit Tests**: Core component functionality with mocked browser APIs
✅ **Type Safety**: Full TypeScript compilation with strict settings
✅ **Build System**: ESM/CJS dual build with proper tree-shaking
✅ **Linting**: ESLint compliance with automated formatting

### Integration Testing
✅ **Network failure/recovery scenarios**: Implemented with offline/online mutation switching
✅ **Multi-tab leader election**: Tested with fake leader election implementation
✅ **Application restart with pending transactions**: Verified transaction replay from storage
- [ ] Storage quota exceeded handling
- [ ] Large transaction volume performance

## Integration Approach

**Opt-in Offline Transactions**: This implementation provides offline capabilities as an explicit opt-in rather than automatically upgrading existing collection operations:

```typescript
// Standard TanStack DB (works as before)
todoCollection.insert({ id: '1', text: 'Buy milk' }) // Online only

// Offline-capable transactions (new pattern)
const offline = startOfflineExecutor({
  collections: { todos: todoCollection },
  mutationFns: { syncTodos: async ({ transaction, idempotencyKey }) => {...} }
})

const tx = offline.createOfflineTransaction({ mutationFnName: 'syncTodos' })
tx.mutate(() => todoCollection.insert({ id: '1', text: 'Buy milk' }))
await tx.commit() // Works offline!
```

This allows incremental adoption - add offline capabilities to critical user flows while keeping less critical operations simple.

## Performance Characteristics

- **Storage Operations**: IndexedDB writes are asynchronous and non-blocking
- **Memory Efficient**: Lazy loading and proper cleanup of completed transactions
- **Storage Optimized**: Automatic transaction pruning after successful execution
- **Network Smart**: Automatic retry coordination with exponential backoff

### DIFF
--- a/.changeset/dark-items-dig.md
+++ b/.changeset/dark-items-dig.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/powersync-db-collection": minor
+---
+
+Initial Release
--- a/docs/collections/powersync-collection.md
+++ b/docs/collections/powersync-collection.md
@@ -1,0 +1,475 @@
+---
+title: PowerSync Collection
+---
+
+# PowerSync Collection
+
+PowerSync collections provide seamless integration between TanStack DB and [PowerSync](https://powersync.com), enabling automatic synchronization between your in-memory TanStack DB collections and PowerSync's SQLite database. This gives you offline-ready persistence, real-time sync capabilities, and powerful conflict resolution.
+
+## Overview
+
+The `@tanstack/powersync-db-collection` package allows you to create collections that:
+
+- Automatically mirror the state of an underlying PowerSync SQLite database
+- Reactively update when PowerSync records change
+- Support optimistic mutations with rollback on error
+- Provide persistence handlers to keep PowerSync in sync with TanStack DB transactions
+- Use PowerSync's efficient SQLite-based storage engine
+- Work with PowerSync's real-time sync features for offline-first scenarios
+- Leverage PowerSync's built-in conflict resolution and data consistency guarantees
+- Enable real-time synchronization with PostgreSQL, MongoDB and MySQL backends
+
+## 1. Installation
+
+Install the PowerSync collection package along with your preferred framework integration.
+PowerSync currently works with Web, React Native and Node.js. The examples below use the Web SDK.
+See the PowerSync quickstart [docs](https://docs.powersync.com/installation/quickstart-guide) for more details.
+
+```bash
+npm install @tanstack/powersync-db-collection @powersync/web @journeyapps/wa-sqlite
+```
+
+### 2. Create a PowerSync Database and Schema
+
+```ts
+import { Schema, Table, column } from "@powersync/web"
+
+// Define your schema
+const APP_SCHEMA = new Schema({
+  documents: new Table({
+    name: column.text,
+    author: column.text,
+    created_at: column.text,
+    archived: column.integer,
+  }),
+})
+
+// Initialize PowerSync database
+const db = new PowerSyncDatabase({
+  database: {
+    dbFilename: "app.sqlite",
+  },
+  schema: APP_SCHEMA,
+})
+```
+
+### 3. (optional) Configure Sync with a Backend
+
+```ts
+import {
+  AbstractPowerSyncDatabase,
+  PowerSyncBackendConnector,
+  PowerSyncCredentials,
+} from "@powersync/web"
+
+// TODO implement your logic here
+class Connector implements PowerSyncBackendConnector {
+  fetchCredentials: () => Promise<PowerSyncCredentials | null>
+
+  /** Upload local changes to the app backend.
+   *
+   * Use {@link AbstractPowerSyncDatabase.getCrudBatch} to get a batch of changes to upload.
+   *
+   * Any thrown errors will result in a retry after the configured wait period (default: 5 seconds).
+   */
+  uploadData: (database: AbstractPowerSyncDatabase) => Promise<void>
+}
+
+// Configure the client to connect to a PowerSync service and your backend
+db.connect(new Connector())
+```
+
+### 4. Create a TanStack DB Collection
+
+There are two main ways to create a collection: using type inference or using schema validation. Type inference will infer collection types from the underlying PowerSync SQLite tables. Schema validation can be used for additional input/output validations and type transforms.
+
+#### Option 1: Using Table Type Inference
+
+The collection types are automatically inferred from the PowerSync schema table definition. The table is used to construct a default standard schema validator which is used internally to validate collection operations.
+
+Collection mutations accept SQLite types and queries report data with SQLite types.
+
+```ts
+import { createCollection } from "@tanstack/react-db"
+import { powerSyncCollectionOptions } from "@tanstack/powersync-db-collection"
+
+const documentsCollection = createCollection(
+  powerSyncCollectionOptions({
+    database: db,
+    table: APP_SCHEMA.props.documents,
+  })
+)
+
+/** Note: The types for input and output are defined as this */
+// Used for mutations like `insert` or `update`
+type DocumentCollectionInput = {
+  id: string
+  name: string | null
+  author: string | null
+  created_at: string | null // SQLite TEXT
+  archived: number | null // SQLite integer
+}
+// The type of query/data results
+type DocumentCollectionOutput = DocumentCollectionInput
+```
+
+The standard PowerSync SQLite types map to these TypeScript types:
+
+| PowerSync Column Type | TypeScript Type  | Description                                                          |
+| --------------------- | ---------------- | -------------------------------------------------------------------- |
+| `column.text`         | `string \| null` | Text values, commonly used for strings, JSON, dates (as ISO strings) |
+| `column.integer`      | `number \| null` | Integer values, also used for booleans (0/1)                         |
+| `column.real`         | `number \| null` | Floating point numbers                                               |
+
+Note: All PowerSync column types are nullable by default.
+
+#### Option 2: SQLite Types with Schema Validation
+
+Additional validations for collection mutations can be performed with a custom schema. The Schema below asserts that
+the `name`, `author` and `created_at` fields are required as input. `name` also has an additional string length check.
+
+Note: The input and output types specified in this example still satisfy the underlying SQLite types. An additional `deserializationSchema` is required if the typing differs. See the examples below for more details.
+
+The application logic (including the backend) should enforce that all incoming synced data passes validation with the `schema`. Failing to validate data will result in inconsistency of the collection data. This is a fatal error! An `onDeserializationError` handler must be provided to react to this case.
+
+```ts
+import { createCollection } from "@tanstack/react-db"
+import { powerSyncCollectionOptions } from "@tanstack/powersync-db-collection"
+import { z } from "zod"
+
+// Schema validates SQLite types but adds constraints
+const schema = z.object({
+  id: z.string(),
+  name: z.string().min(3, { message: "Should be at least 3 characters" }),
+  author: z.string(),
+  created_at: z.string(), // SQLite TEXT for dates
+  archived: z.number(),
+})
+
+const documentsCollection = createCollection(
+  powerSyncCollectionOptions({
+    database: db,
+    table: APP_SCHEMA.props.documents,
+    schema,
+    onDeserializationError: (error) => {
+      // Present fatal error
+    },
+  })
+)
+
+/** Note: The types for input and output are defined as this */
+// Used for mutations like `insert` or `update`
+type DocumentCollectionInput = {
+  id: string
+  name: string
+  author: string
+  created_at: string // SQLite TEXT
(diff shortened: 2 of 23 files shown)


luke-stdev001:
This looks like a great step in the right direction, you guys are shipping some incredible features!

I have some concerns around localStorage persistence due to some browser quirks we've been bitten with in the past (can be reproduced in Chrome, Firefox and Safari), specifically Chrome's issue with not persisting to localStorage in the event of a browser crash or OS crash (this can happen with batteries going flat in field-service type scenarios, or a power cut at a retail store when POS is trying to persist to localStorage, etc.).

https://issues.chromium.org/issues/41172643
https://github.com/odoo/odoo/issues/125037#issue-1756951412
https://github.com/odoo/odoo/issues/125037#issuecomment-2050429170

There are also issues where when the browser comes back online it may take 5-7 seconds - really this is just a guesstimate but roughly lines up with what we've seen - where writing to localStorage will not persist immediately and retries will need to be put into place to continue to retry until you've verified that we've written to localStorage successfully.

My main concern is that this will need to be factored in or it may result in lost transactions in an offline-first scenario where we are trying to write an order, or perhaps a fieldservice visit/notes/pictures.

Synchronous API also blocks the main thread and caused performance issues in our experience when reading or writing to it.

There are quirks we've been burnt by with persistence to IndexedDB that i'll edit this comment with shortly.

Are there any plans to introduce PGLite into the mix for local persistence on the write path, with persistence to disk via OPFS or something similar?

KyleAMathews:
> There are also issues where when the browser comes back online it may take 5-7 seconds - really this is just a guesstimate but roughly lines up with what we've seen - where writing to localStorage will not persist immediately and retries will need to be put into place to continue to retry until you've verified that we've written to localStorage successfully.

Oof! We'll definitely want to build in support for retries, etc. The write is async so we could expose some sort of way to know when a tx is for sure persisted.

> Synchronous API also blocks the main thread and caused performance issues in our experience when reading or writing to it.

Interesting — batching up writes perhaps would help? 

> Are there any plans to introduce PGLite into the mix for local persistence on the write path, with persistence to disk via OPFS or something similar?

PGLite would be a pretty heavy dependency — it's not doing anything special though around how it handles writes so no reason we need to bring it in.

We can definitely add an OPFS storage adapter as well.

luke-stdev001:
Thanks for getting back to me so quickly. Apologies in advance for the essay below.

> > There are also issues where when the browser comes back online it may take 5-7 seconds - really this is just a guesstimate but roughly lines up with what we've seen - where writing to localStorage will not persist immediately and retries will need to be put into place to continue to retry until you've verified that we've written to localStorage successfully.
> 
> Oof! We'll definitely want to build in support for retries, etc. The write is async so we could expose some sort of way to know when a tx is for sure persisted.

Great to hear it's on the radar/being handled. I think if TanStackDB could have sensible defaults around retries, etc. that could be tweaked and configured for those that want that more control this would be ideal.

> > Synchronous API also blocks the main thread and caused performance issues in our experience when reading or writing to it.
> 
> Interesting — batching up writes perhaps would help?

I'll try this in our current implementation and come back to you, that would probably help. As soon as we are throwing large numbers of order records into localStorage though we're having significant performance issues (eg. extended offline periods due to power or Fibre + Cell tower outages)

> > Are there any plans to introduce PGLite into the mix for local persistence on the write path, with persistence to disk via OPFS or something similar?
> 
> PGLite would be a pretty heavy dependency — it's not doing anything special though around how it handles writes so no reason we need to bring it in.

That's fair enough, likely unnecessary for many use-cases. 

> We can definitely add an OPFS storage adapter as well.

It's great to hear this is being considered. It would be ideal for our particular use-case around POS and Field-Service. With our POS we're handling well over 100,000 products, as well as many hundreds of thousands of parts product records. We're dealing with about 2.7 million contact records as well, so being able to squeeze as much out of the local device as possible in terms of r/w performance and being able to ensure persistence in the case the user does a cache flush while offline will definitely be something we want to explore. 

We have been playing with PGLite with OPFS, which has it's pros (things like pg_trgm and potentially pg_search in the future) give us a good-enough search capability locally with minimal work, however if there was a suitable alternative that could work directly on top of OPFS without needing that dependency we could definitely consider living without it. In our case an upfront loading time for first boot of the device to populate the DB with background sync is good enough to make it useful for us. 

For field-service we have extremely patchy cell data support when users are on the road in regional AU and NZ and it is very common to go offline for hours while still needing to support being able to write up reports, create quotes (for BDM use-cases) and do deliveries. This is all with the same data set I mentioned above for products and customers.

I understand ours is an extreme use-case, but I believe with OPFS support it would be entirely possible, and performant enough for us.

KyleAMathews:
That's a lot of data 😂 you'd almost certainly need persistence of data in order to load that offline which will be another design/engineering challenge. But we do want to be able to support millions of rows.

On search, DB's indexes are pluggable and the plan is to add trigram and BM25 eventually.

tigawanna:
Will this work on react native?

luke-stdev001:
> That's a lot of data 😂 you'd almost certainly need persistence of data in order to load that offline which will be another design/engineering challenge. But we do want to be able to support millions of rows.

haha, yes, we've been wrangling with this problem for awhile now and don't have any good solution yet beyond some hacky methods with branch-level shapes and last-touched-on datetime field & rules on postcodes included for that branch for when the contact was last interacted with at a branch or by a BDM, keeping a local hot cache of customer data and then our core product data as a single shape. It would be great to have it all local though, but it will be a challenge.

> On search, DB's indexes are pluggable and the plan is to add trigram and BM25 eventually.

Awesome to hear that Trigram and BM25 will potentially be possible with TanStackDB in the future, this would be a game-changer for sync-first/local-first use-cases. Our main requirement is around product and customer search which is where Trigram or BM25 would be incredibly useful to us. Orders, invoices, etc. can fail gracefully when offline, but product and customer data will need to be queried locally.

viktorbonino:
Do you have an estimated timeframe for when this will be merged?

KyleAMathews:
@viktorbonino next week perhaps

yunusyuksel:
Would this feature also work in an Expo project?

Does TanStack DB require detecting the user’s internet connectivity using packages like @react-native-community/netinfo?

packages/db/src/optimistic-action.ts
```diff
@@ -1,6 +1,15 @@
...
+import { OnMutateMustBeSynchronousError } from "./errors"
 import type { CreateOptimisticActionsOptions, Transaction } from "./types"
 
+function isPromiseLike(value: unknown): value is PromiseLike<unknown> {
+  return (
+    !!value &&
+    (typeof value === `object` || typeof value === `function`) &&
+    typeof (value as { then?: unknown }).then === `function`
+  )
+}
```
samwillis: This should live in a utils somewhere, maybe called "isThenable"?

packages/db/src/transactions.ts
```diff
@@ -15,6 +16,37 @@ import type {
   TransactionWithMutations,
 } from "./types"
 
+/**
+ * Check if we're in a browser top-level window (not a worker, SSR, or iframe).
+ * This helps avoid false positives in environments where multiple instances are legitimate.
+ */
+function isBrowserTopWindow(): boolean {
```
samwillis: Let's move this and the dupe checkered into a more relevant utils file.

samwillis (approved):
Checked files in the main DB package, only a couple of nits.

KyleAMathews:
@yunusyuksel

> Would this feature also work in an Expo project?

It needs an expo storage plugin added but then yes!

> Does TanStack DB require detecting the user’s internet connectivity using packages like @react-native-community/netinfo?

We should also explore adding this — sounds like a good plan

PR #560, opened 2025-09-16
refactor the Collection class into multiple manager classes

samwillis:
`Collection` had grown too large and complex - this moves everything into manager classes.


samwillis:
Copilots review complained of the circular dependancy between `CollecitonImpl` and the manager classes. I don't think this is an issue as its a type only circular dependency (there is no runtime circular dependency), and there isn't really a clean way around it without making the *significantly* code much more complex (a separate Collection interface would still have a circular dependency as it would need to have the manager class instances).

samwillis:
@kevin-dp either as part of your PR, a followup, or this one if it goes in second, we should consider moving `change-events.ts` into `colleciton/changes.ts` along with your refactored stuff.

samwillis:
@kevin-dp thanks! Yep that mix-in pattern is nice. The one drawback is that the inner mix-in doesn't see the types of the props on the outer mix-in (I don't think?). We would need to central abstract interface still. We have a quite a lot of two way dependancies through the separate functionality.

very open to suggestions on this

kevin-dp:
@samwillis You're right that inner ones don't know about state and methods introduced by outer ones. In typical OOP with traits (or abstract classes with multiple inheritance) each trait can define abstract state/methods it requires. We can also leverage abstract properties/methods in the mixin pattern. So, we can have the base class define abstract state/methods and each mixin can be defined as an abstract class (such that TS doesn't complain if they don't implement (all) the abstract properties/methods):
```ts
type Ctor<T = {}> = abstract new (...args: any[]) => T;

abstract class Core {
  abstract hookSchedule(ms: number, f: () => void): void;
}

function Scheduler<TBase extends Ctor<Core>>(Base: TBase) {
  abstract class Scheduler extends Base {
    hookSchedule(ms: number, f: () => void) {
      setTimeout(f, ms);
    }
  };
  return Scheduler
}

function NeedsScheduler<TBase extends Ctor<Core>>(Base: TBase) {
  abstract class NeedsScheduler extends Base {
    doThing() {
      this.hookSchedule(10, () => {/* ... */});
    }
  };
  return NeedsScheduler
}

class Big extends NeedsScheduler(Scheduler(Core)) {}
```

packages/db/src/collection/changes.ts
```diff
@@ -0,0 +1,168 @@
+import { NegativeActiveSubscribersError } from "../errors"
+import { CollectionSubscription } from "./subscription.js"
+import type { StandardSchemaV1 } from "@standard-schema/spec"
+import type { ChangeMessage, SubscribeChangesOptions } from "../types"
+import type { CollectionLifecycleManager } from "./lifecycle.js"
+import type { CollectionSyncManager } from "./sync.js"
+import type { CollectionEventsManager } from "./events.js"
+import type { CollectionImpl } from "./index.js"
+
+// depends on:
```
kevin-dp: Comments seem obsolete

packages/db/src/collection/sync.ts
```diff
@@ -0,0 +1,235 @@
...
+import type { CollectionStateManager } from "./state"
+import type { CollectionLifecycleManager } from "./lifecycle"
+
+export class CollectionSyncManager<
+  TOutput extends object = Record<string, unknown>,
+  TKey extends string | number = string | number,
+  TSchema extends StandardSchemaV1 = StandardSchemaV1,
+  TInput extends object = TOutput,
+> {
+  private collection!: CollectionImpl<TOutput, TKey, any, TSchema, TInput>
```
kevin-dp, line 22: What is the `!` at the end of the name? Is this a TS thing or some convention i don't know?

samwillis: These don't have an initialiser in the constructor - they are set in setDeps, so we either have to mark them as optional (then check in every method), or do this which says "I know I don't have an initialiser but I promise it will be there before I access it".

We can revisit this pattern later if you want and add a check in each method, but I think for this use case in an internal api it's fine.

PR #564, opened 2025-09-17
Move data loading concerns from join/order by to CollectionSubscription

kevin-dp:
Currently, our optimizations for the join and orderBy operators dynamically load data directly from the indexes. However, loading data should be a concern of the collection (subscription) instead of the query operators. Hence, we introduced a new `CollectionSubscription` class that represents a subscription that you get from calling `collection.subscribeChanges`. The subscription exposes `requestSnapshot` and `requestLimitedSnapshot` methods to dynamically load data.

The `requestSnapshot`  method loads a snapshot containing all values that fulfill the where clause you provide it (if you provide any). The `requestLimitedSnapshot` method will only load N items (configurable) that are bigger than a minimum value (optional). This method is used by the orderBy optimization to load only the amount of values it needs.


packages/db/src/collection-subscription.ts
```diff
@@ -0,0 +1,239 @@
...
+    return true
+  }
+
+  /**
+   * Sends a snapshot that is limited to the first `limit` rows that fulfill the `where` clause and are bigger than `minValue`.
+   * Requires a range index to be set with `setOrderByIndex` prior to calling this method.
+   * It uses that range index to load the items in the order of the index.
+   * Note: it does not send keys that have already been sent before.
+   */
+  requestLimitedSnapshot({ limit, minValue }: RequestLimitedSnapshotOptions) {
```
samwillis, line 143: Currently if there is an order by on a joined collection we only perform one of the optimisations? Is it only the join lookup or the orderBy?

This `requestLimitedSnapshot` doesn't take a `where` and so it can't support both a join and a orderBy index lookup at the same time.

kevin-dp: Exactly. If there is both a join and an orderBy (with a limit) we optimise only the orderBy because that's the first one [we check in the code](https://github.com/TanStack/db/blob/eb4e3eb4c5e56dbfe225b253e41da7a0924b7fa6/packages/db/src/query/live/collection-subscriber.ts#L60-L67). The idea is that the limit is usually rather small and so it's better to load only a limited number of rows instead of all rows that match the join.

PR #565, opened 2025-09-17
fix: use preload internally in stateWhenReady and toArrayWhenReady methods

### LINKED ISSUE #563: stateWhenReady should use preload internally
https://github.com/TanStack/db/blob/58534e24838f5712f4adac55f1b9f85dd029800f/packages/db/src/collection.ts#L2267-L2279

A discord user asked why they seem to work differently as `stateWhenReady` wasn't actually fully waiting under some circumstances. So we should just have the whenReady methods use preload under the hood to keep logic the same.

KyleAMathews:
## Summary
- Fixed `stateWhenReady()` and `toArrayWhenReady()` methods to consistently wait for collections to be fully ready
- Both methods now use the `preload()` method internally instead of passively waiting with `onFirstReady()`

## Problem
A Discord user reported that `stateWhenReady` wasn't consistently "fully waiting" in some circumstances. The issue was that these methods would wait passively using `onFirstReady()` without actually triggering the loading process if the collection hadn't started syncing yet.

## Solution
By using `preload()` internally, both methods now:
1. Ensure the collection starts loading if it hasn't already
2. Properly handle the loading state and promise management
3. Avoid duplicate sync starts through preload's promise caching

## Test Plan
- [x] All existing tests pass
- [x] TypeScript build succeeds without errors
- [x] The methods still return immediately when data is already available
- [x] The methods properly trigger loading when collection is idle

Fixes #563

### DIFF
--- a/.changeset/breezy-queens-drive.md
+++ b/.changeset/breezy-queens-drive.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+Fix `stateWhenReady()` and `toArrayWhenReady()` methods to consistently wait for collections to be ready by using `preload()` internally. This ensures the collection starts loading if needed rather than just waiting passively.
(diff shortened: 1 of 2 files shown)


samwillis (approved):
:shipit:

PR #570, opened 2025-09-18
Run graph until no more pending work

kevin-dp:


kevin-dp:
@samwillis I don't think this PR requires a changeset? It's only internal, non-user-facing changes.

packages/db-ivm/src/operators/distinct.ts
```diff
@@ -32,19 +36,23 @@ export class DistinctOperator<T> extends UnaryOperator<T> {
     // Compute the new multiplicity for each value
     for (const message of this.inputMessages()) {
       for (const [value, diff] of message.getInner()) {
+        console.log(`value: `, JSON.stringify(value, null, 2))
+        console.log(
+          `this.#by(value): `,
+          JSON.stringify(this.#by(value), null, 2)
+        )
```
samwillis: stray console.log

PR #571, opened 2025-09-19
refactor joins to use direct implementation or each type rather than composition of inner+anti joins

samwillis:
This removes all the additional state that was needed for each of the composed joins, and the need for the consolidate step. Delivers a father 50% speedup over the index refactor PR #549 this is stacked on.

### DIFF
--- a/packages/db-ivm/src/operators/join.ts
+++ b/packages/db-ivm/src/operators/join.ts
@@ -1,3 +1,53 @@
+/**
+ * # Direct Join Algorithms for Incremental View Maintenance
+ *
+ * High-performance join operations implementing all join types (inner, left, right, full, anti)
+ * with minimal state and optimized performance.
+ *
+ * ## Algorithm
+ *
+ * For each tick, the algorithm processes incoming changes (deltas) and emits join results:
+ *
+ * 1. **Build deltas**: Extract new/changed/deleted rows from input messages
+ * 2. **Inner results**: Emit `ΔA⋈B_old + A_old⋈ΔB + ΔA⋈ΔB` (matched pairs)
+ * 3. **Outer results**: For unmatched rows, emit null-extended tuples:
+ *    - New unmatched rows from deltas (when opposite side empty)
+ *    - Presence transitions: when key goes `0→>0` (retract nulls) or `>0→0` (emit nulls)
+ * 4. **Update state**: Append deltas to indexes and update mass counters
+ *
+ * **Mass tracking** enables O(1) presence checks instead of scanning index buckets.
+ *
+ * ## State
+ *
+ * **Indexes** store the actual data:
+ * - `indexA: Index<K, V1>` - all left-side rows accumulated over time
+ * - `indexB: Index<K, V2>` - all right-side rows accumulated over time
+ *
+ * **Mass maps** track presence efficiently:
+ * - `massA/massB: Map<K, number>` - sum of multiplicities per key
+ * - Used for O(1) presence checks: `mass.get(key) !== 0` means key exists
+ * - Avoids scanning entire index buckets just to check if key has any rows
+ *
+ * ## Join Types
+ *
+ * - **Inner**: Standard delta terms only
+ * - **Outer**: Inner results + null-extended unmatched rows with transition handling
+ * - **Anti**: Unmatched rows only (no inner results)
+ *
+ * ## Key Optimizations
+ *
+ * - **No temp copying**: Uses `(A⊎ΔA)⋈ΔB = A⋈ΔB ⊎ ΔA⋈ΔB` distributive property
+ * - **Early-out checks**: Skip phases when no deltas present
+ * - **Zero-entry pruning**: Keep maps compact, O(distinct keys) memory
+ * - **Final presence logic**: Avoid emit→retract churn within same tick
+ *
+ * ## Correctness
+ *
+ * - **Ordering**: Pre-append snapshots for emissions, post-emit state updates
+ * - **Presence**: Key matched iff mass ≠ 0, transitions trigger null handling
+ * - **Bag semantics**: Proper multiplicity handling including negatives
+ */
+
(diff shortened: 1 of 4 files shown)


packages/db-ivm/src/operators/join.ts
```diff
@@ -1,10 +1,57 @@
...
+ * **Mass tracking** enables O(1) presence checks instead of scanning index buckets.
+ *
+ * ## State
+ *
+ * **Indexes** store the actual data:
+ * - `indexA: Index<K, V1>` - all left-side rows accumulated over time
+ * - `indexB: Index<K, V2>` - all right-side rows accumulated over time
+ *
+ * **Mass maps** track presence efficiently:
+ * - `massA/massB: Map<K, number>` - sum of multiplicities per key
```
kevin-dp: Essentially, this map keeps a consolidated multiset. I had a version of the `Index` that internally keeps such a map instead of an array. I think it makes more sense to modify the index to keep a consolidated map, rather than having the index and then also keeping a map here. This is a concern of the index.

packages/db-ivm/src/operators/join.ts
```diff
@@ -14,66 +61,204 @@ import type { IStreamBuilder, KeyValue, PipedOperator } from "../types.js"
 export type JoinType = `inner` | `left` | `right` | `full` | `anti`
 
 /**
- * Operator that joins two input streams
+ * Helper to build delta index and mass map from messages
+ */
+function buildDelta<K, V>(
```
kevin-dp: This is definitely an index concern and not a join concern. The index should keep the "mass" (btw i think this is a very bad name, it's just (consolidated) multiplicity) and the join operator should simply do `index.addValue(key, [value, multiplicity])` when it's going through the messages. No need for `buildDelta`.

packages/db-ivm/src/operators/join.ts
```diff
@@ -14,66 +61,204 @@ import type { IStreamBuilder, KeyValue, PipedOperator } from "../types.js"
...
+    const hasDeltaB = deltaB.size > 0
+    const hasDeltaMassA = deltaMassA.size > 0
+    const hasDeltaMassB = deltaMassB.size > 0
+
+    // If nothing happened, bail early
+    if (!(hasDeltaA || hasDeltaB || hasDeltaMassA || hasDeltaMassB)) return
+
+    // Precompute mode flags to avoid repeated string comparisons
+    const mode = this.#mode
+    const emitInner =
```
kevin-dp: isn't `emitInner` always true?

kevin-dp: Ah, no it's false when we're doing an anti join. Perhaps `const emitInner = mode !== 'anti'` is more accurate.

packages/db-ivm/src/operators/join.ts
```diff
@@ -14,66 +61,204 @@ import type { IStreamBuilder, KeyValue, PipedOperator } from "../types.js"
...
+          // Skip if presence doesn't flip (0->0, >0->different>0)
+          if ((before === 0) === (after === 0)) continue
+
+          const it = this.#indexA.getIterator(key)
+          const retract = before === 0 // 0->!0 => retract, else (>0->0) emit
+          for (const [value, multiplicity] of it) {
+            if (multiplicity !== 0) {
+              results.add(
+                [key, [value, null]],
+                retract ? -multiplicity : +multiplicity
```
kevin-dp: I don't get the retract logic. If the key used to have multiplicity 0, then `retract` is `true` so we will add the new change to the index but with its negated multiplicity, why?

packages/db-ivm/src/operators/join.ts
```diff
@@ -14,66 +61,204 @@ import type { IStreamBuilder, KeyValue, PipedOperator } from "../types.js"
...
+
+      // 3b) Right-side presence transitions (only if some RHS masses changed)
+      if (hasDeltaMassB) {
+        for (const [key, deltaMass] of deltaMassB) {
+          const before = this.#massB.get(key) || 0
+          if (deltaMass === 0) continue
+          const after = before + deltaMass
+
+          // Skip if presence doesn't flip (0->0, >0->different>0)
+          if ((before === 0) === (after === 0)) continue
```
kevin-dp: Even if the presence doesn't flip, don't we need to update the indexes/mass index ?
Say, that `before` is `5` and the new delta has multiplicity `-3`. Then `after` is `2`. So it is still visible and this won't update the index so in the index the key will still have multiplicity `5`. Next time, if it receives a multiplicity of `-3` the key will still not flip (even though it should at this point!) because it thinks the multiplicity is 5. This looks wrong to me?

packages/db-ivm/src/operators/join.ts
```diff
@@ -14,66 +61,204 @@ import type { IStreamBuilder, KeyValue, PipedOperator } from "../types.js"
...
     inputB: DifferenceStreamReader<[K, V2]>,
-    output: DifferenceStreamWriter<[K, [V1, V2]]>
+    output: DifferenceStreamWriter<any>,
+    mode: JoinType = `inner`
   ) {
     super(id, inputA, inputB, output)
+    this.#mode = mode
   }
 
   run(): void {
```
kevin-dp, line 84: I would introduce helper functions to simplify this method and that way we could also unit test the helper functions in isolation. This method contains 4 steps, each step could be extracted to a helper function.

packages/db-ivm/src/operators/join.ts
samwillis: I've improved the comments and naming to try and explain this better, it's not really retraction, it that we are transitioning from not having any matches, to having matches, and so we need to send a retraction of the old row, with its old multiplicity so that the new rows can be sent.

packages/db-ivm/src/operators/join.ts
samwillis: Updating the multiplicity in the indexes happens regardless of this section of code, at the end of `run()`.

PR #572, opened 2025-09-19
electric-db-collection: Pull in the latest version of Electric client instead of pinning it

lirbank:


KyleAMathews:
We do want it as a direct dependency as most people won't use it directly. We just want to bump the version and add the caret so the latest version is always pulled in

KyleAMathews (approved):

PR #573, opened 2025-09-19
Remove peerDependencies from electric-db-collection

KyleAMathews:
We have @electric-sql/client as a dependency and you don't want both.


lirbank:
Suggestion, remove the dev dev dependency too:

https://github.com/TanStack/db/blob/d1056c2dee335196aaaaf687c1358f55aaddde0a/packages/electric-db-collection/package.json#L13

KyleAMathews:
Good call 👍

Issue #589, opened 2025-09-25
Use requestIdleCallback for collection cleanup as it's not urgent exactly when it happens


KyleAMathews:
This can prevent needless blocking of the UI thread when other stuff is going on.

samwillis:
Complete in #590

PR #590, opened 2025-09-25
feat: implement idle cleanup for collection garbage collection

KyleAMathews:
## Summary
- Implements `requestIdleCallback` for collection cleanup operations to prevent blocking the UI thread
- Addresses issue #589 by scheduling cleanup during browser idle time instead of immediately
- Includes polyfill fallback for older browsers that don't support `requestIdleCallback`

## Key Changes
- Modified `CollectionLifecycleManager` to use `requestIdleCallback()` for non-blocking cleanup
- Added `scheduleIdleCleanup()` method that schedules cleanup during idle periods
- Added `performCleanup()` method that handles actual cleanup operations with idle time budgeting
- Implemented proper callback ID management to prevent race conditions during cleanup rescheduling
- Updated tests to handle the new asynchronous cleanup behavior

## Benefits
- **Non-blocking UI**: Collection cleanup no longer blocks the main thread during active interactions
- **Better performance**: Cleanup happens during browser idle periods when CPU is available  
- **Backward compatibility**: Works in all browsers with automatic fallback to `setTimeout`
- **No breaking changes**: All existing functionality remains intact

## Test Plan
- [x] All existing lifecycle tests pass (16/16)
- [x] Cleanup operations are properly scheduled during idle time
- [x] Fallback works correctly in test environment
- [x] Callback ID management prevents race conditions
- [x] Explicit cleanup calls still work immediately
- [x] No TypeScript errors
- [x] Linting passes

### DIFF
--- a/packages/db/src/collection/lifecycle.ts
+++ b/packages/db/src/collection/lifecycle.ts
@@ -2,6 +2,11 @@ import {
   CollectionInErrorStateError,
   InvalidCollectionStatusTransitionError,
 } from "../errors"
+import {
+  safeCancelIdleCallback,
+  safeRequestIdleCallback,
+} from "../utils/browser-polyfills"
+import type { IdleCallbackDeadline } from "../utils/browser-polyfills"
 import type { StandardSchemaV1 } from "@standard-schema/spec"
 import type { CollectionConfig, CollectionStatus } from "../types"
 import type { CollectionEventsManager } from "./events"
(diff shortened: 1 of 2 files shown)


packages/db/src/collection/lifecycle.ts
```diff
@@ -10,6 +10,46 @@ import type { CollectionChangesManager } from "./changes"
...
+) => number =
+  typeof window !== `undefined` && `requestIdleCallback` in window
+    ? (callback, options) =>
+        (window as any).requestIdleCallback(callback, options)
+    : (callback, _options) => requestIdleCallbackPolyfill(callback)
+
+const safeCancelIdleCallback: (id: number) => void =
+  typeof window !== `undefined` && `cancelIdleCallback` in window
+    ? (id) => (window as any).cancelIdleCallback(id)
+    : cancelIdleCallbackPolyfill
```
samwillis: Can these move to a utils file rather than inline in lifecycle.ts

KyleAMathews: done 👍

PR #595, opened 2025-09-26
Fix joins using conditions with computed values

duailibe:
## Summary
- Fix joins using conditions with computed values like \`concat()\` by preventing optimization that causes type mismatches in followRef
- Add test coverage for joins with computed expressions to ensure proper handling


samwillis:
Thanks @duailibe! This look correct to me but I'm going to ask @kevin-dp to cast his eyes over it as he worked on the optimisations.

PR #532, opened 2025-09-11
fix stuck loading state after an electric must-refetch

samwillis:
also fixes an inconstant state bug.

PR #603, opened 2025-09-29
fix repeated renders when markReady called when collection already ready

### LINKED ISSUE #597: Prevent triggering `subscribeChanges` when long-polling request finishes with `up-to-date`
I noticed in our app that every time a long-polling request finishes, it causes all associated queries on that collection to return a new `data` object. In React, this triggers a lot of unnecessary re-renders.

This appears to be because the snapshot version is bumped whenever `subscribeChanges` is called, even if the list of changes is empty.

https://github.com/TanStack/db/blob/59b406edb57a087f440eace8fc6a75453d91796e/packages/react-db/src/useLiveQuery.ts#L390-L394

When I patch @tanstack/react-db as below, the problem goes away.

Image

samwillis:
This fixes #597

We were emitting an empty change event whenever `markReady` was called, which would happed after each Electric long polling request returned.

The code related to this was changed previous to add this behaviour in #532 to fix a suck loading state. That PR introduced a test case that failed without this previous change. I was a little unsure if my change here would cause a regression, but it has not - I think this is due to other changes in the subscription system that has ensured the race condition is now fixed.

I'm comfortable that the previous test passes and so this is not introducing a regression.

### DIFF
--- a/packages/electric-db-collection/tests/electric-live-query.test.ts
+++ b/packages/electric-db-collection/tests/electric-live-query.test.ts
@@ -386,13 +386,6 @@ describe.each([
 
     // The live query should not have emitted any changes
     // because there were no actual data changes
-    // BUG: Currently emits empty change arrays on up-to-date messages
-    if (changeNotifications.length > 0) {
-      console.log(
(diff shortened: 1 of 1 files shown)


samwillis:
stray, `.only` tricked me into thinking this was passing 😢 - closing so I can investigate and fix properly.

PR #604, opened 2025-09-29
fix repeated renders when markReady called when collection already ready

### LINKED ISSUE #597: Prevent triggering `subscribeChanges` when long-polling request finishes with `up-to-date`

samwillis:
This fixes #597

We were emitting an empty change event whenever `markReady` was called, which would happed after each Electric long polling request returned.

The code that triggers these additional empty events was added in #532 to fix a suck loading state. That PR introduced a test case that failed without this previous change. My change here initially caused a regression, but I went digging and found the *real* case of the stuck loading state and have fixed that 🥳

### DIFF
--- a/.changeset/polite-eels-laugh.md
+++ b/.changeset/polite-eels-laugh.md
@@ -3,4 +3,4 @@
 "@tanstack/db": patch
 ---
 
-Fix repeated renders when markReady called when the collection was already ready. This would occur after each long poll on a Electric collection.
+Fix repeated renders when markReady called when the collection was already ready. This would occur after each long poll on an Electric collection.
--- a/.changeset/polite-eels-laugh.md
+++ b/.changeset/polite-eels-laugh.md
@@ -1,0 +1,6 @@
+---
+"@tanstack/electric-db-collection": patch
+"@tanstack/db": patch
+---
+
+Fix repeated renders when markReady called when the collection was already ready. This would occur after each long poll on a Electric collection.
--- a/packages/db/src/collection/lifecycle.ts
+++ b/packages/db/src/collection/lifecycle.ts
@@ -146,12 +146,12 @@ export class CollectionLifecycleManager<
         this.onFirstReadyCallbacks = []
         callbacks.forEach((callback) => callback())
       }
-    }
 
-    // Always notify dependents when markReady is called, after status is set
-    // This ensures live queries get notified when their dependencies become ready
-    if (this.changes.changeSubscriptions.size > 0) {
-      this.changes.emitEmptyReadyEvent()
+      // Notify dependents when markReady is called, after status is set
+      // This ensures live queries get notified when their dependencies become ready
+      if (this.changes.changeSubscriptions.size > 0) {
+        this.changes.emitEmptyReadyEvent()
+      }
     }
   }
 
--- a/packages/db/src/collection/lifecycle.ts
+++ b/packages/db/src/collection/lifecycle.ts
@@ -96,8 +96,11 @@ export class CollectionLifecycleManager<
     allowReady: boolean = false
   ): void {
     if (newStatus === `ready` && !allowReady) {
+      // setStatus('ready') is an internal method that should not be called directly
+      // Instead, use markReady to transition to ready triggering the necessary events
+      // and side effects.
       throw new CollectionStateError(
-        `Cannot set status to anything other than ready, must use markReady instead`
+        `You can't directly call "setStatus('ready'). You must use markReady instead.`
       )
     }
     this.validateStatusTransition(this.status, newStatus)
--- a/packages/db/src/collection/lifecycle.ts
+++ b/packages/db/src/collection/lifecycle.ts
@@ -1,5 +1,6 @@
 import {
   CollectionInErrorStateError,
+  CollectionStateError,
   InvalidCollectionStatusTransitionError,
 } from "../errors"
 import {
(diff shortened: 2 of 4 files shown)


packages/db/src/collection/state.ts
```diff
@@ -644,7 +644,7 @@ export class CollectionStateManager<
 
         // Ensure listeners are active before emitting this critical batch
         if (this.lifecycle.status !== `ready`) {
-          this.lifecycle.setStatus(`ready`)
+          this.lifecycle.markReady()
```
samwillis, line 647: This change fixed what the *real* cause of the stuck loading status bug from #532. We should always use `markReady` when transitioning to ready.

packages/db/src/collection/lifecycle.ts
```diff
@@ -90,7 +91,15 @@ export class CollectionLifecycleManager<
...
-  public setStatus(newStatus: CollectionStatus): void {
+  public setStatus(
+    newStatus: CollectionStatus,
+    allowReady: boolean = false
+  ): void {
+    if (newStatus === `ready` && !allowReady) {
+      throw new CollectionStateError(
+        `Cannot set status to anything other than ready, must use markReady instead`
+      )
+    }
```
samwillis, line 105: I added this check to ensure we don't make the same mistake again when transitioning to ready - see other comment below.

packages/db/src/collection/lifecycle.ts
```diff
@@ -146,12 +156,11 @@ export class CollectionLifecycleManager<
...
-
-    // Always notify dependents when markReady is called, after status is set
-    // This ensures live queries get notified when their dependencies become ready
-    if (this.changes.changeSubscriptions.size > 0) {
-      this.changes.emitEmptyReadyEvent()
+      // Notify dependents when markReady is called, after status is set
+      // This ensures live queries get notified when their dependencies become ready
+      if (this.changes.changeSubscriptions.size > 0) {
+        this.changes.emitEmptyReadyEvent()
+      }
```
samwillis, line 166: Moving this inside the `if` above ensures we only send this empty change event when we first transition to ready. This is the fix to prevent additional renders.

packages/db/src/collection/state.ts
KyleAMathews: Maybe we just remove setStatus? It's nice to have proper functions where we can put in more logic when needed.

KyleAMathews: not needed for this PR of course but a thought

packages/db/src/collection/lifecycle.ts
```diff
@@ -90,7 +91,15 @@ export class CollectionLifecycleManager<
...
    * @private
    */
-  public setStatus(newStatus: CollectionStatus): void {
+  public setStatus(
+    newStatus: CollectionStatus,
+    allowReady: boolean = false
+  ): void {
+    if (newStatus === `ready` && !allowReady) {
+      throw new CollectionStateError(
+        `Cannot set status to anything other than ready, must use markReady instead`
```
KyleAMathews:
```suggestion
        `You can't directly call "setStatus('ready'). You must use markReady instead.`
```

packages/db/src/collection/lifecycle.ts
```diff
@@ -90,7 +91,15 @@ export class CollectionLifecycleManager<
    * Safely update the collection status with validation
    * @private
    */
-  public setStatus(newStatus: CollectionStatus): void {
+  public setStatus(
+    newStatus: CollectionStatus,
+    allowReady: boolean = false
+  ): void {
```
KyleAMathews, line 97: We should prevent same status calls: `if (newStatus === this.status) return`

packages/db/src/collection/lifecycle.ts
```diff
@@ -90,7 +91,15 @@ export class CollectionLifecycleManager<
...
+    if (newStatus === `ready` && !allowReady) {
```
KyleAMathews, line 98: Add a comment (or add to the error) adding that `setStatus('ready')` is for internal use only.

.changeset/polite-eels-laugh.md
```diff
@@ -0,0 +1,6 @@
+---
+"@tanstack/electric-db-collection": patch
+"@tanstack/db": patch
+---
+
+Fix repeated renders when markReady called when the collection was already ready. This would occur after each long poll on a Electric collection.
```
KyleAMathews:
```suggestion
Fix repeated renders when markReady called when the collection was already ready. This would occur after each long poll on an Electric collection.
```

packages/db/src/collection/lifecycle.ts
samwillis: I'd like to change this in a followup PR as currently at least one test fails if we enforce that.

samwillis: I would like to address this in a follow-up pr, when this is done it triggers at least one test failure.

PR #605, opened 2025-09-29
Remove circular imports

bradleyayers:
## 🎯 Changes

Refactor code to remove circular imports so that the code is compatible with Metro bundler.

- Enabled `import-x/no-cycle` to prevent these issues in the future. This seems like something that should be upstreamed to `@tanstack/config/eslint` but since this is my first contribution I'm not sure what is involved in doing that, and it seems like something can be done in the future but including it now in this package provides immediate value.
- Merged `compiler/joins.ts` into `compiler/index.ts` to remove circular dependency between `compileQuery()` and `processJoins()`.
- Moved `followRef` into `ir.ts`.
- After fixing the configuration for `import-x` to work with TypeScript a number of other import linting errors were fixed as a byproduct (e.g. merging together multiple import statements).

I've done this by 

## ✅ Checklist

## 🚀 Release Impact


samwillis:
Thanks for digging into this @bradleyayers, looks great.

As you kind of guessed in your comment on the issue my one hesitation is the moving of the join compilation into the index.js file. I understand why that was needed to removed the circular import, but I'd like to consider an alternative. In future we are likely to have other operators/methods that can take a sub query and so will also need to recursively compile.

I wander if we can pass a callback to the join compiler function, an onCompileSubquery, that can be used to inject the compile function?

The function signature of join has grown too large, and we need to refactor this up, but for now we could get add it as another arg.

bradleyayers:
Yep can do, are you happy with followRef moving to ir.ts? or is there a better place for it?

samwillis:
Yep, moving `followRef` is fine, ir.ts seems like a good place for it.

bradleyayers:
@samwillis okay I've done that, it was a bit awkward to define the type of the `compileQuery` function because it doesn't take an options object, so I've just gone with the simplest minimal diff I could come up with. The callback approach worked out much better.

bradleybernard:
Awesome, thank you both!

PR #606, opened 2025-09-29
Update starters to TanStack Start RC

KyleAMathews:
## 🎯 Changes

TanStack Start RC had some breaking changes — this upgrades the examples/starters


examples/react/projects/src/routes/_authenticated/index.tsx
```diff
@@ -33,7 +33,7 @@ function IndexRedirect() {
   const navigate = useNavigate()
   const { data: projects } = useLiveQuery((q) => q.from({ projectCollection }))
 
-  useEffect(() => {
+  React.useEffect(() => {
```
samwillis: Is this a change to match the start template / generated code, or an LLM thing?

examples/react/projects/src/routes/login.tsx
```diff
@@ -1,22 +1,22 @@
+import * as React from "react"
```
samwillis: This again seems unlike how most people use React - the code here without the `* and React` was cleaner in my opinion.

samwillis (approved):
Look good, nothing serious.

Only comment on the changing of react imports from `import { useEffect } from "react"` to `import * as React from "react"`. LLM thing or to match a pattern from Start?

PR #617, opened 2025-10-01
Predicate pushdown to sync

kevin-dp:
This PR modifies tanstack DB to push down predicates to the sync layer.

PR #618, opened 2025-10-01
Handle pushed down predicates in Electric collection

kevin-dp:
This PR is a follow up on https://github.com/TanStack/db/pull/617 and modifies the Electric collection to handle predicates that are being pushed down to the Electric collection.

### DIFF
--- a/packages/electric-db-collection/src/electric.ts
+++ b/packages/electric-db-collection/src/electric.ts
@@ -419,7 +419,7 @@ function createElectricSync<T extends Row<unknown>>(
       })
 
       return {
-        onLoadMore: (opts) => onLoadMore(params, opts),
+        onLoadMore: (opts) => onLoadMore(stream, params, opts),
         cleanup: () => {
           // Unsubscribe from the stream
           unsubscribeStream()
@@ -434,6 +434,7 @@ function createElectricSync<T extends Row<unknown>>(
 }
 
 async function onLoadMore<T extends Row<unknown>>(
+  stream: ShapeStream<T>,
   syncParams: Parameters<SyncConfig<T>[`sync`]>[0],
   options: OnLoadMoreOptions
 ) {
@@ -444,7 +445,7 @@ async function onLoadMore<T extends Row<unknown>>(
 
   const snapshotParams = compileSQL<T>(options)
 
-  const snapshot = await requestSnapshot(snapshotParams)
+  const snapshot = await stream.requestSnapshot(snapshotParams)
 
   begin()
 
--- a/packages/electric-db-collection/src/electric.ts
+++ b/packages/electric-db-collection/src/electric.ts
@@ -439,22 +439,15 @@ async function onLoadMore<T extends Row<unknown>>(
   options: OnLoadMoreOptions
 ) {
   const { begin, write, commit } = syncParams
-
-  // TODO: optimize this by keeping track of which snapshot have been loaded already
-  //       and only load this one if it's not a subset of the ones that have been loaded already
-
   const snapshotParams = compileSQL<T>(options)
-
   const snapshot = await stream.requestSnapshot(snapshotParams)
 
   begin()
-
   snapshot.data.forEach((row) => {
     write({
       type: `insert`,
       value: row.value,
     })
   })
-
   commit()
 }
--- a/packages/electric-db-collection/src/electric.ts
+++ b/packages/electric-db-collection/src/electric.ts
@@ -15,12 +15,10 @@ import {
 import { compileSQL } from "./sql-compiler"
 import type {
   BaseCollectionConfig,
-  Collection,
   CollectionConfig,
   DeleteMutationFnParams,
   Fn,
   InsertMutationFnParams,
-  OnLoadMoreOptions,
   SyncConfig,
   UpdateMutationFnParams,
   UtilsRecord,
@@ -420,9 +418,9 @@ function createElectricSync<T extends Row<unknown>>(
       })
 
       return {
-        onLoadMore: (opts) => {
+        onLoadMore: async (opts) => {
           const snapshotParams = compileSQL<T>(opts)
-          return stream.requestSnapshot(snapshotParams)
+          await stream.requestSnapshot(snapshotParams)
         },
         cleanup: () => {
           // Unsubscribe from the stream
@@ -436,4 +434,3 @@ function createElectricSync<T extends Row<unknown>>(
     getSyncMetadata,
   }
 }
-
--- a/packages/electric-db-collection/src/electric.ts
+++ b/packages/electric-db-collection/src/electric.ts
@@ -15,6 +15,7 @@ import {
 import { compileSQL } from "./sql-compiler"
 import type {
   BaseCollectionConfig,
+  Collection,
   CollectionConfig,
   DeleteMutationFnParams,
   Fn,
@@ -419,7 +420,10 @@ function createElectricSync<T extends Row<unknown>>(
       })
 
       return {
-        onLoadMore: (opts) => onLoadMore(stream, params, opts),
+        onLoadMore: (opts) => {
+          const snapshotParams = compileSQL<T>(opts)
+          return stream.requestSnapshot(snapshotParams)
+        },
         cleanup: () => {
           // Unsubscribe from the stream
           unsubscribeStream()
@@ -433,21 +437,3 @@ function createElectricSync<T extends Row<unknown>>(
   }
 }
 
-async function onLoadMore<T extends Row<unknown>>(
-  stream: ShapeStream<T>,
-  syncParams: Parameters<SyncConfig<T>[`sync`]>[0],
-  options: OnLoadMoreOptions
-) {
-  const { begin, write, commit } = syncParams
-  const snapshotParams = compileSQL<T>(options)
-  const snapshot = await stream.requestSnapshot(snapshotParams)
-
-  begin()
-  snapshot.data.forEach((row) => {
-    write({
-      type: `insert`,
-      value: row.value,
-    })
-  })
-  commit()
-}
--- a/packages/electric-db-collection/src/electric.ts
+++ b/packages/electric-db-collection/src/electric.ts
@@ -12,12 +12,14 @@ import {
   ExpectedNumberInAwaitTxIdError,
   TimeoutWaitingForTxIdError,
 } from "./errors"
+import { compileSQL } from "./sql-compiler"
 import type {
   BaseCollectionConfig,
   CollectionConfig,
   DeleteMutationFnParams,
   Fn,
   InsertMutationFnParams,
+  OnLoadMoreOptions,
   SyncConfig,
   UpdateMutationFnParams,
   UtilsRecord,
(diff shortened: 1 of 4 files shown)


packages/electric-db-collection/src/sql-compiler.ts
```diff
@@ -0,0 +1,163 @@
...
+    compileOrderByClause(clause, params)
+  )
+  return compiledOrderByClauses.join(`,`)
+}
+
+function compileOrderByClause(
+  clause: IR.OrderByClause,
+  params: Array<unknown>
+): string {
+  // TODO: what to do with stringSort and locale?
```
samwillis: This raises the question on if we have the wrong default in DB, should we default to lexical sort?

packages/electric-db-collection/tests/electric-live-query.test.ts
```diff
@@ -437,4 +466,444 @@ describe.each([
...
+      })
+    )
+
+    // After receiving Bob and Charlie, the collection now has 3 users (Alice + Bob + Charlie)
+    // but it still requests 2 more... TODO: check if this is correct?
+    expect(callArgs(1)).toMatchObject({
+      limit: 2,
+      orderBy: `age NULLS FIRST`,
+    })
+  })
```
samwillis: @kevin-dp not sure why it does this

packages/electric-db-collection/src/sql-compiler.ts
kevin-dp: Even if we default to lexical sort, what do we do if the user explicitly chooses a locale compare? Translating that into SQL is non-trivial.

packages/electric-db-collection/tests/electric-live-query.test.ts
kevin-dp: You simulated an initial sync with just 1 user (Alice)
```ts
simulateInitialSync([sampleUsers[0]!])
```
Then you make this live query collection:
```ts
createLiveQueryCollection({
  id: `expanding-live-query`,
  startSync: true,
  query: (q) =>
    q
      .from({ user: electricCollection })
      .orderBy(({ user }) => user.age, `asc`)
      .limit(3),
})
```
The live query asks for 3 results, so we request a limited snapshot (limit = 3). `requestLimitedSnapshot` immediately tries to load this data from the collection, so it loads Alice. Then it also asks the collection to sync more but it doesn't await this (L218 in `packages/db/src/collection/subscription.ts`). So this is the original snapshot request with limit 3:
```ts
callArgs(0) is {
  params: {},
  orderBy: 'age NULLS FIRST',
  limit: 3
}
```
At the end of the live query run, we notice that the top 3 is not filled since it contains only 1 user. So we request a snapshot for 2 more users:
```ts
callArgs(1) is {
  params: { '1': '25' },
  where: 'age > $1',
  orderBy: 'age NULLS FIRST',
  limit: 2
}
```
We will only receive Bob and Charlie from this snapshot request (whose result is mocked).

However, notice that we simulated an initial snapshot containing only Alice (she has age 25):
```ts
const sampleUsers: Array<User> = [
  {
    id: 1,
    name: `Alice`,
    age: 25,
    email: `alice@example.com`,
    active: true,
  },
  {
    id: 2,
    name: `Bob`,
    age: 19,
    email: `bob@example.com`,
    active: true,
  },
  {
    id: 3,
    name: `Charlie`,
    age: 30,
    email: `charlie@example.com`,
    active: false,
  },
  {
    id: 4,
    name: `Dave`,
    age: 22,
    email: `dave@example.com`,
    active: true,
  },
]
```
So, since we did a full initial sync, we believe we saw everything up to Alice (age 25) and so the 2nd snapshot request includes where age > 25.

So the logic of this test is a bit flawed. In practice, we should either have synced the user with the lowest age during the initial sync and then the snapshot request should return the next 2 users. Or, we can still sync only Alice during initial sync but then we need to mimick live updates coming in for any users with age smaller than Alice's age.

packages/electric-db-collection/src/electric.ts
```diff
@@ -73,6 +74,24 @@ type InferSchemaOutput<T> = T extends StandardSchemaV1
...
+/**
+ * The mode of sync to use for the collection.
+ * @default `eager`
+ * @description
+ * - `eager`:
+ *   - syncs all data immediately on preload
+ *   - collection will be marked as ready once the sync is complete
+ *   - there is no incremental sync
+ * - `on-demand`:
+ *   - syncs data synced in incremental snapshots as the collection is queried
```
kevin-dp, line 25: "syncs data synced" --> need to remove "syncs" or "synced"

packages/electric-db-collection/src/electric.ts
```diff
@@ -73,6 +74,24 @@ type InferSchemaOutput<T> = T extends StandardSchemaV1
...
+ *   - collection will be marked as ready immediately after the first snapshot is synced
+ * - `progressive`:
+ *   - syncs all data in the shape in the background
```
kevin-dp, line 35: This is the only place in this docstring where we mention shapes and it is a bit confusing that we mention shape here and not in the other modes. I would remove it and say "syncs all data in the background"

packages/electric-db-collection/src/sql-compiler.ts
samwillis: Options:
1. not push down orderBy when it's using not postgres compatible ordering
2. try and convert to something postgres compatible where possible
3. throw and error or warning that its using the wrong ordering

or some combination of above

packages/electric-db-collection/src/electric.ts
```diff
@@ -73,6 +74,24 @@ type InferSchemaOutput<T> = T extends StandardSchemaV1
...
+ *   - uses incremental snapshots during the initial sync to provide a fast path to the data required for queries
+ *   - collection will be marked as ready once the initial sync is complete
```
kevin-dp, line 37: I'm not sure what is meant by "initial sync" here. The previous bullet says:
> uses incremental snapshots during the initial sync to provide **a fast path to the data required for queries**

The part in bolt makes me think that my queries will use the data immediately? Since `on-demand` marks as ready after the first snapshot is synced, i'm expecting this to mark it as ready immediately (even before first snapshit sync)? But then this sentence now says:
> collection will be marked as ready once the initial sync is complete

What does this mean? What is the initial sync in this mode? Is it the first snapshot sync, all snapshot syncs, ...?

packages/electric-db-collection/src/electric.ts
```diff
@@ -812,11 +850,19 @@ function createElectricSync<T extends Row<unknown>>(
         }
       })
 
+      // Only set onLoadMore if the sync mode is not eager, this indicates to the sync
+      // layer can load more data on demand via the requestSnapshot method when,
```
kevin-dp, line 137: nit: missing some words between "layer" and "can". Perhaps:
> this indicates to the sync layer **that it** can load more data on demand ...

packages/electric-db-collection/tests/electric-live-query.test.ts
```diff
@@ -468,14 +468,45 @@ describe.each([
...
+      vi.clearAllMocks()
+
+      const testSubscriber = vi.fn<(messages: Array<Message<User>>) => void>()
+      mockSubscribe.mockImplementation((callback) => {
+        testSubscriber.mockImplementation(callback)
+        return () => {}
+      })
+
+      const testElectricCollection = createCollection({
+        ...electricCollectionOptions({
```
kevin-dp: Why do we need to spread it here?
Can't we just do `createCollection(electricCollectionOptions({ ... }))`?

packages/electric-db-collection/tests/electric-live-query.test.ts
```diff
@@ -468,14 +468,45 @@ describe.each([
...
       mockRequestSnapshot.mockResolvedValue({
         data: [],
       })
 
       // Initial sync with limited data
-      simulateInitialSync()
-      expect(electricCollection.status).toBe(`ready`)
-      expect(electricCollection.size).toBe(4)
+      testSubscriber([
+        ...sampleUsers.map((user) => ({
```
kevin-dp, line 501: Same, spreading is unnecessary here

packages/electric-db-collection/tests/electric-live-query.test.ts
```diff
@@ -468,14 +468,45 @@ describe.each([
   })
   if (autoIndex === `eager`) {
     it(`should load more data via requestSnapshot when creating live query with higher limit`, async () => {
+      // Create a new electric collection with on-demand syncMode for this test
+      vi.clearAllMocks()
+
+      const testSubscriber = vi.fn<(messages: Array<Message<User>>) => void>()
+      mockSubscribe.mockImplementation((callback) => {
```
kevin-dp, line 476: Why do we need to define a `testSubscriber` and wrap it inside the `mockSubscribe`? The wrapping does not do anything extra. I believe we can remove `testSubscriber` in this test and replace all occurences of `testSubscriber` by `mockSubscribe`

packages/electric-db-collection/tests/electric-live-query.test.ts
```diff
@@ -594,3 +625,285 @@ describe.each([
...
+          .from({ user: electricCollection })
+          .orderBy(({ user }) => user.id, `asc`)
+          .limit(3),
+    })
+
+    // Wait for the live query to process
+    await new Promise((resolve) => setTimeout(resolve, 0))
+
+    // Should have requested more data from Electric
+    expect(mockRequestSnapshot).toHaveBeenCalled()
```
kevin-dp: This is a very simple check. Would be better to actually check which snapshots have been requested.

packages/electric-db-collection/tests/electric-live-query.test.ts
samwillis: @kevin-dp and I chatted about this, it is doing what we would expect at this stage as there is no deduplication of snapshot requests yet. Once we implement that there will not be this additional request.

packages/electric-db-collection/src/electric.ts
samwillis: Will change to "full sync"

As of #658 live queries run eagerly while data is loading, so we still don't mark the collection as "ready" till its fully synced, but it can use snapshots during that time to fulfil queries. Users will have to handle the fast the `isReady=false` themselves on the queries. We may yet want to iterate on the status of live queries in this case.

packages/electric-db-collection/tests/electric-live-query.test.ts
samwillis: This spread enables us to add the up-to-date at the end of the message batch.

packages/electric-db-collection/src/electric.ts
```diff
@@ -281,6 +303,8 @@ export function electricCollectionOptions(
 } {
   const seenTxids = new Store<Set<Txid>>(new Set([]))
   const seenSnapshots = new Store<Array<PostgresSnapshot>>([])
+  const syncMode = config.syncMode ?? `eager`
+  const finalSyncMode = syncMode === `progressive` ? `on-demand` : syncMode
```
kevin-dp: nit: would call it `internalSyncMode`

packages/electric-db-collection/src/electric.ts
```diff
@@ -653,6 +681,15 @@ function createElectricSync<T extends Row<unknown>>(
 
       const stream = new ShapeStream({
         ...shapeOptions,
+        // In on-demand mode, we only want to sync changes, so we set the log to `changes_only`
+        log: syncMode === `on-demand` ? `changes_only` : undefined,
+        // In on-demand mode, we only need the changes from the point of time the collection was created
+        // so we default to `now` when there is no saved offset.
+        offset: shapeOptions.offset
```
kevin-dp: The first ternary can be replaced by `??`:
```ts
offset: shapeOptions.offset ?? syncMode === `on-demand` ? `now` : undefined
```

packages/electric-db-collection/src/sql-compiler.ts
kevin-dp: TODO: leave a fixme, we will need to do this soon-ish -> locale per collection

PR #620, opened 2025-10-01
docs: add tanstack-db-pglite

letstri:
## 🎯 Changes

Added community tanstack-db-pglite package.


KyleAMathews (approved):
Thanks for sharing!

PR #621, opened 2025-10-01
Fix a bug that could result in a duplicate delete event for a row

### LINKED ISSUE #609: Sums are updated twice in live queries
Hello 👋 

I'm using Tanstack DB with a custom collection [I created based on tRPC](https://github.com/fuegoio/trpc-db-collection). I use a `sum` in a live query and I noticed that if I have an action on delete (implementing an `onDelete` basically), the sum amount is updated twice.

Here is a CodeSandbox that reproduces the issue: https://codesandbox.io/p/sandbox/dark-worker-8y8342

Basically, when the optimistic update is triggered, the sum is correctly updated. But if there is a possible rollback, ie on `onDelete` or `onUpdate` is implemented, when the optimistic update should be discarded it is not and the sum amount is updated again.

Minimal repro code here:

```tsx
import {
  useLiveQuery,
  sum,
  localOnlyCollectionOptions,
  createCollection,
} from "@tanstack/react-db";

export const numbersCollection = createCollection(
  localOnlyCollectionOptions({
    id: "numbers",
    getKey: (item) => item.id,
    initialData: [
      { id: 0, number: 15 },
      { id: 1, number: 15 },
      { id: 2, number: 15 },
    ],
      const { original, modified } = transaction.mutations[0];
      return modified;
    },
  })
);

  const query = useLiveQuery((q) =>
    q.from({ numbers: numbersCollection }).select(({ numbers }) => ({
      totalNumber: sum(numbers.number),
    }))
  );

  const deleteItem = () => {
    numbersCollection.delete(0);
  };

  return (
    <div className="App">
      <h1>Total: {query.data.at(0).totalNumber}</h1>
      <button onClick={deleteItem}>Delete item</button>
    </div>
  );
}
```

The sum is first 45, but when clicking on the button it goes first to 30 after the optimistic update, and then 15.

Thanks a lot your for time and work!

samwillis:
fixes #609

We had code in place that filtered out redundant change events from sync operations, but it failed to filter delete events, only filtering insert/update.

CI run with test failure showing the repo before the fix is applied: https://github.com/TanStack/db/actions/runs/18163302240/job/51698976831?pr=621

### DIFF
--- a/packages/db/tests/collection-subscribe-changes.test.ts
+++ b/packages/db/tests/collection-subscribe-changes.test.ts
@@ -1428,22 +1428,12 @@ describe(`Collection.subscribeChanges`, () => {
       includeInitialState: true,
     })
 
-    console.log({ calls: callback.mock.calls.length })
-    const changes = callback.mock.calls[0]![0] as ChangesPayload<TestItem>
-
-    console.log(changes)
-
     callback.mockReset()
 
     // Delete item 0
     collection.delete(0)
 
     await new Promise((resolve) => setTimeout(resolve, 10))
-
-    console.log({ calls: callback.mock.calls.length })
-    for (const call of callback.mock.calls) {
-      console.log(call[0]![0])
-    }
 
     expect(callback.mock.calls.length).toBe(1)
     expect(callback.mock.calls[0]![0]).toEqual([
@@ -1490,7 +1480,6 @@ describe(`Collection.subscribeChanges`, () => {
       },
       onDelete: async ({ transaction }) => {
         const { original } = transaction.mutations[0]
-        console.log({ original })
 
         // Simulate waiting for some async operation
         await new Promise((resolve) => setTimeout(resolve, 0))
@@ -1509,22 +1498,12 @@ describe(`Collection.subscribeChanges`, () => {
       includeInitialState: true,
     })
 
-    console.log({ calls: callback.mock.calls.length })
-    const changes = callback.mock.calls[0]![0] as ChangesPayload<TestItem>
-
-    console.log(changes)
-
     callback.mockReset()
 
     // Delete item 0
     collection.delete(0)
 
     await new Promise((resolve) => setTimeout(resolve, 10))
-
-    console.log({ calls: callback.mock.calls.length })
-    for (const call of callback.mock.calls) {
-      console.log(call[0]![0])
-    }
 
     expect(callback.mock.calls.length).toBe(1)
     expect(callback.mock.calls[0]![0]).toEqual([


packages/db/tests/collection-subscribe-changes.test.ts
```diff
@@ -1379,4 +1379,162 @@ describe(`Collection.subscribeChanges`, () => {
...
+        callCommit()
+      },
+    })
+
+    // Subscribe to changes
+    const subscription = collection.subscribeChanges(callback, {
+      includeInitialState: true,
+    })
+
+    console.log({ calls: callback.mock.calls.length })
```
KyleAMathews: console.log

packages/db/src/collection/state.ts
```diff
@@ -708,10 +708,23 @@ export class CollectionStateManager<
...
-          newVisibleValue !== undefined &&
-          deepEquals(completedOp.value, newVisibleValue)
+        let isRedundantSync = false
+
+        if (completedOp) {
+          if (
+            completedOp.type === `delete` &&
+            previousVisibleValue !== undefined &&
+            newVisibleValue === undefined &&
+            deepEquals(completedOp.value, previousVisibleValue)
```
KyleAMathews, line 718: I don't love really that we do all these deepEquals — they're fairly expensive — but I suppose cheaper than the UI framework re-rendering the same bits of UI 🤷

KyleAMathews: In this particular case, we don't actually need a deepEqual right? Since a delete is a delete is a delete. If the optimistic mutation got rid of it already, we can just immediately assume the synced delete is the same?

samwillis: The deepEquals are not really about stoping re-renders, they are deduplicating events in the same batch that could result in an invalid state. This is particularly important when feeding into the d2 graph as aggregates could become wrong, or the multiplicity could and we then track inserts/deletes later incorrectly.

I agree, I would prefer not to have them, but I think we should approach that in a general redesign of this reconciliation process - I don't particularly like how it currently works.

> In this particular case, we don't actually need a deepEqual right?
Maybe, but I don't really want to bet on it. It's still about removing duplicates and there could be multiple deletes for the same key in the same batch. We just want to ensure the state inside d2 is correct. I don't want to make an assumption, I would prefer to take a belt and braces at this point before we further refactor.

Do note that these `deepEquals` are only happening on optimistic mutations, not on every change that is synced, and not on every emitted initial state into the live query engine. As overhead it's tiny compared to the other things happening in D2.

KyleAMathews: > Do note that these deepEquals are only happening on optimistic mutations, not on every change that is synced

Ah ok, so very little generally 👍

KyleAMathews (approved):
:shipit:

Issue #616, opened 2025-10-01
Queries on date has a mutating impact on other queries


fuegoio:
Hello again 👋 

I have a bug where I use an `orderBy` and a `limit` to get most recent items by a date. But weirdingly, when there is another query before this one that just filters on this date, the result is different and wrong.

Codesandbox that reproduces the issue: https://codesandbox.io/p/devbox/quirky-rui-mh7ypn

Minimal reproduction code:

```tsx
export const numbersCollection = createCollection(
  localOnlyCollectionOptions({
    id: "dates",
    getKey: (item) => item.id,
    initialData: [
      {
        id: "1",
        date: new Date("2025-09-15"),
      },
      {
        id: "2",
        date: new Date("2025-09-10"),
      },
    ],
  })
);

export default function App() {
  const today = new Date();
  // Comment this query and the result will be different
  const firstQueryThatPoisonTheRest = useLiveQuery((q) => {
    return q
      .from({ numbers: numbersCollection })
      .where(({ numbers }) =>
        and(gt(numbers.date, today), lt(numbers.date, today))
      );
  });

  const query = useLiveQuery((q) =>
    q
      .from({ numbers: numbersCollection })
      .orderBy(({ numbers }) => numbers.date, "desc")
      .limit(1)
  );

  return (
    <div className="App">
      <h1>Id: {query.data.at(0).id}</h1>
    </div>
  );
}
```

When the `firstQueryThatPoisonTheRest` is commented, the result is `Id: 1` which is what we want ("2025-09-15" > "2025-09-10"). But when this query runs first, the result is `Id: 2`, which is wrong.

Happy to help on this if needed, and thanks again for your work!

PR #623, opened 2025-10-01
asc/desc index bug

samwillis:
repo of #616

### DIFF
--- a/packages/db/tests/query/order-by.test.ts
+++ b/packages/db/tests/query/order-by.test.ts
@@ -1186,6 +1186,190 @@ function createOrderByTests(autoIndex: `off` | `eager`): void {
         const results = Array.from(collection.values())
         expect(results).toHaveLength(0)
       })
+
+      it(`can use use orderBy in both ascending and descending order on the same column`, async () => {
+        type DateItem = {
+          id: string
+          date: Date
+        }
+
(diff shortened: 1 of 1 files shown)


kevin-dp:
The bug occurs because we store an index for a given column but we don't track the comparison options (direction, nulls, stringSort) that were used to construct the index. Hence, when we do a first query we may create an index on a given column (say in ascending order) and then later another query may order by the same column but in descending order but it will re-use the same index because the index lookup function only checks the column name and not the comparison options. I solved this bug by adding the comparison options to the index' state and making sure that `ensureIndexForField` and `findIndexForField` also check the comparison options.

I added some additional unit tests and modified the unit tests for stringSort and nulls such that they also break (because the wrong index is used). With the fix, these tests now pass. However, the tests showed that there is another bug when the indexed column contains null values and the query uses orderBy + limit + eager indexing mode. Under those circumstances the returned results become empty.

With the current fix, we would create 2 indexes: one for ASC order and one for DESC order. That is suboptimal. As a follow up we should re-use the existing index and walk it in reverse order. We can write a `reverseIndex(index)` function that takes an index and returns a new index that wraps the old one but traverses it in reverse.

samwillis:
This is all good and ready to merge. I can't approve as I originally opened the PR with the tests... **"I approve"**

PR #625, opened 2025-10-01
change to a subscription per collection alias rather than collection inside a live query

samwillis:
This fixes this bug report from Discord: https://discord.com/channels/719702312431386674/1369767025723052134/1422935844175614072

That report identified that a self query with a where clause on the main alias would result in the join not seeing the whole collection. It also tries to tidy up the terminology we use, dropping "table" and using source & collection.

This has grown in to a larger refactor of the compiler that changes it to do a subscription per source alias rather than a single subscription per collection.

---

# Per-Alias Subscriptions: Fix Self-Joins and Enable Independent Filtering

## The Problem

Live queries previously subscribed **once per collection ID**. This caused critical bugs when the same collection appeared multiple times in a query with different aliases:

```typescript
// This query would FAIL before this PR
const query = db.query((q) => 
  q.from({ employee: employeesCollection })
   .join({ manager: employeesCollection }, 
     ({ employee, manager }) => eq(employee.managerId, manager.id))
   .where(({ employee }) => eq(employee.department, 'Engineering'))
)
```

**What went wrong:**
- Both `employee` and `manager` aliases mapped to the same collection
- The live query layer created only ONE subscription for `employeesCollection`
- The WHERE filter meant for `employee` was incorrectly applied to `manager` too
- Lazy loading couldn't distinguish between the two aliases
- Result: managers were filtered by department when they shouldn't be

## The Solution

**Subscribe once per source alias, not once per collection.**

Now `employee` and `manager` each get their own independent subscription with:
- Their own filters
- Their own lazy loading state  
- Their own input streams into the query pipeline

### 1. **Compiler Output Tracking**

Added two new fields to `CompilationResult`:

```typescript
interface CompilationResult {
  // NEW: Maps every alias to its collection ID
  // Example: { employee: 'employees-id', manager: 'employees-id' }
  aliasToCollectionId: Record<string, string>
  
  // NEW: Maps outer subquery aliases to inner aliases
  // Example: { activeUser: 'user' } when join alias differs from subquery's internal alias
  aliasRemapping: Record<string, string>
}
```

### 2. **Alias-Keyed Everything**

**Before:**
```typescript
inputs[collection.id]  // One input per collection
subscriptions[collection.id]  // One subscription per collection
```

**After:**
```typescript
inputs[alias]  // One input per alias: { employee: input1, manager: input2 }
subscriptions[alias]  // One subscription per alias
```

### 3. **Subquery Alias Resolution**

When a subquery uses different aliases than the parent query:
```typescript
const activeUsers = q.from({ user: usersCollection })
                     .where(({ user }) => eq(user.active, true))

q.from({ post: posts })
 .join({ author: activeUsers }, ...)  // Outer: 'author', Inner: 'user'
```

The compiler now tracks: `aliasRemapping['author'] = 'user'` so lazy loading can find the correct subscription.

### 4. **Better Error Messages**

```typescript
// Before
throw new Error(`Input for collection "users" not found`)

// After  
throw new CollectionInputNotFoundError(
  'manager',  // The missing alias
  'employees-col-id',  // The collection it refers to
  ['employee', 'post']  // Available aliases for debugging
)
// Error: Input for alias "manager" (collection "employees-col-id") not found in inputs map. 
//        Available keys: employee, post
```

### 5. **Terminology Consistency**

- `collectionWhereClauses` → `sourceWhereClauses`
- "table alias" → "source alias" throughout comments
- `tables` → `sources` in pipeline code

### 6. **Code Simplification**

Removed unnecessary two-phase compilation that was guarding against optimizer-generated aliases (which never actually happened). The compiler now runs once with cleaner logic.

### ✅ Self-Joins
```typescript
// Employees who manage other employees
q.from({ employee: employeesCollection })
 .join({ manager: employeesCollection }, 
   ({ employee, manager }) => eq(employee.managerId, manager.id))
 .where(({ employee }) => eq(employee.department, 'Engineering'))
 .select(({ employee, manager }) => ({
   employeeName: employee.name,
   managerName: manager.name  // ✅ Manager not filtered by department
 }))
```

### ✅ Multiple Aliases with Independent Filters
```typescript
q.from({ current: usersCollection })
 .join({ previous: usersCollection }, ...)
 .where(({ current }) => eq(current.active, true))
 .where(({ previous }) => eq(previous.active, false))
 // ✅ Each alias has its own filter
```

### ✅ Subquery Alias Resolution
```typescript
const admins = q.from({ user: usersCollection })
                .where(({ user }) => eq(user.role, 'admin'))

q.from({ post: posts })
 .join({ author: admins }, ...)  
 // ✅ Compiler maps 'author' → 'user' for subscription resolution
```

### ✅ Lazy Loading in Self-Joins
```typescript
// Only loads managers for employees that pass the WHERE clause
q.from({ employee: employeesCollection })
 .join({ manager: employeesCollection }, ...)
 .where(({ employee }) => eq(employee.active, true))
 // ✅ Lazy loads only needed managers, using correct alias subscription
```

## Breaking Changes

**None for users.** This is entirely internal to the query compilation and live layer.

The only "breaking" aspect is that previously broken queries (self-joins) now work correctly.

## Performance Impact

**Neutral to positive:**
- Avoids incorrect filter application across aliases
- Enables more precise lazy loading (only load what each alias needs)
- Slightly more memory (one subscription per alias vs per collection)
- Opens door for future optimization: deduplicate identical alias pipelines

## Testing

All existing tests pass. The fact that tests passed throughout development confirms:
- No optimizer actually generates new aliases (validated with logging)
- All aliases come from user declarations
- The defensive checks never trigger in practice

Self-join scenarios that previously failed now work correctly.

## Files Changed

- `packages/db/src/query/compiler/index.ts` - Core compilation changes
- `packages/db/src/query/compiler/joins.ts` - Join processing with alias tracking
- `packages/db/src/query/live/collection-config-builder.ts` - Per-alias subscription creation
- `packages/db/src/query/live/collection-subscriber.ts` - Alias-aware subscriptions
- `packages/db/src/errors.ts` - New error classes with better context

## Migration Guide

**No migration needed.** Existing queries work unchanged. Self-joins that were broken now work.

### DIFF
--- a/packages/db/src/query/live/collection-config-builder.ts
+++ b/packages/db/src/query/live/collection-config-builder.ts
@@ -517,8 +517,9 @@ export class CollectionConfigBuilder<
       return loadMore
     })
 
-    // Combine all loaders into a single callback that attempts to load more data
-    // from any source that needs it (returns true when done loading)
+    // Combine all loaders into a single callback that initiates loading more data
+    // from any source that needs it. Returns true once all loaders have been called,
+    // but the actual async loading may still be in progress.
     const loadMoreDataCallback = () => {
       loaders.map((loader) => loader())
       return true


packages/db/src/query/compiler/index.ts
```diff
@@ -78,12 +86,24 @@ export function compileQuery(
...
-  const tables: Record<string, KeyedStream> = {}
+  // Track alias to collection id relationships discovered during compilation so
+  // the live layer can subscribe to every alias the optimizer introduces.
+  const aliasToCollectionId: Record<string, string> = {}
 
-  // Process the FROM clause to get the main table
+  // Track alias remapping for subqueries (outer alias → inner alias)
+  // e.g., when .join({ activeUser: subquery }) where subquery uses .from({ user: collection })
+  // we store: aliasRemapping['activeUser'] = 'user'
+  const aliasRemapping: Record<string, string> = {}
```
kevin-dp, line 116: Not sure how this works. What if the subquery is also a query with a join?
Does that matter for this alias?

packages/db/src/query/compiler/index.ts
```diff
@@ -31,32 +31,51 @@ import type {
...
-  collectionWhereClauses: Map<string, BasicExpression<boolean>>
+
+  /** Map of source aliases to their WHERE clauses for index optimization */
+  sourceWhereClauses: Map<string, BasicExpression<boolean>>
+
+  /**
+   * Maps each source alias to its collection ID. Enables per-alias subscriptions for self-joins.
+   * Example: `{ employee: 'employees-col-id', manager: 'employees-col-id' }`
+   */
+  aliasToCollectionId: Record<string, string>
```
samwillis, line 50: This is a mapping of the alias used in a `from` or `join` to the collection id that is to be used for that source.

packages/db/src/query/compiler/index.ts
```diff
@@ -31,32 +31,51 @@ import type {
...
+
+  /**
+   * Maps outer alias to inner alias for subqueries (e.g., `{ activeUser: 'user' }`).
+   * Used to resolve subscriptions during lazy loading when aliases differ.
+   */
+  aliasRemapping: Record<string, string>
```
samwillis, line 66: In this query below the alias mapping is used to track that `activeUser` is an alias of the `user` alias.
```
q.from(issue: issueCollection)
  .join(
    { activeUser: q.from({ user: userCollection }) },
    ({ issue }) => ....
```

kevin-dp: What about shadowing? Imagine a subquery that defines an alias that is also defined in the outer query:
```ts
q.from(issue: issueCollection)
  .join(
    { user: q.from({ user: userCollection }) }, // <--
    ({ issue }) => ...
```

packages/db/src/query/live/collection-config-builder.ts
```diff
@@ -400,44 +461,71 @@ export class CollectionConfigBuilder<
...
+      const loadMore = collectionSubscriber.loadMoreIfNeeded.bind(
+        collectionSubscriber,
+        subscription
+      )
+
+      return loadMore
+    })
+
+    // Combine all loaders into a single callback that attempts to load more data
+    // from any source that needs it (returns true when done loading)
```
kevin-dp: > returns true when done loading
That's slightly inacurate. It returns true once all loaders have been initiated but the loading itself could be async, we don't await that.

packages/db/tests/query/compiler/subquery-caching.test.ts
```diff
@@ -5,12 +5,24 @@ import { CollectionRef, PropRef, QueryRef } from "../../../src/query/ir.js"
 import type { QueryIR } from "../../../src/query/ir.js"
 import type { CollectionImpl } from "../../../src/collection/index.js"
 
+// Helper to create a minimal mock collection for compiler tests
+function createMockCollection(id: string): CollectionImpl {
+  return {
+    id,
+    autoIndex: `off`,
+    config: {
+      autoIndex: `off`,
```
kevin-dp, line 14: How come we need `autoIndex` in the top level object and also in the `config` property?

packages/db/src/query/compiler/index.ts
samwillis: It always follows to the main path following the `from`.

packages/db/tests/query/compiler/subquery-caching.test.ts
samwillis: This is a mock of a collection object, and autoIndex is exposed as a prop on the top level as well as in the config prop.

PR #627, opened 2025-10-02
Reverse index

kevin-dp:
https://github.com/TanStack/db/pull/623 fixes a bug with the indexes. As a result, we create separate indexes for ASC and DESC queries. This is suboptimal. This PR introduces a ReverseIndex such that if an index exist but it is in the wrong direction we reverse it. As a result, there's only one underlying index.


samwillis (approved):
:shipit:

PR #628, opened 2025-10-02
Add a scheduler that ensures single batch of changes from live query due to a transaction that touches multiple source collections

samwillis:
stacked on #625

# Live Query Scheduler Overview

## Overview
This change introduces a scoped, dependency-aware scheduler that guarantees every live-query builder runs at most once per transaction, even when multiple source collections or derived queries fire in the same `mutate` call. The scheduler groups work by transaction id, dedupes entries by the `CollectionConfigBuilder` instance, tracks dependencies between builders, and flushes synchronously when `Transaction.mutate` exits. The net effect: optimistic updates from a single transaction coalesce into a single graph run for each live query, so downstream frameworks see one change batch per transaction.

## How scheduling works
- Each `CollectionSubscriber` immediately forwards changes to the D2 input and calls `CollectionConfigBuilder.scheduleGraphRun`.
- `scheduleGraphRun` records which upstream builders the current builder depends on (based on the collections it subscribed to) and hands a job to the shared scheduler with:
  - `contextId`: the transaction id, grouping all work triggered by that transaction.
  - `jobId`: the builder instance; repeated schedules for the same builder merge into a single entry.
  - `dependencies`: the set of upstream builders that must finish before this builder can run.
- The scheduler maintains, per transaction, a queue, entry map, dependency map, and “completed” set. When flushing, it only executes a job once all dependencies are marked completed, deferring the job otherwise. It loops until no work remains and throws if it detects an impossible cycle.

## When the graph actually runs
- Inside `Transaction.mutate`, we call `registerTransaction` before the user’s callback and `unregisterTransaction` in a `finally` block afterward.
- `registerTransaction` clears any stale entries from previous failed scopes.
- `unregisterTransaction` calls `scheduler.flush(tx.id)`. `flush` now loops while a context map exists, so if a job enqueues additional work during its own execution (e.g., the join scheduling itself after its parents run), the scheduler immediately picks up the new entry before leaving the transaction scope.
- If there is no active transaction, `scheduleGraphRun` runs `maybeRunGraph` immediately (backward compatible behaviour).

## Nested live queries (the “diamond” cases)
- Builders register themselves when a live-query collection is created. Whenever a builder subscribes to another live query, it records that dependency.
- During a transaction, updates enqueue jobs for the parent builders. When those builders finish, they mark themselves as completed, which in turn allows child jobs (such as joins) to run exactly once at the end of the flush.
- A hybrid variant (join between `liveQueryA` and raw `collectionB`) behaves the same way: even if `collectionB` fires first, the join job is deferred until `liveQueryA` has finished.
- Because dependencies are tracked explicitly, there are no deadlocks. If the scheduler ever loops without making progress, it throws a clear “dependency cycle” error.

### Order-by loaders and batching
- The `maybeRunGraph` loop keeps calling `graph.run()` while `pendingWork()` is true. If an order-by loader requests more rows, it pushes those changes into the same D2 input, triggering `pendingWork()` again.
- During this loop `CollectionConfigBuilder.isGraphRunning` is `true`, so any nested call to `scheduleGraphRun` is ignored; the loader’s extra rows are consumed by the ongoing loop.
- As a result, even hungry top-K queries emit a single batch per synchronous transaction run. Multiple batches only occur when a loader deliberately yields results asynchronously (for example, the incremental sync features we’re adding)—in that case the UI should expect staged updates.

## Tests
`scheduler.test.ts` now covers:

1. Basic single run per transaction (two collections mutated inside one `mutate`).
2. Nested transactions.
3. Rollback cleanup.
4. Multiple subscribers receiving exactly one batch.
5. Loader deduping.
6. Diamond dependency (live query joining two parents) – verifies parents run first and the join runs once (tracked via `getRunCount`).
7. Hybrid diamond (join of a live query with a base collection) – same single-run guarantee, even when the raw collection fires first.

Each of the diamond tests mutates once and then performs another transaction with updates, demonstrating that we still emit exactly one additional batch—no duplicates, no missed runs—and that `getRunCount()` only increments once per transaction.

## Summary
- Scheduler coalesces work by transaction + builder.
- `Transaction.mutate` clears before and flushes after every scope, so everything runs synchronously in the optimistic phase.
- `flush` loops until no jobs remain, ensuring child live queries (joins) run exactly once per transaction after their parents.
- Comprehensive tests cover simple, nested, rollback, multi-subscriber, loader, diamond, and hybrid patterns; asynchronous loaders will still produce multiple batches by design.

### DIFF
--- a/.changeset/olive-crews-love.md
+++ b/.changeset/olive-crews-love.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+Add a scheduler that ensures that if a transaction touches multiple collections that feed into a single live query, the live query only emits a single batch of updates. This fixes an issue where multiple renders could be triggered from a live query under this situation.
--- a/packages/db/src/query/live-query-collection.ts
+++ b/packages/db/src/query/live-query-collection.ts
@@ -1,5 +1,10 @@
 import { createCollection } from "../collection/index.js"
 import { CollectionConfigBuilder } from "./live/collection-config-builder.js"
+import {
+  getBuilderFromConfig,
+  registerCollectionBuilder,
+} from "./live/collection-registry.js"
+import type { LiveQueryCollectionUtils } from "./live/collection-config-builder.js"
 import type { LiveQueryCollectionConfig } from "./live/types.js"
 import type { InitialQueryBuilder, QueryBuilder } from "./builder/index.js"
 import type { Collection } from "../collection/index.js"
@@ -55,7 +60,9 @@ export function liveQueryCollectionOptions<
   TResult extends object = GetResult<TContext>,
 >(
   config: LiveQueryCollectionConfig<TContext, TResult>
-): CollectionConfigForContext<TContext, TResult> {
+): CollectionConfigForContext<TContext, TResult> & {
+  utils: LiveQueryCollectionUtils
+} {
   const collectionConfigBuilder = new CollectionConfigBuilder<
     TContext,
     TResult
@@ -63,7 +70,7 @@ export function liveQueryCollectionOptions<
   return collectionConfigBuilder.getConfig() as CollectionConfigForContext<
     TContext,
     TResult
-  >
+  > & { utils: LiveQueryCollectionUtils }
 }
 
 /**
@@ -106,7 +113,9 @@ export function createLiveQueryCollection<
   TResult extends object = GetResult<TContext>,
 >(
   query: (q: InitialQueryBuilder) => QueryBuilder<TContext>
-): CollectionForContext<TContext, TResult>
+): CollectionForContext<TContext, TResult> & {
+  utils: LiveQueryCollectionUtils
+}
 
 // Overload 2: Accept full config object with optional utilities
 export function createLiveQueryCollection<
@@ -115,7 +124,9 @@ export function createLiveQueryCollection<
   TUtils extends UtilsRecord = {},
 >(
   config: LiveQueryCollectionConfig<TContext, TResult> & { utils?: TUtils }
-): CollectionForContext<TContext, TResult>
+): CollectionForContext<TContext, TResult> & {
+  utils: LiveQueryCollectionUtils & TUtils
+}
 
 // Implementation
 export function createLiveQueryCollection<
@@ -126,7 +137,9 @@ export function createLiveQueryCollection<
   configOrQuery:
     | (LiveQueryCollectionConfig<TContext, TResult> & { utils?: TUtils })
     | ((q: InitialQueryBuilder) => QueryBuilder<TContext>)
-): CollectionForContext<TContext, TResult> {
+): CollectionForContext<TContext, TResult> & {
+  utils: LiveQueryCollectionUtils & TUtils
+} {
   // Determine if the argument is a function (query) or a config object
   if (typeof configOrQuery === `function`) {
     // Simple query function case
@@ -139,7 +152,7 @@ export function createLiveQueryCollection<
     return bridgeToCreateCollection(options) as CollectionForContext<
       TContext,
       TResult
-    >
+    > & { utils: LiveQueryCollectionUtils & TUtils }
   } else {
     // Config object case
     const config = configOrQuery as LiveQueryCollectionConfig<
@@ -147,10 +160,16 @@ export function createLiveQueryCollection<
       TResult
     > & { utils?: TUtils }
     const options = liveQueryCollectionOptions<TContext, TResult>(config)
-    return bridgeToCreateCollection({
-      ...options,
-      utils: config.utils,
-    }) as CollectionForContext<TContext, TResult>
+
+    // Merge custom utils if provided, preserving the getBuilder() method for dependency tracking
+    if (config.utils) {
+      options.utils = { ...options.utils, ...config.utils }
+    }
+
+    return bridgeToCreateCollection(options) as CollectionForContext<
+      TContext,
+      TResult
+    > & { utils: LiveQueryCollectionUtils & TUtils }
   }
 }
 
@@ -162,12 +181,18 @@ function bridgeToCreateCollection<
   TResult extends object,
   TUtils extends UtilsRecord = {},
 >(
-  options: CollectionConfig<TResult> & { utils?: TUtils }
+  options: CollectionConfig<TResult> & { utils: TUtils }
 ): Collection<TResult, string | number, TUtils> {
-  // This is the only place we need a type assertion, hidden from user API
-  return createCollection(options as any) as unknown as Collection<
+  const collection = createCollection(options as any) as unknown as Collection<
     TResult,
     string | number,
-    TUtils
+    LiveQueryCollectionUtils
   >
+
+  const builder = getBuilderFromConfig(options)
+  if (builder) {
+    registerCollectionBuilder(collection, builder)
+  }
+
+  return collection as unknown as Collection<TResult, string | number, TUtils>
 }
--- a/packages/db/src/query/live/collection-config-builder.ts
+++ b/packages/db/src/query/live/collection-config-builder.ts
@@ -2,7 +2,11 @@ import { D2, output } from "@tanstack/db-ivm"
 import { compileQuery } from "../compiler/index.js"
 import { buildQuery, getQueryIR } from "../builder/index.js"
 import { MissingAliasInputsError } from "../../errors.js"
+import { transactionScopedScheduler } from "../../scheduler.js"
+import { getActiveTransaction } from "../../transactions.js"
 import { CollectionSubscriber } from "./collection-subscriber.js"
+import { getCollectionBuilder } from "./collection-registry.js"
+import type { SchedulerContextId } from "../../scheduler.js"
 import type { CollectionSubscription } from "../../collection/subscription.js"
 import type { RootStreamBuilder } from "@tanstack/db-ivm"
 import type { OrderByOptimizationInfo } from "../compiler/order-by.js"
@@ -12,6 +16,7 @@ import type {
   KeyedStream,
   ResultStream,
   SyncConfig,
+  UtilsRecord,
 } from "../../types.js"
 import type { Context, GetResult } from "../builder/types.js"
 import type { BasicExpression, QueryIR } from "../ir.js"
@@ -23,6 +28,15 @@ import type {
   SyncState,
 } from "./types.js"
 import type { AllCollectionEvents } from "../../collection/events.js"
+
+export type LiveQueryCollectionUtils = UtilsRecord & {
+  getRunCount: () => number
+  getBuilder: () => CollectionConfigBuilder<any, any>
+}
+
+type PendingGraphRun = {
+  loadCallbacks: Set<() => boolean>
+}
 
 // Global counter for auto-generated collection IDs
 let liveQueryCollectionCounter = 0
@@ -52,12 +66,42 @@ export class CollectionConfigBuilder<
   private readonly compare?: (val1: TResult, val2: TResult) => number
 
   private isGraphRunning = false
+  private runCount = 0
+
+  // Current sync session state (set when sync starts, cleared when it stops)
+  // Public for testing purposes (CollectionConfigBuilder is internal, not public API)
+  public currentSyncConfig:
+    | Parameters<SyncConfig<TResult>[`sync`]>[0]
+    | undefined
+  public currentSyncState: FullSyncState | undefined
 
   // Error state tracking
   private isInErrorState = false
 
   // Reference to the live query collection for error state transitions
   private liveQueryCollection?: Collection<TResult, any, any>
+
+  private readonly aliasDependencies: Record<
+    string,
+    Array<CollectionConfigBuilder<any, any>>
+  > = {}
+
+  private readonly builderDependencies = new Set<
+    CollectionConfigBuilder<any, any>
+  >()
+
+  // Pending graph runs per scheduler context (e.g., per transaction)
+  // The builder manages its own state; the scheduler just orchestrates execution order
+  // Only stores callbacks - if sync ends, pending jobs gracefully no-op
+  private readonly pendingGraphRuns = new Map<
+    SchedulerContextId,
+    PendingGraphRun
+  >()
+
+  // Unsubscribe function for scheduler's onClear listener
+  // Registered when sync starts, unregistered when sync stops
+  // Prevents memory leaks by releasing the scheduler's reference to this builder
+  private unsubscribeFromSchedulerClears?: () => void
 
   private graphCache: D2 | undefined
   private inputsCache: Record<string, RootStreamBuilder<unknown>> | undefined
@@ -107,7 +151,9 @@ export class CollectionConfigBuilder<
     this.compileBasePipeline()
   }
 
-  getConfig(): CollectionConfigSingleRowOption<TResult> {
+  getConfig(): CollectionConfigSingleRowOption<TResult> & {
+    utils: LiveQueryCollectionUtils
+  } {
     return {
       id: this.id,
       getKey:
@@ -122,6 +168,10 @@ export class CollectionConfigBuilder<
       onDelete: this.config.onDelete,
       startSync: this.config.startSync,
       singleResult: this.query.singleResult,
+      utils: {
+        getRunCount: this.getRunCount.bind(this),
+        getBuilder: () => this,
+      },
     }
   }
 
@@ -159,12 +209,8 @@ export class CollectionConfigBuilder<
   // That can happen because even though we load N rows, the pipeline might filter some of these rows out
   // causing the orderBy operator to receive less than N rows or even no rows at all.
   // So this callback would notice that it doesn't have enough rows and load some more.
-  // The callback returns a boolean, when it's true it's done loading data.
-  maybeRunGraph(
-    config: SyncMethods<TResult>,
-    syncState: FullSyncState,
-    callback?: () => boolean
-  ) {
+  // The callback returns a boolean, when it's true it's done loading data and we can mark the collection as ready.
+  maybeRunGraph(callback?: () => boolean) {
     if (this.isGraphRunning) {
       // no nested runs of the graph
       // which is possible if the `callback`
@@ -172,10 +218,18 @@ export class CollectionConfigBuilder<
       return
     }
 
+    // Should only be called when sync is active
+    if (!this.currentSyncConfig || !this.currentSyncState) {
+      throw new Error(
+        `maybeRunGraph called without active sync session. This should not happen.`
+      )
+    }
+
     this.isGraphRunning = true
 
     try {
-      const { begin, commit } = config
+      const { begin, commit } = this.currentSyncConfig
+      const syncState = this.currentSyncState
 
       // Don't run if the live query is in an error state
       if (this.isInErrorState) {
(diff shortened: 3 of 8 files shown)


packages/db/src/query/live/collection-config-builder.ts
```diff
@@ -22,6 +30,16 @@ import type {
   SyncState,
 } from "./types.js"
 
+export type RunCountUtils = UtilsRecord & {
+  getRunCount: () => number
```
samwillis, line 33: I have added a `utils.getRunCount()` in order to check this in the tests easily. We should extend this with more run metadata and timing, will be useful for dev tools.

.changeset/olive-crews-love.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+Add a scheduler that ensures that if a transaction touches multiple collections that feed into a single live query, the liver query only emits a single batch of updates. This fixes an issue where multiple renders could be triggered from a live query under this situation.
```
kevin-dp: typo: the live**r** query

packages/db/src/query/live-query-collection.ts
```diff
@@ -113,35 +118,50 @@ export function createLiveQueryCollection<
...
     const options = liveQueryCollectionOptions<TContext, TResult>(config)
-    return bridgeToCreateCollection({
-      ...options,
-      utils: config.utils,
-    })
+
+    const collection = bridgeToCreateCollection(options)
+
+    if (config.utils) {
+      Object.assign(collection.utils, config.utils)
```
kevin-dp: This is weird. Why do we need to mutate it after the facts instead of passing it as an option when calling `bridgeToCreateCollection` as we used to do before.

packages/db/src/query/live-query-collection.ts
```diff
@@ -113,35 +118,50 @@ export function createLiveQueryCollection<
...
  */
-function bridgeToCreateCollection<
-  TResult extends object,
-  TUtils extends UtilsRecord = {},
->(
-  options: CollectionConfig<TResult> & { utils?: TUtils }
-): Collection<TResult, string | number, TUtils> {
-  // This is the only place we need a type assertion, hidden from user API
-  return createCollection(options as any) as unknown as Collection<
+function bridgeToCreateCollection<TResult extends object>(
```
kevin-dp: Why removing the `TUtils` type parameter? Because of that the return type is now less precise and we have to explicitly cast it on L140.

packages/db/src/query/live/collection-config-builder.ts
```diff
@@ -54,6 +73,15 @@ export class CollectionConfigBuilder<
     | Map<string, BasicExpression<boolean>>
     | undefined
 
+  private readonly aliasDependencies: Record<
+    string,
+    Array<CollectionConfigBuilder<any, any>>
+  > = Object.create(null)
```
kevin-dp: Why do we need `Object.create(null)` ? Can't we just use an object literal `{}`?

packages/db/src/query/live/collection-config-builder.ts
```diff
@@ -176,13 +209,149 @@ export class CollectionConfigBuilder<
...
+
+                  for (const loader of state.loadCallbacks) {
+                    try {
+                      const result = loader()
+                      if (result === false) {
+                        allDone = false
+                      }
+                    } catch (error) {
+                      allDone = false
+                      if (firstError === undefined) {
```
kevin-dp: Can simplify to `firstError ??= error`

packages/db/src/query/live/collection-config-builder.ts
```diff
@@ -412,6 +581,14 @@ export class CollectionConfigBuilder<
       const collection =
         this.collectionByAlias[alias] ?? this.collections[collectionId]!
 
+      const dependencyBuilder = getCollectionBuilder(collection)
+      if (dependencyBuilder && dependencyBuilder !== this) {
+        this.aliasDependencies[alias] = [dependencyBuilder]
```
kevin-dp, line 729: If it's an array, should we check if there is already an array for `this.aliasDependencies[alias]` and if so push this builder in it? Because right now, it would overwrite any exisisting aliasDependencies.

packages/db/src/query/live/collection-config-builder.ts
```diff
@@ -176,13 +209,149 @@ export class CollectionConfigBuilder<
...
+        state,
+        run: () => {
+          this.incrementRunCount()
+          const combinedLoader =
+            state.loadCallbacks.size > 0
+              ? () => {
+                  let allDone = true
+                  let firstError: unknown
+
+                  for (const loader of state.loadCallbacks) {
```
kevin-dp: We can simplify this `for` loop:
```ts
for (const loader of state.loadCallbacks) {
  try {
    allDone = loader() && allDone
  } catch (error) {
    allDone = false
    firstError ??= error
  }
}
```

packages/db/src/query/live/collection-config-builder.ts
```diff
@@ -176,13 +209,149 @@ export class CollectionConfigBuilder<
...
+      if (callback) {
+        state.loadCallbacks.add(callback)
+      }
+
+      return {
+        state,
+        run: () => {
+          this.incrementRunCount()
+          const combinedLoader =
+            state.loadCallbacks.size > 0
```
kevin-dp: Why doing a ternary if test here?
We could just create the function that loops over `state.loadCallbacks`, if it's empty it will simply return `allDone` which is `true`.

kevin-dp: I would refactor it this way:
```ts
const combinedLoader = () => {
  let allDone = true
  let firstError: unknown
  state.loadCallbacks.forEach((loader) => {
    try {
      allDone = loader() && allDone
    } catch (error) {
      allDone = false
      firstError ??= error
    }
  })
  if (firstError) {
    throw firstError
  }
  // Returning false signals that callers should schedule another pass.
  return allDone
}
```

packages/db/src/query/live/collection-config-builder.ts
```diff
@@ -176,13 +209,149 @@ export class CollectionConfigBuilder<
...
+      jobId?: unknown
+      alias?: string
+      dependencies?: Array<CollectionConfigBuilder<any, any>>
+    }
+  ) {
+    const contextId = options?.contextId ?? getActiveTransaction()?.id
+    // Use the builder instance as the job ID for deduplication. This is memory-safe
+    // because the scheduler's context Map is deleted after flushing (no long-term retention).
+    const jobId = options?.jobId ?? this
+    const dependencyBuilders = (() => {
```
kevin-dp: nit: `dependencyBuilders` reads as if this is a function that builds dependencies. This is not the case because actually this holds the array of dependencies so a better name would be `dependentBuilders`.

packages/db/src/query/live/collection-config-builder.ts
```diff
@@ -412,6 +581,14 @@ export class CollectionConfigBuilder<
       const collection =
         this.collectionByAlias[alias] ?? this.collections[collectionId]!
 
+      const dependencyBuilder = getCollectionBuilder(collection)
```
kevin-dp, line 727: nit: `dependentBuilder` would be a more accurate name

packages/db/src/query/live/collection-config-builder.ts
```diff
@@ -176,13 +209,149 @@ export class CollectionConfigBuilder<
...
+
+      deps.delete(this)
+
+      return Array.from(deps)
+    })()
+    // We intentionally scope deduplication to the builder instance. Each instance
+    // owns caches and compiled pipelines, so sharing an entry across instances that
+    // merely reuse the same string id would bind the wrong `this` in the run closure.
+
+    const createEntry = () => {
```
kevin-dp: Why does this need to be a function? It doesn't take any arguments and it's only called once by the scheduler. We could as well just construct the object that is returned by this function and pass that object to `transactionScopedScheduler.schedule`:
```ts
const state: PendingGraphRun<TResult> = {
  config,
  syncState,
  loadCallbacks: callback ? new Set([callback]) : new Set(),
}

const entry = {
  state,
  run: () => {
    ...
  }
}

transactionScopedScheduler.schedule({
  contextId,
  jobId,
  dependencies: dependencyBuilders,
  entry,
  updateEntry,
})
```

packages/db/src/query/live/collection-config-builder.ts
```diff
@@ -176,13 +209,149 @@ export class CollectionConfigBuilder<
...
+            this.maybeRunGraph(state.config, state.syncState, combinedLoader)
+          } finally {
+            // Clear callbacks after run to avoid carrying stale closures across transactions
+            state.loadCallbacks.clear()
+          }
+        },
+      }
+    }
+
+    const updateEntry = (entry: { state: PendingGraphRun<TResult> }) => {
```
kevin-dp: I don't like that we are creating a function that closes over the `config`, `syncState`, and `callback` to then pass that closure to the `schedule` method which then will call it on the new entry or the entry it fetched. There's a lot of indirection here and it's not intuitive at first. I much prefer functions that take their dependencies as arguments instead of closing over them:
```ts
// Define this in an utility file or in the Scheduler file
function updateEntry(entry, config, syncState, callback) {
  entry.state.config = config
  entry.state.syncState = syncState
  if (callback) {
    entry.state.loadCallbacks.add(callback)
  }
}

// The here in `scheduleGraphRun`
transactionScopedScheduler.schedule({
  contextId,
  jobId,
  dependencies: dependencyBuilders,
  entry, // see my other comment
  // now pass the state explicitly instead of passing a function that closes over it
  config,
  syncState,
  callback
})
```

packages/db/src/scheduler.ts
```diff
@@ -0,0 +1,204 @@
...
+    if (!entry) {
+      entry = createEntry()
+      context.entries.set(jobId, entry)
+      context.queue.push(jobId)
+    }
+
+    updateEntry?.(entry)
+
+    if (dependencies) {
+      const depSet = new Set<unknown>()
```
kevin-dp: This constant and the for loop below can be simplified:
```ts
const depSet = new Set<unknown>(dependencies)
depSet.delete(jobId)
```

packages/db/src/scheduler.ts
```diff
@@ -0,0 +1,204 @@
...
+        if (!entry) {
+          dependencies.delete(jobId)
+          completed.delete(jobId)
+          continue
+        }
+
+        const deps = dependencies.get(jobId)
+        const ready =
+          !deps ||
+          deps.size === 0 ||
```
kevin-dp: No need to check that the size is 0, if the size is zero the condition on the line below this one will be true

packages/db/src/scheduler.ts
```diff
@@ -0,0 +1,204 @@
...
+          [...deps].every((dep) => dep === jobId || completed.has(dep))
```
kevin-dp: Why are we making a copy of deps here with `[...deps]`? The `every` function does not mutate the array, no need to make a copy.

packages/db/src/scheduler.ts
```diff
@@ -0,0 +1,204 @@
...
+   * Flush all queued work for a context. Jobs with unmet dependencies are retried.
+   * Throws if a pass completes without running any job (dependency cycle).
+   */
+  flush(contextId: SchedulerContextId): void {
+    const context = this.contexts.get(contextId)
+    if (!context) return
+
+    const { queue, entries, dependencies, completed } = context
+
+    while (queue.length > 0) {
```
kevin-dp, line 104: This is quite inefficient because a `queue` is the wrong data structure for this. The crux of the inneficiency comes from the fact that the order of the elements in the queue does not represent the order in which the elements need to be processed. In the worst case, the elements are actually in reverse order. Say that we have 4 jobs, numbered 1 to 4, and their dependencies are such that they need to be executed in this order: 1, 2, 3, 4. Now imagine that in the queue they are in reverse order: [4, 3, 2, 1]. The while loop below will handle them as follows:
```
4, 3, 2, 1
   3, 2, 1, 4 // because 4 has unmet dependencies
      2, 1, 4, 3 // because 3 has unmet dependencies
         1, 4, 3, 2 // because 2 has unmet dependencies
-- We had to loop through the entire list just to process the first element
            4, 3, 2 // finally processed 1
               3, 2, 4 // because 4 has unmet dependencies
                  2, 4, 3 // because 3 has unmet dependencies
-- We had to loop through the entire list again to process the 2nd element
                     4, 3 // finally processed 2
                        3, 4 // because 4 has unmet dependencies
-- We had to loop through the entire list again to process the 3nd element
                           4 // finally processed 3
-- And now we proces the last element
```
As we can see from this example, this is in `O(n^2)` because each element may require traversing the entire queue.

If we analyse this closer, we can only process jobs that have no more dependencies (or all dependencies are met). Conceptually, it would be a DAG and we want to continuously process sinks (nodes having no outgoing edges) and after processing a sink we remove it from the DAG. Practically, a full-blown DAG would probably be too heavy but i'm pretty sure we can build a custom solution here.

packages/db/src/scheduler.ts
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    this.contexts.delete(contextId)
+  }
+
+  /**
+   * Flush all contexts with pending work. Useful during tear-down.
+   */
+  flushAll(): void {
+    for (const contextId of Array.from(this.contexts.keys())) {
```
kevin-dp, line 158: Given that `flush` is `O(n^2)` and this is calling `flush` for each `contextId` it becomes even more expensive.

packages/db/src/scheduler.ts
```diff
@@ -0,0 +1,204 @@
...
+
+  /** Clear all scheduled jobs for a context. */
+  clear(contextId: SchedulerContextId): void {
+    this.contexts.delete(contextId)
+  }
+
+  /** Check if a context has pending jobs. */
+  hasPendingJobs(contextId: SchedulerContextId): boolean {
+    const context = this.contexts.get(contextId)
+    if (!context) return false
```
kevin-dp: These 2 lines can be simplified to:
```ts
return context && context.entries.size > 0
```

packages/db/src/query/live/collection-subscriber.ts
```diff
@@ -11,6 +11,8 @@ import type { OrderByOptimizationInfo } from "../compiler/order-by.js"
 import type { CollectionConfigBuilder } from "./collection-config-builder.js"
 import type { CollectionSubscription } from "../../collection/subscription.js"
 
+const loadMoreCallbackSymbol = Symbol(`tanstack.db.loadMore`)
```
kevin-dp: nit: I propose we change this one to `@tanstack/db.loadMore` to be consistent with the naming we used in `collection-registry.ts`:
```ts
const BUILDER_SYMBOL = Symbol.for(`@tanstack/db.collection-config-builder`)
```

packages/db/src/query/live/collection-subscriber.ts
```diff
@@ -80,18 +82,13 @@ export class CollectionSubscriber<
       this.collection.config.getKey
     )
 
-    // Do not provide the callback that loads more data
-    // if there's no more data to load
-    // otherwise we end up in an infinite loop trying to load more data
```
kevin-dp, line 88: This comment is useful to keep

packages/db/src/query/live/collection-subscriber.ts
```diff
@@ -80,18 +82,13 @@ export class CollectionSubscriber<
...
     const dataLoader = sentChanges > 0 ? callback : undefined
 
-    // We need to call `maybeRunGraph` even if there's no data to load
-    // because we need to mark the collection as ready if it's not already
-    // and that's only done in `maybeRunGraph`
```
kevin-dp: This comment is also useful to keep but may need to be changed slightly (at least the references to `maybeRunGraph` need to be changed to `scheduleGraphRun`)

packages/db/src/query/live/collection-subscriber.ts
```diff
@@ -201,10 +198,22 @@ export class CollectionSubscriber<
...
+    // This ensures we pass the same function instance to the scheduler each time,
+    // allowing it to deduplicate callbacks when multiple changes arrive during a transaction.
+    type SubscriptionWithLoader = CollectionSubscription & {
+      [loadMoreCallbackSymbol]?: () => boolean
+    }
+
+    const subscriptionWithLoader = subscription as SubscriptionWithLoader
+
+    const boundLoader =
+      subscriptionWithLoader[loadMoreCallbackSymbol] ??
```
kevin-dp: If i'm not mistaked, the trick we do with ?? here is exactly what `??=` does so we can rewrite this as:
```ts
subscriptionWithLoader[loadMoreCallbackSymbol] ??=
  this.loadMoreIfNeeded.bind(this, subscription)
```

packages/db/src/query/live/collection-subscriber.ts
```diff
@@ -201,10 +198,22 @@ export class CollectionSubscriber<
     }
 
     const trackedChanges = this.trackSentValues(changes, orderByInfo.comparator)
-    this.sendChangesToPipeline(
-      trackedChanges,
-      this.loadMoreIfNeeded.bind(this, subscription)
-    )
+
+    // Cache the bound loader on the subscription using a symbol property.
```
kevin-dp: "bound loader" is quite cryptic.

kevin-dp (changes requested):
I had a close look at this PR and left some comments. My main concern is the time complexity of the scheduler's `flush` method.

packages/db/src/scheduler.ts
samwillis: I think I would be concurred if there thousands of tasks, but usually a transaction is only going to trigger updates in a handful of live queries. The intention this is to remove the very expensive re-renders from React and other frameworks.
Also note that the queue will naturally be partially (if not completely) sorted due to the triggering of child queries scheduling their own task and it being appended to the end.
This schedular is only there to catch the edge cases where a live query may join in an order that could result in a partial order rather than complete order to the tasks.
The worse case is a vanishingly small likelihood.

packages/db/src/query/live/collection-config-builder.ts
samwillis: ~~createEntry is a function for lazy initialization: the scheduler only calls it when creating a new entry. If this job was already scheduled (same jobId), the scheduler reuses the existing entry and only calls updateEntry to merge state.~~
~~A job can be scheduled multiple times in a transaction when multiple source collections change (e.g., a live query joins users & tasks; both users.insert() and tasks.insert() in the same transaction each schedule this query, but it should only run once).~~

samwillis: Scratch this, I'm refactoring the scheduler.

samwillis:
@kevin-dp I've largely refactored this PR based on your feedback, removing the closures and hopefully simplifying it. Summary for cursor:

## Overview

This refactoring addressed PR feedback about excessive closure usage in the live query scheduler and builder, while also fixing a type error, preventing memory leaks, and improving overall code clarity, testability, and performance.

## Problems Addressed

### 1. Type Error
- **Problem**: `liveQueryJoin.utils.getRunCount is not a function` when running tests
- **Root Cause**: `createLiveQueryCollection` was overwriting `LiveQueryCollectionUtils` when custom utils were provided via `config.utils`
- **Fix**: Changed to merge utils using spread syntax: `options.utils = { ...options.utils, ...config.utils }`

### 2. Excessive Closure Usage
- **Problem**: PR review feedback indicated too many closures, particularly in the scheduler and how it interacted with the builder
- **Previous Architecture**: 
  - Scheduler managed state via `createEntry`/`updateEntry` callbacks that closed over `config`, `syncState`, and callbacks
  - Complex indirection: builder created closures → passed to scheduler → scheduler called closures to manage state
  - State was scattered between scheduler's generic state management and builder's specific needs

### 3. Memory Leaks
- **Problem 1**: `pendingGraphRuns` map entries could leak when sync ended before transactions flushed
- **Problem 2**: Scheduler's `onClear` listener held strong reference to builder, preventing garbage collection
- **Impact**: Stale callbacks and builder instances accumulated in long-lived apps that create/destroy live queries

## Solution: Separation of Concerns

### New Architecture Principles

1. **Scheduler only orchestrates execution order**
   - Simplified to just managing job dependencies and execution sequencing
   - No longer manages caller-specific state
   - Jobs provide a simple `run: () => void` function
   - Deduplication happens by replacing the `run` function when the same job is scheduled multiple times

2. **Builder manages its own state**
   - Introduced `currentSyncConfig` and `currentSyncState` as public instance properties
   - Manages `pendingGraphRuns` map internally (per scheduler context/transaction)
   - Clear lifecycle: properties set when sync starts, cleared when sync stops
   - Made public for testability (builder is internal, not public API)

3. **Lifecycle Coordination**
   - Builder registers a `clearPendingGraphRun` listener via `scheduler.onClear()` when sync starts
   - When scheduler clears a context (e.g., transaction rollback), builder cleans up its pending state
   - Listener is unregistered when sync stops to prevent memory leaks
   - Graceful degradation: if sync ends while transaction is in-flight, execution returns early without error

## Key Changes

### Scheduler (`packages/db/src/scheduler.ts`)

**Before:**
```typescript
interface ScheduleOptions<TState> {
  contextId?: SchedulerContextId
  jobId: unknown
  dependencies?: Iterable<unknown>
  createEntry: () => SchedulerEntry<TState>
  updateEntry: (entry: SchedulerEntry<TState>) => void
  run: (state: TState) => void
}
```

**After:**
```typescript
interface ScheduleOptions {
  contextId?: SchedulerContextId
  jobId: unknown
  dependencies?: Iterable<unknown>
  run: () => void
}
```

**Changes:**
- Removed `createEntry`, `updateEntry`, and generic `TState` parameter
- Simplified state management to just storing `run` functions
- Added `onClear()` listener mechanism for cleanup coordination
- State is now: `{ jobs: Map<unknown, () => void>, ... }` instead of `{ entries: Map<unknown, SchedulerEntry<TState>>, ... }`

**Performance Optimizations:**
- Dependency set creation: `new Set(dependencies); depSet.delete(jobId)` instead of manual filtering
- Changed from `[...deps].every()` to `for...of` loop to avoid intermediate array allocation
- Simplified `hasPendingJobs` to single return statement: `return !!context && context.jobs.size > 0`

### CollectionConfigBuilder (`packages/db/src/query/live/collection-config-builder.ts`)

**Key Additions:**
```typescript
export class CollectionConfigBuilder<TContext, TResult> {
  // Current sync session state - set when sync starts, cleared when stops
  // Public for testing (builder is internal, not public API)
  public currentSyncConfig: Parameters<SyncConfig<TResult>['sync']>[0] | undefined
  public currentSyncState: FullSyncState | undefined

  // Builder manages its own pending state per scheduler context
  private readonly pendingGraphRuns = new Map<SchedulerContextId, PendingGraphRun>()
  
  // Scheduler listener lifecycle management (prevents memory leaks)
  private unsubscribeFromSchedulerClears?: () => void
}
```

**Flow Changes:**

1. **`scheduleGraphRun()`**:
   - Uses instance properties `currentSyncConfig` and `currentSyncState` 
   - Manages `pendingGraphRuns` state directly (get or create, accumulate callbacks)
   - Passes simple `run` function to scheduler: `run: () => this.executeGraphRun(contextId, pendingToPass)`
   - No closures over `config`/`syncState` passed to scheduler

2. **`executeGraphRun()`**:
   - **Removes pending state from map FIRST** (before checking sync state) to prevent memory leaks
   - Returns early without error if sync has ended (graceful degradation)
   - Re-schedules during execution create fresh state (map entry already removed)
   - Increments run count and executes graph with combined loader callbacks

3. **`clearPendingGraphRun()`**:
   - New public method called by scheduler's `onClear` listener
   - Ensures `pendingGraphRuns` map stays in sync when scheduler clears a context
   - Prevents stale callbacks from surviving transaction rollbacks

4. **`maybeRunGraph()`**:
   - Signature simplified from `maybeRunGraph(config, syncState, callback)` to `maybeRunGraph(callback)`
   - Uses instance properties `currentSyncConfig` and `currentSyncState` directly

5. **`syncFn()`** lifecycle:
   - **Start**: 
     - Sets `currentSyncConfig` and `currentSyncState` instance properties
     - Registers `onClear` listener with scheduler
   - **Stop**: 
     - Unregisters `onClear` listener (prevents builder memory leak)
     - Clears `currentSyncConfig` and `currentSyncState`
     - Clears all `pendingGraphRuns` (prevents transaction entry leaks)
     - Resets caches, clears lazy source state, unsubscribes and clears subscriptions

### CollectionSubscriber (`packages/db/src/query/live/collection-subscriber.ts`)

**Before:**
```typescript
constructor(
  private alias: string,
  private collectionId: string,
  private collection: Collection,
  private config: Parameters<SyncConfig<TResult>['sync']>[0],
  private syncState: FullSyncState,
  private collectionConfigBuilder: CollectionConfigBuilder<TContext, TResult>
)
```

**After:**
```typescript
constructor(
  private alias: string,
  private collectionId: string,
  private collection: Collection,
  private collectionConfigBuilder: CollectionConfigBuilder<TContext, TResult>
)
```

**Changes:**
- Removed `config` (was never used) and `syncState` parameters
- Accesses sync state via `collectionConfigBuilder.currentSyncState` instead
- Consistent with the new architecture: state lives on the builder

**Code Quality Improvements:**
- Renamed "bound loader" to "loadMoreCallback" for clarity
- Simplified callback caching using `??=` operator instead of manual `??` pattern
- Improved code comments to explain callback purpose rather than implementation details

## Benefits

### 1. **Eliminated Closure Anti-patterns**
- Scheduler no longer accepts or manages closures over state
- Builder state management is explicit through instance properties
- Clearer data flow: state lives where it's used, not captured in closures

### 2. **Memory Leak Prevention**
- Fixed leak where `pendingGraphRuns` entries weren't cleaned up when sync ended before transaction flush
- Fixed leak where scheduler's listener Set held strong reference to builder after teardown
- Defense-in-depth cleanup: per-transaction cleanup, full map clear on sync stop, listener unregistration
- Prevents accumulation of stale callbacks and builder instances across subscription cycles

### 3. **Performance Optimizations**
- Eliminated intermediate array allocations in scheduler's dependency checking
- More efficient dependency set construction using Set constructor
- Simplified conditional logic throughout (fewer branches, clearer intent)

### 4. **Improved Testability**
- Instance properties `currentSyncConfig`, `currentSyncState`, `scheduleGraphRun`, and `clearPendingGraphRun` are public
- Tests can set state directly without `any` casts
- Internal class boundaries respected (builder is not public API, so public members are fine)

### 5. **Better Separation of Concerns**
- **Scheduler**: Pure orchestration of execution order and dependencies
- **Builder**: Owns all live query state and lifecycle
- Clear interfaces and responsibilities

### 6. **Lifecycle Safety**
- Explicit state lifecycle tied to sync session (set on start, cleared on stop)
- Graceful handling of edge cases (sync ends during transaction, transaction rollback)
- Listener registration/unregistration follows sync session lifecycle

### 7. **Simpler Mental Model**
- State is stored where it's used (instance properties) rather than captured in closures
- Execution flow is more linear and easier to trace
- Less indirection between scheduler and builder

## Testing

- All existing tests pass (1095 passed, 3 skipped)
- Updated `scheduler.test.ts` to work with new public properties
- Mock signatures updated to reflect simplified `maybeRunGraph` signature
- No behavioral changes to live query functionality

## Migration Notes

This is an internal refactoring with no public API changes:
- `Scheduler` class is internal
- `CollectionConfigBuilder` is internal
- `CollectionSubscriber` is internal
- `createLiveQueryCollection` public API unchanged
- Live query behavior unchanged from user perspective
- Existing code using the public API requires no modifications

## Summary

The refactoring successfully addresses PR feedback by eliminating problematic closure patterns while improving code clarity, testability, and lifecycle management. The new architecture draws clear boundaries: the scheduler orchestrates execution order, while the builder manages its own state through the lifecycle of a sync session.

**Key Achievements:**
- ✅ Eliminated closure anti-patterns (primary goal)
- ✅ Fixed multiple memory leaks in transaction/sync lifecycle
- ✅ Improved performance through allocation reduction
- ✅ Enhanced testability with explicit state management
- ✅ Better code clarity with improved naming and simplified logic
- ✅ No breaking changes to public API
- ✅ All tests passing

packages/db/src/query/live/collection-subscriber.ts
```diff
@@ -201,10 +201,20 @@ export class CollectionSubscriber<
...
+    // Cache the loadMoreIfNeeded callback on the subscription using a symbol property.
+    // This ensures we pass the same function instance to the scheduler each time,
+    // allowing it to deduplicate callbacks when multiple changes arrive during a transaction.
+    type SubscriptionWithLoader = CollectionSubscription & {
+      [loadMoreCallbackSymbol]?: () => boolean
+    }
+
+    const subscriptionWithLoader = subscription as SubscriptionWithLoader
+
+    const loadMoreCallback = (subscriptionWithLoader[loadMoreCallbackSymbol] ??=
```
kevin-dp: Assigning within an assignment is a bit ugly. Perhaps just assign to `subscriptionWithLoader[loadMoreCallbackSymbol]` using `??=` and then pass `subscriptionWithLoader[loadMoreCallbackSymbol]` to `this.sendChangesToPipeline`?

packages/db/src/query/live/collection-subscriber.ts
kevin-dp: I still think this comment is useful ;D

packages/db/src/query/live/collection-subscriber.ts
kevin-dp: Still think the comment is useful

Issue #478, opened 2025-09-01
🐛 Bug: ⁠queryCollection stops re-rendering after rapid inserts with delayed ⁠onInsert


treyhoover:
When adding artificial latency to ⁠onInsert and inserting two items in quick succession, the UI stops re-rendering further updates.

A ⁠`setInterval` log of the collection state confirms it is correct in memory (but not before the last render is logged, hinting at a race condition).
Removing ⁠`orderBy` in the live query fixes the render but throws a different error (see below).

```ts
export const todosCollection = createCollection(
  queryCollectionOptions({
    queryKey: ["todos"],
    queryFn: async () => {
      const todos = await supabase.from("todos").select("*");
      if (todos.error) throw todos.error;

      return todos.data;
    },
    queryClient,
    getKey: (item) => item.id,
    onInsert: async ({ transaction }): Promise<void> => {
      await sleep(1000);

      await Promise.all(
        transaction.mutations.map(async (item) => {
          const todo = await supabase.from("todos").insert(item.modified);
          if (todo.error) throw todo.error;
        }),
      );
    },

    onUpdate: async ({ transaction }): Promise<void> => {
      await Promise.all(
        transaction.mutations.map(async (item) => {
          const todo = await supabase.from("todos").update(item.modified).eq("id", item.modified.id);

          if (todo.error) throw todo.error;

          return todo.data;
        }),
      );
    },

    onDelete: async ({ transaction }): Promise<void> => {
      const ids = transaction.mutations.map((m) => m.key);
      const res = await supabase.from("todos").delete().in("id", ids);
      if (res.error) throw res.error;
    },
  }),
);
```

```ts
  const todosQuery = useLiveQuery(
    (q) => q.from({ todo: todosCollection }).orderBy(({ todo }) => todo.created_at, "asc"),
    [user.id],
  );
```

If I remove `orderBy` on the live query it renders correctly and surfaces this error:
`Uncaught (in promise) TransactionError: You can no longer call .rollback() as the transaction is already completed`

PR #629, opened 2025-10-02
fix concurrent mutation bug

samwillis:
test reproducing bug in #478

failed test run showing the bug in the tests: https://github.com/TanStack/db/actions/runs/18202939375/job/51826196817?pr=629

Description of bug from GPT5:

### Root cause

Two separate problems were hiding behind the “live query stops updating” report:

1. Batched change flush never finished.
  When a mutation completed while a sync transaction was pending, `CollectionStateManager.onTransactionStateChange` flipped `this._changes.shouldBatchEvents` to `true` so that “optimistic clean‑up” events would be buffered until the server response arrived. If the eventual sync commit produced no visible diff (the optimistic state already matched the authoritatively synced data), `commitPendingTransactions` invoked `emitEvents([], true)`. The previous implementation returned early for an empty change set and never cleared `shouldBatchEvents`. From then on the collection stayed in “batching” mode and every subsequent live-query change was silently dropped.

2. Redundant inserts surfaced as duplicate-key errors once the first bug was fixed.
  After unsticking the batching flag, the test revealed that the live-query sync pipeline can legitimately replay the same insert the optimistic layer already applied. The existing sync adapter treated that as a protocol violation and threw `DuplicateKeySyncError`, which is just noise if the payload matches.

### Fix

- Reset the batching state whenever `emitEvents` is called with `forceEmit = true`, even if the combined change list is empty. This guarantees the collection exits batching mode at the end of every sync commit, so live-query subscribers continue to receive future changes.
- Teach the sync bridge to coalesce redundant inserts into updates: when we see an insert for a key that already exists and the stored value is identical, rewrite the change to update instead of throwing. That lets harmless “echo” writes flow through without triggering an error.

Together these changes restore the expected behavior for concurrent optimistic inserts: the live query keeps emitting events, and the commit phase no longer crashes on duplicate-key checks.

### DIFF
--- a/.changeset/plain-pets-fix.md
+++ b/.changeset/plain-pets-fix.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+Fix live query stalls with concurrent optimistic inserts by clearing stale batching state and tolerating redundant sync echoes.


packages/db/src/collection/sync.ts
```diff
@@ -7,6 +7,7 @@ import {
   SyncTransactionAlreadyCommittedError,
   SyncTransactionAlreadyCommittedWriteError,
 } from "../errors"
+import { deepEquals } from "../utils"
```
samwillis, line 10: @KyleAMathews going to blame GPT5 for another `deepEquals` 😂 

Again, this is very really going to be triggered.

samwillis:
Review from ChatGPT5:

Here’s a focused code review of that PR.

## TL;DR

Good fix. It addresses two real issues behind “live query stalls on concurrent inserts”:

1. batching flag getting stuck after a force flush with no changes, and
2. duplicate-key errors when a sync “echo” replays an insert that matches the optimistic write.

The implementation is sensible and the tests exercise the behavior well. I have a few suggestions around performance, edge cases, and test coverage.

---

## What changed & why it helps

### 1) Unsticking the batching state

In `CollectionChangesManager.emitEvents`, the PR now clears both `batchedEvents` and `shouldBatchEvents` whenever `forceEmit` is true — even if there are no events to emit. This prevents the collection from remaining in “batching mode” and dropping future change events when a commit is effectively a no-op. ([[GitHub](https://github.com/TanStack/db/pull/629/files)][1])

**Looks good:** This is the correct place to end the batching cycle; it’s low risk and addresses the stall.

**Minor nits / suggestions**

* Consider adding a short comment that explicitly explains the problematic scenario (commit yields no diff → batching never ends) to avoid future regressions. The current comment is close; one extra sentence would make intent crystal clear. ([[GitHub](https://github.com/TanStack/db/pull/629/files)][1])

### 2) Tolerating redundant insert “echoes”

In `CollectionSyncManager`, when an incoming `insert` targets a key that already exists, the PR checks if the stored value is **deep-equal** to the incoming value. If so, it rewrites the change to an `update` instead of throwing `DuplicateKeySyncError`; otherwise it continues to throw. ([[GitHub](https://github.com/TanStack/db/pull/629/files)][1])

**I like it:** This is a pragmatic way to let harmless duplicates through and keeps the pipeline stable during optimistic → authoritative reconciliation.

**Concerns / ideas**

* **Performance:** `deepEquals` on larger records or hot paths could be costly. If this code runs frequently, consider:

  * Early-outs with a shallow/shape hash (if available) before falling back to deep compare.
  * Comparing only the projected fields the collection tracks (if you have that notion) to avoid scanning large blobs. ([[GitHub](https://github.com/TanStack/db/pull/629/files)][1])
* **Equality semantics:** Make sure `deepEquals` treats `undefined` vs. “missing property” consistently with how collection values are serialized. Subtle mismatches here could turn benign echoes into hard errors. If not guaranteed, documenting the expected semantics would help. ([[GitHub](https://github.com/TanStack/db/pull/629/files)][1])
* **Pending deletes / truncates:** Your logic still throws if there’s no pending delete and it isn’t a truncate, unless values are equal. That’s correct; consider an explicit unit test for the “existing but different value” path (i.e., still throws). ([[GitHub](https://github.com/TanStack/db/pull/629/files)][1])

---

## Tests

The new `concurrent-mutations.test.ts` does a nice job reproducing and guarding the behavior:

* Uses fake timers to simulate latency and calls the captured sync hooks.
* Covers with and without `orderBy`.
* Asserts batching flag reset (`shouldBatchEvents === false`). ([[GitHub](https://github.com/TanStack/db/pull/629/files)][1])

**Suggested additions**

1. **Echo insert with identical value**: Explicitly assert that no error is thrown and the op is rewritten to `update` (you could spy on the operations list to confirm `type` was adjusted). ([[GitHub](https://github.com/TanStack/db/pull/629/files)][1])
2. **Echo insert with *different* value**: Assert that `DuplicateKeySyncError` is still thrown. This ensures the guardrail remains. ([[GitHub](https://github.com/TanStack/db/pull/629/files)][1])
3. **Pending delete scenario**: When a delete is pending for the key, verify the existing throw path still occurs. ([[GitHub](https://github.com/TanStack/db/pull/629/files)][1])
4. **No-op commit flushing batching**: A tiny unit test directly exercising `emitEvents([], /*force*/ true)` to prove batching gets cleared even with empty changes (this isolates the fix). ([[GitHub](https://github.com/TanStack/db/pull/629/files)][1])

---

## DX / maintainability

* The changeset note succinctly describes the externally visible behavior (“Fix live query stalls… tolerate redundant sync echoes”). Good. Consider also referencing the internal batching flag fix in a code comment near `emitEvents` so future maintainers don’t regress it. ([[GitHub](https://github.com/TanStack/db/pull/629/files)][1])
* A teammate commented about “another `deepEquals`.” If this codepath is rare, it’s fine. If not, you may want to centralize an optimized equality check for synced values, possibly with a cheap pre-check (e.g., pointer or version compare) before deep diff. ([[GitHub](https://github.com/TanStack/db/pull/629/files)][1])

---

## Merge readiness

* ✅ Behavior fix is targeted and minimal.
* ✅ Tests cover the main regression and the batching flag.
* ⚠️ I’d add the couple of edge-case tests above, but I wouldn’t block merge if this is urgent; they can follow-up.

**Verdict:** Approve, with optional follow-ups on tests and a note to keep an eye on `deepEquals` cost in hot paths. ([[GitHub](https://github.com/TanStack/db/pull/629/files)][1])

If you want, I can draft those additional tests right now in the style of `concurrent-mutations.test.ts`.

[1]: https://github.com/TanStack/db/pull/629/files "fix concurrent mutation bug by samwillis · Pull Request #629 · TanStack/db · GitHub"

samwillis:
Closing in favour of #631 which builds on this

PR #638, opened 2025-10-03
Add acceptMutations utility for local collections in manual transactions

### LINKED ISSUE #446: Bug: Transactions do not work when using localOnlyCollection
Example:
```js
// transaction-test.js
// Run with: node transaction-test.js

import { createCollection, createTransaction } from '@tanstack/db';
import { localOnlyCollectionOptions } from '@tanstack/db';

console.log('🧪 Starting TanStack DB LocalOnly Transaction Test\n');

const testCollection = createCollection(
  localOnlyCollectionOptions({
    id: 'test-collection',
    getKey: (item) => item.id,
    initialData: [
      { id: 'initial1', name: 'Initial Item 1', value: 100 },
      { id: 'initial2', name: 'Initial Item 2', value: 200 },
    ],
      const inserted = transaction.mutations[0].modified;
      console.log('  📦 onInsert triggered:', inserted.name);
    },
      const { original, modified } = transaction.mutations[0];
      console.log(
        '  📦 onUpdate triggered:',
        original.name,
        '->',
        modified.name
      );
    },
      const deleted = transaction.mutations[0].original;
      console.log('  📦 onDelete triggered:', deleted.name);
    },
  })
);

console.log('1️⃣ Initial collection state:');
console.log('   Array:', testCollection.toArray);
console.log('   Count:', testCollection.size, '\n');

// Test 1: Direct operations (should work fine)
console.log('2️⃣ Testing direct operations:');
testCollection.insert({ id: 'direct1', name: 'Direct Insert', value: 300 });
console.log('   After direct insert - Count:', testCollection.size);
console.log('   Inserted item:', testCollection.get('direct1'));

const firstItem = testCollection.toArray[0];
if (firstItem) {
  testCollection.update(firstItem.id, (draft) => {
    draft.name = 'Updated Direct';
    draft.value = 999;
  });
  console.log('   After direct update - Count:', testCollection.size);
  console.log('   Updated item:', testCollection.get(firstItem.id));
}
console.log();

// Test 2: Manual transaction
console.log('3️⃣ Testing manual transaction:');
console.log('   Before transaction - Count:', testCollection.size);
console.log('   Array:', testCollection.toArray);

const transaction = createTransaction({
  autoCommit: false,
  mutationFn: async ({ transaction }) => {
    console.log(
      '   🎯 Transaction mutationFn called with',
      transaction.mutations.length,
      'mutations'
    );
  },
});

console.log('   Creating transaction mutate block...');
transaction.mutate(() => {
  console.log('   🔄 Inside mutate block');
  console.log('Inserting item tx1');
  testCollection.insert({
    id: 'tx1',
    name: 'Transaction Insert 1',
    value: 400,
  });
  console.log('Inserting item tx2');
  testCollection.insert({
    id: 'tx2',
    name: 'Transaction Insert 2',
    value: 500,
  });

  console.log('Updating item initial2');
  const itemToUpdate = testCollection.get('initial2');
  if (itemToUpdate) {
    testCollection.update('initial2', (draft) => {
      draft.name = 'Updated via Transaction';
      draft.value = 777;
    });
  }
});

console.log('   Before commit - Count:', testCollection.size);
console.log(
  '   Items:',
  testCollection.toArray.map((i) => ({ id: i.id, name: i.name }))
);

console.log('   Committing transaction...');
try {
  await transaction.commit();
  console.log('   ✅ Transaction committed successfully');
} catch (error) {
  console.log('   ❌ Transaction commit failed:', error.message);
}

console.log('   After commit - Count:', testCollection.size);
console.log(
  '   Final items:',
  testCollection.toArray.map((i) => ({
    id: i.id,
    name: i.name,
    value: i.value,
  }))
);
```

Output:
```
🧪 Starting TanStack DB LocalOnly Transaction Test

1️⃣ Initial collection state:
   Array: [
  { id: 'initial1', name: 'Initial Item 1', value: 100 },
  { id: 'initial2', name: 'Initial Item 2', value: 200 }
]
   Count: 2 

2️⃣ Testing direct operations:
  📦 onInsert triggered: Direct Insert
   After direct insert - Count: 3
   Inserted item: { id: 'direct1', name: 'Direct Insert', value: 300 }
  📦 onUpdate triggered: Initial Item 1 -> Updated Direct
   After direct update - Count: 3
   Updated item: { id: 'initial1', name: 'Updated Direct', value: 999 }

3️⃣ Testing manual transaction:
   Before transaction - Count: 3
   Array: [
  { id: 'initial1', name: 'Updated Direct', value: 999 },
  { id: 'initial2', name: 'Initial Item 2', value: 200 },
  { id: 'direct1', name: 'Direct Insert', value: 300 }
]
   Creating transaction mutate block...
   🔄 Inside mutate block
Inserting item tx1
Inserting item tx2
Updating item initial2
   Before commit - Count: 5
   Items: [
  { id: 'initial1', name: 'Updated Direct' },
  { id: 'initial2', name: 'Updated via Transaction' },
  { id: 'direct1', name: 'Direct Insert' },
  { id: 'tx1', name: 'Transaction Insert 1' },
  { id: 'tx2', name: 'Transaction Insert 2' }
]
   Committing transaction...
   🎯 Transaction mutationFn called with 3 mutations
   ✅ Transaction committed successfully
   After commit - Count: 3
   Final items: [
  { id: 'initial1', name: 'Updated Direct', value: 999 },
  { id: 'initial2', name: 'Initial Item 2', value: 200 },
  { id: 'direct1', name: 'Direct Insert', value: 300 }
]
```

Expected behavior: Committing a transaction applies the changes as a batch update.
Actual behavior: Committing a transaction erases the changes.

KyleAMathews:
## Summary
- Fixes #446
- Adds `utils.acceptMutations(transaction, collection)` to local-only and local-storage collections
- Enables manual transactions to work with local collections by providing explicit mutation persistence

## Context
Manual transactions with local-only and local-storage collections were losing changes because these collections rely on their mutation handlers (`onInsert`/`onUpdate`/`onDelete`) being called to persist data. When operations occur inside `tx.mutate()`, handlers aren't automatically called - only the outer transaction's `mutationFn` runs.

## Solution
Following the pattern from `@tanstack/query-db-collection`, both collection types now expose a `utils.acceptMutations(transaction, collection)` method that users explicitly call in their transaction's `mutationFn` to persist local mutations:

```typescript
const localData = createCollection(
  localOnlyCollectionOptions({
    getKey: (item) => item.id,
  })
)

const tx = createTransaction({
  mutationFn: async ({ transaction }) => {
    // Persist local-only mutations
    localData.utils.acceptMutations(transaction, localData)
    
    // Then make API call
    await api.save(...)
  }
})

tx.mutate(() => {
  localData.insert({ id: 1, data: 'metadata' })
  apiCollection.insert({ id: 2, data: 'main data' })
})

await tx.commit()
```

## Changes
- Added `acceptMutations` to `LocalOnlyCollectionUtils` interface
- Added `acceptMutations` to `LocalStorageCollectionUtils` interface
- Implemented mutation filtering and persistence logic
- Included JSON serialization validation for local-storage
- Updated JSDoc with manual transaction examples

## Test plan
- [ ] Test manual transactions with local-only collections
- [ ] Test manual transactions with local-storage collections
- [ ] Verify mutations are persisted correctly
- [ ] Verify JSON serialization validation works
- [ ] Verify existing direct operations still work

### DIFF
--- a/packages/db/src/local-only.ts
+++ b/packages/db/src/local-only.ts
@@ -266,7 +266,9 @@ export function localOnlyCollectionOptions<
     }
 
     // Persist the mutations through sync
-    syncResult.confirmOperationsSync(collectionMutations)
+    syncResult.confirmOperationsSync(
+      collectionMutations as Array<PendingMutation<T>>
+    )
   }
 
   return {


packages/db/src/local-only.ts
```diff
@@ -33,9 +34,32 @@ export interface LocalOnlyCollectionConfig<
...
+   *     // Persist local-only mutations
+   *     localData.utils.acceptMutations(transaction, localData)
+   *     // Then make API call
+   *     await api.save(...)
+   *   }
+   * })
+   */
+  acceptMutations: (
+    transaction: { mutations: Array<PendingMutation<Record<string, unknown>>> },
+    collection: unknown
```
samwillis: I don't understand why you have to pass the collection in here, that seems unnecessary.

packages/db/tests/local-only.test.ts
samwillis, line 1: There really should be tests for this new functionality

samwillis (commented):
Conceptional I think this is good, however if this is becoming a generic pattern it shouldn't be on the `utils` namespace, but on the top level.

Could we just have a `collection.acceptMutations(transaction)` method that forwards to the relevant `onInsert/Update/Delete` method when available?

That would work with all collections then.

packages/db/src/local-only.ts
KyleAMathews: hmm I thought I'd gotten rid of all of those. Will remove

KyleAMathews:
> however if this is becoming a generic pattern it shouldn't be on the utils namespace, but on the top level.

It's not — this is only intended for collections with no backend. It'd be very odd to do this e.g. for the query/electric collections as then nothing would be written to the backend.

samwillis:
> It's not — this is only intended for collections with no backend

But any collection with `onInsert/onUpdate/onDelete` does have a "backend". So if you have configured those, and now want to use this collection in a transaction with another, if `collection.acceptMutations(transaction)` was standard they could just do that?

if you try and use it on a collection that does not have the appropriate `onInsert/onUpdate/onDelete` then it would throw.

KyleAMathews:
Ok, I see where you're going — letting people rely on their onInsert/onUpdate/onDelete logic even for actions/custom transactions — I'm not a fan of that tbh — the great thing about collection handlers is the logic is very focused/simple as you know it's just coming from collection mutators. So keeping a clean separation between collection direct state changes & everything else is worthwhile. People commonly use api clients & helpers to abstract out any shared logic.

packages/db/src/local-only.ts
```diff
@@ -187,13 +241,34 @@ export function localOnlyCollectionOptions(
...
   return {
     ...restConfig,
     sync: syncResult.sync,
-    onInsert: wrappedOnInsert,
-    onUpdate: wrappedOnUpdate,
-    onDelete: wrappedOnDelete,
-    utils: {} as LocalOnlyCollectionUtils,
+    onInsert: wrappedOnInsert as any,
+    onUpdate: wrappedOnUpdate as any,
+    onDelete: wrappedOnDelete as any,
```
samwillis: nit: I don't like the cast to `any`, but not important right now.

pnpm-lock.yaml
samwillis, line 1: There are no package.json changes, and so it's unexpected to see a ppm-lock.yaml change.

packages/db/src/local-only.ts
KyleAMathews: heh, I keep telling Claude not to do this 😬 😆

pnpm-lock.yaml
KyleAMathews: There was a conflict I believe so it re-ran it — perhaps it deleted it first though — I can try reverting it

PR #646, opened 2025-10-06
Handle pushed down predicates in query collection

kevin-dp:
This PR modifies the query collection such that it creates a new query when predicates are pushed down. It passes the predicates to the query key builder function and also to the query function via a new `tsDb` namespace on the `meta` property of the context object.

An example:
```ts
import { createCollection } from "@tanstack/react-db"
import { queryCollectionOptions } from "@tanstack/query-db-collection"

const itemsPerPage = 20

  queryCollectionOptions({
    // Derive page number from limit, encode into the query key
    queryKey: ({ limit, orderBy: _orderBy, where: _where }) => {
      const page = computePageNumber(limit ?? itemsPerPage)
      return ["todos", { page }]
    },

    queryFn: async (ctx) => {
      const page = computePageNumber(ctx.meta?.tsDb?.limit ?? itemsPerPage)
      const res = await fetch(`/api/todos?page=${page}`)
      if (!res.ok) throw new Error("Failed to fetch todos")
      return res.json()
    },

    getKey: (item) => item.id,
    schema: todoSchema,
  })
)

function computePageNumber(limit: number): number {
  return Math.max(1, Math.ceil(limit / itemsPerPage))
}
```


KyleAMathews:
Closing this PR and reopening with a rebased branch to remove duplicate commits from #669 that are already in main. New PR incoming...

PR #648, opened 2025-10-06
electric-db-collection: support for snapshots in awaitTxid

samwillis:
### Summary

This PR enhances the `awaitTxId` utility in `@tanstack/electric-db-collection` to resolve transaction IDs based on PostgreSQL snapshot metadata from `snapshot-end` messages, enabling transaction matching on the initial snapshot at the start of a new shape.

### Motivation

When an Electric shape starts, the initial snapshot includes a `snapshot-end` control message with PostgreSQL snapshot metadata (xmin, xmax, xip_list). Previously, `awaitTxId` could only match against explicit txid arrays sent with individual change messages. This meant that transactions completed before the snapshot started couldn't be matched, even though they were visible in the snapshot data.

By supporting snapshot-based visibility checks, we can now correctly identify which transactions are included in the initial snapshot, improving transaction confirmation reliability.

### Changes

**Implementation (`electric.ts`):**
- Added `seenSnapshots` store to track PostgreSQL snapshots alongside `seenTxids`
- Added `isSnapshotEndMessage` and `parseSnapshotMessage` helper functions
- Enhanced `awaitTxId` to check transaction visibility against stored snapshots using `isVisibleInSnapshot` from `@electric-sql/client`
- Added dual subscription logic to resolve on either explicit txid match or snapshot visibility match
- Process and store snapshot metadata from `snapshot-end` control messages

**Tests (`electric.test.ts`):**
- Test snapshot metadata parsing and visibility matching
- Test txids visible in snapshots (< xmin, in range [xmin, xmax) excluding xip_list)
- Test txids NOT visible in snapshots (in xip_list, >= xmax)
- Test awaiting txids that arrive later via snapshot
- Test handling multiple snapshots
- Test that explicit txids and snapshot matching work together

### Dependencies

- Upgraded `@electric-sql/client` from `^1.0.10` to `^1.0.14` to access the `isVisibleInSnapshot` utility and `PostgresSnapshot` type

### Future Work

This implementation currently handles snapshots at the beginning of shape logs. Upcoming PRs for incremental sync will extend this to handle multiple snapshots over time and may introduce snapshot retention policies (they are kept unbounded for now).

### DIFF
--- a/packages/electric-db-collection/src/electric.ts
+++ b/packages/electric-db-collection/src/electric.ts
@@ -201,6 +201,7 @@ export function electricCollectionOptions(
     return new Promise((resolve, reject) => {
       const timeoutId = setTimeout(() => {
         unsubscribeSeenTxids()
+        unsubscribeSeenSnapshots()
         reject(new TimeoutWaitingForTxIdError(txId))
       }, timeout)
 
--- a/packages/electric-db-collection/src/electric.ts
+++ b/packages/electric-db-collection/src/electric.ts
@@ -2,6 +2,7 @@ import {
   ShapeStream,
   isChangeMessage,
   isControlMessage,
+  isVisibleInSnapshot,
 } from "@electric-sql/client"
 import { Store } from "@tanstack/store"
 import DebugModule from "debug"
@@ -27,6 +28,7 @@ import type {
   ControlMessage,
   GetExtensions,
   Message,
+  PostgresSnapshot,
   Row,
   ShapeStreamOptions,
 } from "@electric-sql/client"
@@ -37,6 +39,16 @@ const debug = DebugModule.debug(`ts/db:electric`)
  * Type representing a transaction ID in ElectricSQL
  */
 export type Txid = number
+
+/**
+ * Type representing the result of an insert, update, or delete handler
+ */
+type HandlerResult =
+  | {
+      txid?: Txid | Array<Txid>
+    }
+  | undefined
+  | null
 
 // The `InferSchemaOutput` and `ResolveType` are copied from the `@tanstack/db` package
 // but we modified `InferSchemaOutput` slightly to restrict the schema output to `Row<unknown>`
@@ -78,6 +90,22 @@ function isMustRefetchMessage<T extends Row<unknown>>(
   message: Message<T>
 ): message is ControlMessage & { headers: { control: `must-refetch` } } {
   return isControlMessage(message) && message.headers.control === `must-refetch`
+}
+
+function isSnapshotEndMessage<T extends Row<unknown>>(
+  message: Message<T>
+): message is ControlMessage & { headers: { control: `snapshot-end` } } {
+  return isControlMessage(message) && message.headers.control === `snapshot-end`
+}
+
+function parseSnapshotMessage(
+  message: ControlMessage & { headers: { control: `snapshot-end` } }
+): PostgresSnapshot {
+  return {
+    xmin: message.headers.xmin,
+    xmax: message.headers.xmax,
+    xip_list: message.headers.xip_list,
+  }
 }
 
 // Check if a message contains txids in its headers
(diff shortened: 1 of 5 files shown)


packages/electric-db-collection/src/electric.ts
```diff
@@ -38,6 +40,16 @@ const debug = DebugModule.debug(`ts/db:electric`)
  */
 export type Txid = number
 
+/**
+ * Type representing the result of an insert, update, or delete handler
+ */
+type HandlerResult =
```
kevin-dp: nit: i prefer a descriptive name like `MaybeTxId`

packages/electric-db-collection/src/electric.ts
```diff
@@ -80,6 +92,22 @@ function isMustRefetchMessage<T extends Row<unknown>>(
...
 }
 
+function isSnapshotEndMessage<T extends Row<unknown>>(
+  message: Message<T>
+): message is ControlMessage & { headers: { control: `snapshot-end` } } {
+  return isControlMessage(message) && message.headers.control === `snapshot-end`
+}
+
+function parseSnapshotMessage(
+  message: ControlMessage & { headers: { control: `snapshot-end` } }
```
kevin-dp: nit: perhaps worth introducing a helper type `type SnapshotEndMessage = ControlMessage & { headers: { control: `snapshot-end` } }` since we used this type twice already

packages/electric-db-collection/src/electric.ts
```diff
@@ -158,20 +188,46 @@ export function electricCollectionOptions(
       throw new ExpectedNumberInAwaitTxIdError(typeof txId)
     }
 
+    // First check if the txid is in the seenTxids store
     const hasTxid = seenTxids.state.has(txId)
     if (hasTxid) return true
 
+    // Then check if the txid is in any of the seen snapshots
+    const hasSnapshot = seenSnapshots.state.some((snapshot) =>
```
kevin-dp, line 201: How big are we envisioning this `seenSnapshots` list could become? Worst case we're looping over all snapshots and for each snapshot we check `isVisibleInSnapshot` which in the worst case loop over the snapshot's `xip_list`. So i'm wondering here if this could become a performance issue. If it is, we could as well keep a set of all txIds we've seen in snapshots and then we could just do a set lookup (which is in constant time).

packages/electric-db-collection/src/electric.ts
```diff
@@ -158,20 +188,46 @@ export function electricCollectionOptions(
...
           debug(`awaitTxId found match for txid %o`, txId)
           clearTimeout(timeoutId)
-          unsubscribe()
+          unsubscribeSeenTxids()
+          unsubscribeSeenSnapshots()
+          resolve(true)
+        }
+      })
+
+      const unsubscribeSeenSnapshots = seenSnapshots.subscribe(() => {
```
kevin-dp, line 223: iiuc this subscription is racing against the `seenTxids`subscription. The first one to find the txId will resolve the promise.

packages/electric-db-collection/src/electric.ts
```diff
@@ -413,6 +476,17 @@ function createElectricSync<T extends Row<unknown>>(
...
             return clonedSeen
           })
+
+          // Always commit snapshots when we receive up-to-date, regardless of transaction state
+          seenSnapshots.setState((currentSnapshots) => {
+            const clonedSeen = currentSnapshots.slice()
+            newSnapshots.forEach((snapshot) => {
+              debug(`new snapshot synced from pg %o`, snapshot)
+              clonedSeen.push(snapshot)
+            })
```
kevin-dp: Can we rewrite these 5 lines using the spread operator?
`const snapshots = [...currentSnapshots, ...newSnapshots]`

kevin-dp (approved):
LGTM, left a few nits.

packages/electric-db-collection/src/electric.ts
samwillis: Initially it should only be one, the initial snapshot, but with incremental sync it could grow. I intend to revisit both this and the seenTxids in a future PR

packages/electric-db-collection/src/electric.ts
samwillis: Yes

PR #650, opened 2025-10-06
fix race condition that resulted in live queries becoming stuck

### LINKED ISSUE #645: Functional where queries break collection sync when multiple optimistic updates are queued
I've made a repro of an issue I've been seeing in my app where collections with [functional where queries](https://tanstack.com/db/latest/docs/guides/live-queries#functional-where) can stop syncing when multiple optimistic updates are queued simultaneously. The page needs to be fully reloaded when this happens, so it's a fairly serious issue for me.

## Reproduction

[https://stackblitz.com/edit/tanstack-db-init-ewe1mn4x](https://stackblitz.com/edit/tanstack-db-init-ewe1mn4x?file=src%2FApp.tsx) Click the "Insert" button to see the issue:

* The collection without functional where queries continues displaying updates each second. ✓
* The collection with a functional where query throws errors which you can see in the console, and stops syncing. ✗

Image

## Logic

Both share the same simulated logic:

1. Insert two items simultaneously
2. Wait for server sync before returning from `onInsert`
3. The (simulated) server then sends updates to each item once per second. Only the collection that has a functional where query registered breaks.

I've had difficulty tracing the internal flow here, so any help/advice if I'm doing something wrong in how I'm calling the APId would be appreciated. 🙏

samwillis:
fixes #645 

The original reproduction from that issue updated to the preview package from this PR: https://stackblitz.com/edit/tanstack-db-init-jm98exso?file=src%2FApp.tsx

## Summary

The bug was in the `capturePreSyncVisibleState()` method in `state.ts`. Here's what was happening:

**The Race Condition:**
1. When the first transaction (T1) completed, `capturePreSyncVisibleState()` captured the visible state for key "0" (with optimistic data)
2. Then `recomputeOptimisticState()` cleared T1's optimistic state
3. `commitPendingTransactions()` couldn't run yet because T2 was still persisting
4. When T2 completed, `capturePreSyncVisibleState()` was called again and **cleared** all previous captures with `this.preSyncVisibleState.clear()`
5. At this point, key "0" was no longer in the optimistic state (cleared when T1 completed), so it wasn't recaptured
6. When `commitPendingTransactions()` finally ran, it had no `previousVisibleValue` for key "0", causing it to emit INSERT instead of UPDATE

**The Fix:**
I modified `capturePreSyncVisibleState()` to:
1. Remove the `this.preSyncVisibleState.clear()` call to preserve previously captured state
2. Add a check `if (!this.preSyncVisibleState.has(key))` before capturing to avoid overwriting earlier captures

This ensures that when transactions complete at different times (due to async delays in `onInsert`), each key's visible state is captured at the correct moment - when its corresponding transaction completes and before its optimistic state is cleared. The state is then preserved until all sync operations can be committed together.

### DIFF
--- a/.changeset/nasty-stars-take.md
+++ b/.changeset/nasty-stars-take.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+Fixed race condition which could result in a live query throwing and becoming stuck after multiple mutations complete asynchronously.


obeattie:
Just tried this in my codebase and it definitely fixes the issue I was seeing. You guys rock.

PR #654, opened 2025-10-07
stop pushing down where clauses that match a select projection in a subquery

samwillis:
we will want to revisit this in a future refactor

### DIFF
--- a/packages/db/src/query/optimizer.ts
+++ b/packages/db/src/query/optimizer.ts
@@ -782,14 +782,18 @@ function optimizeFromWithTracking(
     return new QueryRefClass(subQuery, from.alias)
   }
 
-  // Must be queryRef due to type system
-
   // SAFETY CHECK: Only check safety when pushing WHERE clauses into existing subqueries
   // We need to be careful about pushing WHERE clauses into subqueries that already have
   // aggregates, HAVING, or ORDER BY + LIMIT since that could change their semantics
   if (!isSafeToPushIntoExistingSubquery(from.query, whereClause, from.alias)) {
     // Return a copy without optimization to maintain immutability
     // Do NOT mark as optimized since we didn't actually optimize it
+    return new QueryRefClass(deepCopyQuery(from.query), from.alias)
+  }
+
+  // Skip pushdown when a clause references a field that only exists via a renamed
+  // projection inside the subquery; leaving it outside preserves the alias mapping.
+  if (referencesAliasWithRemappedSelect(from.query, whereClause, from.alias)) {
     return new QueryRefClass(deepCopyQuery(from.query), from.alias)
   }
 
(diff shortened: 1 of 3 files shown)


packages/db/src/query/optimizer.ts
```diff
@@ -943,6 +947,70 @@ function whereReferencesComputedSelectFields(
...
+ * TODO: in future we should handle this by rewriting the clause to use the subquery's
+ * internal field references, but it likely needs a wider refactor to do cleanly.
+ */
+function referencesAliasWithRemappedSelect(
+  subquery: QueryIR,
+  whereClause: BasicExpression<boolean>,
+  outerAlias: string
+): boolean {
+  // Only care about clauses that actually reference the outer alias.
+  if (!collectRefs(whereClause).some((ref) => ref.path[0] === outerAlias)) {
```
kevin-dp: nit: usually easier to read positive predicates instead of negated predicates:
```ts
if (collectRefs(whereClause).every((ref) => ref.path[0] !== outerAlias))
```

packages/db/src/query/optimizer.ts
```diff
@@ -943,6 +947,70 @@ function whereReferencesComputedSelectFields(
...
+    return true
+  }
+
+  const select = subquery.select
+  // Without an explicit SELECT the clause still refers to the original collection.
+  if (!select) {
+    return false
+  }
+
+  for (const ref of collectRefs(whereClause)) {
```
kevin-dp: We also called collectRefs on L963. Worth introducing a constant on L962: `const refs = collectRefs(whereClause)` and use that constant on L963 and here

packages/db/src/query/optimizer.ts
```diff
@@ -943,6 +947,70 @@ function whereReferencesComputedSelectFields(
...
+    const path = ref.path
+    // Need at least alias + field to matter.
+    if (path.length < 2) continue
+    if (path[0] !== outerAlias) continue
+
+    const projected = select[path[1]!]
+    // Unselected fields mean the alias was reshaped; treat as unsafe.
```
kevin-dp: Does this comment actually match the line below it?
If `projected` is false we `continue`, so if we `continue` we treat it as safe but the comment is saying we treat it as unsafe. Or am i mixing this up here?

packages/db/src/query/optimizer.ts
```diff
@@ -943,6 +947,70 @@ function whereReferencesComputedSelectFields(
...
+    if (!projected) continue
+
+    // Non-PropRef projections are computed values; cannot push down.
+    if (!(projected instanceof PropRef)) {
+      return true
+    }
+
+    // Require the projection to carry both alias and field info for comparison.
+    if (projected.path.length < 2) {
```
kevin-dp, line 996: Not sure about this one. If the length is smaller than 2, it doesn't have an alias and field and so we mark it as unsafe. Is that correct? I'm not sure what it means that the projection doesn't have an alias and field.

PR #655, opened 2025-10-07
add collection id to important logs

marbemac:
Follow-up to https://github.com/TanStack/db/pull/641#issuecomment-3378787317

Prefixes some of the more important logs & errors with collection `id` if it is available. This helped me a lot when trying to trace what was going on with txids.

The lint command isn't working locally for me for whatever reason, so we'll see if the GH workflows pass.

Screenshot 2025-10-07 at 4 17 33 PM


KyleAMathews (approved):
Very nice! Thanks!

KyleAMathews:
Some tests need updated

PR #658, opened 2025-10-08
eagerly run live queries while collections are loading

samwillis:
### Summary
Refactors the live query execution model to support eager materialization. Live queries now process and display data as it arrives from source collections during sync, rather than waiting for all source collections to reach a "ready" state before executing.

### Key Changes
- **Eager Execution**: Live queries now run immediately when triggered via `.preload()` or `.startSync()`, processing data as it arrives from source collections
- **Separated Status & Data Management**: Status updates now use `on('status:change')` event listeners, while `subscribeChanges` is used solely for data updates
- **Error Propagation**: Live queries automatically transition to error state when source collections error or are manually cleaned up
- **Optimistic Transactions**: Local optimistic mutations are immediately visible in live queries even while source collections are still in "loading" state, with proper rollback behavior when mutations aren't persisted via sync
- **Lifecycle Simplification**: Removed `initialCommit` status from collection lifecycle
- **Auto-indexing**: Enabled auto-indexing during `loading` state to support eager query execution
- **Public API**: Made `_lifecycle` property public on Collection interface for internal library use

### Testing
- Added comprehensive test suite for eager execution in `query-while-syncing.test.ts` (24 tests covering basic queries, joins, multiple sources, error handling, and both `.startSync()` and `.preload()` patterns)
- Added test validating optimistic transaction visibility and rollback behavior during loading state
- Added tests validating multiple sync transactions before `markReady()` in collection tests
- Refactored tests to use modern vitest `describe.each` pattern for parameterized testing
- All tests use fake timers for faster, more deterministic execution

### User-Facing Impact
Users will see live query results update in real-time as data loads, providing immediate feedback during sync operations. This includes both synced data and local optimistic updates, creating a more responsive user experience even before the underlying data is fully loaded. They can use the collection status to show loading state while the data streams in.

### DIFF
--- a/.changeset/ready-bats-call.md
+++ b/.changeset/ready-bats-call.md
@@ -1,0 +1,6 @@
+---
+"@tanstack/react-db": patch
+"@tanstack/db": patch
+---
+
+Refactored live queries to execute eagerly during sync. Live queries now materialize their results immediately as data arrives from source collections, even while those collections are still in a "loading" state, rather than waiting for all sources to be "ready" before executing.
--- a/packages/db/src/collection/index.ts
+++ b/packages/db/src/collection/index.ts
@@ -217,7 +217,7 @@ export class CollectionImpl<
   // Managers
   private _events: CollectionEventsManager
   private _changes: CollectionChangesManager<TOutput, TKey, TSchema, TInput>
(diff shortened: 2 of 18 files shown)


marbemac:
❤️  works like a charm - closing #641

PR #660, opened 2025-10-08
Add getOptimisticInfo() method to track optimistic state per record

### LINKED ISSUE #20: Track Optimistic Status Per Record for Local-Only Query Differentiation
## Problem

Currently, there is no explicit way to distinguish between records that exist optimistically on the client and those that have been confirmed by the server.  
This can cause ambiguity in UI rendering, state reconciliation, and user feedback flows.

## Proposed Solution

Introduce a system-reserved field `__optimistic` on each record.

- When a record is created optimistically, set `__optimistic: true`.
- Upon successful server confirmation, set `__optimistic: false`.
- Records without an explicit `__optimistic` flag or with `__optimistic: false` are treated as fully server-validated.

Example record shape:

```typescript
interface Record {
  id: string;
  fields: Record<string, any>;
  __optimistic: boolean;
}

### PR DESCRIPTION
**Pull request by KyleAMathews**
## Summary

Implements `collection.getOptimisticInfo(id)` to provide rich metadata about optimistic state for a given record.

**Returns:**
- `isOptimistic`: boolean flag indicating if record has active mutations
- `original`: pre-mutation state (for updates)
- `modified`: current state including optimistic changes  
- `changes`: delta of modifications (for updates)
- `mutations`: array of all active mutations affecting the record

**Usage:**
```typescript
const info = collection.getOptimisticInfo(todoId)
if (info?.isOptimistic) {
  return <Badge>Saving...</Badge>
}

// Or for diff views:
if (info?.changes) {
  showDiff(info.original, info.modified, info.changes)
}
```

This provides everything needed for UI states like loading badges, diff views, and rollback UI while keeping records pure (no `__optimistic` field polluting user data).

Closes #20

## Test plan
- ✅ Comprehensive test coverage (12 new tests)
- ✅ All existing tests pass (69 test files, 1289 tests)

### DIFF
--- a/.changeset/cool-cups-argue.md
+++ b/.changeset/cool-cups-argue.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+Add `getOptimisticInfo()` method to track optimistic state per record. Returns metadata including `isOptimistic` flag, original/modified states, changes delta, and active mutations array for building UI features like loading badges and diff views.


KyleAMathews:
Closing this for now — @samwillis pointed out that using this in a component is problematic as the component doesn't necessarily re-render when a record goes from optimistic to synced so there needs to be a way to query a `$status` field on the record (or some similar design) so it's properly reactive.

This isn't a priority atm so we'll pick this up later.

Issue #630, opened 2025-10-02
Deleting RxDB rows does not persist to localStorage


daveycodez:
I just made a very simple example where I am using only RxDB with TanStackDB, I'm able to add rows just fine. The only issue is when you delete a row, it deletes properly, but if you refresh the page, the row comes back, and it is impossible to delete it. This only occurs with localStorage, not memory adapter.

This must be an issue with creating an rxdbCollection, since it breaks as soon as you refresh the page and it brings back old deleted rows from RxDB. Maybe related to not correctly filtering out _deleted: true rows?

daveycodez:
@kevin-dp 

initialFetch() the query selector needs to have _deleted: false

```tsx
let query: FilledMangoQuery<Row>
if (cursor) {
    query = {
        selector: {
            $or: [
                { "_meta.lwt": { $gt: cursor._meta.lwt } },
                {
                    "_meta.lwt": cursor._meta.lwt,
                    [primaryPath]: {
                        $gt: getKey(cursor),
                    },
                },
            ],
            _deleted: false,
        } as any,
        sort: [
            { "_meta.lwt": `asc` },
            { [primaryPath]: `asc` } as any,
        ],
        limit: syncBatchSize,
        skip: 0,
    }
} else {
    query = {
        selector: {
            _deleted: false,
        },
        sort: [
            { "_meta.lwt": `asc` },
            { [primaryPath]: `asc` } as any,
        ],
        limit: syncBatchSize,
        skip: 0,
    }
}
```

daveycodez:
The other option is to address this ->

```ts
/**
 * Instead of doing a RxCollection.query(),
 * we directly query the storage engine of the RxCollection so we do not use the
 * RxCollection document cache because it likely wont be used anyway
 * since most queries will run directly on the tanstack-db side.
 */
```


If you skip RxCollection.find().exec() and use storageInstance directly, it will skip all the checks for deleted documents.


So another fix would be instead of checking for _deleted in the selector to do this:

```tsx
const docs = await rxCollection
  .find(preparedQuery.query)
  .exec()

cursor = lastOfArray(docs)
if (docs.length === 0) {
    ready = true
    break
}

docs.forEach((d) => {
    write({
        type: `insert`,
        value: stripRxdbFields(clone(d.toJSON())),
    })
})
```

I probably prefer this method @KyleAMathews

daveycodez:
I also cleaned up 99% of the "any" types.


rxdb.ts
```tsx
[...]

PR #667, opened 2025-10-10
rxdb-db-collection fixes

daveycodez:
This PR addresses the majority of "any" type usage in the rxdb-db-collection adapter as well as fixes for only querying documents where _deleted is false, which fixes issues when using localStorage or other persistent storage mechanisms.

Fixes https://github.com/TanStack/db/issues/630

### DIFF
--- a/.changeset/smooth-candies-visit.md
+++ b/.changeset/smooth-candies-visit.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/rxdb-db-collection": patch
+---
+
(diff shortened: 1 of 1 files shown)


daveycodez:
> Thanks for this @daveycodez
> 
> Would it be possible to add a test that validates the delete behaviour?

Seems like a complex test that would need to run in the web browser and then refresh the page to verify. It requires access to localStorage. 

The issue was that it loads the data and skips rxdb query and runs directly on the storage instance which skips rxdb filtering out deleted documents.

PR #668, opened 2025-10-11
Predicate Comparison and Merging Utilities for Predicate Push-Down

samwillis:
stacked on #669

## Summary

Implements utilities for comparing and merging predicates (`where` clauses, `orderBy`, and `limit`) to support predicate push-down in collection sync operations. Provides a complete solution for tracking loaded data and preventing redundant server requests.

**Key Features:**
- ✅ Logical subset checking for `where` clauses (AND, OR, comparisons, IN)
- ✅ Smart predicate merging with intersection (AND) and union (OR) semantics
- ✅ **Predicate difference** - Compute `A AND NOT(B)` with simplification
- ✅ **Automatic deduplication wrapper** - Eliminates redundant data fetches
- ✅ Complete `Date` object support (equality, ranges, IN clauses)
- ✅ Contradiction detection (returns `false` literal for impossible predicates)
- ✅ **Performance optimized** for large primitive IN predicates (100-1250x speedup via Set-based lookups)
- ✅ **149 comprehensive tests** with extensive coverage

## Motivation

The `onLoadMore` callback needs to:
1. **Check if data is already loaded** - Determine if a new predicate is covered by previously loaded predicates
2. **Track total coverage** - Merge multiple load operations to understand complete data coverage
3. **Prevent redundant fetches** - Automatically deduplicate concurrent and sequential requests

Without these utilities, the sync layer cannot efficiently track loaded data ranges or prevent duplicate network requests.

### Core Predicate Functions

**Where Clause Operations:**
- `isWhereSubset(subset, superset)` - Checks if one where clause logically implies another
- `intersectWherePredicates(predicates)` - Combines predicates with AND logic (most restrictive)
- `unionWherePredicates(predicates)` - Combines predicates with OR logic (least restrictive)
- `minusWherePredicates(from, subtract)` - Computes `from AND NOT(subtract)` with simplification

**OrderBy & Limit Operations:**
- `isOrderBySubset(subset, superset)` - Validates ordering requirements via prefix matching
- `isLimitSubset(subset, superset)` - Compares limit constraints

**Complete Predicate Operations:**
- `isPredicateSubset(subset, superset)` - Checks all components (where + orderBy + limit)
- `intersectPredicates(predicates)` - Merges predicates with intersection semantics
- `unionPredicates(predicates)` - Merges predicates with union semantics

### DeduplicatedLoadSubset Class

A production-ready wrapper that automatically deduplicates `loadSubset` calls:

**Features:**
- **Smart subset detection** - Uses predicate logic to avoid redundant fetches
- **In-flight request sharing** - Concurrent identical/subset requests share the same promise
- **Separate tracking** - Handles unlimited vs limited queries differently
- **Reset support** - Clear state with generation counter to prevent repopulation
- **Auto-bound methods** - Safe to pass as callbacks without binding issues

**How it works:**
- Tracks all unlimited queries in a combined `where` predicate via union
- Tracks limited queries (with `orderBy`/`limit`) separately for exact matching
- Checks incoming requests against tracked state using subset logic
- Shares in-flight requests when new requests are subsets of pending ones

### Examples

```typescript
// Subset checking
isWhereSubset(
  gt(ref('age'), val(20)),     // age > 20
  gt(ref('age'), val(10))      // age > 10
) // → true (20 > 10, so more restrictive)

// Intersection (AND logic)
intersectWherePredicates([
  gt(ref('age'), val(10)),
  lt(ref('age'), val(50))
]) // → age > 10 AND age < 50

// Union (OR logic)
unionWherePredicates([
  eq(ref('age'), val(5)),
  eq(ref('age'), val(10))
]) // → age IN [5, 10]

// Predicate difference
minusWherePredicates(
  gt(ref('age'), val(10)),     // Requested: age > 10
  gt(ref('age'), val(20))      // Already loaded: age > 20
) // → age > 10 AND age <= 20 (simplified)

// Contradiction detection
intersectWherePredicates([
  eq(ref('age'), val(5)),
  eq(ref('age'), val(6))
]) // → {type: 'val', value: false}

// Automatic deduplication
const dedupe = new DeduplicatedLoadSubset(myLoadSubset)

// First call - fetches data
await dedupe.loadSubset({ where: gt(ref('age'), val(10)) })

// Second call - returns true immediately (subset of first)
await dedupe.loadSubset({ where: gt(ref('age'), val(20)) })

// Reset state when data store is cleared
dedupe.reset()
```

### Logical Subset Checking

Uses recursive descent to check logical implications:

```typescript
// AND handling
(A AND B) ⊆ C  if  (A ⊆ C) OR (B ⊆ C)

// OR handling
(A OR B) ⊆ C  if  (A ⊆ C) AND (B ⊆ C)
```

### Range Simplification

**Intersection:** Takes most restrictive constraints
- `age > 10 AND age > 20` → `age > 20`
- `age = 5 AND age = 6` → `false` literal
- `age IN [1,2] AND age IN [2,3]` → `age IN [2]`

**Union:** Takes least restrictive constraints
- `age > 10 OR age > 20` → `age > 10`
- `age = 5 OR age = 10` → `age IN [5, 10]`

**Difference:** Simplifies same-field predicates
- `age > 10 MINUS age > 20` → `age > 10 AND age <= 20`
- `age IN [1,2,3] MINUS age IN [2,4]` → `age IN [1,3]`

### Value Type Support

**✅ Supported:**
- Primitives: strings, numbers, booleans, null, undefined
- Date objects: equality, ranges, IN clauses (compared by timestamp)

**❌ Not Supported:**
- Arrays/objects

### Performance Optimizations

For large primitive IN predicates (>10 elements):
- **Smart Set Construction** - Builds Sets once and reuses them
- **O(1) Lookups** - Uses `Set.has()` instead of array scans
- **Cached Metadata** - Stores `areAllPrimitives` and `primitiveSet` on extraction
- **Pre-simplified IN values** - Removes duplicates when building primitive sets

| Operation | Without | With | Speedup |
|-----------|---------|------|---------|
| `eq = X` vs `IN [1000 items]` | O(1000) scan | O(1) lookup | ~1000x |
| `IN [100]` ⊆ `IN [10000]` | O(1M) comparisons | O(10,100) ops | ~100x |
| Intersect 3 `IN [5000]` clauses | O(75M) comparisons | O(60K) ops | ~1250x |

### Deduplication Architecture

**State Tracking:**
- `unlimitedWhere` - Combined OR of all unlimited predicates
- `limitedCalls[]` - Array of all limited queries for exact matching
- `inflightCalls[]` - Active requests with their predicates
- `generation` - Counter to invalidate stale in-flight handlers after reset

**Request Flow:**
1. Check if data already loaded (via `isPredicateSubset`)
2. Check if in-flight request covers this (via subset logic)
3. If not covered, make request and track it
4. On completion, update tracking state (unless reset was called)

## What This Covers

✅ All operators supported by collection index system: `eq`, `gt`, `gte`, `lt`, `lte`, `in`, `and`, `or`  
✅ Date object support (equality, ranges, IN clauses)  
✅ Conflict detection (contradictory equalities, empty IN intersections)  
✅ Predicate difference for incremental loading  
✅ Production-ready deduplication wrapper  
✅ Concurrent request handling with subset matching  
✅ State reset with generation counter safety  
✅ **149 tests** covering edge cases, Date handling, performance optimizations, and deduplication  

## What This Does NOT Cover

❌ **Range contradiction detection** - `age > 20 AND age < 10` is preserved as-is (could detect and return `false`)  
❌ **Property-to-property comparisons** - Assumes pattern: `field op value`  
❌ **Advanced OR simplification** - Complex nested OR/AND kept as-is for safety  
❌ **NOT operator** - Not supported by collection index system  
❌ **State persistence** - DeduplicatedLoadSubset is in-memory only (persistence hooks planned for future)

**Why Conservative?** Correctness over optimization—false negatives (missed optimizations) are better than false positives (incorrect results).

## Files Changed

**New Files:**
- `packages/db/src/query/predicate-utils.ts` (1,544 lines) - 10 exported functions with JSDoc
- `packages/db/src/query/subset-dedupe.ts` (244 lines) - DeduplicatedLoadSubset class
- `packages/db/tests/predicate-utils.test.ts` (1,342 lines) - 130 tests
- `packages/db/tests/subset-dedupe.test.ts` (326 lines) - 19 tests

**Modified Files:**
- `packages/db/src/query/index.ts` - Export new utilities and DeduplicatedLoadSubset class

### Basic Predicate Operations

```typescript
import { isPredicateSubset, intersectPredicates } from '@tanstack/db'

const loadedPredicates: LoadSubsetOptions[] = []

function onLoadMore(requested: LoadSubsetOptions) {
  // Check if already loaded
  const alreadyLoaded = loadedPredicates.some(loaded =>
    isPredicateSubset(requested, loaded)
  )
  
  if (alreadyLoaded) {
    console.log('Data already loaded, using cache')
    return true
  }
  
  // Fetch from server and track
  await fetchFromServer(requested)
  loadedPredicates.push(requested)
  
  // Compute total coverage
  const totalCoverage = intersectPredicates(loadedPredicates)
  
  // Check for contradictions
  if (totalCoverage.where?.type === 'val' && 
      (totalCoverage.where as any).value === false) {
    console.warn('Contradictory predicates detected!')
  }
}
```

### Automatic Deduplication

```typescript
import { DeduplicatedLoadSubset } from '@tanstack/db'

// Wrap your loadSubset function
const dedupe = new DeduplicatedLoadSubset(
  async (options: LoadSubsetOptions) => {
    const data = await fetchFromServer(options)
    updateLocalCache(data)
  }
)

// Use in sync config - sync is a function that returns an object
export const myCollection = db.collection({
  name: 'users',
  sync: () => ({
    // Pass the auto-bound method directly
    loadSubset: dedupe.loadSubset,
  })
})

// Concurrent requests automatically deduplicated:
await Promise.all([
  dedupe.loadSubset({ where: gt(ref('age'), val(10)) }), // Fetches
  dedupe.loadSubset({ where: gt(ref('age'), val(20)) }), // Waits for first
  dedupe.loadSubset({ where: gt(ref('age'), val(30)) }), // Waits for first
]) // Only one network request made!

// Clear state when data store is reset
function clearAllData() {
  clearLocalCache()
  dedupe.reset() // Clear deduplication state
}
```

### Computing Incremental Loads

```typescript
import { minusWherePredicates } from '@tanstack/db'

const alreadyLoaded = gt(ref('age'), val(20)) // age > 20
const requested = gt(ref('age'), val(10))      // age > 10

const stillNeeded = minusWherePredicates(requested, alreadyLoaded)
// Result: age > 10 AND age <= 20

if (stillNeeded.type === 'val' && stillNeeded.value === false) {
  console.log('All requested data already loaded!')
} else {
  await fetchFromServer({ where: stillNeeded })
}
```

## Testing

**149 tests passing:**

**predicate-utils.test.ts (130 tests):**
- Basic subset comparisons (5 tests)
- Comparison operators (12 tests)
- IN operator edge cases (4 tests)
- AND/OR combinations (9 tests)
- Date support (12 tests)
- Conflict detection (6 tests)
- Range simplifications (14 tests)
- OrderBy/Limit (11 tests)
- Complete predicate operations (28 tests)
- Predicate difference operations (29 tests)

**subset-dedupe.test.ts (19 tests):**
- Basic deduplication (10 tests)
- Concurrent request handling (3 tests)
- Options mutation protection (2 tests)
- Failed request retry (1 test)
- Reset behavior (2 tests)
- Unbound callback safety (1 test)

## Type Safety

- No `null` returns where `BasicExpression<boolean>` is expected
- Empty sets represented as concrete false literals: `{type: 'val', value: false}`
- Proper handling of `undefined` vs constrained predicates
- Auto-bound methods for safe callback usage without `this` binding issues
- All type assertions validated by TypeScript strict mode

## Breaking Changes

None - purely additive functionality.

### DIFF
--- a/packages/db/src/query/predicate-utils.ts
+++ b/packages/db/src/query/predicate-utils.ts
@@ -769,14 +769,15 @@ export function isLimitSubset(
   subset: number | undefined,
   superset: number | undefined
 ): boolean {
-  // No limit requirement is always satisfied
-  if (subset === undefined) {
-    return true
-  }
-
   // Unlimited superset satisfies any limit requirement
   if (superset === undefined) {
     return true
+  }
+
+  // If requesting all data (no limit), we need unlimited data to satisfy it
+  // But we know superset is not unlimited so we return false
+  if (subset === undefined) {
+    return false
   }
 
   // Otherwise, subset must be less than or equal to superset
--- a/packages/db/tests/predicate-utils.test.ts
+++ b/packages/db/tests/predicate-utils.test.ts
@@ -621,8 +621,12 @@ describe(`isOrderBySubset`, () => {
 })
 
 describe(`isLimitSubset`, () => {
-  it(`should return true for undefined subset`, () => {
-    expect(isLimitSubset(undefined, 10)).toBe(true)
+  it(`should return false for undefined subset with limited superset (requesting all data but only have limited)`, () => {
+    expect(isLimitSubset(undefined, 10)).toBe(false)
+  })
+
+  it(`should return true for undefined subset with undefined superset (requesting all data and have all data)`, () => {
+    expect(isLimitSubset(undefined, undefined)).toBe(true)
   })
 
   it(`should return true for undefined superset`, () => {


samwillis:
Something I considered while implementing this was to normalise the predicates into a [DNF](https://en.wikipedia.org/wiki/Disjunctive_normal_form) form, but it would potentially explode the size of the predicates, and in a real world situation the predicates are simple and repetitive in structure so I think this implementation makes the right tradeoff.

.changeset/light-phones-flash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+Add predicate comparison and merging utilities (isWhereSubset, intersectWherePredicates, unionWherePredicates, and related functions) to support predicate push-down in collection sync operations, enabling efficient tracking of loaded data ranges and preventing redundant server requests. Includes performance optimizations for large primitive IN predicates (100-1250x speedup via Set-based lookups) and full support for Date objects in equality, range, and IN clause comparisons.
```
kevin-dp: > Includes performance optimizations for large primitive IN predicates (100-1250x speedup via Set-based lookups

This wording is a bit misleading. I understand it comes from the described speedup in the PR description but that speedup section is off. It essentially describes linear O(n) behavior vs constant-time O(1) lookup. We could equally say a speedup of 10 billion if your list would have contained 10 billion elements vs a constant-time lookup. But that's not really a correct way to express speedups as the speedup is arbitrary (depends on the size of the array since it's linear...).

PS: In Big Oh analysis, O(any_constant) = O(1)  so O(1_000_000) = O(1). What matters is the relation to the size of the input: O(1), O(n), O(n^2), etc.

packages/db/src/query/predicate-utils.ts
```diff
@@ -0,0 +1,1356 @@
...
+  }
+
+  // Handle subset being an AND: (A AND B) implies both A and B
+  if (subset.type === `func` && subset.name === `and`) {
+    // Special case: if superset is also AND, check if all conjuncts in superset are implied by subset
+    if (superset.type === `func` && superset.name === `and`) {
+      // For (A AND B) ⊆ (C AND D), we need every conjunct in superset to be implied by subset
+      // For each conjunct in superset, at least one conjunct in subset must be a subset of it
+      // OR the entire subset implies it
+      return superset.args.every((superArg) => {
```
kevin-dp: Why do we need to handle the case where both the subset and superset are AND clauses explicitly?
I would expect to handle the case where subset is an AND by doing:
```ts
return subset.args.some((subArg) => isWhereSubsetInternal(subArg, superset)
```
And then there would be a separate (distinct) case that checks if `superset` is and AND clause like you already have on L80. That should work just fine because when you have an AND clause in both subset and superset, it will first be handled by the case for the subset which splits the AND clause into N calls to isWhereSubsetInternal (1 call per conjunct of the subset). And then in each call the subset is no longer an AND clause but the superset is so that is handled by the case on L80.

kevin-dp: So in this case we only need L72-74

kevin-dp: Note how you don't special case the cause where both subset and superset are OR clauses so that confirms that for ANDs you should not have to do that either.

packages/db/src/query/predicate-utils.ts
```diff
@@ -0,0 +1,1356 @@
...
+      return isComparisonSubset(
+        subsetFunc,
+        subsetField.value,
+        supersetFunc,
+        supersetField.value
+      )
+    }
+
+    // Handle eq vs in
+    if (subsetFunc.name === `eq` && supersetFunc.name === `in`) {
```
kevin-dp, line 154: We don't handle the case where subset is IN (with no elements or a single element) and superset is EQ. Those are corner cases but could also occur e.g. `age IN [ 18 ]` is a subset of `age = 18` because they are actually the same.

Also, we don't handle subset that is IN/EQ vs superset that is a range (i.e. comparison operators, e.g. >= 18). Should be fairly straightforward to determine that `age = 18` is a subset of `age >= 18` and `age IN [ 18, 19, 20]` is a subset of `age >= 18`. For IN you just need to recursively check each element of the array against the superset.

kevin-dp: Again, i don't think we should explicitly handle cases where subset is EQ and superset is IN, and cases where both subset and superset use IN. Those should be treated individually. So, if subset is an `IN` then we should just recursively call `isWhereSubsetInternal` over the elements of the array:
```ts
inElements.every(elem => isWhereSubsetInternal(elem, superset))
```
Again, that will lead to 1 call per element that is part of the IN clause. In fact, a subset where `a IN [ x, y , z]` is equivalent to `a = x OR a = y OR a = z` so perhaps we should not handle IN specially at all and just have 1 case that handles both OR and IN (in the subset). Similarly, for `IN` in the superset (treat it like OR in superset).

packages/db/src/query/predicate-utils.ts
```diff
@@ -0,0 +1,1356 @@
...
+ * unionWherePredicates([gt(ref('age'), val(10)), gt(ref('age'), val(20))]) // age > 10
+ *
+ * @example
+ * // Combine equals into IN
+ * unionWherePredicates([eq(ref('age'), val(5)), eq(ref('age'), val(10))]) // age IN [5, 10]
+ *
+ * @param predicates - Array of where predicates to union
+ * @returns Combined predicate representing the union
+ */
+export function unionWherePredicates(
```
kevin-dp, line 295: This function is almost identical to `intersectWherePredicates`. Let's extract the common pattern to a helper function. That will simplify code understanding, evolution, maintainability, reviewing, etc.

packages/db/src/query/predicate-utils.ts
```diff
@@ -0,0 +1,1356 @@
...
+ * @param subset - The limit requirement to check
+ * @param superset - The limit that might satisfy the requirement
+ * @returns true if subset is satisfied by superset
+ */
+export function isLimitSubset(
+  subset: number | undefined,
+  superset: number | undefined
+): boolean {
+  // No limit requirement is always satisfied
+  if (subset === undefined) {
```
kevin-dp: If the subset has no limit but the superset has a limit, then it should return `false` iiuc but this returns `true`. I think this function should be:
```ts
return superset === undefined || subset <= superset
```

packages/db/src/query/predicate-utils.ts
```diff
@@ -0,0 +1,1356 @@
...
+  // If any predicate is unlimited, the intersection is limited by the others
+  const limits = predicates
+    .map((p) => p.limit)
+    .filter((l): l is number => l !== undefined)
+  const mergedLimit =
+    limits.length === 0
+      ? undefined // All unlimited = result unlimited
+      : Math.min(...limits) // Take most restrictive
+
+  return {
```
kevin-dp: I'm not sure about these semantics, whether this is really what we expect. Let me explain with an example. Imagine 2 queries:
```sql
-- Query 1
WHERE age >= 18 LIMIT 1
-- Query 2
WHERE age >= 20 LIMIT 3
```
Imagine we have 3 users: Alice aged 18, Bob aged 19, Charlie aged 20. Based on this data query 1 should return 1 user: Alice. And query 2 should return also 1 user: Charlie.
So:
```
Query 1: { Alice }
Query 2: { Charlie }
Intersection of Query 1 and Query 2: { } (empty)
```
But if we look at `intersectPredicates` it will generate this predicate:
```sql
WHERE age >= 20 LIMIT 1
```
It takes age >= 20 because that's the most restrictive clause and it takes limit 1 because that's also the most restrictive. Now based on the data from our example, this intersected query will return 1 user: Charlie. So, the result of this intersected query is different from the intersection of the results... Don't think that we want those semantics?

packages/db/src/query/predicate-utils.ts
```diff
@@ -0,0 +1,1356 @@
...
+  }
+}
+
+/**
+ * Merge multiple predicates by unioning their where clauses.
+ *
+ * @param predicates - Array of predicates to merge
+ * @returns Combined predicate
+ */
+export function unionPredicates(
```
kevin-dp: Again, very similar to the `intersectPredicates`. Let's extract the common flow into a helper function.

samwillis:
We need to track the disjoint loaded sets by the upper and lower bound in columns (exclusive of upper bound when loaded with a limit), then use the indexes to check that within those bounds we have enough rows to answer the query without going to the server.

Marking this as draft as it needs rework. The where expression utils are good, anything using a limit needs changing.

packages/db/src/query/subset-dedupe.ts
```diff
@@ -0,0 +1,243 @@
...
+import type { BasicExpression } from "./ir.js"
+import type { LoadSubsetOptions } from "../types.js"
+
+/**
+ * Deduplicated wrapper for a loadSubset function.
+ * Tracks what data has been loaded and avoids redundant calls by applying
+ * subset logic to predicates.
+ *
+ * @example
+ * const dedupe = new DeduplicatedLoadSubset(myLoadSubset)
```
kevin-dp, line 16: It does feel a bit inconsistent that we have to instantiate a class whereas loadSubset would just be a function.
If we want we can create a function that hides this:
```ts
function dedup(loadSubset) {
  return new DeduplicatedLoadSubset(loadSubset)
}
```

KyleAMathews:
These utilities add a fair bit of code — could we lazy load them so they don't get loaded/parsed until needed?

PR #669, opened 2025-10-12
Add loadSubset State Tracking and On-Demand Sync Mode

samwillis:
## Overview

This PR adds comprehensive loading status tracking to collections and live queries, enabling UIs to display loading indicators when more data is being fetched. It also introduces a new `syncMode` configuration option to control when collections load data - either eagerly on initial sync or on-demand as queries request it.

## Problem

1. **No Loading Indicators**: When live queries pushed predicates down to source collections via `syncMore`, there was no indication that data was being loaded. This made it impossible to show proper loading states in the UI.

2. **Always-Eager Loading**: Collections always loaded all data immediately during initial sync, even when using predicate pushdown. There was no way to configure collections to only load data as it was requested by queries.

## Solution

This PR implements a multi-layered loading state tracking system with configurable sync modes:

### Loading State Tracking

1. **Collection Subscriptions**: Track their own loading state when requesting snapshots, with a public `Subscription` interface for external consumers
2. **Collections**: Track pending load promises and expose an `isLoadingSubset` property
3. **Live Queries**: Automatically reflect the loading state of their source collection subscriptions

**Key Isolation Property**: Each live query maintains its own loading state based on *its own subscriptions*. When a live query triggers loading via predicate pushdown, only that live query's `isLoadingSubset` becomes `true`. Other live queries that share the same source collection but don't need that specific snapshot remain unaffected.

### Sync Modes

Collections can now be configured with two sync modes:

- **`eager` (default)**: Loads all data immediately during initial sync. `loadSubset` calls are bypassed.
- **`on-demand`**: Only loads data as it's requested via `loadSubset` calls. Requires a `loadSubset` handler.

### 1. Renamed `syncMore`/`loadMore` to `loadSubset`

**Files:** `packages/db/src/collection/sync.ts`, `packages/db/src/types.ts`, `packages/db/src/collection/subscription.ts`

- Renamed `syncMore` → `loadSubset` (collection method)
- Renamed `onLoadMore` → `loadSubset` (sync config property)
- Renamed `OnLoadMoreOptions` → `LoadSubsetOptions` (type)
- Renamed `syncOnLoadMoreFn` → `syncLoadSubsetFn` (internal)
- Renamed `pendingLoadMorePromises` → `pendingLoadSubsetPromises` (internal)

**Rationale**: "loadSubset" better reflects that this operation loads a filtered/limited subset of data, not just "more" data.

### 2. Added `syncMode` Configuration

**Files:** `packages/db/src/types.ts`, `packages/db/src/collection/sync.ts`

New `syncMode` property on collection config:

```typescript
type SyncMode = 'eager' | 'on-demand'

interface CollectionConfig {
  syncMode?: SyncMode // defaults to 'eager'
  // ... other properties
}
```

**Behavior:**
- **`eager` (default)**: `loadSubset` is bypassed and returns `undefined`. All data is loaded during initial sync.
- **`on-demand`**: `loadSubset` calls proceed normally. Data is only loaded as requested.

**Validation:**
- If `syncMode: 'on-demand'` is set but no `loadSubset` handler is provided, throws `CollectionConfigurationError` with a helpful message
- Validation occurs when the sync function is first called

### 3. Standardized `loadSubset` Return Type

**Files:** `packages/db/src/collection/sync.ts`, `packages/db/src/types.ts`

- Changed to consistently return `Promise<void> | undefined`
- Returns `undefined` when `syncMode` is `'eager'` or no sync implementation is configured
- Wraps synchronous `loadSubset` results in `Promise.resolve()`
- Updated `SyncConfigRes['loadSubset']` type signature

### 4. CollectionSubscription Status Tracking & Events

**File:** `packages/db/src/collection/subscription.ts`

Added comprehensive status tracking:

- **Status Property**: `status: 'ready' | 'loadingSubset'` (readonly getter with private mutable field)
- **Concurrent Tracking**: `pendingLoadSubsetPromises: Set<Promise<void>>`
- **Events**: 
  - `status:change` - Emitted when status transitions
  - `status:ready` - Emitted when entering ready state
  - `status:loadingSubset` - Emitted when entering loading state
  - `unsubscribed` - New event emitted when subscription is destroyed
- **Error Handling**: Status returns to `ready` even on promise rejection
- **Cleanup**: `unsubscribe()` emits `unsubscribed` event *before* clearing listeners

### 5. Generic `isLoadingSubset` for All Collections

**Files:** `packages/db/src/collection/sync.ts`, `packages/db/src/collection/index.ts`, `packages/db/src/collection/events.ts`

All collections now track their loading state:

- **Property**: `isLoadingSubset` (boolean getter, not a method)
- **Tracking**: `pendingLoadSubsetPromises: Set<Promise<void>>` in `CollectionSyncManager`
- **Events**: `loadingSubset:change` event when state transitions
- **API**: `trackLoadPromise(promise: Promise<void>)` method for internal coordination
- **Access**: Made `_sync` public on `Collection` for internal use

### 6. Live Query Integration

**Files:** `packages/db/src/query/live/collection-subscriber.ts`, `packages/db/src/query/live/collection-config-builder.ts`

Live queries reflect loading state from their subscriptions:

- `CollectionSubscriber` subscribes to subscription `status:change` events
- Creates deferred promises when subscriptions enter `loadingSubset` state
- Passes promises to result collection via `liveQueryCollection.trackLoadPromise()`
- Result collection's `isLoadingSubset` reflects subscription loading states
- Proper cleanup on unsubscribe (resolves pending promises)

**Loading State Isolation**: Each live query's subscriptions are independent. Query A triggering `loadSubset` doesn't affect Query B's `isLoadingSubset` status.

### 7. Subscription Interface for External Consumers

**File:** `packages/db/src/types.ts`

New public `Subscription` interface:

```typescript
export interface Subscription extends EventEmitter<SubscriptionEvents> {
  readonly status: SubscriptionStatus
}

export type LoadSubsetOptions = {
  where?: BasicExpression<boolean>
  orderBy?: OrderBy
  limit?: number
  subscription?: Subscription // Optional, for sync implementation use
}
```

**Purpose**: Allows sync implementations to:
- Track which subscription triggered a `loadSubset` call
- Subscribe to subscription lifecycle events (e.g., `unsubscribed`)
- Implement advanced caching/ref-counting based on subscription lifecycle

### 8. Reusable Event Emitter

**File:** `packages/db/src/event-emitter.ts` (new)

Extracted event emitter logic into a reusable base class:

- **Type-Safe**: Generic `EventEmitter<TEvents>` with full type safety
- **Methods**: `on`, `once`, `off`, `waitFor`
- **Protected**: `emitInner` (for subclass use), `clearListeners`
- **Error Handling**: Re-throws listener errors via `queueMicrotask`
- **Usage**: 
  - `CollectionEventsManager` extends it and wraps `emitInner` with public `emit`
  - `CollectionSubscription` extends it and uses `emitInner` internally

### 9. Fixed Local-Only Collection Types

**File:** `packages/db/src/local-only.ts`

Fixed mutation function typing issues:

- Wrapper functions now accept broader `UtilsRecord` type parameters
- Properly handles contravariance in function parameter types
- Removed unnecessary type casts in return statement

### 10. Comprehensive Test Coverage

**Files:** `packages/db/tests/collection-subscription.test.ts`, `packages/db/tests/collection.test.ts`, `packages/db/tests/query/live-query-collection.test.ts`

Added extensive test suites:

- **CollectionSubscription**: Status tracking, event emission, concurrent promises, error handling, cleanup
- **Collection.isLoadingSubset**: Property tracking, event emission, concurrent loads, error handling  
- **Live Query Integration**: Result collection reflects subscription states, isolation between queries
- **All tests using `loadSubset`**: Updated to use `syncMode: 'on-demand'`

### 11. Enhanced `setWindow` with Loading Awareness

**File:** `packages/db/src/query/live/collection-config-builder.ts`

The `setWindow` utility function now returns a value that indicates whether subset loading was triggered, allowing callers to wait for data loading to complete:

**Return Type**: `true | Promise<void>`

- Returns `true` when `isLoadingSubset` is `false` after calling the window function - no loading was triggered
- Returns `Promise<void>` when `isLoadingSubset` is `true` - loading was triggered and the promise resolves when loading completes

**Implementation Details**:
- After calling `windowFn()` and `maybeRunGraphFn()`, checks `this.liveQueryCollection?.isLoadingSubset`
- If loading is active, subscribes to the `loadingSubset:change` event
- Returns a promise that resolves when `isLoadingSubset` becomes `false`
- Automatically unsubscribes from the event once loading completes

**Usage Pattern**:

```typescript
const result = liveQuery.utils.setWindow({ offset: 10, limit: 20 })

// Type guard pattern
if (result !== true) {
  // Loading was triggered, wait for it to complete
  await result
  console.log('Data loaded and window updated')
} else {
  // No loading needed, window updated immediately
  console.log('Window updated synchronously')
}

// Or simply always await (works with both return types)
const result = liveQuery.utils.setWindow({ offset: 10, limit: 20 })
if (result !== true) {
  await result
}
```

**Test Coverage**:
- Test that `setWindow` returns `true` when no loading is triggered
- Integration test that validates the full async flow with fake timers:
  - Verifies `Promise<void>` is returned when loading is triggered
  - Confirms `isLoadingSubset` becomes `true` during loading
  - Validates promise doesn't resolve until loading completes
  - Checks `isLoadingSubset` returns to `false` after resolution
  - Verifies correct data is loaded in the new window

### Sync Mode Configuration

```typescript
// Eager mode (default) - loads all data immediately
const eagerCollection = createCollection({
  getKey: (item) => item.id,
  syncMode: 'eager', // optional, this is the default
  sync: {
      // Load all data here
      begin()
      allData.forEach(item => write({ type: 'insert', value: item }))
      commit()
      markReady()
    }
  }
})

// On-demand mode - only loads data as requested
const onDemandCollection = createCollection({
  getKey: (item) => item.id,
  syncMode: 'on-demand',
  sync: {
    sync: ({ markReady }) => {
      markReady() // Don't load data initially
      
      return {
        // Required for on-demand mode
        loadSubset: async (options) => {
          const { where, limit, orderBy, subscription } = options
          // Load only the requested subset
          const data = await fetchDataSubset(where, limit, orderBy)
          // ... apply data to collection
        }
      }
    }
  }
})
```

### Collection Subscription

```typescript
const subscription = collection.subscribeChanges(callback, options)

// Status property
console.log(subscription.status) // 'ready' | 'loadingSubset'

// Event listeners
subscription.on('status:change', (event) => {
  console.log(`Status: ${event.previousStatus} → ${event.status}`)
})

subscription.on('status:loadingSubset', (event) => {
  console.log('Loading more data...')
})

subscription.on('status:ready', (event) => {
  console.log('Data loaded')
})

subscription.on('unsubscribed', (event) => {
  console.log('Subscription destroyed')
})

// Cleanup
subscription.unsubscribe()
```

### Collection

```typescript
// All collections have isLoadingSubset property
console.log(collection.isLoadingSubset) // boolean

// Listen for loading state changes
collection.on('loadingSubset:change', (event) => {
  console.log(`Loading: ${event.isLoadingSubset}`)
})
```

### Live Query

```typescript
const liveQuery = createLiveQueryCollection({
  query: (q) => q.from({ users: userCollection })
    .where(({ users }) => users.active)
})

// Result collection automatically tracks loading from subscriptions
console.log(liveQuery.isLoadingSubset) // boolean

liveQuery.on('loadingSubset:change', (event) => {
  if (event.isLoadingSubset) {
    showLoadingSpinner()
  } else {
    hideLoadingSpinner()
  }
})
```

### Renamed Methods and Types

- `syncMore` → `loadSubset` (collection method)
- `onLoadMore` → `loadSubset` (sync config property)
- `OnLoadMoreOptions` → `LoadSubsetOptions` (type)

**Migration:**

```typescript
// Before
const result = await collection.syncMore({ where: expr })
sync: {
  sync: () => ({
    onLoadMore: (options) => { /* ... */ }
  })
}

// After
const result = await collection._sync.loadSubset({ where: expr })
sync: {
  sync: () => ({
    loadSubset: (options) => { /* ... */ }
  })
}
```

Note: `loadSubset` is now called via `collection._sync.loadSubset()` as it's an internal coordination API, not for general public use.

### For Collection Users

No changes required if you're just using collections - `isLoadingSubset` is automatically available.

### For Sync Implementers

1. Rename `onLoadMore` to `loadSubset`:
```typescript
// Before
return { onLoadMore: (options) => { /* ... */ } }

// After  
return { loadSubset: (options) => { /* ... */ } }
```

2. (Optional) Use the new `subscription` parameter for advanced use cases:
```typescript
return {
  loadSubset: (options) => {
    // Track which subscription triggered this
    const sub = options.subscription
    
    // Can subscribe to unsubscribe event for cleanup
    sub?.on('unsubscribed', () => {
      // Clean up resources for this subscription
    })
  }
}
```

3. (Optional) Configure `syncMode` for on-demand loading:
```typescript
createCollection({
  syncMode: 'on-demand', // Only load data as requested
  sync: {
    sync: ({ markReady }) => {
      markReady()
      return {
        loadSubset: async (options) => { /* load subset */ }
      }
    }
  }
})
```

### DIFF
--- a/.changeset/smooth-goats-ring.md
+++ b/.changeset/smooth-goats-ring.md
@@ -1,0 +1,58 @@
+---
+"@tanstack/react-db": patch
+---
+
+Add `useLiveInfiniteQuery` hook for infinite scrolling with live updates.
+
+The new `useLiveInfiniteQuery` hook provides an infinite query pattern similar to TanStack Query's `useInfiniteQuery`, but with live updates from your local collection. It uses `liveQueryCollection.utils.setWindow()` internally to efficiently paginate through ordered data without recreating the query on each page fetch.
+
+**Key features:**
+
+- Automatic live updates as data changes in the collection
+- Efficient pagination using dynamic window adjustment
+- Peek-ahead mechanism to detect when more pages are available
+- Compatible with TanStack Query's infinite query API patterns
+
+**Example usage:**
+
+```tsx
+import { useLiveInfiniteQuery } from "@tanstack/react-db"
+
+function PostList() {
+  const { data, pages, fetchNextPage, hasNextPage, isLoading } =
+    useLiveInfiniteQuery(
+      (q) =>
+        q
+          .from({ posts: postsCollection })
+          .orderBy(({ posts }) => posts.createdAt, "desc"),
+      {
+        pageSize: 20,
+        getNextPageParam: (lastPage, allPages) =>
+          lastPage.length === 20 ? allPages.length : undefined,
+      }
+    )
+
+  if (isLoading) return <div>Loading...</div>
+
+  return (
+    <div>
+      {pages.map((page, i) => (
+        <div key={i}>
+          {page.map((post) => (
+            <PostCard key={post.id} post={post} />
+          ))}
+        </div>
+      ))}
+      {hasNextPage && (
+        <button onClick={() => fetchNextPage()}>Load More</button>
+      )}
+    </div>
+  )
+}
+```
+
+**Requirements:**
+
+- Query must include `.orderBy()` for the window mechanism to work
+- Returns flattened `data` array and `pages` array for flexible rendering
+- Automatically detects new pages when data is synced to the collection
--- a/packages/db/src/query/index.ts
+++ b/packages/db/src/query/index.ts
@@ -56,3 +56,4 @@ export {
 } from "./live-query-collection.js"
 
 export { type LiveQueryCollectionConfig } from "./live/types.js"
+export { type LiveQueryCollectionUtils } from "./live/collection-config-builder.js"
--- a/packages/react-db/src/index.ts
+++ b/packages/react-db/src/index.ts
@@ -1,5 +1,6 @@
 // Re-export all public APIs
 export * from "./useLiveQuery"
+export * from "./useLiveInfiniteQuery"
 
 // Re-export everything from @tanstack/db
 export * from "@tanstack/db"
--- a/packages/react-db/src/useLiveInfiniteQuery.ts
+++ b/packages/react-db/src/useLiveInfiniteQuery.ts
@@ -1,0 +1,185 @@
+import { useCallback, useEffect, useMemo, useRef, useState } from "react"
+import { useLiveQuery } from "./useLiveQuery"
+import type {
+  Context,
+  InferResultType,
+  InitialQueryBuilder,
+  LiveQueryCollectionUtils,
+  QueryBuilder,
+} from "@tanstack/db"
+
+/**
+ * Type guard to check if utils object has setWindow method (LiveQueryCollectionUtils)
+ */
+function isLiveQueryCollectionUtils(
+  utils: unknown
+): utils is LiveQueryCollectionUtils {
+  return typeof (utils as any).setWindow === `function`
+}
+
+export type UseLiveInfiniteQueryConfig<TContext extends Context> = {
+  pageSize?: number
+  initialPageParam?: number
+  getNextPageParam: (
+    lastPage: Array<InferResultType<TContext>[number]>,
+    allPages: Array<Array<InferResultType<TContext>[number]>>,
+    lastPageParam: number,
+    allPageParams: Array<number>
+  ) => number | undefined
+}
+
+export type UseLiveInfiniteQueryReturn<TContext extends Context> = Omit<
+  ReturnType<typeof useLiveQuery<TContext>>,
+  `data`
+> & {
+  data: InferResultType<TContext>
+  pages: Array<Array<InferResultType<TContext>[number]>>
+  pageParams: Array<number>
+  fetchNextPage: () => void
+  hasNextPage: boolean
+  isFetchingNextPage: boolean
+}
+
+/**
+ * Create an infinite query using a query function with live updates
+ *
+ * Uses `utils.setWindow()` to dynamically adjust the limit/offset window
+ * without recreating the live query collection on each page change.
+ *
+ * @param queryFn - Query function that defines what data to fetch. Must include `.orderBy()` for setWindow to work.
+ * @param config - Configuration including pageSize and getNextPageParam
+ * @param deps - Array of dependencies that trigger query re-execution when changed
+ * @returns Object with pages, data, and pagination controls
+ *
+ * @example
+ * // Basic infinite query
+ * const { data, pages, fetchNextPage, hasNextPage } = useLiveInfiniteQuery(
+ *   (q) => q
+ *     .from({ posts: postsCollection })
+ *     .orderBy(({ posts }) => posts.createdAt, 'desc')
+ *     .select(({ posts }) => ({
+ *       id: posts.id,
+ *       title: posts.title
+ *     })),
+ *   {
+ *     pageSize: 20,
+ *     getNextPageParam: (lastPage, allPages) =>
+ *       lastPage.length === 20 ? allPages.length : undefined
+ *   }
+ * )
+ *
+ * @example
+ * // With dependencies
+ * const { pages, fetchNextPage } = useLiveInfiniteQuery(
+ *   (q) => q
+ *     .from({ posts: postsCollection })
+ *     .where(({ posts }) => eq(posts.category, category))
+ *     .orderBy(({ posts }) => posts.createdAt, 'desc'),
+ *   {
+ *     pageSize: 10,
+ *     getNextPageParam: (lastPage) =>
+ *       lastPage.length === 10 ? lastPage.length : undefined
+ *   },
+ *   [category]
+ * )
+ */
+export function useLiveInfiniteQuery<TContext extends Context>(
+  queryFn: (q: InitialQueryBuilder) => QueryBuilder<TContext>,
+  config: UseLiveInfiniteQueryConfig<TContext>,
+  deps: Array<unknown> = []
+): UseLiveInfiniteQueryReturn<TContext> {
+  const pageSize = config.pageSize || 20
+  const initialPageParam = config.initialPageParam ?? 0
+
+  // Track how many pages have been loaded
+  const [loadedPageCount, setLoadedPageCount] = useState(1)
+  const [isFetchingNextPage, setIsFetchingNextPage] = useState(false)
+
+  // Stringify deps for comparison
+  const depsKey = JSON.stringify(deps)
+  const prevDepsKeyRef = useRef(depsKey)
+
+  // Reset page count when dependencies change
+  useEffect(() => {
+    if (prevDepsKeyRef.current !== depsKey) {
+      setLoadedPageCount(1)
+      prevDepsKeyRef.current = depsKey
+    }
+  }, [depsKey])
+
+  // Create a live query with initial limit and offset
+  // The query function is wrapped to add limit/offset to the query
+  const queryResult = useLiveQuery(
+    (q) => queryFn(q).limit(pageSize).offset(0),
+    deps
+  )
+
+  // Update the window when loadedPageCount changes
+  // We fetch one extra item to peek if there's a next page
+  useEffect(() => {
+    const newLimit = loadedPageCount * pageSize + 1 // +1 to peek ahead
+    const utils = queryResult.collection.utils
+    // setWindow is available on live query collections with orderBy
+    if (isLiveQueryCollectionUtils(utils)) {
+      const result = utils.setWindow({ offset: 0, limit: newLimit })
+      // setWindow returns true if data is immediately available, or Promise<void> if loading
+      if (result !== true) {
+        setIsFetchingNextPage(true)
+        result.then(() => {
+          setIsFetchingNextPage(false)
+        })
+      } else {
+        setIsFetchingNextPage(false)
+      }
+    }
+  }, [loadedPageCount, pageSize, queryResult.collection])
+
+  // Split the data array into pages and determine if there's a next page
+  const { pages, pageParams, hasNextPage, flatData } = useMemo(() => {
+    const dataArray = queryResult.data as InferResultType<TContext>
+    const totalItemsRequested = loadedPageCount * pageSize
+
+    // Check if we have more data than requested (the peek ahead item)
+    const hasMore = dataArray.length > totalItemsRequested
+
+    // Build pages array (without the peek ahead item)
+    const pagesResult: Array<Array<InferResultType<TContext>[number]>> = []
+    const pageParamsResult: Array<number> = []
+
+    for (let i = 0; i < loadedPageCount; i++) {
+      const pageData = dataArray.slice(i * pageSize, (i + 1) * pageSize)
+      pagesResult.push(pageData)
+      pageParamsResult.push(initialPageParam + i)
+    }
+
+    // Flatten the pages for the data return (without peek ahead item)
+    const flatDataResult = dataArray.slice(
+      0,
+      totalItemsRequested
+    ) as InferResultType<TContext>
+
+    return {
+      pages: pagesResult,
+      pageParams: pageParamsResult,
+      hasNextPage: hasMore,
+      flatData: flatDataResult,
+    }
+  }, [queryResult.data, loadedPageCount, pageSize, initialPageParam])
+
+  // Fetch next page
+  const fetchNextPage = useCallback(() => {
+    if (!hasNextPage || isFetchingNextPage) return
+
+    setLoadedPageCount((prev) => prev + 1)
+  }, [hasNextPage, isFetchingNextPage])
+
+  return {
+    ...queryResult,
+    data: flatData,
+    pages,
+    pageParams,
+    fetchNextPage,
+    hasNextPage,
+    isFetchingNextPage,
+  }
+}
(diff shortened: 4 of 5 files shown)


packages/db/src/collection/sync.ts
```diff
@@ -239,16 +256,73 @@ export class CollectionSyncManager<
...
+  public loadSubset(options: LoadSubsetOptions): Promise<void> | undefined {
+    // Bypass loadSubset when syncMode is 'eager'
+    if (this.syncMode === `eager`) {
+      return undefined
+    }
+
+    if (this.syncLoadSubsetFn) {
+      const result = this.syncLoadSubsetFn(options)
+
+      // If the result is void (synchronous), wrap in Promise.resolve()
```
KyleAMathews: We talked about collections needing to be able to return `true` or a promise — true if the data is loaded already. We need a sync response as otherwise useLiveQuery can't run sync

packages/db/tests/query/live-query-collection.test.ts
```diff
@@ -937,4 +937,93 @@ describe(`createLiveQueryCollection`, () => {
...
+      })
+
+      const liveQuery = createLiveQueryCollection({
+        query: (q) => q.from({ item: sourceCollection }),
+        startSync: true,
+      })
+
+      await liveQuery.preload()
+
+      // Calling loadSubset directly on source collection sets its own isLoadingMore
```
KyleAMathews: I thought the plan here was isLoadingMore is true only if a live query calls updatePredicate? Otherwise a live query would be see there `isLoadingMore` set to true when e.g. a joined collection needs to grab an object, etc. So any UI set to this would be flickering on and off seemingly randomly w/o any way to control it by the dev.

packages/db/src/collection/sync.ts
samwillis: hmm need to check why it's wrapping it there...

Returning void/undefined (synchronicity) makes it easy to check if a promise was returned as it's thruthy. Returning undefined if there is noting to do makes the code simpler than true - no typeof checks - but happy to change.

packages/db/tests/query/live-query-collection.test.ts
samwillis: This bit of the code all dates to when I was working over the weekend. Will update...

I do wander if we need to expose that it is grabbing data from the server though, joins lazy appearing without the dev being able to put a placeholder/spinner does feel messy.

Maybe we need to separate flags?

Also not that this version from the weekend made the isLoadingMore prop a standard on all collections. For base collections it's true when their own loadSubset is pending, for live query collections it's when a subscription trigger a loadSubset that returns a promise.

samwillis: Maybe we need an isLoadingMore and a isLoadingSubset, the latter whenever any subset triggered by the collection is loading, and isLoadingMore for when it's triggered by an offset/limit change (which do not isn't complete yet, and not in this PR)

packages/db/src/collection/sync.ts
samwillis: I've changed to return `Promise<void> | true` and check synchronously before setting status.

packages/db/tests/query/live-query-collection.test.ts
samwillis: To fully implement this as only triggered when the offset/limit is changed we will have to rebase this PR on #663 or wait till thats merged. I suggest we leave the implementation as is now (setting isLoading on any subscription triggering a `loadSubset`) and fix in a future PR.

https://github.com/TanStack/db/blob/9ad11697a32145476ec3a7caf3cebd31bed1711f/packages/db/src/query/live/collection-subscriber.ts#L77-L84

The question is if `collection. isLoadingMore ` should be `collection.isLoadingSubset` and that `isLoadingMore` is a live query only thing?

packages/db/src/collection/sync.ts
```diff
@@ -239,16 +256,70 @@ export class CollectionSyncManager<
...
+  /**
+   * Tracks a load promise for isLoadingMore state.
+   * @internal This is for internal coordination (e.g., live-query glue code), not for general use.
+   */
+  public trackLoadPromise(promise: Promise<void>): void {
+    const wasLoading = this.isLoadingMore
+    this.pendingLoadSubsetPromises.add(promise)
+    const isLoadingNow = this.isLoadingMore
+
+    if (!wasLoading && isLoadingNow) {
```
kevin-dp: This is a very confusing condition. I had to carefully look at this code to understand what it's checking. `isLoadingNow` will always be true.

It would be way more intuitive to store the length of `this.pendingLoadSubsetPromises` before adding the promise. Then just check if it was empty (meaning that now we need to start tracking).

packages/db/src/collection/sync.ts
```diff
@@ -239,16 +256,70 @@ export class CollectionSyncManager<
...
+      this._events.emit(`loadingMore:change`, {
+        type: `loadingMore:change`,
+        collection: this.collection,
+        isLoadingMore: true,
```
kevin-dp: Passing `isLoadingMore` and `previousIsLoadingMore` requires the user to do some annoying checks to know whether it started loading more or whether it finished loading more. Why not do that logic here and provide a more meaningful description in the event, e.g. `loadingState: 'startedLoading' | 'finishedLoading'`.

packages/db/src/collection/sync.ts
```diff
@@ -239,16 +256,70 @@ export class CollectionSyncManager<
...
+      })
+    }
+
+    promise.finally(() => {
+      // Check loading state BEFORE removing the promise
+      const wasLoadingBeforeRemoval = this.isLoadingMore
+      this.pendingLoadSubsetPromises.delete(promise)
+      const stillLoading = this.isLoadingMore
+
+      if (wasLoadingBeforeRemoval && !stillLoading) {
```
kevin-dp: Same here, condition would be more intuitive if we just check that the size of the set is 1 and that the only element it contains it `promise`

packages/db/src/collection/sync.ts
```diff
@@ -239,16 +256,70 @@ export class CollectionSyncManager<
...
+        this._events.emit(`loadingMore:change`, {
+          type: `loadingMore:change`,
+          collection: this.collection,
+          isLoadingMore: false,
```
kevin-dp: Would prefer to provide `loadingState: 'finishedLoading'` here

packages/db/src/collection/index.ts
```diff
@@ -364,15 +372,11 @@ export class CollectionImpl<
...
 
   /**
-   * Requests the sync layer to load more data.
-   * @param options Options to control what data is being loaded
-   * @returns If data loading is asynchronous, this method returns a promise that resolves when the data is loaded.
-   *          If data loading is synchronous, the data is loaded when the method returns.
+   * Tracks a load promise for isLoadingMore state.
+   * @internal This is for internal coordination (e.g., live-query glue code), not for general use.
    */
-  public syncMore(options: OnLoadMoreOptions): void | Promise<void> {
```
kevin-dp, line 372: I would rather keep this method such that we can keep the `_sync` property private

packages/db/src/collection/subscription.ts
```diff
@@ -121,10 +174,24 @@ export class CollectionSubscription {
...
     // Request the sync layer to load more data
     // don't await it, we will load the data into the collection when it comes in
-    this.collection.syncMore({
+    const syncResult = this.collection._sync.loadSubset({
       where: stateOpts.where,
+      subscription: this,
     })
 
+    // Track the promise if it's actually a promise (async work)
+    if (syncResult instanceof Promise) {
```
kevin-dp: Should we also check that the promise is not yet resolved? Because if it is already resolved then there is no need to do the below as that may lead to some flickering in the UI if one would e.g. show a spinner when more is loading.

packages/db/src/collection/subscription.ts
```diff
@@ -215,11 +282,25 @@ export class CollectionSubscription {
...
-    this.collection.syncMore({
+    const syncResult = this.collection._sync.loadSubset({
       where: whereWithValueFilter,
       limit,
       orderBy,
+      subscription: this,
     })
+
+    // Track the promise if it's actually a promise (async work)
+    if (syncResult instanceof Promise) {
```
kevin-dp: This code is identical to L183. Let's extract it to a private helper method.

packages/db/src/types.ts
```diff
@@ -313,6 +380,16 @@ export type CollectionStatus =
   /** Collection has been cleaned up and resources freed */
   | `cleaned-up`
 
+/**
```
kevin-dp: We already have this docstring on the `syncMode` property of the `BaseCollectionConfig` interface. IMO it makes sense to have it on the property but it is not needed on the type.

packages/db/src/collection/index.ts
samwillis: I disagree, `syncMore` is an internal method that shouldering be presented on the front he public api. The managers having the underscore marks them as "internal implementation but exposed for debugging and internal communication". I would prefer to not present `syncMore` on the prompts when people are interacting with a collection.

packages/db/src/collection/subscription.ts
samwillis: Unfortunately there is no api to do this, thats why loadSubset can return `true` to indicate that it has the data synchronously.

Issue #634, opened 2025-10-03
refactor the reconciliation process to be simpler and more maintainable


samwillis:
# Reconciliation Refactor

The current reconciliation pipeline has accumulated complexity after several rounds of fixes for truncate behaviour, optimistic rebasing, and duplicate-suppression bugs. Maintaining it has become risky: much of the flow is synchronous/CPU-bound but still entwined with batching flags and per-key delete replays for truncates, making the code brittle even without heavy IO. This plan aligns the system around a simpler contract—explicit truncate events, a single optimistic rebase path, and a shared change stream—so we can reason about the data flow, reduce footguns, and unblock future improvements.

## 1. Goals & Scope

- Reduce reconciliation complexity while preserving correctness guarantees: no missing/phantom events, consistent indexes, stable lifecycle state.
- Unify the change stream for live queries and index maintenance, adding explicit `truncate` semantics.
- Keep optimistic UI responsive with a predictable, easy-to-audit rebasing routine.
- Make the observable behaviour for `insert`/`update`/`delete` compatible with today’s tests, while documenting any intentional relaxations (e.g. allowing extra `update` noise when values truly change).

Out of scope: rewriting transaction orchestration, changing persistence APIs, or altering subscription filtering beyond what the new event contract requires.

## 2. Change-Event Contract

1. Extend `ChangeMessage.type` to `insert | update | delete | truncate`.
   - `truncate` has shape `{ type: 'truncate', metadata? }` with **no** `key`, `value`, or `previousValue`.
   - `insert`/`update`/`delete` retain their current fields.
2. `CollectionChangesManager.emitEvents` and `CollectionSubscription.filterAndFlipChanges` treat `truncate` as a stream-wide reset:
   - emit it immediately (bypass deferral—don't wait for pending sync transactions),
   - skip key-based flipping logic,
   - allow subsequent inserts in the same batch (from the same commit),
   - expect subscriptions to clear caches and live queries to resnapshot on receiving `truncate`.
3. Indexes gain a `reset()` entry point so a single `truncate` clears their internal state before replaying subsequent inserts.
4. Live-query consumers clear their local caches and resnapshot on `truncate` rather than depending on synthetic delete batches. Add regression tests for this behaviour.

## 3. Core State & Simplifications

- Retain `syncedData`, `syncedMetadata`, `transactions`, `optimisticUpserts`, `optimisticDeletes`, and `size`.
[...]

PR #671, opened 2025-10-13
Fix error propagation from sync clients to collections

KyleAMathews:
## Summary

Fixes critical error propagation issues where errors from Electric SQL and TanStack Query weren't being propagated to collections, causing:

1. **`preload()` hanging indefinitely** - When sync errors occurred, apps waiting for `preload()` would block forever
2. **Collections marked ready with no data** - Collections would show `status: 'ready'` even when they had failed to sync any data

### Electric-db-collection (`packages/electric-db-collection/src/electric.ts`)
- Implement 10-second grace period before setting error status, allowing Electric's built-in retry logic to recover from transitory network issues
- Remove premature `markReady()` call that was incorrectly marking collections ready before data synced
- Clear error timer on successful recovery

### Query-db-collection (`packages/query-db-collection/src/query.ts`)  
- Set error status immediately after TanStack Query exhausts all retry attempts
- No grace period needed since TanStack Query handles retries internally

### preload() fix (`packages/db/src/collection/sync.ts`)
- Listen for `status:error` events and reject the promise
- Prevents indefinite hangs when sync fails

### Collection lifecycle improvements (`packages/db/src/collection/`)
- Added `markError()` method for sync implementations to set error state
- Made `events` property public (was `_events` private)
- `markReady()` now auto-handles error recovery by transitioning through loading state (`error → loading → ready`)
- Valid state transitions updated to support recovery: `error` can transition to `idle` or `loading`

## Test plan

- ✅ All 66 tests passing in `@tanstack/electric-db-collection`
- ✅ All 58 tests passing in `@tanstack/query-db-collection`  
- ✅ New tests verify:
  - Error status set after 10 seconds of continuous errors (electric)
  - `preload()` rejects after 10 seconds of errors
  - Error timer cleared on recovery within grace period
  - Transitory errors recover naturally without setting error status


packages/db/src/collection/index.ts
```diff
@@ -215,9 +215,9 @@ export class CollectionImpl<
   public utils: Record<string, Fn> = {}
 
   // Managers
-  private _events: CollectionEventsManager
+  public events: CollectionEventsManager
```
samwillis, line 218: the manager classes are intended to be a private api, but have to be public so that they can see each other, and we can introspect them from tests. This should change it to public, but maintain the underscore on the name.

The "public" face of the events api is a set of methods on the Collection class itself that forward to this manager.

```suggestion
  public _events: CollectionEventsManager
```

packages/db/src/collection/lifecycle.ts
```diff
@@ -143,9 +144,14 @@ export class CollectionLifecycleManager<
    * @private - Should only be called by sync implementations
    */
   public markReady(): void {
+    // If recovering from error, transition to loading first
+    if (this.status === `error`) {
+      this.setStatus(`loading`)
+    }
```
samwillis, line 150: I don't like this, `markReady` should only move you to a ready state not `loading`.

We do not have a way in live queries to handle a `ready` => `loading`

How much handling of errors and moving between states to we expect the sync implementation to do? It seems to me there are two options for a collection:
1. handle error recovery with no braking of the synced state, as they do with a truncate, and so a `markReady` to go from `error -> ready` makes sense.
2. catastrophic "unrecoverable" error and the sync needs to be restarted. In that case a `restartSync()` method that the sync handler can call that:
  - calls `.cleanup()`, doing a gc and moves to `cleaned-up`
  - then calls `.startSync()`, which moves to `loading` and finally `ready` once the sync marks it as such.

Option 2 has some nuances arround subscriptions we would need to handle. Existing subscription would also need to be restarted, but we had not implemented that yet, it would be the truncate message from issue #634 proposing a refactor of the reconciliation process.

.beads/electric-error-propegate.db
samwillis, line 1: Whats this?

mccoyadd (approved):

samwillis:
@KyleAMathews what is the status of this PR? Do you plan to work through it soon or should we mark it as Draft for now?

PR #681, opened 2025-10-15
Handle pushed down predicates in query collection

KyleAMathews:
### Overview

This PR extends Query Collections to support predicate pushdown from live queries by enabling multiple concurrent queries with different predicates/filters. When live queries push down predicates via `loadSubset`, Query Collections now create separate TanStack Query instances for each unique set of options, pass those options to both the query key builder and query function, and manage the lifecycle of multiple queries with proper reference counting and garbage collection.

### Problem

When live queries push down predicates (via `loadSubset`), Query Collections need to:
1. Pass `LoadSubsetOptions` (predicates, limits, ordering) to the query key builder and query function
2. Create separate TanStack Query instances for each unique set of options
3. Track which rows belong to which queries (reference counting)
4. Only remove rows when no active queries reference them
5. Handle query garbage collection when queries are no longer needed
6. Return promises that resolve when query data is first available

Without this, Query Collections couldn't properly support the on-demand sync mode introduced in #669.

### Solution

This PR implements a comprehensive multi-query management system that flows predicates from live queries through to your TanStack Query implementation:

#### Predicate Flow

When a live query calls `loadSubset` with predicates, those options flow through the system:

1. **Live Query** → calls `collection._sync.loadSubset(options)`
2. **Query Collection** → calls `createQueryFromOpts(options)`
3. **Query Key Builder** → receives `options` to create unique query key
4. **Query Function** → receives `options` via `context.meta.loadSubsetOptions`

#### Example Usage

```typescript
import { createCollection } from "@tanstack/react-db"
import { queryCollectionOptions } from "@tanstack/query-db-collection"

const itemsPerPage = 20

  queryCollectionOptions({
    syncMode: 'on-demand', // Enable predicate pushdown
    
    // Query key builder receives LoadSubsetOptions
    queryKey: ({ limit, orderBy, where }) => {
      const page = computePageNumber(limit ?? itemsPerPage)
      return ["todos", { page, where, orderBy }]
    },

    // Query function receives options via context.meta.loadSubsetOptions
    queryFn: async (ctx) => {
      const { limit, where, orderBy } = ctx.meta?.loadSubsetOptions ?? {}
      const page = computePageNumber(limit ?? itemsPerPage)
      
      const params = new URLSearchParams({
        page: page.toString(),
        ...(where && { filter: JSON.stringify(where) }),
        ...(orderBy && { sort: JSON.stringify(orderBy) }),
      })
      
      const res = await fetch(\`/api/todos?\${params}\`)
      if (!res.ok) throw new Error("Failed to fetch todos")
      return res.json()
    },

    getKey: (item) => item.id,
    schema: todoSchema,
  })
)

function computePageNumber(limit: number): number {
  return Math.max(1, Math.ceil(limit / itemsPerPage))
}
```

In this example:
- The `queryKey` function builds different cache keys based on page/filters
- The `queryFn` receives the same options via `ctx.meta.loadSubsetOptions` to fetch the right data
- Each unique combination of predicates creates a separate TanStack Query
- Rows are reference-counted across all queries

#### 1. Dynamic Query Keys

**Type:** Added `TQueryKeyBuilder<TQueryKey>` type

```typescript
type TQueryKeyBuilder<TQueryKey> = (opts: LoadSubsetOptions) => TQueryKey

interface QueryCollectionConfig {
  queryKey: TQueryKey | TQueryKeyBuilder<TQueryKey>
  // ... other properties
}
```

The `queryKey` config option now accepts either:
- A static query key (for eager mode with no predicates)
- A function that builds a query key from `LoadSubsetOptions` (for on-demand mode with predicate pushdown)

**LoadSubsetOptions Structure:**
```typescript
interface LoadSubsetOptions {
  where?: any      // Filter predicates
  orderBy?: any    // Ordering directives
  limit?: number   // Result limit
  offset?: number  // Result offset
}
```

#### 2. Meta Property Extension

When creating a query, the collection merges `LoadSubsetOptions` into the query's meta:

```typescript
const extendedMeta = { ...meta, loadSubsetOptions: opts }

const observerOptions = {
  queryKey: key,
  queryFn: queryFn,
  meta: extendedMeta,  // Contains loadSubsetOptions
  // ... other options
}
```

Your `queryFn` can then access these options via `context.meta.loadSubsetOptions` to fetch the appropriate data.

#### 3. Multi-Query Tracking System

Implemented comprehensive state tracking using Maps:

```typescript
// hashedQueryKey → QueryKey
const hashToQueryKey = new Map<string, QueryKey>()

// hashedQueryKey → Set<RowKey> (which rows belong to which query)
const queryToRows = new Map<string, Set<string | number>>()

// RowKey → Set<hashedQueryKey> (which queries reference each row)
const rowToQueries = new Map<string | number, Set<string>>()

// hashedQueryKey → QueryObserver (active query observers)
const observers = new Map<string, QueryObserver>()

// hashedQueryKey → unsubscribe function
const unsubscribes = new Map<string, () => void>()
```

**Reference Counting:** Rows are only deleted from the collection when their reference count drops to zero (no queries reference them anymore).

#### 4. `createQueryFromOpts` Function

New internal function that creates or reuses queries based on `LoadSubsetOptions`:

**Return Type:** `true | Promise<void>`
- Returns `true` synchronously if query data is already available
- Returns `Promise<void>` that resolves when query data loads
- Returns `Promise<void>` that rejects if query encounters an error

**Behavior:**
- Hashes the query key (built from options) to check for existing queries
- Reuses existing `QueryObserver` instances when the same predicates are requested
- Creates new observers for unique predicate combinations
- Passes options to both query key builder and query function (via `meta.loadSubsetOptions`)
- Automatically subscribes if sync has started or collection has subscribers

#### 5. Query Garbage Collection

Listens to TanStack Query's cache events to handle query removal:

```typescript
queryClient.getQueryCache().subscribe((event) => {
  if (event.type === 'removed') {
    cleanupQuery(hashedKey)
  }
})
```

**Cleanup Process:**
1. Decrements reference counts for all rows in the removed query
2. Deletes rows when reference count reaches zero
3. Cleans up observer and unsubscribe function
4. Removes query from all tracking Maps

#### 6. Sync Mode Integration

**Eager Mode** (default):
- Creates single initial query with empty options (`{}`)
- Bypasses `loadSubset` (returns `undefined`)
- Static query key (no builder function needed)

**On-Demand Mode:**
- No initial query created
- Calls `markReady()` immediately since there's nothing to wait for
- Creates queries dynamically as `loadSubset(options)` is called
- Requires query key builder function to handle different predicates
- Returns `createQueryFromOpts` directly as the `loadSubset` implementation

**Sync Started Tracking:**
Added `syncStarted` flag to determine when to subscribe to new queries:
- Set to `true` when sync begins (via `preload()`, `startSync`, or first subscriber)
- Used instead of checking `config.startSync` to handle all sync scenarios

### Changes

**Files Modified:**
- `packages/query-db-collection/src/query.ts` - Core implementation (+354 lines)
- `packages/query-db-collection/tests/query.test.ts` - Comprehensive test suite (+567 lines)
- `.changeset/silent-trains-tell.md` - Changeset entry

**Test Coverage:**
- Added 4 new tests for Query Garbage Collection scenarios
- Added test for `preload()` in on-demand mode
- All 64 tests passing
- Code coverage: 88.66% lines

### Key Features

✅ **Predicate Pushdown** - Pass `LoadSubsetOptions` from live queries to TanStack Query  
✅ **Multiple Concurrent Queries** - Manage multiple TanStack Query instances with different predicates  
✅ **Reference Counting** - Track which queries reference which rows  
✅ **Automatic Garbage Collection** - Clean up queries and rows when no longer needed  
✅ **Promise-Based Loading** - Return promises that resolve when data is available  
✅ **Sync Mode Support** - Works with both eager and on-demand sync modes  
✅ **Immediate Ready State** - On-demand collections transition to ready immediately  

### Breaking Changes

None - this is a backward-compatible extension. Existing Query Collections with static query keys continue to work as before.

### Migration Guide

If you want to enable predicate pushdown:

1. Set `syncMode: 'on-demand'`
2. Change `queryKey` from a static value to a builder function
3. Access predicates in `queryFn` via `context.meta.loadSubsetOptions`

**Before:**
```typescript
queryCollectionOptions({
  queryKey: ['todos'],
  queryFn: async () => fetch('/api/todos').then(r => r.json()),
  // ...
})
```

**After:**
```typescript
queryCollectionOptions({
  syncMode: 'on-demand',
  queryKey: ({ where, limit, orderBy }) => ['todos', { where, limit, orderBy }],
  queryFn: async (ctx) => {
    const { where, limit } = ctx.meta?.loadSubsetOptions ?? {}
    const params = new URLSearchParams()
    if (where) params.set('filter', JSON.stringify(where))
    if (limit) params.set('limit', limit.toString())
    return fetch(\`/api/todos?\${params}\`).then(r => r.json())
  },
  // ...
})
```

### Related

- Builds on #669 (loadSubset state tracking and on-demand sync mode)
- Enables live queries to efficiently push down predicates to Query Collections
- Prepares for framework hook integration in follow-up PRs

---

**Note:** This PR replaces #646 with a rebased branch to remove duplicate commits from #669 that were already merged into main.

### DIFF
--- a/packages/query-db-collection/src/query.ts
+++ b/packages/query-db-collection/src/query.ts
@@ -568,6 +568,13 @@ export function queryCollectionOptions(
         subscribeToQuery(localObserver, hashedQueryKey)
       }
 
+      // Tell tanstack query to GC the query when the subscription is unsubscribed
+      // The subscription is unsubscribed when the live query is GCed.
+      const subscription = opts.subscription
+      subscription?.once(`unsubscribed`, () => {
+        queryClient.removeQueries({ queryKey: key, exact: true })
+      })
+
       return readyPromise
     }
 
@@ -829,10 +836,6 @@ export function queryCollectionOptions(
             loadSubset: createQueryFromOpts,
             onDeduplicate: createLocalQuery,
           })
-
-    // TODO: run the tests, probably some will fail bc different requests are made now
-    //       so fix the test expectations
-    //       then also add all the new dedup tests that Sam also added to the Electric collection
 
     return {
       loadSubset: loadSubsetDedupe?.loadSubset,
--- a/packages/db/src/query/subset-dedupe.ts
+++ b/packages/db/src/query/subset-dedupe.ts
@@ -239,10 +239,5 @@ export class DeduplicatedLoadSubset {
  * would reflect the mutated values rather than what was actually loaded.
  */
 export function cloneOptions(options: LoadSubsetOptions): LoadSubsetOptions {
-  return {
-    where: options.where,
-    orderBy: options.orderBy,
-    limit: options.limit,
-    // Note: We don't clone subscription as it's not part of predicate matching
-  }
+  return { ...options }
 }
--- a/packages/query-db-collection/tests/query.test.ts
+++ b/packages/query-db-collection/tests/query.test.ts
@@ -1,8 +1,12 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
 import { QueryClient } from "@tanstack/query-core"
-import { createCollection } from "@tanstack/db"
+import {
+  createCollection,
+  createLiveQueryCollection,
+  eq,
+  or,
+} from "@tanstack/db"
 import { queryCollectionOptions } from "../src/query"
-import { Func, PropRef, Value } from "../../db/src/query/ir"
 import type { QueryFunctionContext } from "@tanstack/query-core"
 import type {
   CollectionImpl,
@@ -2846,6 +2850,16 @@ describe(`QueryCollection`, () => {
   })
 
   describe(`Query Garbage Collection`, () => {
+    const isCategory = (category: `A` | `B` | `C`, where: any) => {
+      return (
+        where &&
+        where.type === `func` &&
+        where.name === `eq` &&
+        where.args[0].path[0] === `category` &&
+        where.args[1].value === category
+      )
+    }
+
     it(`should delete all rows when a single query is garbage collected`, async () => {
       const queryKey = [`single-query-gc-test`]
       const items: Array<TestItem> = [
@@ -2882,12 +2896,7 @@ describe(`QueryCollection`, () => {
       expect(collection.has(`3`)).toBe(true)
 
       // Simulate query garbage collection by removing the query from the cache
-      queryClient.removeQueries({ queryKey })
-
-      // Wait for GC to process
-      await vi.waitFor(() => {
-        expect(collection.size).toBe(0)
-      })
+      await collection.cleanup()
 
       // Verify all items are removed
       expect(collection.has(`1`)).toBe(false)
@@ -2904,8 +2913,11 @@ describe(`QueryCollection`, () => {
         const loadSubsetOptions = meta?.loadSubsetOptions ?? {}
         const { where } = loadSubsetOptions
 
+        console.log(`In queryFn:\n`, JSON.stringify(where, null, 2))
+
         // Query 1: items 1, 2, 3 (where: { category: 'A' })
-        if (where?.category === `A`) {
+        if (isCategory(`A`, where)) {
+          console.log(`Is category A`)
           return Promise.resolve([
             { id: `1`, name: `Item 1` },
             { id: `2`, name: `Item 2` },
@@ -2914,7 +2926,7 @@ describe(`QueryCollection`, () => {
         }
 
         // Query 2: items 2, 3, 4 (where: { category: 'B' })
-        if (where?.category === `B`) {
+        if (isCategory(`B`, where)) {
           return Promise.resolve([
             { id: `2`, name: `Item 2` },
             { id: `3`, name: `Item 3` },
@@ -2923,7 +2935,7 @@ describe(`QueryCollection`, () => {
         }
 
         // Query 3: items 3, 4, 5 (where: { category: 'C' })
-        if (where?.category === `C`) {
+        if (isCategory(`C`, where)) {
           return Promise.resolve([
             { id: `3`, name: `Item 3` },
             { id: `4`, name: `Item 4` },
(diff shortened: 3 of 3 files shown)


kevin-dp:
I pushed a unit test for limited ordered queries. The test currently fails because the query collection uses `currentStateAsChanges` to execute the query locally when it is deduplicated but the `currentStateAsChanges` function currently ignores the `orderBy` and `limit` options and so it contains all the query results instead of just the ones that fulfil the limit. Something that i'm going to fix in a separate PR.

kevin-dp:
I added support for `orderBy` and `limit` in `currentStateAsChanges` in https://github.com/TanStack/db/pull/701 and rebased the chain of PRs. All tests are green now.

packages/query-db-collection/src/query.ts
```diff
@@ -421,190 +428,427 @@ export function queryCollectionOptions(
...
+  }
+
+  const internalSync: SyncConfig<any>[`sync`] = (params) => {
+    const { begin, write, commit, markReady, collection } = params
+
+    // Track whether sync has been started
+    let syncStarted = false
+
+    const createQueryFromOpts = (
+      opts: LoadSubsetOptions,
```
kevin-dp, line 484: `opts` has `collectionSubscription` and it has an `unsubscribe` event that we can use to trigger GC.
On `unsubscribe` we should call tanstack to tell it to GC the query (see how it's done in the unit tests).

kevin-dp:
@samwillis I fixed the tanstack query GC such that it GCs when the live query subscription is unsubscribed. Also fixed the tests. I had to change the `cloneOptions` to also include the subscription because otherwise the `loadSubsetOptions` that are passed to the live query doesn't contain the subscription (because the query collection uses the deduper which clones the options and was omitting the subscription).

PR #684, opened 2025-10-16
Add Pre-created Collection Support to useLiveInfiniteQuery

samwillis:
## Summary
Extends `useLiveInfiniteQuery` to accept pre-created `LiveQueryCollection` instances in addition to query functions, enabling router loader patterns where collections can be created, awaited for preload, and then passed to route components.

### Core Functionality
- **`useLiveInfiniteQuery`** now accepts either:
  - A query function: `(q) => q.from(...).orderBy(...).limit(...)`
  - A pre-created collection: `createLiveQueryCollection({ query: ... })`
- Added automatic window validation and adjustment for pre-created collections
- Window adjustments happen in effects to avoid React subscription timing issues
- Pagination state resets when collection instance changes

### New API: `getWindow()` on LiveQueryCollectionUtils
- Added `getWindow()` method to inspect current window settings
- Returns `{ offset: number, limit: number } | undefined`
- Returns `undefined` for non-windowed queries
- Tracks window state internally in `CollectionConfigBuilder`

### Validation & Error Handling
- Pre-created collections **must** have an `orderBy` clause (throws error if missing)
- Input validation ensures first argument is either a `CollectionImpl` or function
- Logs warning when pre-created collection's initial window doesn't match expected values
- Warning only appears once per collection instance

### Type Safety
- Added overload signatures for type inference based on input type
- Proper type narrowing for both collection and query function paths
- Uses `instanceof CollectionImpl` for reliable runtime type checking

### Router Loader Pattern
```typescript
// In router loader
export async function loader() {
  const postsQuery = createLiveQueryCollection({
    query: (q) => q
      .from({ posts: postsCollection })
      .orderBy(({ posts }) => posts.createdAt, 'desc')
      .limit(20)
      .offset(0)
  })
  
  await postsQuery.preload()
  return { postsQuery }
}

// In component
function PostsList() {
  const { postsQuery } = useLoaderData()
  
  const { data, pages, hasNextPage, fetchNextPage } = useLiveInfiniteQuery(
    postsQuery,
    { pageSize: 20 }
  )
  
  // ... render
}
```

### Traditional Query Function (unchanged)
```typescript
const { data, pages, hasNextPage, fetchNextPage } = useLiveInfiniteQuery(
  (q) => q
    .from({ posts: postsCollection })
    .orderBy(({ posts }) => posts.createdAt, 'desc'),
  { pageSize: 20 }
)
```

## Testing
- ✅ 24 tests passing (7 new tests for pre-created collections)
- ✅ No linter errors
- ✅ No React warnings
- Tests cover: preloaded collections, pagination, instance changes, error cases, window validation, live updates, and router loader patterns

## Implementation Notes
- Window adjustment happens in `useEffect` after `useLiveQuery` subscribes to avoid triggering subscription callbacks during render which React will log warnings about
- Uses `hasValidatedCollectionRef` to track first-time validation and only log warnings once
- Delegates as much work as possible to the underlying `useLiveQuery` hook
- Both collection and query function code paths use a unified window adjustment effect

### DIFF
--- a/.changeset/brown-otters-grab.md
+++ b/.changeset/brown-otters-grab.md
@@ -1,0 +1,6 @@
+---
+"@tanstack/react-db": patch
+"@tanstack/db": patch
+---
+
+Add support for pre-created live query collections in useLiveInfiniteQuery, enabling router loader patterns where live queries can be created, preloaded, and passed to components.
--- a/packages/db/src/query/live/collection-config-builder.ts
+++ b/packages/db/src/query/live/collection-config-builder.ts
@@ -43,6 +43,12 @@ export type LiveQueryCollectionUtils = UtilsRecord & {
    * @returns `true` if no subset loading was triggered, or `Promise<void>` that resolves when the subset has been loaded
    */
   setWindow: (options: WindowOptions) => true | Promise<void>
+  /**
+   * Gets the current window (offset and limit) for an ordered query.
+   *
+   * @returns The current window settings, or `undefined` if the query is not windowed
+   */
+  getWindow: () => { offset: number; limit: number } | undefined
 }
 
 type PendingGraphRun = {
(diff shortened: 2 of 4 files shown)


kevin-dp (approved):
Great work!

PR #694, opened 2025-10-20
Callback to inform about loadSubset deduplication

kevin-dp:
Adds an `onDeduplicate` callback to the `DeduplicatedLoadSubset` class from https://github.com/TanStack/db/pull/668. Some collections (like the query collection) need this callback such that they are informed if a query is deduplicated. This is important because e.g. the query collection needs to manually reference count rows such that it can GC rows when they are not referenced by any query anymore. But this means they need to be able to track rows also for deduplicated queries.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -10,7 +10,7 @@ overrides:
   '@tanstack/react-db': workspace:*
   '@tanstack/offline-transactions': workspace:*
 
-pnpmfileChecksum: mas3tltraedwzq3bkgnzvfq2za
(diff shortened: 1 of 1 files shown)


packages/db/src/query/subset-dedupe.ts
```diff
@@ -53,9 +64,11 @@ export class DeduplicatedLoadSubset {
   private generation = 0
 
   constructor(
-    loadSubset: (options: LoadSubsetOptions) => true | Promise<void>
+    loadSubset: (options: LoadSubsetOptions) => true | Promise<void>,
```
samwillis: Object params

KyleAMathews (approved):
:shipit:

PR #696, opened 2025-10-20
feat: Add in-memory fallback for localStorage collections in SSR environments

### LINKED ISSUE #691: Prevent "LocalStorageCollectionError" when local-storage collections are imported on the server
We shouldn't throw an error unless someone tries to make a collection. Sometimes client code is pulled into server code but instantiating the module shouldn't throw an error.

KyleAMathews:
Prevents LocalStorageCollectionError when localStorage collections are imported on the server by automatically falling back to an in-memory store. This addresses issue #691 where importing client code with localStorage collections in server environments would throw errors during module initialization.

Changes:
- Add createInMemoryStorage() to provide in-memory StorageApi fallback
- Add createNoOpStorageEventApi() for server environments without window
- Update localStorageCollectionOptions to use fallbacks instead of throwing errors
- Remove unused NoStorageAvailableError and NoStorageEventApiError imports
- Update tests to verify fallback behavior instead of expecting errors
- Update JSDoc to document fallback behavior

This allows isomorphic JavaScript applications to safely import localStorage collection modules without errors, though data will not persist across reloads or sync across tabs when using the in-memory fallback.

Fixes #691

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/.changeset/in-memory-fallback-for-ssr.md
+++ b/.changeset/in-memory-fallback-for-ssr.md
@@ -4,7 +4,7 @@
 
 Add in-memory fallback for localStorage collections in SSR environments
 
-Prevents `LocalStorageCollectionError` when localStorage collections are imported on the server by automatically falling back to an in-memory store. This allows isomorphic JavaScript applications to safely import localStorage collection modules without errors during module initialization.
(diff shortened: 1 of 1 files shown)


KyleAMathews:
It should only happen in non-web environments in which case this is fine — so probably don't need any `debug`.

PR #697, opened 2025-10-20
 Add `useLiveSuspenseQuery` hook for React Suspense support

### LINKED ISSUE #692: Support suspense in React
**Current Situation**
Right now, there is no way to handle the data loading aspect of `useLiveQuery` with `Suspense` in React or any other integrations as far as I could tell.

**Expected behavior**
We should be able to opt into suspense or have a new hook called `useLiveSuspenseQuery` like `tanstack/query`.

**Additional context**
This was requested in [this](https://discord.com/channels/719702312431386674/1428722430498832566) discord chat, and I've also encountered the need to do this.

**Implementation**
I've done a rough implementation on top of the current `useLiveQuery` [here](https://github.com/MAST1999/db/tree/feat_add_react_suspense_query) which I don't think is ideal, but I've hit a blocker.

The `use(collection.preload())` promise is never resolving, or it's not correctly notified. I can't figure it out right now.

KyleAMathews:
Introduces a new `useLiveSuspenseQuery` hook that provides declarative data loading with React Suspense, following TanStack Query's `useSuspenseQuery` pattern.

**Key features:**
- React 18+ compatible using the throw promise pattern
- Type-safe API with guaranteed data (never undefined)
- Automatic error handling via Error Boundaries
- Reactive updates after initial load via useSyncExternalStore
- Support for dependency-based re-suspension
- Works with query functions, config objects, and pre-created collections

**Example usage:**

```tsx
import { Suspense } from 'react';
import { useLiveSuspenseQuery } from '@tanstack/react-db';

function TodoList() {
  // Data is guaranteed to be defined - no isLoading needed
  const { data } = useLiveSuspenseQuery((q) =>
    q.from({ todos: todosCollection })
      .where(({ todos }) => eq(todos.completed, false))
  );

  return (
    <ul>
      {data.map(todo => <li key={todo.id}>{todo.text}</li>)}
    </ul>
  );
}

function App() {
  return (
    <Suspense fallback={<div>Loading...</div>}>
      <TodoList />
    </Suspense>
  );
}
```

**Implementation details:**
- Throws promises when collection is loading (caught by Suspense)
- Throws errors when collection fails (caught by Error Boundary)
- Reuses promise across re-renders to prevent infinite loops
- Detects dependency changes and creates new collection/promise
- Same TypeScript overloads as useLiveQuery for consistency

Resolves #692

### DIFF
--- a/.changeset/suspense-query-hook.md
+++ b/.changeset/suspense-query-hook.md
@@ -4,7 +4,7 @@
 
 Add `useLiveSuspenseQuery` hook for React Suspense support
 
-Introduces a new `useLiveSuspenseQuery` hook that provides declarative data loading with React Suspense, following TanStack Query's `useSuspenseQuery` pattern.
+Introduces a new `useLiveSuspenseQuery` hook that integrates with React Suspense and Error Boundaries, following TanStack Query's `useSuspenseQuery` pattern.
 
 **Key features:**
 
@@ -54,5 +54,12 @@ function App() {
 - Reuses promise across re-renders to prevent infinite loops
 - Detects dependency changes and creates new collection/promise
 - Same TypeScript overloads as useLiveQuery for consistency
+
+**Documentation:**
+
+- Comprehensive guide in live-queries.md covering usage patterns and when to use each hook
+- Comparison with useLiveQuery showing different approaches to loading/error states
+- Router loader pattern recommendation for React Router/TanStack Router users
+- Error handling examples with Suspense and Error Boundaries
 
 Resolves #692
--- a/docs/guides/error-handling.md
+++ b/docs/guides/error-handling.md
@@ -123,6 +123,36 @@ Collection status values:
 - `error` - In error state
 - `cleaned-up` - Cleaned up and no longer usable
 
+### Using Suspense and Error Boundaries (React)
+
+For React applications, you can handle loading and error states with `useLiveSuspenseQuery`, React Suspense, and Error Boundaries:
+
+```tsx
+import { useLiveSuspenseQuery } from "@tanstack/react-db"
+import { Suspense } from "react"
+import { ErrorBoundary } from "react-error-boundary"
+
+const TodoList = () => {
+  // No need to check status - Suspense and ErrorBoundary handle it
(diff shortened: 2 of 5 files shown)


KyleAMathews:
Yeah, agree it's not great it duplicates things – I'll ask it to not

KyleAMathews:
Ok updated

mhsnook:
Excellent! I am switching my tanstack router SPA from react-query to react-db and this will be very helpful for me :)

packages/react-db/src/useLiveSuspenseQuery.ts
```diff
@@ -0,0 +1,178 @@
...
+    throw new Error(
+      `useLiveSuspenseQuery does not support disabled queries. Use useLiveQuery instead for conditional queries.`
+    )
+  }
+
+  // Only throw errors during initial load (before first ready)
+  // After success, errors surface as stale data (matches TanStack Query behavior)
+  if (result.status === `error` && !hasBeenReadyRef.current) {
+    promiseRef.current = null
+    throw new Error(`Collection "${result.collection.id}" failed to load`)
```
samwillis, line 161: This is good for now, but with the plan for collection to hold a reference to their last error object #671 we should rethrow that here.

packages/react-db/src/useLiveSuspenseQuery.ts
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  if (result.status === `loading` || result.status === `idle`) {
+    // Create or reuse promise for current collection
+    if (!promiseRef.current) {
+      promiseRef.current = result.collection.preload()
+    }
+    // THROW PROMISE - React Suspense catches this (React 18+ compatible)
```
samwillis: Do we need to check react version, or just accept that in <18 it will show as an error in an error boundary?

samwillis:
@KyleAMathews are you picking this up to finish it?

packages/react-db/src/useLiveSuspenseQuery.ts
KyleAMathews: yeah, no need to worry about older versions of React as suspense doesn't work there

Issue #706, opened 2025-10-22
writeDelete silently failing


pavle99:
- [x] I've validated the bug against the latest version of DB packages

**Describe the bug**
The `writeDelete` method for directly changing the sync store silently fails and stops further execution.

**To Reproduce**

1. Clone the example project from tanstack DB: https://github.com/TanStack/db/tree/main/examples/react/projects
2. Add the following below this [line](https://github.com/TanStack/db/blob/a03b5a5a619a448357fa1145bdd9733d755e366d/examples/react/projects/src/lib/collections.ts#L120):
```ts
console.log("todoCollection size1", todoCollection.size)

todoCollection.utils.writeDelete(deletedTodo.id)

console.log("todoCollection size2", todoCollection.size)

return { refetch: false }
```
3. Follow the README.md file to start the dev server
4. Start the server and login as any user
5. Add a couple of todos
6. Try deleting one
7. Notice that it will briefly flicker out (deleted) and then reappear and also notice that the `"todoCollection size2"` doesn't get logged to the console meaning the `onDelete` handlers execution stops at the line with `.writeDelete` but no errors get shown anywhere

**Expected behavior**

The todo gets deleted from the store and the execution continues after the `writeDelete` call.

**Desktop (please complete the following information):**

- OS: Windows 11
- Browser Brave
- Version 1.83.118

PR #708, opened 2025-10-22
Manual writes should validate against the synced store, not the combined synced + optimistic store

KyleAMathews:
Fixes https://github.com/TanStack/db/issues/706

Previously we were validating calls to write* calls against the combined sync+optimistic view — which failed if you'd already made some types of optimistic changes. This PR fixes validation to only look at the synced store as that's where write* calls are going to.

### DIFF
--- a/.changeset/fix-write-delete-in-handlers.md
+++ b/.changeset/fix-write-delete-in-handlers.md
@@ -1,0 +1,11 @@
+---
+"@tanstack/query-db-collection": patch
+---
+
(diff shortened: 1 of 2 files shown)


samwillis:
~~Good spot. Note that this will reject updates/deletes to inserted items until they have been confirmed by the server, that's likely a safe place to be, but does conflict with persisted offline mutations.~~

Nope, misunderstood it, it's the manure sync methods, this looks good to me

PR #713, opened 2025-10-23
Fix bug with orderBy on duplicate values

kevin-dp:
There was a bug in `orderBy` that could cause duplicate values to be skipped. For example, if we had this data: `1, 2, 3, 3, 3, 4, 5`. And we would request LIMIT 3 then we would load `1, 2, 3` and remember that the max value we've loaded is `3`. Then on the next load request we would ask for `data > 3` which would skip over the duplicate values `3` and immediately return `4, 5` thereby missing the two other occurences of the value `3`.

This PR modifies `requestLimitedSnapshot` such that it loads 2 subsets: all the occurences of `maxValue` and all data `> maxValue`. We can't do it in 1 query which is `>= maxValue LIMIT N` because the duplicate values would count towards the limit but we may already have loaded some of them. e.g. `>= 3 LIMIT 3` would give `3, 3, 3` but we had already seen the first occurence of `3`, so that's wrong because we actually expect to get `3, 3, 4`  (the next two occurences of 3 and the value `4`).

TODO:
- I found a bug with the liveQueryCollection which always returns `true` from `setWindow` instead of returning the promise of the underlying collection. Need to fix this in a follow up PR and then we can update the unit tests here to use `await` instead of `sleep()`.

### DIFF
--- a/packages/db/src/collection/changes.ts
+++ b/packages/db/src/collection/changes.ts
@@ -108,7 +108,7 @@ export class CollectionChangesManager<
     })
 
     if (options.includeInitialState) {
-      subscription.requestSnapshot()
+      subscription.requestSnapshot({ trackLoadSubsetPromise: false })
     }
 
     // Add to batched listeners
--- a/packages/db/src/collection/subscription.ts
+++ b/packages/db/src/collection/subscription.ts
@@ -20,6 +20,7 @@ import type { CollectionImpl } from "./index.js"
 type RequestSnapshotOptions = {
   where?: BasicExpression<boolean>
   optimizedOnly?: boolean
+  trackLoadSubsetPromise?: boolean
 }
 
 type RequestLimitedSnapshotOptions = {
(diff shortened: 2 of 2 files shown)


samwillis:
@kevin-dp I think this is the bug we were discussing before I went away and we talked about having a composite custom: `[maxValue, lastPK]` and always orderbying by the PK after the requested field?

We need to enable using indexes to order by `[fieldOne, fieldTwo]` anyway, so having it always append the PK (`[fieldOne, fieldTwo, PK]`) if it is not already used makes sense. The cursor is always the final value of `[fieldOne, fieldTwo, ..., PK]`then.

This will then be correctly pushed down to the sync layer as a stable cursor for paging through the data on the server.

kevin-dp:
@samwillis 

> I think this is the bug we were discussing before I went away and we talked about having a composite custom: [maxValue, lastPK] and always orderbying by the PK after the requested field?

Correct, it's that bug. I looked into making all `orderBy` keys composites by appending the PK if not already in the `orderBy` key, however, it requires a lot of changes to the code base that i wasn't fully comfortable with. After discussing with @KyleAMathews we decided not to go that route and instead come up with a minimal fix for this bug that can be shipped immediately.

> We need to enable using indexes to order by [fieldOne, fieldTwo] anyway, so having it always append the PK ([fieldOne, fieldTwo, PK]) if it is not already used makes sense. The cursor is always the final value of [fieldOne, fieldTwo, ..., PK]then.

Now that we have a fix for this bug, we can take our time to properly refactor the code base to support composite `orderBy` keys.

PR #715, opened 2025-10-23
fix: dedupe filtering for non-optimistic mutations

mpotter:
## 🎯 Changes

Fix `commitPendingTransactions` so that completed, non-optimistic mutations no longer suppress matching synced events, allowing the initiating client to observe the server-sourced mutation.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/.changeset/full-meals-ask.md
+++ b/.changeset/full-meals-ask.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+Fix synced propagation when preceding mutation was non-optimistic


KyleAMathews:
Can you share your code that wasn't working before this fix?

mpotter:
> Can you share your code that wasn't working before this fix?

It works for all clients that didn't originate the transaction. For the client that initiates the transaction, the local collection is not updated.

```typescript
export const chatThreadsCollection = createCollection(
  electricCollectionOptions({
    id: "chat-threads",
    shapeOptions: { ... },
    getKey: (row) => row.id,
    schema: ChatThreadSchema,

      const { modified } = transaction.mutations[0]!;
      const { id: _id, created_at: _c, updated_at: _u, ...payload } = modified;
      const response = await api.chat.threads.create(payload);
      return { txid: response.txid };
    },

      const { original, changes } = transaction.mutations[0]!;
      const { id: _id, created_at: _c, updated_at: _u, ...payload } = changes;
      const response = await api.chat.threads.update(original.id, payload);
      return { txid: response.txid };
    },

      const { original } = transaction.mutations[0]!;
      const response = await api.chat.threads.destroy(original.id);
      return { txid: response.txid };
    },
  })
);
```

```typescript
const tx = chatThreadsCollection.delete(id, { optimistic: false });
await tx.isPersisted.promise;
```

KyleAMathews:
Even after the synced tx comes back? It seems like you might have a txid matching issue perhaps? https://tanstack.com/db/latest/docs/collections/electric-collection#debugging

mpotter:
I don't believe I do. From the console logs immediately following delete:

```
ts/db:electric [chat-threads] awaitTxId called with txid 1782380 +9s
ts/db:electric [chat-messages] new txids synced from pg Array(1) +40ms
ts/db:electric [chat-threads] new txids synced from pg Array(1) +30ms
ts/db:electric [chat-threads] awaitTxId found match for txid 1782380 +0ms
```

My syncing is working everywhere else. It's only in this instance where I'm trying to use non-optimistic for a delete that it is broken.

KyleAMathews:
Ah ok I see. This PR prevents non-optimistic mutations from being added to completedOptimisticOps. The synced data is coming back and the optimistic mutations are being dropped but the collection thinks nothing has changed so doesn't emit anything but it has because it wasn't optimistic.

KyleAMathews (approved):
Thanks for the fix!

mpotter:
@KyleAMathews thanks for the quick merge 🙏

PR #717, opened 2025-10-24
feat: add runtime validation to prevent custom getKey with joined queries

KyleAMathews:
Implements fail-fast validation to catch the custom getKey + joins bug
at collection creation time instead of during sync.

Changes:
- Added CustomGetKeyWithJoinError to provide clear error message
- Added hasJoins() method that recursively checks query tree for joins
- Validation runs in CollectionConfigBuilder constructor
- Updated tests to verify error is thrown correctly
- Added test for nested subquery join detection

The error message guides users to:
- Remove custom getKey for joined queries
- Use array methods like .toArray.find() instead of .get()

Prevents the CollectionOperationError and TransactionError that occurred
when sync tried to insert with composite keys "[key1,key2]" while the
collection expected simple keys from custom getKey.

### DIFF
--- a/docs/reference/classes/AggregateFunctionNotInSelectError.md
+++ b/docs/reference/classes/AggregateFunctionNotInSelectError.md
@@ -5,7 +5,7 @@ title: AggregateFunctionNotInSelectError
 
 # Class: AggregateFunctionNotInSelectError
 
-Defined in: [packages/db/src/errors.ts:531](https://github.com/TanStack/db/blob/main/packages/db/src/errors.ts#L531)
+Defined in: [packages/db/src/errors.ts:547](https://github.com/TanStack/db/blob/main/packages/db/src/errors.ts#L547)
 
 ## Extends
 
@@ -19,7 +19,7 @@ Defined in: [packages/db/src/errors.ts:531](https://github.com/TanStack/db/blob/
 new AggregateFunctionNotInSelectError(functionName): AggregateFunctionNotInSelectError;
 ```
 
-Defined in: [packages/db/src/errors.ts:532](https://github.com/TanStack/db/blob/main/packages/db/src/errors.ts#L532)
+Defined in: [packages/db/src/errors.ts:548](https://github.com/TanStack/db/blob/main/packages/db/src/errors.ts#L548)
 
 #### Parameters
 
--- a/docs/reference/classes/AggregateNotSupportedError.md
+++ b/docs/reference/classes/AggregateNotSupportedError.md
@@ -5,7 +5,7 @@ title: AggregateNotSupportedError
 
 # Class: AggregateNotSupportedError
 
-Defined in: [packages/db/src/errors.ts:647](https://github.com/TanStack/db/blob/main/packages/db/src/errors.ts#L647)
+Defined in: [packages/db/src/errors.ts:663](https://github.com/TanStack/db/blob/main/packages/db/src/errors.ts#L663)
 
 Error thrown when aggregate expressions are used outside of a GROUP BY context.
 
@@ -21,7 +21,7 @@ Error thrown when aggregate expressions are used outside of a GROUP BY context.
 new AggregateNotSupportedError(): AggregateNotSupportedError;
 ```
 
-Defined in: [packages/db/src/errors.ts:648](https://github.com/TanStack/db/blob/main/packages/db/src/errors.ts#L648)
+Defined in: [packages/db/src/errors.ts:664](https://github.com/TanStack/db/blob/main/packages/db/src/errors.ts#L664)
 
 #### Returns
 
--- a/docs/reference/classes/CannotCombineEmptyExpressionListError.md
+++ b/docs/reference/classes/CannotCombineEmptyExpressionListError.md
@@ -5,7 +5,7 @@ title: CannotCombineEmptyExpressionListError
 
 # Class: CannotCombineEmptyExpressionListError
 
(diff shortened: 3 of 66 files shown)


samwillis:
@KyleAMathews could you fill me in on your thinking here? The intention is that a user can set a getKey on a joined query to specify a natural key to use rather than the generated on. We should fix the underlying issue not prevent this option as it was intended to be possible.

samwillis:
It may be that there is a bug somewhere where it is not using the getKey when materialising.

KyleAMathews:
@samwillis this was the original bug report https://discord.com/channels/933657521581858818/933657523049885698/1431144066099183656:

```
client.ts:1032 Uncaught CollectionOperationError: Cannot insert document with key "693892a1-4d2e-4f8d-9c05-9e9c5c667ca1" from sync because it already exists in the collection "live-query-1
 
client.ts:1032 Uncaught TransactionError: The pending sync transaction is already committed, you can't commit it again.
```

Which seemed like a duplicate key problem caused by their custom `getKey` that was just returning the id from one of the joined collections.

samwillis:
@KyleAMathews This is a slightly tough one to solve in a generic way - an example of a query where a user is likely to want a custom getKey is joins users to a comments collection. They know with certainty there is going to be a single row per comment and so a getKey that sets it to the comments ID is ideal, they don't really want to key that collection by `[commentId,userId]`. I don't think we can particularly easily infer these type of situations from the query though.

Maybe the solution is that a LiveQueryCollection overloads the `CollectionOperationError` error for inserts and producers a better error message when there is a custom `getKey` + join. Have it indicate that their getKey unexpectedly produced a duplicate key and they likely need to generate a composite key or refactor the join to do create duplicates.

KyleAMathews:
@samwillis ok ready for another review

packages/db/src/collection/sync.ts
```diff
@@ -127,7 +127,12 @@ export class CollectionSyncManager<
                   // throwing a duplicate-key error during reconciliation.
                   messageType = `update`
                 } else {
-                  throw new DuplicateKeySyncError(key, this.id)
+                  // Check if this is a live query with custom getKey and joins for better error messaging
+                  const utils = this.config.utils as any
```
samwillis: I don't like the `as any`. Can we have a `LiveQueryCollectionUtils` type: https://github.com/TanStack/db/blob/e0c4e2d7942727ed17f0d07e857bc5cf76527b16/packages/db/src/query/live/collection-config-builder.ts#L36-L37

Can you do something like:


```suggestion
                  const utils = this.config.utils as Partial<LiveQueryCollectionUtils>
```

packages/db/src/query/live/collection-config-builder.ts
```diff
@@ -173,6 +173,36 @@ export class CollectionConfigBuilder<
...
+    // Recursively check subqueries in join clauses
+    if (query.join) {
+      for (const joinClause of query.join) {
+        if (joinClause.from.type === `queryRef`) {
+          if (this.hasJoins(joinClause.from.query)) {
+            return true
+          }
+        }
+      }
+    }
```
samwillis: I'm not sure why we are recessively checking if joined sub queries have joins themselves? I think we only need to recurse down the `from`.

```suggestion
```

packages/db/src/query/live/collection-config-builder.ts
```diff
@@ -195,6 +225,9 @@ export class CollectionConfigBuilder<
         getBuilder: () => this,
         setWindow: this.setWindow.bind(this),
         getWindow: this.getWindow.bind(this),
+        // Metadata for error handling (as functions to match UtilsRecord type)
```
samwillis: redundant llm comment

packages/db/src/query/live/collection-config-builder.ts
```diff
@@ -195,6 +225,9 @@ export class CollectionConfigBuilder<
...
+        _hasCustomGetKey: () => !!this.config.getKey,
+        _hasJoins: () => this.hasJoins(this.query),
```
samwillis: `getBuilder` above isn't really a public api, but doesn't have an underscore prefix. We should decide on one pattern to follow for these and stick to it.

I also would not that currently we force the utils namespace to only have methods, but it would probably make sense to allow props and getters too. We should change that so we can make cleaner apis.

packages/db/tests/query/live-query-collection.test.ts
```diff
@@ -1706,4 +1706,45 @@ describe(`createLiveQueryCollection`, () => {
...
+              name: b.name,
+              value: r?.value,
+            })),
+        getKey: (item) => item.id, // Valid for 1:1 joins with unique keys
+      })
+
+      await liveQuery.preload()
+      expect(liveQuery.size).toBe(1)
+      // If duplicate keys occurred, an enhanced error would be thrown
+      // with guidance about composite keys
```
samwillis: We don't have a tests for if a duplicate occurs, we should add one.

packages/db/src/query/live/collection-config-builder.ts
samwillis: I still don't love these, maybe they can be namespaces behind a symbol or something? `getBuilder` could be moved too?

KyleAMathews: https://github.com/TanStack/db/pull/717/commits/9290765636a52b70af09f537481222d99914685a symbol namespace is fun

PR #719, opened 2025-10-24
fix: validate against duplicate collection aliases in subqueries

KyleAMathews:
## Summary

Fixes a Discord bug where using the same collection alias in both a parent query and subquery causes empty results or incorrect aggregation values.

## Problem

When both parent and subquery use the same alias for a direct collection reference:

```typescript
const locksAgg = q
  .from({ lock: c.locksCollection })
  .join({ vote: c.votesCollection }, ...)  // Uses "vote"

return q
  .from({ vote: c.votesCollection })  // Also uses "vote" directly
  .join({ lock: locksAgg }, ...)
```

**Result:**
- Empty query results
- Incorrect aggregation (values from individual rows instead of aggregates)

**Root Cause:** Both queries share the same input stream, causing interference.

## Solution

Added validation that throws a clear `DuplicateAliasInSubqueryError` when this pattern is detected.

**Implementation:**
- New error type with helpful message suggesting to rename the alias
- `collectDirectCollectionAliases()`: Identifies conflicting aliases
- `validateSubqueryAliases()`: Validates before compiling subqueries
- Only validates DIRECT collection references (allows legitimate subquery wrapping)

**User Fix:**
Simply rename the alias in either the parent or subquery:

```typescript
const locksAgg = q
  .from({ lock: c.locksCollection })
  .join({ v: c.votesCollection }, ...)  // Renamed "vote" to "v"

return q
  .from({ vote: c.votesCollection })  // No conflict
  .join({ lock: locksAgg }, ...)
```

## Testing

- Added `discord-alias-bug.test.ts` with comprehensive test coverage
- Validates error is thrown for conflicting aliases
- Verifies workaround works correctly

## Related

Fixes Discord bug reported by JustTheSyme.


packages/db/tests/query/validate-aliases.test.ts
samwillis, line 1: This test file is badly named, and the tests in it reference the "Discord Bug". The assertions should describe the expected behaviour, and maybe the test file should be named something like "validate-aliases.test.ts"

samwillis (changes requested):

samwillis (approved):
:shipit:

samwillis:
@KyleAMathews nudge, this is approved

PR #723, opened 2025-10-25
fix: manual transaction support in local-only collections

### LINKED ISSUE #722: Local collection manual transaction does not work

**Describe the bug**

The newly inserted item is removed on next tick.

**To Reproduce**

1. Open `packages/db/tests/local-only.test.ts`
2. Add the below to the end of the `should accept and persist mutations from manual transactions` test

```ts
await new Promise((resolve) => setTimeout(resolve, 10))

expect(collection.get(100)).toEqual({ id: 100, name: `Manual Tx Insert` })
```

marbemac:
fixes #722 

## 🎯 Changes

Fixed an issue where acceptMutations() wasn't working for manual transactions with local-only collections. The data would appear optimistically but then disappear after the transaction committed.

When filtering mutations to find ones belonging to the collection, it was comparing against `syncResult.collection`, which is `null` at the time `acceptMutations` is called. This is because the collection reference gets set during the sync callback, but `utils` aren't assigned to the collection until after construction completes. So all mutations were being filtered out and nothing was persisted.

To resolve this, I updated the mutation filtering to compare function references instead.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/db/src/local-only.ts
+++ b/packages/db/src/local-only.ts
@@ -250,7 +250,7 @@ export function localOnlyCollectionOptions<
   }) => {
     // Filter mutations that belong to this collection
     const collectionMutations = transaction.mutations.filter(
-      (m) => m.collection.id === collectionId
+      (m) => m.collection.id === collectionId,
     )
 
     if (collectionMutations.length === 0) {


marbemac:
Actually this might affect local only collection mutations in general (not just via manual transactions).

KyleAMathews:
So it turns out we solved this in a different way in local-storage.ts — which seems a bit less fragile than this:

```
// local-storage.ts:503-510
const collectionMutations = transaction.mutations.filter((m) => {
  // Try to match by collection reference first
  if (sync.collection && m.collection === sync.collection) {
    return true
  }
  // Fall back to matching by collection ID
  return m.collection.id === collectionId
})
```

marbemac:
> So it turns out we solved this in a different way in local-storage.ts

Yeah I noticed this, but this works because local storage has a much stronger guarantee of a unique id to operate off of. Since `storageKey` is required, it falls back to that if the user did not provide the optional collection id. The local collection has no such guarantee.

```
// Default id to a pattern based on storage key if not provided
const collectionId = id ?? `local-collection:${config.storageKey}`
```

KyleAMathews:
All collections have stable ids though — if one isn't passed, a UUID is generated.

marbemac:
Ah ok I missed that! Will take a look today

packages/db/src/local-only.ts
```diff
@@ -179,7 +179,10 @@ export function localOnlyCollectionOptions<
 ): LocalOnlyCollectionOptionsResult<T, TKey, TSchema> & {
   schema?: StandardSchemaV1
 } {
-  const { initialData, onInsert, onUpdate, onDelete, ...restConfig } = config
+  const { initialData, onInsert, onUpdate, onDelete, id, ...restConfig } =
+    config
+
+  const collectionId = id ?? crypto.randomUUID()
```
marbemac, line 185: @KyleAMathews ok updated with the uuid approach. As far as I can tell, the id automatically generated by the base collection impl is not available in time for use in the collection filtering below, so I had to generate it here.

kevin-dp:
Thanks @marbemac for your PR. I've forked your branch and opened a new PR because i couldn't make changes here. So i'm closing in favor of: https://github.com/TanStack/db/pull/1253

PR #730, opened 2025-10-30
Feat: Add support for custom parsers/serializers in LocalStorage collections

sadkebab:
## 🎯 Changes

This PR adds an option for a custom parser/serializer like superjson to be passed to a LocalStorage collection.

Since JSON.stringify and JSON.parse do not support serialization and deserialization of classes like Date, Map, Set or BigInt you can easily slip into runtime bugs because the objects retrieved from the collection don't match the types infered by the validation schema.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/db/src/local-storage.ts
+++ b/packages/db/src/local-storage.ts
@@ -44,7 +44,7 @@ interface StoredItem<T> {
   data: T
 }
 
-interface Parser {
+export interface Parser {
   parse: (data: string) => unknown
   stringify: (data: unknown) => string
 }


sadkebab:
The link for the `CONTRIBUTING.md` file is broken, I have read the one from `@tanstack/query` tho since I assume they would be the same.

packages/db/src/local-storage.ts
```diff
@@ -44,6 +44,11 @@ interface StoredItem<T> {
   data: T
 }
 
+interface Parser {
```
KyleAMathews: we should also export this as it's part of the exported `LocalStorageCollectionConfig` type

KyleAMathews (changes requested):
Neat! Just one little change and let's get this in

KyleAMathews (approved):
:shipit:

packages/db/src/local-storage.ts
sadkebab: done!

Issue #445, opened 2025-08-25
Seeing 40%+ slowness in simple useLiveQuery vs redux selectors when rendering many items


skidrowe:
We have an app that uses redux/re-reselect to render many items on the screen at a time. Switching the code to use tanstack db/useLiveQuery causes 40% or more speed reduction. A simple runnable example is attached...

[test.zip](https://github.com/user-attachments/files/21974367/test.zip)

skidrowe:
Updated to latest and added dependency array. 

[test2.zip](https://github.com/user-attachments/files/23372044/test2.zip)

KyleAMathews:
@skidrowe you still see a 40% slowdown? Or has that changed?

PR #732, opened 2025-10-30
Fix: Optimizer Missing Final Step - Combine Remaining WHERE Clauses

KyleAMathews:
## Overview

Fixes issue #445 - performance issue when using multiple `.where()` calls. The root cause was broader than initially identified: **the optimizer was missing "step 3"** (combining remaining WHERE clauses), affecting both queries with and without joins.

### The Optimizer's Intended Process

1. **Split**: Split WHERE clauses with AND at the root level into separate clauses
2. **Push down**: Push single-source clauses to subqueries (for queries with joins)
3. **Combine**: Combine all remaining WHERE clauses back into a single AND expression

**Step 3 was completely missing**, causing multiple `filter()` operations in the query pipeline.

#### 1. Queries WITHOUT Joins (Reported in Issue #445)

```javascript
useLiveQuery((q) =>
  q
    .from({ item: orderCollection })
    .where(({ item }) => eq(item.gridId, gridId))
    .where(({ item }) => eq(item.rowId, rowId))
    .where(({ item }) => eq(item.side, side))
)
```

The optimizer was skipping these entirely, leaving **3 separate WHERE clauses** → **3 filter operators** in the pipeline.

#### 2. Queries WITH Joins (Broader Issue)

```javascript
q.from({ stats: subqueryWithGroupBy })  // Can't push WHERE into GROUP BY
  .join({ posts: postsCollection }, ...)
  .where(({ stats }) => gt(stats.count, 5))     // Can't push down (safety check)
  .where(({ posts }) => gt(posts.views, 100))   // Can push down ✓
  .where(({ stats, posts }) => eq(stats.id, posts.author_id))  // Multi-source
```

After predicate pushdown:

- Posts clause: pushed down ✓
- Stats clause + multi-source clause: **2 separate WHERE clauses remain** → **2 filter operators** ✗

### Performance Impact

Each filter operator adds overhead. Data flows through N filter stages instead of 1 combined evaluation, causing unnecessary performance degradation especially when rendering many items.

## Solution

Implemented "step 3" in two places:

### Fix #1: `applySingleLevelOptimization` (queries without joins)

```typescript
if (!query.join || query.join.length === 0) {
  if (query.where.length > 1) {
    const splitWhereClauses = splitAndClauses(query.where)
    const combinedWhere = combineWithAnd(splitWhereClauses)
    return { ...query, where: [combinedWhere] }
  }
  return query
}
```

### Fix #2: `applyOptimizations` (queries with joins)

```typescript
// After predicate pushdown, combine remaining WHERE clauses
const finalWhere: Array<Where> =
  remainingWhereClauses.length > 1
    ? [combineWithAnd(remainingWhereClauses.map(getWhereExpression))]
    : remainingWhereClauses
```

## Testing

- ✅ All 43 optimizer tests pass
- ✅ Added test: "should combine multiple WHERE clauses for queries without joins"
- ✅ Added test: "should combine multiple remaining WHERE clauses after optimization"
- ✅ Updated 5 existing tests to expect combined WHERE clauses

## Before vs After

**Before (Multiple filter operators):**

```
FROM collection
→ filter(gridId = x)
→ filter(rowId = y)
→ filter(side = z)
```

**After (Single combined filter):**

```
FROM collection
→ filter(AND(gridId = x, rowId = y, side = z))
```

## Benefits

1. **Single Pipeline Operator**: Only 1 filter operation regardless of how many WHERE clauses
2. **Consistent Performance**: Chaining `.where()` now performs identically to using `.where(and(...))`
3. **Semantically Equivalent**: Multiple WHERE clauses still ANDed together
4. **Universal Application**: Works for all query types (with/without joins, simple/complex)
5. **Preserves Optimizations**: Still performs predicate pushdown for queries with joins

## Files Changed

- `packages/db/src/query/optimizer.ts` - Added WHERE combining logic (2 locations)
- `packages/db/tests/query/optimizer.test.ts` - Added tests and updated existing ones
- `.changeset/optimize-multiple-where-clauses.md` - Changeset describing the fix
- `ISSUE-445-INVESTIGATION.md` - Detailed investigation report

## Credits

Thanks to colleague feedback for catching that step 3 was missing from the optimizer!

---

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/packages/db/src/query/optimizer.ts
+++ b/packages/db/src/query/optimizer.ts
@@ -331,7 +331,7 @@ function applySingleLevelOptimization(query: QueryIR): QueryIR {
   }
 
   // For queries without joins, combine multiple WHERE clauses into a single clause
-  // to avoid creating multiple filter operators in the pipeline (performance optimization for issue #445)
+  // to avoid creating multiple filter operators in the pipeline
   if (!query.join || query.join.length === 0) {
     // Only optimize if there are multiple WHERE clauses to combine
     if (query.where.length > 1) {


packages/db/src/query/optimizer.ts
```diff
@@ -330,9 +330,22 @@ function applySingleLevelOptimization(query: QueryIR): QueryIR {
     return query
   }
 
-  // Skip optimization if there are no joins - predicate pushdown only benefits joins
-  // Single-table queries don't benefit from this optimization
+  // For queries without joins, combine multiple WHERE clauses into a single clause
+  // to avoid creating multiple filter operators in the pipeline (performance optimization for issue #445)
```
samwillis: remove ref to the issue from comment.

samwillis (approved):
Looks, good, just a nit on a comment.

I'm intrigued how much of a perf improvement this could make. Would love to know. It certainly does remove some allocations!

KyleAMathews:
Re perf, I had the same question. This is what Claude said — seems to make sense?

---

Great question! Let me explain by showing you what actually happens in the compiler:

Read
/home/user/db/packages/db/src/query/compiler/index.ts
Read 20 lines
The key is in line 187-195 - look at the for loop. For each WHERE clause, it creates a separate .pipe(filter(...)) operation in the D2 IVM pipeline.

The Performance Difference
Before (3 WHERE clauses):

pipeline
  .pipe(filter(row => row.item.gridId === gridId))    // Filter #1
  .pipe(filter(row => row.item.rowId === rowId))      // Filter #2
  .pipe(filter(row => row.item.side === side))        // Filter #3
After (1 combined WHERE clause):

pipeline
  .pipe(filter(row => 
    row.item.gridId === gridId && 
    row.item.rowId === rowId && 
    row.item.side === side
  ))  // Single filter
Why Multiple Filters Are Slower
Function Call Overhead: Each .pipe(filter(...)) wraps the predicate in a closure. For every row:

3 separate function calls (filter wrapper → compiled predicate)
vs. 1 function call with AND short-circuiting
Iterator/Stream Overhead: The D2 IVM library implements reactive streams. Each .pipe() operator:

Creates a new stream node in the dependency graph
Adds indirection for each item flowing through
More pipeline stages = more bookkeeping
Cache Locality: Data flows through 3 separate stages:

After filter #1, intermediate results are passed to filter #2
More memory accesses, worse cache behavior
vs. evaluating all conditions while row is hot in cache
No Short-Circuit Optimization: With separate filters:

Row fails filter #1 → still creates intermediate result → passes to filter #2 → passes to filter #3
With combined AND:

Row fails first condition → immediately returns false (short-circuit)
Real-World Impact
For useLiveQuery in React rendering 1000 rows:

Before: 3000 filter function calls + 3 pipeline stages per row
After: 1000 filter function calls + 1 pipeline stage per row
It's like the classic JavaScript optimization:

// Slower
arr.filter(x => x.age > 18)
   .filter(x => x.dept === 1)
   .filter(x => x.active)

// Faster
arr.filter(x => x.age > 18 && x.dept === 1 && x.active)
But amplified because D2 IVM's reactive pipeline has additional per-operator overhead that vanilla arrays don't have.

PR #734, opened 2025-10-30
Add stable viewKey API to prevent UI re-renders during ID transitions

### LINKED ISSUE #19: Support for Stable ViewKeys to Prevent UI Re-renders on ID Mapping
## Problem

Currently, optimistic inserts use temporary IDs (`tempid`) until the server responds with the authoritative ID (`realid`). When the ID changes, React re-renders the keyed element, causing UI issues:

- Focus loss
- Component state reset
- Input interruption
- Flickering

Because React treats the `key` as identity, replacing the ID forces React to unmount and remount the component. In many cases this is acceptable, but in advanced usage (e.g., editable lists, interactive forms) it causes UX degradation.

## Proposed Solution

Introduce **optional developer control** to maintain a stable `viewKey` that survives the tempid → realid mapping.

- Each record will have a `viewKey` field generated at optimistic creation time.
- `viewKey` remains constant across ID upgrades.
- UI components should key off `viewKey`, not `id`, if stability is required.
- Expose a manual API for developers to map `tempid` to `realid` in the persist function.

Example usage:

```typescript
persist: (transaction) => {
  const result = api.insert(data);
  transaction.mapViewKey({ tempId: tempid, realId: result.id });
}
```

- Mapping must occur after server success but before transaction completion.
- System updates internal references.
- No UI re-renders triggered by ID changes.
- If no mapping is performed, fallback to current behavior (tempid replaced by realid, React re-renders).

## Design Principles
- Opt-in only. No automatic viewKey generation unless explicitly requested.
- Defaults stay simple for developers who don't need stability.
- Advanced users can maintain seamless UX where needed.
- Keeps transaction model deterministic and explicit.

## Future Work
- Consider offering a helper to auto-generate viewKey at insert time if requested.
- Potentially expose viewKey defaults in the collection or transaction configuration for convenience.

KyleAMathews:
## Summary

Fixes #19

Adds built-in support for stable view keys that prevent UI flicker when optimistic inserts transition from temporary IDs to real server-generated IDs.

## The Problem

When inserting items with temporary IDs (e.g., negative numbers) that are later replaced by real server IDs, React treats the key change as a component identity shift, causing:

1. **UI Flicker** - Components unmount and remount, resetting state
2. **Lost Focus** - Input fields lose focus during ID transition
3. **Visual Jank** - Animations restart, scroll position resets

Previously, developers had to manually maintain an external mapping from IDs to stable keys.

## The Solution

Collections can now be configured with a `viewKey` function to automatically generate and track stable keys:

```typescript
const todoCollection = createCollection({
  getKey: (item) => item.id,
  viewKey: () => crypto.randomUUID(), // ← Auto-generate stable keys
    const tempId = transaction.mutations[0].modified.id
    const response = await api.create(...)

    // Link temp ID to real ID - they share the same viewKey
    todoCollection.mapViewKey(tempId, response.id)
    await todoCollection.utils.refetch()
  },
})

// Use stable keys in React - no more flicker!
{todos.map((todo) => (
  <li key={todoCollection.getViewKey(todo.id)}>
    {todo.text}
  </li>
))}
```

### New Configuration Option

- **`viewKey?: (item: T) => string`** - Function to generate stable view keys for inserted items

### New Collection Methods

- **`getViewKey(key: TKey): string`** - Returns stable viewKey for any key (temporary or real). Falls back to `String(key)` if no viewKey is configured.

- **`mapViewKey(tempKey: TKey, realKey: TKey): void`** - Links temporary and real IDs to share the same stable viewKey

### Type Changes

- Added `viewKey?: string` to `PendingMutation` interface
- Added `viewKey?: string` to `ChangeMessage` interface

## Implementation Details

1. **Storage**: Added `viewKeyMap: Map<TKey, string>` to `CollectionStateManager` to track stable keys
2. **Generation**: ViewKeys are automatically generated during `insert()` if configured
3. **Linking**: `mapViewKey()` creates bidirectional mapping from both temp and real IDs to the same viewKey
4. **Events**: All change events (insert/update/delete) now include viewKey for subscribers
5. **Persistence**: ViewKeys kept indefinitely (tiny memory overhead ~50 bytes per item)

## Backward Compatibility

✅ **Fully backward compatible** - All changes are opt-in:
- Collections without `viewKey` config work exactly as before
- `getViewKey()` returns `String(key)` when no viewKey is configured
- No breaking changes to existing APIs

## Documentation

Updated `/docs/guides/mutations.md` to replace the manual workaround with the new built-in API, including:
- Complete usage example
- How it works explanation
- Best practices

## Design Decisions

1. **Opt-in via configuration** - Only active when explicitly enabled
2. **Function instead of object** - Simple `viewKey: () => uuid()` instead of `{ generate: () => uuid() }`
3. **Explicit linking** - Manual `mapViewKey()` call for reliability (vs auto-detection which would be fragile)
4. **Collection-level storage** - ViewKeys stored in collection metadata, not polluting item data
5. **Indefinite retention** - Mappings kept forever for consistency (negligible memory impact)

---

## Related

- Original issue: https://github.com/TanStack/db/issues/19
- Mentioned in mutations.md documentation

### DIFF
--- a/packages/db/src/collection/state.ts
+++ b/packages/db/src/collection/state.ts
@@ -383,13 +383,10 @@ export class CollectionStateManager<
         previousDeletes
       )
 
-      // Get viewKey if available
-      const viewKey = this.viewKeyMap.get(key)
-
       if (previousValue !== undefined && currentValue === undefined) {
-        events.push({ type: `delete`, key, value: previousValue, viewKey })
+        events.push({ type: `delete`, key, value: previousValue })
       } else if (previousValue === undefined && currentValue !== undefined) {
-        events.push({ type: `insert`, key, value: currentValue, viewKey })
+        events.push({ type: `insert`, key, value: currentValue })
       } else if (
         previousValue !== undefined &&
         currentValue !== undefined &&
@@ -400,7 +397,6 @@ export class CollectionStateManager<
           key,
           value: currentValue,
           previousValue,
-          viewKey,
         })
       }
     }
@@ -519,13 +515,7 @@ export class CollectionStateManager<
               truncateOptimisticSnapshot?.upserts.get(key) ||
               this.syncedData.get(key)
             if (previousValue !== undefined) {
-              const viewKey = this.viewKeyMap.get(key)
-              events.push({
-                type: `delete`,
-                key,
-                value: previousValue,
-                viewKey,
-              })
+              events.push({ type: `delete`, key, value: previousValue })
             }
           }
 
@@ -632,12 +622,10 @@ export class CollectionStateManager<
               }
             }
             if (!foundInsert) {
-              const viewKey = this.viewKeyMap.get(key)
-              events.push({ type: `insert`, key, value, viewKey })
+              events.push({ type: `insert`, key, value })
             }
           } else {
-            const viewKey = this.viewKeyMap.get(key)
-            events.push({ type: `insert`, key, value, viewKey })
+            events.push({ type: `insert`, key, value })
           }
         }
 
@@ -754,7 +742,6 @@ export class CollectionStateManager<
         }
 
         if (!isRedundantSync) {
-          const viewKey = this.viewKeyMap.get(key)
           if (
             previousVisibleValue === undefined &&
             newVisibleValue !== undefined
@@ -763,7 +750,6 @@ export class CollectionStateManager<
               type: `insert`,
               key,
               value: newVisibleValue,
-              viewKey,
             })
           } else if (
             previousVisibleValue !== undefined &&
@@ -773,7 +759,6 @@ export class CollectionStateManager<
               type: `delete`,
               key,
               value: previousVisibleValue,
-              viewKey,
             })
           } else if (
             previousVisibleValue !== undefined &&
@@ -785,7 +770,6 @@ export class CollectionStateManager<
               key,
               value: newVisibleValue,
               previousValue: previousVisibleValue,
-              viewKey,
             })
           }
         }
--- a/packages/db/src/types.ts
+++ b/packages/db/src/types.ts
@@ -269,8 +269,6 @@ export interface ChangeMessage<
   previousValue?: T
   type: OperationType
   metadata?: Record<string, unknown>
-  /** Stable view key for rendering (survives ID transitions) */
-  viewKey?: string
 }
 
 export interface OptimisticChangeMessage<


kevin-dp:
Nit: i would rename `mapViewKey` to `setViewKey` or `storeViewKey`.

packages/db/src/collection/index.ts
```diff
@@ -450,6 +450,62 @@ export class CollectionImpl<
...
+   * @example
+   * // Use in React components for stable keys during ID transitions
+   * {todos.map((todo) => (
+   *   <li key={collection.getViewKey(todo.id)}>
+   *     {todo.text}
+   *   </li>
+   * ))}
+   */
+  public getViewKey(key: TKey): string {
+    const viewKey = this._state.viewKeyMap.get(key)
```
kevin-dp, line 470: This is a leaky abstraction. The caller (i.e. `getViewKey`) here should not be aware of the internal representation of the view keys. I'd prefer moving this `getViewKey` method to the `CollectionStateManager` and then here define a method that forwards the call:
```ts
public getViewKey(key: TKey): string {
  return this._state.getViewKey(key)
}
```

packages/db/src/collection/index.ts
```diff
@@ -450,6 +450,62 @@ export class CollectionImpl<
...
+    return viewKey ?? String(key)
```
kevin-dp, line 471: Keys are of type `TKey = string | number`. If we have no `viewKey` we turn the key into a string and use it as the `viewKey`. Is there any reason to have keys as `string | number` but `viewKey`s always as `string`? Would it make sense to also allow `viewKey`s to be of type `string | number` ?

packages/db/src/collection/index.ts
```diff
@@ -450,6 +450,62 @@ export class CollectionImpl<
...
+   *   const response = await api.todos.create(mutation.modified)
+   *   const realId = response.id
+   *
+   *   // Link the IDs so they share the same viewKey
+   *   collection.mapViewKey(tempId, realId)
+   *
+   *   await collection.utils.refetch()
+   * }
+   */
+  public mapViewKey(tempKey: TKey, realKey: TKey): void {
```
kevin-dp, line 496: Also a leaky abstraction. Move this implementation to the `CollectionStateManager` and have this public method delegate the call to the state manager's method.

packages/db/src/collection/state.ts
```diff
@@ -50,6 +50,9 @@ export class CollectionStateManager<
   public optimisticUpserts = new Map<TKey, TOutput>()
   public optimisticDeletes = new Set<TKey>()
 
+  // ViewKey mapping for stable rendering keys across ID transitions
+  public viewKeyMap = new Map<TKey, string>()
```
kevin-dp, line 54: No need to describe the type of the data structure in the field's name, i'd simplify the name to `viewKeys`.

packages/db/src/collection/mutations.ts
```diff
@@ -176,6 +176,14 @@ export class CollectionMutationsManager<
       }
       const globalKey = this.generateGlobalKey(key, item)
 
+      // Generate viewKey if configured
+      let viewKey: string | undefined
+      if (this.config.viewKey) {
+        viewKey = this.config.viewKey(validatedData)
+        // Store viewKey mapping
+        this.state.viewKeyMap.set(key, viewKey)
```
kevin-dp, line 184: Leaky abstraction. If the data structure ever changes, this call needs to be changed too.

kevin-dp (approved):
Overall looks good to me. Left a few minor comments that i would like to see addressed.

samwillis:
@KyleAMathews code all looks ok with @kevin-dp suggestions. I do however want a little time to mull over the api, I'm not sure if I'm 100% happy with it (I may turn out to be). Can we wait till tomorrow before merging so I have a little time to think about it.

In brief, and absolutely not my final thoughts:
1. it would be nice to use the actual PK when there is one.
2. I think the view keys will grow un-bound even when using a collection that GC's rows.
3. does this have any impact on persistence when that comes, do we need to persist the view keys (writing that out, I don't think we do...).

KyleAMathews:
> I think the view keys will grow un-bound even when using a collection that GC's rows.

We should clean these up when the collection is GCed. But also this only runs when something is optimistically inserted — so it'll be very rare that people will insert thousands of rows client-side & be using this system. If they do happen to hit this edge case where there's a memory leak, we can politely ask them to invent their own system that they can constrain the memory 😆 

> does this have any impact on persistence when that comes, do we need to persist the view keys (writing that out, I don't think we do...).

Hmm perhaps? Though this is more a problem for the offline transactions package which does persist optimistic mutations. That's an edge case though that doesn't seem particularly urgent to solve.

packages/db/src/collection/state.ts
samwillis, line 54: needs to be emptied in the cleanup at the bottom of this manger class.

Issue #417, opened 2025-08-18
Fix flaky react test


samwillis:
We have a flaky test in the react package:

```
 ⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯
packages/react-db test:  FAIL   @tanstack/react-db  tests/useLiveQuery.test.tsx > Query Collections > optimistic state is dropped after commit
packages/react-db test: AssertionError: expected undefined to match object { id: 'temp-key', …(2) }
packages/react-db test: - Expected: 
packages/react-db test: {
packages/react-db test:   "id": "temp-key",
packages/react-db test:   "name": "John Doe",
packages/react-db test:   "title": "New Issue",
packages/react-db test: }
packages/react-db test: + Received: 
packages/react-db test: undefined
packages/react-db test:  ❯ tests/useLiveQuery.test.tsx:758:54
packages/react-db test:     756|       expect(result.current.state.size).toBe(4)
packages/react-db test:     757|     })
packages/react-db test:     758|     expect(result.current.state.get(`[temp-key,1]`)).toMatchObject({
packages/react-db test:        |                                                      ^
packages/react-db test:     759|       id: `temp-key`,
packages/react-db test:     760|       name: `John Doe`,
packages/react-db test: ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯
packages/react-db test:  Test Files  1 failed (1)
packages/react-db test:       Tests  1 failed | 15 passed (16)
packages/react-db test: Type Errors  no errors
packages/react-db test:    Start at  14:44:30
```

PR #735, opened 2025-10-30
Fix flaky test

KyleAMathews:
Fixes https://github.com/TanStack/db/issues/417

The test "optimistic state is dropped after commit" was flaky because it had a race condition:
1. The test would wait for state size to become 4
2. Then immediately check that the temp-key exists
3. However, the async mutation (with only 10ms delay) could complete between steps 1 and 2

Fixed by moving all assertions into the same waitFor() block, ensuring they execute atomically. This prevents the mutation from completing between the size check and the temp-key verification.

## ✅ Checklist

## 🚀 Release Impact


samwillis (approved):
LGTM :shipit:

PR #741, opened 2025-10-31
Improve schema documentation

KyleAMathews:
This was quite thin before.

### DIFF
--- a/docs/guides/schemas.md
+++ b/docs/guides/schemas.md
@@ -118,6 +118,33 @@ const todoSchema = z.object({
 
 The schema acts as a **boundary** that transforms TInput → TOutput.
 
+### Critical Design Principle: TInput Must Be a Superset of TOutput
+
+When using transformations, **TInput must accept all values that TOutput contains**. This is essential for updates to work correctly.
+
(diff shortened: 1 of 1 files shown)


docs/guides/schemas.md
```diff
@@ -0,0 +1,1060 @@
...
+
+> **Important:** Schemas validate **client changes only** - data you insert or update via `collection.insert()` and `collection.update()`. They do not automatically validate data loaded from your server or sync layer. If you need to validate server data, you must do so explicitly in your integration layer.
+
+### What are TInput and TOutput?
+
+When you define a schema with transformations, it has two types:
+
+- **TInput**: The type users provide when calling `insert()` or `update()`
+- **TOutput**: The type stored in the collection and returned from queries
+
```
samwillis, line 107: I think it's worth documenting that it's best practice (or even essential) to make `TInput` and superset of `TOutput`. This is particularly important due to the `draft` on `update()` calls, the draft is typed as `TInput` but will contain the data from that will be in from the output of the schemas.

We made this tradeoff in the design to keep things simple and understandable, at the expense of having to document it.

We looked at trying to make the draft `TOutput | TInput` but that has to be computed recursively though the structures - its not just a simple union of the types, but a union of each prop recursively. Documenting that one is a superset of the other seemed better.

KyleAMathews: Good call — added.

PR #742, opened 2025-10-31
Add QueryObserver state utility methods to QueryCollectionUtils

KyleAMathews:
Adds new utility methods to QueryCollectionUtils to expose TanStack Query's QueryObserver state, addressing user request for visibility into sync status:

New utils exposed:
- isFetching() - Check if query is currently fetching (initial or background)
- isRefetching() - Check if query is refetching in background
- isLoading() - Check if query is loading for first time
- dataUpdatedAt() - Get timestamp of last successful data update
- fetchStatus() - Get current fetch status ('fetching' | 'paused' | 'idle')

This allows users to:
- Show loading indicators during background refetches
- Implement "Last updated X minutes ago" UI patterns
- Understand sync behavior beyond just error states

Converted error state utilities from methods to getters:
- lastError() → lastError
- isError() → isError
- errorCount() → errorCount

Resolves user request from Discord where status is always 'ready' after initial load, making it impossible to know if background refetches are happening.

### DIFF
--- a/packages/query-db-collection/src/query.ts
+++ b/packages/query-db-collection/src/query.ts
@@ -207,8 +207,8 @@ interface QueryCollectionState {
 }
 
 /**
- * Implementation class for QueryCollectionUtils that maintains proper closure
- * and compatibility with testing frameworks
+ * Implementation class for QueryCollectionUtils with explicit dependency injection
+ * for better testability and architectural clarity
  */
 class QueryCollectionUtilsImpl {
   private state: QueryCollectionState


packages/db/src/types.ts
```diff
@@ -35,9 +35,9 @@ export type TransactionState = `pending` | `persisting` | `completed` | `failed`
 export type Fn = (...args: Array<any>) => any
 
 /**
- * A record of utility functions that can be attached to a collection
+ * A record of utilities (functions or getters) that can be attached to a collection
  */
-export type UtilsRecord = Record<string, Fn>
+export type UtilsRecord = Record<string, any>
```
samwillis, line 40: 🫶

packages/query-db-collection/src/query.ts
```diff
@@ -710,25 +749,80 @@ export function queryCollectionOptions(
       }
     : undefined
 
+  // Create utils object with getters using a class pattern
+  // This ensures proper closure and compatibility with testing frameworks like vitest
+  class QueryCollectionUtilsImpl {
```
samwillis: Bit strange to have a class inside a function as a closure, I think this should re refactored out with the required state passed in to the constructor - you did add a nice `state` object thats perfect for this!

samwillis (changes requested):
Looking very good, much prefer this to the previous and the widening of the types for the utils.

Just a comment above on the class being a closure.

packages/query-db-collection/src/query.ts
```diff
@@ -181,6 +194,93 @@ export interface QueryCollectionUtils<
...
+  lastError: any
+  errorCount: number
+  lastErrorUpdatedAt: number
+  queryObserver:
+    | QueryObserver<Array<any>, any, Array<any>, Array<any>, any>
+    | undefined
+}
+
+/**
+ * Implementation class for QueryCollectionUtils that maintains proper closure
```
samwillis: > that maintains proper closure

its not a colure any more...

PR #745, opened 2025-11-03
Remove `flushSync` from svelte-db to avoid breaking async compiler mode

### LINKED ISSUE #744: svelte-db: flushSync inside effect breaks Svelte 5 async compiler mode

**Describe the bug**
TanStack Svelte DB is incompatible with Svelte 5's async compiler option due to calling `flushSync()` inside an effect, which violates its reactivity constraints:

https://github.com/TanStack/db/blob/48b8e8fc08ba8b0589476b5fd294f44b09f8ce8c/packages/svelte-db/src/useLiveQuery.svelte.ts#L357

The bug that this triggers:
`Uncaught Svelte error: flush_sync_in_effect
Cannot use `flushSync` inside an effect`

See https://svelte.dev/e/flush_sync_in_effect

**To Reproduce**

1. Enable async compiler in `svelte.config.js`
2. Use `useLiveQuery` in any component
3. Navigate to the page → Error occurs during collection initialization

- OS: iOS
- Browser chrome

KyleAMathews:
Fixes #744

The flushSync() call inside the onFirstReady callback was breaking Svelte 5's async compiler mode. The async compiler enforces a rule that flushSync cannot be called inside effects, as documented at svelte.dev/e/flush_sync_in_effect.

The fix removes the flushSync call and updates status directly. Svelte's reactivity system handles the update automatically without needing synchronous flushing. This matches the pattern used in Vue's implementation.

Changes:
- Removed flushSync() wrapper from onFirstReady callback
- Removed unused flushSync import
- Updated comment to explain why flushSync cannot be used
- All 23 existing tests pass, confirming no regression

This fix is backward compatible:
- For users WITHOUT async mode (current default): Works as before
- For users WITH async mode: Now works instead of throwing error
- Future-proof: async mode will be default in Svelte 6

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/docs/framework/react/reference/functions/useLiveSuspenseQuery.md
+++ b/docs/framework/react/reference/functions/useLiveSuspenseQuery.md
@@ -1,0 +1,490 @@
+---
+id: useLiveSuspenseQuery
+title: useLiveSuspenseQuery
+---
+
+# Function: useLiveSuspenseQuery()
+
+## Call Signature
+
+```ts
+function useLiveSuspenseQuery<TContext>(queryFn, deps?): object;
+```
+
+Defined in: [useLiveSuspenseQuery.ts:76](https://github.com/TanStack/db/blob/main/packages/react-db/src/useLiveSuspenseQuery.ts#L76)
+
+Create a live query with React Suspense support
+
+### Type Parameters
+
+#### TContext
+
+`TContext` *extends* `Context`
+
+### Parameters
+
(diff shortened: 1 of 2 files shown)


KyleAMathews:
@wobsoriano does this seem reasonable?

KyleAMathews:
Ah good call — just added an actual test.

PR #749, opened 2025-11-03
Fix findOne type issue with joins in useLiveQuery

KyleAMathews:
Fix type inference for findOne() when used with join operations

Previously, using `findOne()` with join operations (leftJoin, innerJoin, etc.) resulted in the query type being inferred as `never`, breaking TypeScript type checking:

```typescript
const query = useLiveQuery(
  (q) =>
    q
      .leftJoin({ todoOptions: todoOptionsCollection }, ...)
      .findOne() // Type became 'never'
)
```

**The Fix:**

Fixed the `MergeContextWithJoinType` type definition to conditionally include the `singleResult` property only when it's explicitly `true`, avoiding type conflicts when `findOne()` is called after joins:

```typescript
// Before (buggy):
singleResult: TContext['singleResult'] extends true ? true : false

// After (fixed):
type PreserveSingleResultFlag<TFlag> = [TFlag] extends [true]
  ? { singleResult: true }
  : {}

// Used as:
} & PreserveSingleResultFlag<TContext['singleResult']>
```

**Why This Works:**

By using a conditional intersection that omits the property entirely when not needed, we avoid type conflicts. Intersecting `{} & { singleResult: true }` cleanly results in `{ singleResult: true }`, whereas the previous approach created conflicting property types resulting in `never`. The tuple wrapper (`[TFlag]`) ensures robust behavior even if the flag type becomes a union in the future.

**Impact:**

- ✅ `findOne()` now works correctly with all join types
- ✅ Type inference works properly in `useLiveQuery` and other contexts
- ✅ Both `findOne()` before and after joins work correctly
- ✅ All tests pass with no breaking changes (8 new type tests added)

### DIFF
--- a/docs/reference/powersync-db-collection/classes/PowerSyncTransactor.md
+++ b/docs/reference/powersync-db-collection/classes/PowerSyncTransactor.md
@@ -1,0 +1,240 @@
+---
+id: PowerSyncTransactor
+title: PowerSyncTransactor
+---
+
+# Class: PowerSyncTransactor
+
+Defined in: [PowerSyncTransactor.ts:51](https://github.com/TanStack/db/blob/main/packages/powersync-db-collection/src/PowerSyncTransactor.ts#L51)
+
+Applies mutations to the PowerSync database. This method is called automatically by the collection's
+insert, update, and delete operations. You typically don't need to call this directly unless you
+have special transaction requirements.
+
+## Example
+
+```typescript
+// Create a collection
+const collection = createCollection(
+  powerSyncCollectionOptions<Document>({
+    database: db,
+    table: APP_SCHEMA.props.documents,
+  })
+)
+
+const addTx = createTransaction({
+  autoCommit: false,
+  mutationFn: async ({ transaction }) => {
+    await new PowerSyncTransactor({ database: db }).applyTransaction(transaction)
+  },
+})
+
+addTx.mutate(() => {
+  for (let i = 0; i < 5; i++) {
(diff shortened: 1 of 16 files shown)


packages/db/tests/query/findone-joins.test-d.ts
```diff
@@ -0,0 +1,75 @@
...
+          .findOne(),
+    })
+
+    // Verify query result is properly typed (not never)
+    type QueryData = typeof query.toArray
+    type IsNever<T> = [T] extends [never] ? true : false
+    type DataIsNever = IsNever<QueryData>
+
+    expectTypeOf<DataIsNever>().toEqualTypeOf<false>()
+    expectTypeOf(query.toArray).not.toBeNever()
```
samwillis: Can we assert that the data is exactly the correct type, not that it's not never.

packages/db/tests/query/join.test-d.ts
```diff
@@ -918,3 +918,189 @@ describe(`Join with ArkType Schemas`, () => {
     >()
   })
 })
+
+describe(`findOne() with joins - Type Safety`, () => {
+  test(`findOne() with leftJoin should not result in never type`, () => {
```
samwillis: All of the new tests in this file are still asserting not-never and have that in the name.

We also have the new `findone-joins.test-d.ts` file with what look like sensible tests, these may be redundant?

samwillis (changes requested):

samwillis (approved):
:shipit:

PR #750, opened 2025-11-03
Make clear in mutation doc that using other ways of writing to backend is ok

KyleAMathews:
Address feedback from user who thought they *had* to use DB's built-in mutations system.

### DIFF
--- a/docs/guides/mutations.md
+++ b/docs/guides/mutations.md
@@ -203,6 +203,8 @@ Custom actions provide the cleanest way to capture specific types of mutations a
 
 If you already have mutation logic in an existing system and don't want to rewrite it, you can **completely bypass** TanStack DB's mutation system and use your existing patterns.
 
+With this approach, you write to the server like normal using your existing logic, then use your collection's mechanism for refetching or syncing data to await the server write. After the sync completes, the collection will have the updated server data and you can render the new state, hide loading indicators, show success messages, navigate to a new page, etc.
+
 ```tsx
 // Call your backend directly with your existing logic
 const handleUpdateTodo = async (todoId, changes) => {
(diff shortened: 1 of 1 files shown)


docs/guides/mutations.md
```diff
@@ -197,7 +197,37 @@ Custom actions provide the cleanest way to capture specific types of mutations a
...
+// Call your backend directly with your existing logic
+const handleUpdateTodo = async (todoId, changes) => {
+  await api.todos.update(todoId, changes)
+
+  // Wait for sync back
+  await todoCollection.utils.refetch()
+}
+
+// With Electric
+const handleUpdateTodo = async (todoId, changes) => {
```
thruflo, line 220: Is this just an example of any non-related-to-tanstack-db mutation function? If so, what's the benefit of waiting on the txid -- is it so that your existing system can remove optimistic state after the tx has synced back? If so, perhaps it's worth clarifying?

docs/guides/mutations.md
```diff
@@ -197,7 +197,37 @@ Custom actions provide the cleanest way to capture specific types of mutations a
 
 - **Collection-level mutations** (`collection.update`): Simple CRUD operations on a single collection
 - **`createOptimisticAction`**: Intent-based operations, multi-collection mutations, immediately committed
-- **`createTransaction`**: Fully custom transactions, delayed commits, multi-step workflows
```
thruflo, line 200: Is `createTransaction` documented in a subsequent section? Or does this remove all references to it?

KyleAMathews: Yes, I noticed it was odd we mentioned it here despite not talking about it. Manual transactions are generally reached for so can be buried a bit

docs/guides/mutations.md
KyleAMathews: It's so you only return from the handler when you know the data from the server is back and ready to be rendered e.g. it's common to show a loader until this point. But yes, this needs clarified.

thruflo (approved):

PR #758, opened 2025-11-04
update generated docs & add CI job to fail PRs that need to re-run it

KyleAMathews:
As it turns out, we haven't updated them since the initial PR.


samwillis (approved):
:shipit:

PR #762, opened 2025-11-05
Add optional compareOptions to the Collection Config which act as col…

kevin-dp:
This PR adds support for collection-wide string ordering/locale configuration. It is possible to override these in a specific query by providing specific `compareOptions` in the query's `orderBy` clause. The priority is as follows (high to low):
- query provided `compareOptions`
- collection's `compareOptions` if none specified in query
- default compare options (machine's locale) if the user didn't pass `compareOptions` when creating the collection

The idea behind providing collection-wide locale configuration is that most applications should use the same locale configuration as their backend. This is needed because tanstack DB does a mix of loading ordered data from the backend and loading ordered data that is available locally. Hence, the local ordering must be the same as the one on the backend otherwise we get inconsistent behavior.

### High-level

- We store the collection-wide `StringSortOpts` in the collection itself. If none are provided, we store the default locale configuration.
- Modified the builder such that it no longer hydrates the `orderBy` clause with defaults for the locale, if no locale is specified, we keep them undefined in the `orderBy` clause
- At runtime, when processing an ordered query, we build the locale configuration from the `orderBy` clause's config and the collection's config. If the `orderBy` clause specifies locale configuration, we use that one. Otherwise, we use the locale configuration from the collection.

### Details

- Indexes are sensitive to locale options (i.e. different locale options require different indexes).
- We had to change all index creations and index lookups such that they keep into account the locale configuration.
  - When optimising ordered queries, we need to build a locale configuration (take the locale config from the orderBy clause and if not provided then use the locale config from the collection) and pass that config when we create and/or lookup indexes
  - Other places in the code were passing the default locale configuration when creating/looking up indexes. We had to change this to pass the collection's locale configuration instead

### DIFF
--- a/docs/reference/classes/CollectionImpl.md
+++ b/docs/reference/classes/CollectionImpl.md
@@ -129,14 +129,14 @@ Defined in: [packages/db/src/collection/index.ts:215](https://github.com/TanStac
 #### Get Signature
 
 ```ts
-get compareOptions(): StringSortOpts;
+get compareOptions(): StringCollationConfig;
 ```
 
 Defined in: [packages/db/src/collection/index.ts:516](https://github.com/TanStack/db/blob/main/packages/db/src/collection/index.ts#L516)
 
 ##### Returns
 
-[`StringSortOpts`](../../type-aliases/StringSortOpts.md)
+[`StringCollationConfig`](../../type-aliases/StringCollationConfig.md)
 
 ***
 
--- a/docs/reference/index.md
+++ b/docs/reference/index.md
@@ -194,7 +194,7 @@ title: "@tanstack/db"
 - [StorageEventApi](../type-aliases/StorageEventApi.md)
 - [Strategy](../type-aliases/Strategy.md)
 - [StrategyOptions](../type-aliases/StrategyOptions.md)
-- [StringSortOpts](../type-aliases/StringSortOpts.md)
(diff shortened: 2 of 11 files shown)


packages/db/src/types.ts
```diff
@@ -582,6 +613,14 @@ export interface BaseCollectionConfig<
    */
   onDelete?: DeleteMutationFn<T, TKey, TUtils, TReturn>
 
+  /**
+   * Specifies how to compare data in the collection.
+   * This should be configured to match data ordering on the backend.
+   * E.g., when using the Electric DB collection these options
+   *       should match the database's collation settings.
+   */
+  compareOptions?: StringSortOpts
```
samwillis: @KyleAMathews this is a new collection config option, you may have some to say on naming.

samwillis:
@kevin-dp as discussed we need to traverse the `from` in live queries to determine the compare options for the live query.
Also remember to pull as I rebased!

packages/db/src/types.ts
KyleAMathews: It's a bit ambiguous at the collection level *what* is compared — `defaultCollation` or `defaultStringCollation` is more clear.

kevin-dp: Ok, went for `defaultStringCollation` and renamed `StringSortOpts` type to `StringCollationConfig` (cc @KyleAMathews @samwillis)

PR #766, opened 2025-11-05
Don't make @tanstack/db a direct dependency to avoid having multiple versions of it bundled

KyleAMathews:
Changed @tanstack/db from regular dependencies to peerDependencies in:
- @tanstack/offline-transactions
- @tanstack/react-db
- @tanstack/query-db-collection

This fixes bundling issues where multiple versions of @tanstack/db were being installed. By declaring it as a peerDependency, consumers can provide their own version, ensuring only one instance is used across the application.

The @tanstack/db package is moved to devDependencies to allow local development and testing while preventing it from being installed as a transitive dependency for consumers.

## ✅ Checklist

## 🚀 Release Impact


KyleAMathews:
The opentelemetry stuff should have been removed before the PR was merged — so it's not really a change per se as they don't do anything on main either.

PR #771, opened 2025-11-06
Use versions in examples for workspace packages along with pnpm overrides

KyleAMathews:
When running examples in the monorepo, we want to ensure they use they actual workspace code and don't pull from npm — we use pnpm overrides for this. This PR adds all workspace packages to the overrides list.

When running examples outside of the monorepo (we suggest `gitpick` for this). We want proper versions that have the example pull the latest published version from npm. We had some packages with `workspace:*` versions which wouldn't work anymore outside of the monorepo. So this PR moves all versions to use ^0.0.0 to get the latest version in the 0 range.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -5,10 +5,19 @@ settings:
   excludeLinksFromLockfile: false
 
 overrides:
+  '@tanstack/angular-db': workspace:*
   '@tanstack/db': workspace:*
+  '@tanstack/db-ivm': workspace:*
+  '@tanstack/electric-db-collection': workspace:*
+  '@tanstack/offline-transactions': workspace:*
+  '@tanstack/powersync-db-collection': workspace:*
   '@tanstack/query-db-collection': workspace:*
   '@tanstack/react-db': workspace:*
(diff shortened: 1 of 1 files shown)


kevin-dp (changes requested):
`^0.0.0` does not work as you expect. It restricts to exactly version `0.0.0`.
From the [docs](https://github.com/npm/node-semver?tab=readme-ov-file#caret-ranges-123-025-004):
> Caret Ranges ^1.2.3 ^0.2.5 ^0.0.4
Allows changes that do not modify the left-most non-zero element in the [major, minor, patch] tuple. In other words, this allows patch and minor updates for versions 1.0.0 and above, patch updates for versions 0.X >=0.1.0, and **no updates for versions 0.0.X.**

I think you want `*` or `latest`

KyleAMathews:
Ok this looks good then https://jubianchi.github.io/semver-check/#/%3E%3D0.0.0%20%3C1.0.0/1.0.0 — then we can update to `^1.0.0` when we ship v1.

Screenshot 2025-11-06 at 7 45 08 AM

PR #779, opened 2025-11-07
Hash small Uint8Arrays (≤128 bytes) by content rather than reference 

KyleAMathews:
Small Uint8Arrays (≤128 bytes) are now hashed by their content rather than by reference, enabling proper equality comparisons for binary IDs like ULIDs (16 bytes). Large arrays (>128 bytes) continue to be hashed by reference to avoid performance costs.

This allows the `eq` expression function to correctly compare binary IDs without forcing users to use the more expensive functional expression variant.

Changes:
- Added UINT8ARRAY_CONTENT_HASH_THRESHOLD constant (128 bytes)
- Implemented hashUint8Array() function for content-based hashing
- Modified hashObject() to check byteLength and hash small arrays by content
- Updated existing tests to reflect new behavior for small and large arrays

Fixes issue where binary ULIDs couldn't be compared using `eq` due to reference-based hashing.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/db/src/utils/comparison.ts
+++ b/packages/db/src/utils/comparison.ts
@@ -127,6 +127,14 @@ function areUint8ArraysEqual(a: Uint8Array, b: Uint8Array): boolean {
 }
 
 /**
+ * Threshold for normalizing Uint8Arrays to string representations.
+ * Arrays larger than this will use reference equality to avoid memory overhead.
+ * 128 bytes is enough for common ID formats (ULIDs are 16 bytes, UUIDs are 16 bytes)
+ * while avoiding excessive string allocation for large binary data.
+ */
+const UINT8ARRAY_NORMALIZE_THRESHOLD = 128
+
+/**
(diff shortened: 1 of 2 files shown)


packages/db-ivm/src/hashing/hash.ts
```diff
@@ -78,6 +91,20 @@ function hashDate(input: Date): number {
...
+  hasher.update(UINT8ARRAY_MARKER)
+  // Hash the byte length first to differentiate arrays of different sizes
+  hasher.update(input.byteLength)
+  // Hash each byte in the array
+  for (let i = 0; i < input.byteLength; i++) {
+    // Use _writeByte to hash the actual byte values
+    // @ts-expect-error - _writeByte is private but we need to use it here
+    hasher._writeByte(input[i]!)
+  }
+  return hasher.digest()
```
samwillis, line 103: We can make this method public rather than use the comment to ignore the type error.

PR #773, opened 2025-11-06
E2E Test Suite for Query-Driven Sync

samwillis:
**Stacked on #786**

## Summary

Implements a comprehensive end-to-end test suite for the query-driven sync feature with on-demand collection loading (RFC #676, PR #763). This PR adds a new shared test infrastructure package and integrates it with both Electric and Query collection implementations.

### New Package: `@tanstack/db-collection-e2e`

A reusable e2e test infrastructure package providing:
- 8 comprehensive test suites (95 test scenarios for Electric, 89 for Query)
- Docker orchestration (Postgres + Electric)
- Seed data generation (300 records across 3 tables)
- Vitest fixtures following Electric's proven patterns
- Custom assertions and utilities

### Test Suites Implemented

| Suite | Tests | Coverage |
|-------|-------|----------|
| **Predicates** | 20 | `eq`, `gt`, `gte`, `lt`, `lte`, `inArray`, `isNull`, boolean logic |
| **Pagination** | 15 | `orderBy`, `limit`, `offset`, edge cases |
| **Joins** | 12 | 2-way and 3-way joins, mixed syncModes, predicate pushdown |
| **Deduplication** | 8 | Concurrent queries, overlapping predicates, race conditions |
| **Collation** | 8 | Default/custom collation, case sensitivity, locale support |
| **Mutations** | 10 | Insert, update, delete, soft delete pattern |
| **Live Updates** | 8 | Reactive updates, subscription lifecycle (Electric only) |
| **Regressions** | 5+ | Known bugs from early testing (#7214245, #9874949) |

**Total: 86+ test scenarios**

### Integration

**Electric Collection** (`packages/electric-db-collection/e2e/`):
- Real Electric collections syncing from Postgres
- All 8 test suites running
- Tests against actual HTTP shape API

**Query Collection** (`packages/query-db-collection/e2e/`):
- Real Query collections with TanStack Query
- 7 test suites (excluding Electric-specific Live Updates)
- Mock backend with queryFn

### Electric Collection E2E
```
✅ Tests: 95 total
✅ Infrastructure: Working (Docker, Postgres, Electric)
✅ Real sync: Using electricCollectionOptions() with actual Electric server
⚠️ Status: 58 passing, 37 failing
```

**Failures are expected** - tests are exposing real Electric sync timing issues where `preload()` returns before data is synced. This is valuable feedback for the query-driven sync implementation.

### Query Collection E2E
```
✅ Tests: 89 total
✅ Infrastructure: Working (TanStack Query integration)
[...]

PR #786, opened 2025-11-10
feat: add expression helpers for parsing LoadSubsetOptions in queryFn

KyleAMathews:
## Summary

Adds reusable expression parsing utilities to help developers translate TanStack DB predicates (where, orderBy, limit) into their API's format when implementing query collections with predicate push-down.

### New Expression Helpers (`packages/query-db-collection/src/expression-helpers.ts`)

- **`parseWhereExpression`**: Parse where clauses with custom handlers for each operator
- **`parseOrderByExpression`**: Parse order by into simple array format
- **`extractSimpleComparisons`**: Extract simple AND-ed filters for basic use cases
- **`parseLoadSubsetOptions`**: Convenience function to parse all options at once
- **`walkExpression`**, **`extractFieldPath`**, **`extractValue`**: Lower-level helpers for custom parsing

### Documentation

Added comprehensive section to `docs/collections/query-collection.md`:
- How LoadSubsetOptions are passed via `ctx.meta.loadSubsetOptions`
- Expression helper usage with REST API examples
- GraphQL example
- API reference for all helper functions
- Query key builder pattern
- Tips and best practices

### Tests

35 test cases with 93.65% coverage:
- All helper functions tested
- Various operator combinations (eq, lt, gt, and, or, etc.)
- Edge cases (null/undefined, nested expressions)
- Integration test with complex real-world query

## Why This Matters

Without these helpers, developers had to manually parse expression AST trees:

```typescript
// Before - complex and error-prone
if (where && where.type === 'func' && where.name === 'eq') {
  const field = where.args[0].path[0]
  const value = where.args[1].value
  // ... more traversal logic
}
```

With helpers, it's straightforward:

```typescript
// After - simple and maintainable
const filters = parseWhereExpression(where, {
  handlers: {
    eq: (field, value) => ({ [field.join('.')]: value }),
    lt: (field, value) => ({ [`${field.join('.')}_lt`]: value }),
    and: (...conditions) => Object.assign({}, ...conditions)
  }
})
```

## Example Usage

```typescript
queryFn: async (ctx) => {
  const { where, orderBy, limit } = ctx.meta.loadSubsetOptions

  // Parse into your API format
  const parsed = parseLoadSubsetOptions({ where, orderBy, limit })

  // Build API request
  const params = new URLSearchParams()
  parsed.filters.forEach(({ field, operator, value }) => {
    if (operator === 'eq') {
      params.set(field.join('.'), String(value))
    }
  })

  return fetch(`/api/products?${params}`).then(r => r.json())
}
```

### DIFF
--- a/packages/db/src/query/expression-helpers.ts
+++ b/packages/db/src/query/expression-helpers.ts
@@ -1,5 +1,5 @@
 /**
- * Expression Helpers for Query Collections
+ * Expression Helpers for TanStack DB
  *
  * These utilities help parse LoadSubsetOptions (where, orderBy, limit) from TanStack DB
  * into formats suitable for your API backend. They provide a generic way to traverse
@@ -7,7 +7,7 @@
  *
  * @example
  * ```typescript
- * import { parseWhereExpression, parseOrderByExpression } from '@tanstack/query-db-collection'
+ * import { parseWhereExpression, parseOrderByExpression } from '@tanstack/db'
  *
  * queryFn: async (ctx) => {
  *   const { limit, where, orderBy } = ctx.meta?.loadSubsetOptions ?? {}
@@ -27,7 +27,7 @@
  * ```
  */
 
-import type { IR, OperatorName } from "@tanstack/db"
+import type { IR, OperatorName } from "../index.js"
 
 type BasicExpression<T = any> = IR.BasicExpression<T>
 type OrderBy = IR.OrderBy
--- a/.changeset/expression-helpers-queryfn.md
+++ b/.changeset/expression-helpers-queryfn.md
@@ -1,10 +1,11 @@
 ---
+"@tanstack/db": patch
 "@tanstack/query-db-collection": patch
 ---
 
 Add expression helper utilities for parsing LoadSubsetOptions in queryFn.
 
-When using `syncMode: 'on-demand'`, query collections now provide helper functions to easily parse where clauses, orderBy, and limit predicates into your API's format:
+When using `syncMode: 'on-demand'`, TanStack DB now provides helper functions to easily parse where clauses, orderBy, and limit predicates into your API's format:
 
 - `parseWhereExpression`: Parse where clauses with custom handlers for each operator
 - `parseOrderByExpression`: Parse order by into simple array format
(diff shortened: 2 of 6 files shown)


packages/db/src/query/expression-helpers.ts
```diff
@@ -0,0 +1,385 @@
...
+   * - gte: greater than or equal (>=)
+   * - lt: less than (<)
+   * - lte: less than or equal (<=)
+   * - and: logical AND
+   * - or: logical OR
+   * - in: IN clause
+   */
+  handlers: {
+    [operator: string]: (...args: Array<any>) => T
+  }
```
samwillis, line 71: can we improve the type here such that it knows the operators that the IR supports?

packages/query-db-collection/src/expression-helpers.ts
```diff
@@ -0,0 +1,385 @@
...
+
+import type { IR } from "@tanstack/db"
+
+type BasicExpression<T = any> = IR.BasicExpression<T>
+type OrderBy = IR.OrderBy
+
+/**
+ * Represents a simple field path extracted from an expression
+ */
+export type FieldPath = Array<string>
```
samwillis: We may want to support accessing elements in a array:

```suggestion
export type FieldPath = Array<string | number>
```

packages/db/src/query/expression-helpers.ts
```diff
@@ -0,0 +1,385 @@
...
+ * @example
+ * ```typescript
+ * const comparisons = extractSimpleComparisons(where)
+ * // Returns: [
+ * //   { field: ['category'], operator: 'eq', value: 'electronics' },
+ * //   { field: ['price'], operator: 'lt', value: 100 }
+ * // ]
+ * ```
+ */
+export function extractSimpleComparisons(
```
samwillis, line 323: We may want this to throw if if finds something it can't extract such as an `or`

packages/query-db-collection/src/expression-helpers.ts
```diff
@@ -0,0 +1,385 @@
...
+      throw new Error(
+        `ORDER BY expression must be a field reference, got: ${clause.expression.type}`
+      )
+    }
+
+    return {
+      field,
+      direction: clause.compareOptions.direction,
+      nulls: clause.compareOptions.nulls,
+    }
```
samwillis: we need to return the local / string comparison options too

samwillis:
Mostly there except for the above comments. I will rebase #773 on this and use it.

samwillis:
@KyleAMathews there are a few tests failing too, looks like just types

KyleAMathews:
Cool! BTW, I've been thinking we'll probably want to move something like these utilities into a shared package as other collections will surely want to do something similar. The babel of the IR haha.

PR #795, opened 2025-11-11
Fix returning an array of txids from an Electric collection handler

### LINKED ISSUE #793: Electric collection mutations fail to wait for array of txids

**Describe the bug**
I'm experiencing an issue with awaiting multiple txids in an electric collection mutation handler. I'm getting a `TimeoutWaitingForTxIdError` as soon as one of the txids are seen on the stream, even though I'm well below the timeout.

**To Reproduce**

Define a collection like this:
```ts
  electricCollectionOptions({
    id: 'todos',
    getKey: (item) => item.id,
    shapeOptions: {
      url: '/api/todos',
      params: { table: 'todos' },
    },
      const txids = await Promise.all(
        transaction.mutations.map((mutation) =>
          api.todos.create(mutation.modified),
        )
      );

      return { txid: txids };
    }
  })
)
```

Run a mutation (single or multiple entries):
```ts
// ❌ Throws TimeoutWaitingForTxIdError before 5 second timeout
todosCollection.insert({ id: crypto.randomUUID(), title: 'New todo' })

// ❌ Throws TimeoutWaitingForTxIdError before 5 second timeout
todosCollection.insert([{ id: crypto.randomUUID(), title: 'New todo 1' }, { id: crypto.randomUUID(), title: 'New todo 2' }])
```

**Expected behavior**
I expect the mutations not to throw `TimeoutWaitingForTxIdError` unless the transaction takes > 5 seconds to appear on stream.

- OS: MacOS
- Browser: Chrome
- Version: 141.0.7390.123

**Additional context**
The waiting works fine when returning a single txid instead of a txid array in mutation handler.

This modification to [processMatchingStrategy](https://github.com/TanStack/db/blob/main/packages/electric-db-collection/src/electric.ts#L502) fixes the issue:
```
await Promise.all(result.txid.map((txid) => awaitTxId(txid)))
```

KyleAMathews:
Fixes #793

### DIFF
--- a/packages/electric-db-collection/src/electric.ts
+++ b/packages/electric-db-collection/src/electric.ts
@@ -506,7 +506,6 @@ export function electricCollectionOptions(
     if (result && `txid` in result) {
       // Handle both single txid and array of txids
       if (Array.isArray(result.txid)) {
-        // Use arrow function to avoid passing map's index as timeout parameter
         await Promise.all(result.txid.map((txid) => awaitTxId(txid)))
       } else {
         await awaitTxId(result.txid)


packages/electric-db-collection/src/electric.ts
```diff
@@ -506,7 +506,8 @@ export function electricCollectionOptions(
     if (result && `txid` in result) {
       // Handle both single txid and array of txids
       if (Array.isArray(result.txid)) {
-        await Promise.all(result.txid.map(awaitTxId))
+        // Use arrow function to avoid passing map's index as timeout parameter
```
samwillis: llm comment

PR #800, opened 2025-11-12
Auto append to static queryKey when in on-demand mode for Query Collection 

KyleAMathews:
When using a static queryKey with syncMode: 'on-demand', the system now automatically appends serialized LoadSubsetOptions to create unique cache keys for different predicate combinations.

This fixes an issue where static queryKeys in on-demand mode would cause all live queries with different predicates to share the same cache entry, defeating the purpose of predicate push-down.

Changes:
- Added serialization utilities for LoadSubsetOptions (serializeLoadSubsetOptions, serializeExpression, serializeValue)
- Modified createQueryFromOpts to automatically append serialized predicates when queryKey is static and syncMode is 'on-demand'
- Function-based queryKeys continue to work as before
- Eager mode with static queryKeys unchanged (no automatic serialization)

Tests:
- Added comprehensive test suite for static queryKey with on-demand mode
- Tests verify different predicates create separate cache entries
- Tests verify identical predicates reuse the same cache entry
- Tests verify eager mode behavior unchanged
- All existing tests pass

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/query-db-collection/src/query.ts
+++ b/packages/query-db-collection/src/query.ts
@@ -7,12 +7,12 @@ import {
   QueryKeyRequiredError,
 } from "./errors"
 import { createWriteUtils } from "./manual-sync"
+import { serializeLoadSubsetOptions } from "./serialization"
 import type {
   BaseCollectionConfig,
   ChangeMessage,
   CollectionConfig,
   DeleteMutationFnParams,
-  IR,
   InsertMutationFnParams,
   LoadSubsetOptions,
   SyncConfig,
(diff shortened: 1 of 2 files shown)


packages/query-db-collection/src/query.ts
```diff
@@ -304,6 +305,119 @@ class QueryCollectionUtilsImpl {
   }
 }
 
+/**
+ * Serializes LoadSubsetOptions into a stable, hashable format for query keys
+ * @internal
+ */
+function serializeLoadSubsetOptions(
```
kevin-dp: Let's move all the serialization functions into a utility file

packages/query-db-collection/src/query.ts
```diff
@@ -304,6 +305,119 @@ class QueryCollectionUtilsImpl {
...
+
+  if (options.where) {
+    result.where = serializeExpression(options.where)
+  }
+
+  if (options.orderBy?.length) {
+    result.orderBy = options.orderBy.map((clause) => ({
+      expression: serializeExpression(clause.expression),
+      direction: clause.compareOptions.direction,
+      nulls: clause.compareOptions.nulls,
```
kevin-dp: This object only contains the `direction` and `nulls` but is missing the other options from `StringCollationConfig`:
```ts
export type CompareOptions = StringCollationConfig & {
  direction: OrderByDirection
  nulls: `first` | `last`
}
```

packages/query-db-collection/src/query.ts
```diff
@@ -304,6 +305,119 @@ class QueryCollectionUtilsImpl {
   }
 }
 
+/**
+ * Serializes LoadSubsetOptions into a stable, hashable format for query keys
+ * @internal
+ */
+function serializeLoadSubsetOptions(
+  options: LoadSubsetOptions | undefined
+): unknown {
```
kevin-dp: Why is the return type typed as `unknown` ? Shouldn't this be `string | null`.
Actually, i'm wondering why we return `null` instead of undefined. I think the code would be slightly simplier if we return `undefined.
So the return type would be `string | undefined`.

PR #808, opened 2025-11-13
Fix: don't purge Direct Writes (Query Collection) when query results change

KyleAMathews:
Fixes issue where items inserted via Direct Writes (writeInsert/writeUpsert) were incorrectly purged when query results changed and didn't include those items.

**Root Cause:**
The reference counting system (queryToRows/rowToQueries) tracks which items are in each query result. When a query result doesn't include an item and the item's reference count is 0, it gets deleted. Direct Write items were never added to this tracking system, so they always had a reference count of 0 and were purged when not in query results.

**Solution:**
- Added `directWriteKeys` Set to track items inserted via Direct Writes
- Modified purging logic in `handleQueryResult()` and `cleanupQuery()` to skip items in the `directWriteKeys` Set
- Updated `performWriteOperations()` to add/remove keys from `directWriteKeys`:
  - Insert operations add the key
  - Delete operations remove the key
  - Upsert operations add the key if inserting new item

**Changes:**
- query.ts: Added directWriteKeys Set and modified purging checks
- manual-sync.ts: Track direct write keys in performWriteOperations
- query.test.ts: Added test case to verify Direct Writes are protected

This ensures directly-written items persist until explicitly deleted via writeDelete(), regardless of query result changes.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/.changeset/fix-direct-write-purging.md
+++ b/.changeset/fix-direct-write-purging.md
@@ -1,5 +1,5 @@
 ---
-'@tanstack/query-db-collection': patch
+"@tanstack/query-db-collection": patch
 ---
 
 Fix Direct Write items being purged when query results change
@@ -7,14 +7,16 @@ Fix Direct Write items being purged when query results change
 Items inserted via Direct Writes (`writeInsert`, `writeUpsert`) are now protected from automatic deletion when query results change. Previously, when using `syncMode: 'on-demand'` or dynamic query keys, directly-written items would be incorrectly purged if they didn't appear in new query results.
 
 **What changed:**
+
 - Direct Write items now persist until explicitly deleted via `writeDelete()`
 - Items are tracked separately from query results to prevent premature garbage collection
 - This ensures EventSource updates, real-time data, and manual writes remain in the collection
 
 **Example scenario that's now fixed:**
+
 ```typescript
 // Add item via Direct Write (e.g., from EventSource)
-collection.utils.writeInsert({ id: '4', message: 'New event' })
+collection.utils.writeInsert({ id: "4", message: "New event" })
 
 // Query result changes (different filter/pagination)
 await collection.utils.refetch()


packages/query-db-collection/src/manual-sync.ts
```diff
@@ -38,6 +38,7 @@ export interface SyncContext<
   begin: () => void
   write: (message: Omit<ChangeMessage<TRow>, `key`>) => void
   commit: () => void
+  directWriteKeys?: Set<TKey>
```
kevin-dp, line 41: How come this is optional? Don't we always need it for query collections?

packages/query-db-collection/src/manual-sync.ts
```diff
@@ -195,6 +200,8 @@ export function performWriteOperations<
             type: `insert`,
             value: resolved,
           })
+          // Track this key as a direct write (only for new inserts)
```
kevin-dp, line 203: This means that an optimistic update will be GCed if the row is deleted on the backend. Is that the behaviour we want? i.e. backend-delete wins over optimistic update. I'm flagging this up because for inserts this PR ensures that the insert wins, but for updates it's still delete wins.

packages/query-db-collection/src/query.ts
```diff
@@ -780,7 +784,9 @@ export function queryCollectionOptions(
             const newItem = newItemsMap.get(key)
             if (!newItem) {
               const needToRemove = removeRow(key, hashedQueryKey) // returns true if the row is no longer referenced by any queries
-              if (needToRemove) {
+              // Don't remove items that were inserted via direct writes
+              // They should only be removed via explicit writeDelete calls
+              if (needToRemove && !directWriteKeys.has(key)) {
```
kevin-dp, line 789: @claude So now when we do an optimistic insert and the backend deleted that row, we keep the item in the collection and don't actually delete it. Will this row ever be deleted? The only delete i could find is `ctx.directWriteKeys?.delete(op.key)` in `manual-sync.ts` when processing an optimistic delete. So i think this breaks GC in this case:

1. Server has a row (say row A) and this row is synced to the client
2. The client optimistically inserts row A (inserting it into directWriteKeys)
3. The server deletes row A
4. The client's queryFn runs again and notices the deleted row. Ref count drops to 0 but row is *not* GCed because row A is in the directWriteKeys
5. The optimistic insert of row A reaches the backend and row A is inserted again in the DB
6. The client's queryFn notices row A again and updates the ref count to 1
7. The server deletes row A
8. The client's queryFn notices the delete of row A and the ref count drops to 0 but row A still cannot be deleted from the local collection because it is still in the directWriteKeys (it was never removed)...

kevin-dp: I was looking into another PR and realized that these direct writes aren't optimistic writes but are writes directly into the synced data. What's the use case for this? If we write it directly into synced data i'm a bit puzzled about how to handle GC properly. It feels like it somehow needs to integrate with the ref counting system but i'm not sure how.

byudaniel: In my case I'm getting updates from Server Sent Events and writing it directly into the DB. The issue I was having before this PR is my updates were getting garbage collected.

samwillis:
I had a chat with @KyleAMathews about this PR and we are going to close it for now.

This would change the api contract - the intention is that the server is the source of truth and that the direct write api is an optimisation to skip the refetch, but a later refetch should return all the data written with the direct writes. direct writes are not intended as an alternative to returning the data from the queryFn.

@byudaniel for your use case it would be better to drop down a level and implement your own collection directly - there are docs here https://tanstack.com/db/latest/docs/guides/collection-options-creator - but essentially you could do this:

```ts
  // ... other options ...
  sync: {
    sync: ({begin, write, commit, markReady}) => {
      const evtSource = new EventSource("my/api/");
      evtSource.onmessage = (e) => {
        begin();
        write(e.data);
        commit();
      };
      return () => {
        evtSource.close()
      }
    }
  }
})
```

byudaniel:
@samwillis Dang, I get the dilemma here and will look into making my own collection.

From an API perspective it may make sense to throw an error if the user attempts a direct write where the `queryKey` is dynamic. My use-case worked well until I tried enabling `syncMode: 'on-demand'` with a dynamic `queryKey` and then it failed in a surprising way.

PR #809, opened 2025-11-13
Expose the full response from queryFn on utils.queryData

KyleAMathews:


packages/query-db-collection/src/query.ts
```diff
@@ -156,6 +156,7 @@ export interface QueryCollectionUtils<
   TKey extends string | number = string | number,
   TInsertInput extends object = TItem,
   TError = unknown,
+  TQueryData = any,
```
kevin-dp, line 159: Should this be `unknown` rather than `any`?

packages/query-db-collection/src/query.ts
```diff
@@ -206,6 +226,7 @@ interface QueryCollectionState {
   lastError: any
   errorCount: number
   lastErrorUpdatedAt: number
+  queryData: any
```
kevin-dp, line 229: Shouldn't this be `TQueryData` instead of `any`?

packages/query-db-collection/src/query.ts
```diff
@@ -574,6 +599,7 @@ export function queryCollectionOptions(
     lastError: undefined as any,
     errorCount: 0,
     lastErrorUpdatedAt: 0,
+    queryData: undefined as any,
```
kevin-dp, line 602: Do we really need to type it as `any`?

docs/collections/query-collection.md
```diff
@@ -135,13 +295,35 @@ This is useful when:
...
+
+- `queryData`: The full response from `queryFn`, including metadata (see [Working with API Responses](#working-with-api-responses))
+- `lastError`: The last error encountered (if any)
+- `isError`: Whether the collection is in an error state
+- `errorCount`: Number of consecutive sync failures
+- `isFetching`: Whether the query is currently fetching
+- `isRefetching`: Whether the query is refetching in the background
+- `isLoading`: Whether the query is loading for the first time
+- `dataUpdatedAt`: Timestamp of the last successful data update
+- `fetchStatus`: Current fetch status (`'fetching'`, `'paused'`, or `'idle'`)
```
samwillis, line 325: With us having 1:n collection to query the `isXX` return true when any of the queries are in that state, `dataUpdatedAt` is the most recent query update time, and `fetchStatus` is an array of statuses.

KyleAMathews:
> note that this exposes the data from the last query function. So when used in on-demand it will be the last response for the most recent loadSubset call. I think we need to consider if this is the right API with the 1:n query:collection design.

Hmm yeah you're right. We need to all download all these ideas about query collection & step back and have a rethink.

PR #816, opened 2025-11-14
use `public readonly` to make `compareOptions` readable but not writable

KyleAMathews:
Probably fixes a typescript error.

### DIFF
--- a/packages/query-db-collection/tests/query.test-d.ts
+++ b/packages/query-db-collection/tests/query.test-d.ts
@@ -12,8 +12,8 @@ import { queryCollectionOptions } from "../src/query"
 import type {
   DeleteMutationFnParams,
   InsertMutationFnParams,
-  UpdateMutationFnParams,
   StringCollationConfig,
+  UpdateMutationFnParams,
 } from "@tanstack/db"
 
 describe(`Query collection type resolution tests`, () => {
@@ -459,20 +459,23 @@ describe(`Query collection type resolution tests`, () => {
 
     it(`should have compareOptions property accessible with schema transform`, () => {
       // Reproduces the exact scenario from Discord bug report
-      const schema = z
+      const _schema = z
         .object({
           name: z.string().min(1),
         })
         .transform((item) => ({ ...item, id: -1, blubb: `blubb` }))
 
+      type SchemaOutput = z.infer<typeof _schema>
+
       const options = queryCollectionOptions({
         queryKey: [`local-test-array`],
-        schema,
-        queryFn: async () => {
+        // When using schema with transform, we need to provide the explicit type
+        queryFn: async (): Promise<Array<SchemaOutput>> => {
           return [
             {
               name: `test`,
               id: 0,
+              blubb: `blubb`,
             },
           ]
         },


kevin-dp (changes requested):
After looking into this issue and the proposed fixed, it's not clear to me why the error occurs and why this fixes it. We should add a unit test that reproduces the type error. I tried adding one, but i was not able to reproduce this type error.

I don't like this fix. I would rather turn the `compareOptions` getter into a true readonly property on the `CollectionImpl` class instead of having to add it to the extends clause as is done in this fix.

KyleAMathews:
@kevin-dp we could ask the user to share their code? Also did you try looking at one of the examples?

IARI:
@KyleAMathews a few days ago on discord you mentioned, that my problem would be related to this.

If that is in fact the case, the TS issue I have is very easy to reproduce @kevin-dp.
Here's a minimal example:

```typescript
import { z } from 'zod'
import { queryCollectionOptions } from '@tanstack/query-db-collection'
import { QueryClient } from '@tanstack/query-core'

export const queryClient = new QueryClient()

const schema = z
  .object({
    name: z.string().min(1)
  })
  .transform((item) => ({ ...item, id: -1, blubb: 'blubb' }))

const options = queryCollectionOptions({
  queryKey: ['local-test-array'],
  schema,
  queryFn: async () => {
    return [{
        name: 'test',
        id: 0
    }]
  },
  getKey: (item) => item.id,
  queryClient
})
```

excerpt from `pnpm-lock.yaml`:
```yaml
'@tanstack/query-core':
  specifier: ^5.90.10
  version: 5.90.10
'@tanstack/query-db-collection':
  specifier: ^1.0.1
  version: 1.0.1(@tanstack/db@0.5.1(typescript@5.9.3))(@tanstack/query-core@5.90.10)(typescript@5.9.3)
zod:
  specifier: ^4.1.12
  version: 4.1.12
```

image

KyleAMathews:
@IARI this PR should now fix your issue 👍 I've pushed some changes since @kevin-dp reviewed it earlier

kevin-dp:
The type error that was reported initially seems to be different from the one reported by @IARI. In the future, let's fix separate bugs in separate PRs.

Regarding the original type error, i indeed tried to reproduce it but i couldn't:
```ts
const todoSchema = z.object({
  id: z.string(),
  title: z.string(),
  completed: z.boolean(),
})

type Todo = z.infer<typeof todoSchema>

const queryClient = new QueryClient()

  queryCollectionOptions({
    queryKey: ["todos"],
    queryFn: async () => {
      const response = await fetch("/api/todos")
      return response.json() as Promise<Todo[]>
    },
    queryClient,
    getKey: (item) => item.id,
  })
)
```
This type checks without errors. Note that i has to change the OP's typecast to `return response.json as Promise<Todo[]>` because `.json()` returns a promise.

PS: When fixing a bug let's always introduce a unit test reproducing the bug such that we can be sure the fix actually solves the bug and we don't introduce regressions. In this case, i think we were chasing a bug that does no(t) (longer) exist.

PR #817, opened 2025-11-14
exclude 'persisting' transactions from having their optimistic state re-applied during commitPendingTransactions()

### LINKED ISSUE #814: Query Collection "Incremental Updates" example never updates with server-sent data.

**Describe the bug**
[Query Collection Example Incremental Updates](https://tanstack.com/db/latest/docs/collections/query-collection#example-incremental-updates) does not ever update the collection with the new server data. The idea is that the server data is supposed to load in to the collection once the insert is complete but in practice the collection only retains the old, client-only data.

**To Reproduce**
I've created a codesandbox that illustrates the issue clearly. You can see that the "id" never switches from the client-side negative number to the server-side positive number.

https://codesandbox.io/p/sandbox/tanstack-db-bug-7txp3s?file=%2Fsrc%2FApp.tsx

**Expected behavior**
I would expect the collection data to correctly update with the server-sent data. (Or, does the documentation need to be changed with this no longer being the correct way to achieve this functionality?) 

Thanks!

KyleAMathews:
Fixes #814

When writeInsert() is called from within an onInsert handler, the transaction is in 'persisting' state. Previously, commitPendingTransactions() would re-apply optimistic state from persisting transactions after syncing server data, causing server-generated fields to be overwritten with optimistic client-side values.

This fix excludes 'persisting' transactions from having their optimistic state re-applied during commitPendingTransactions(). This ensures that when writeInsert() syncs server data (e.g., server-generated IDs), that data is not immediately overwritten by the optimistic state.

The flow now works correctly:
1. User inserts item with temporary ID -> optimistic state shows temp ID
2. Transaction commits -> onInsert handler called (state = 'persisting')
3. Handler calls writeInsert(serverItem) with real ID
4. commitPendingTransactions() syncs server data
5. Optimistic state from persisting transaction is NOT re-applied
6. UI now shows server data with real ID
7. Transaction completes -> optimistic state is cleared

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/db/src/collection/state.ts
+++ b/packages/db/src/collection/state.ts
@@ -669,22 +669,20 @@ export class CollectionStateManager<
       // after the truncate snapshot are preserved.
       for (const transaction of this.transactions.values()) {
         if (![`completed`, `failed`].includes(transaction.state)) {
+          // Skip re-applying optimistic state for persisting transactions entirely
+          // (not just for changed keys). When a transaction is persisting, its mutation
+          // handler may write server data via sync functions. That synced data should be
+          // visible immediately, not masked by stale optimistic state.
+          // EXCEPT during truncate operations where optimistic state should always be preserved.
+          if (!hasTruncateSync && transaction.state === `persisting`) {
+            continue
+          }
+
           for (const mutation of transaction.mutations) {
             if (
               this.isThisCollection(mutation.collection) &&
               mutation.optimistic
             ) {
-              // Skip re-applying optimistic state for persisting transactions
-              // if the mutation key was just synced (prevents overwriting fresh server data)
-              // EXCEPT during truncate operations where optimistic state should always be preserved
-              if (
-                !hasTruncateSync &&
-                transaction.state === `persisting` &&
-                changedKeys.has(mutation.key)
-              ) {
-                continue
-              }
-
               switch (mutation.type) {
                 case `insert`:
                 case `update`:


samwillis:
@claude can you add a test to validate that this change does what is intended. See the changeset for an example.

samwillis:
Having said that... on second thoughts I think this does break some expectations. Take an Electric collection as an example:
- the user mutates key A and the transaction is `persisting` while we send data to the server and await the transaction from come back via sync
- concurrently someone else has changed key A on the server and it syncs down with a state from *before* our mutation
- in this case we enter this new branch `state=='persisting'` no truncate, and this key has a change that is synced, and will then not show the optimistic state.

We would expect that optimistic state is kept untill the mutation is complete.

@claude can you try that in a test? and can you check the original issue #814 and consider what that was and if this actually fixes the issue. I'm suspicious that the optimistic state is not removed after the mutation ends, not that the synced state is not applied. It could be a race where the local fake synced mutation is happening during the mutation fn. I can see in the code sandbox there is a bug.

samwillis:
@claude I have pasted the full reproduction code below. What I dint think we have assertaind is exactly what is happening. When the async `onInsert` function returns, the collection should remove the optimistic sate from that change - its transaction will be marked as `complete`. We should then be showing **all** synced state, even that applied during the mutation function. So its not that we need to apply the "synced" state when it is first written using the `utils.writeInsert`, its that it should become visible when the `onInsert` function promise resolves and its transaction is complete. Can you look into that?

```ts
import { QueryClient } from "@tanstack/query-core";
import { queryCollectionOptions } from "@tanstack/query-db-collection";
import { createCollection, useLiveQuery } from "@tanstack/react-db";
import "./styles.css";

const queryClient = new QueryClient();

async function sleep(timeMs: number) {
  return new Promise((resolve) => {
    setTimeout(resolve, timeMs);
  });
}

type Todo = {
  id: number;
  slug: string;
  checked: boolean;
  title: string;
  createdAt: string;
};
let nextId = 1;
const todos: Todo[] = [];
async function fetchTodos() {
  await sleep(500);
  return todos;
}
async function createTodos(newTodos: Todo[]) {
  await sleep(300);

  const savedTodos = newTodos.map((todo) => {
    const savedTodo: Todo = {
      ...todo,
      id: nextId++,
      createdAt: new Date().toISOString(),
    };
    return savedTodo;
  });
  todos.push(...savedTodos);
  return savedTodos;
}

  queryCollectionOptions({
    queryKey: ["todos"],
    queryFn: fetchTodos,
    queryClient,
    getKey: (item) => item.slug,

      const newItems = transaction.mutations.map((m) => m.modified);

      // Send to server and get back items with server-computed fields
      const serverItems = await createTodos(newItems);

      // Sync server-computed fields (like server-generated IDs, timestamps, etc.)
      // to the collection's synced data store
      todosCollection.utils.writeBatch(() => {
        serverItems.forEach((serverItem) => {
          console.log("writing insert from server item", serverItem);
          todosCollection.utils.writeInsert(serverItem);
        });
      });

      // Skip automatic refetch since we've already synced the server response
      // (optimistic state is automatically replaced when handler completes)
      return { refetch: false };
    },
  })
);

  const { data } = useLiveQuery((q) =>
    q
      .from({ todo: todosCollection })
      .orderBy(({ todo }) => todo.createdAt, "asc")
  );

  return (
    <div className="App">
      <h1>Todos</h1>
      <div>
        <p>
          Notice how the "id" never changes from a negative number to the
          correct, server generated number...
        </p>
        <p>Open console to see the server generated data.</p>
        <p>
          Note: with how the "backend" is being simulated here, there can be
          issues with live refresh. You may need to reload the sandbox if adding
          a todo doesn't work.
        </p>
      </div>
      <ul>
        {data.map((todo) => (
          <li key={todo.slug}>
            id: {todo.id}, slug: {todo.slug}
          </li>
        ))}
      </ul>
      <button
        type="button"
        onClick={() =>
          todosCollection.insert({
            id: -Math.round(Math.random() * 5000),
            title: "Task",
            slug: crypto.randomUUID(),
            checked: false,
            createdAt: new Date().toISOString(),
          })
        }
      >
        Add Todo
      </button>
    </div>
  );
}

```

samwillis:
Um @claude, thats not even worse than before... see my previous comments about an electric collection. We absolutely *must* apply *all* optimistic state thats in persisting state. The bug you have not fixed is that we are not reapplying the synced state or removing the optimistic state after the transition becomes complete.

Last try, if you can't fix this I will close this PR and work on it locally.

samwillis:
replaced by #879

PR #826, opened 2025-11-15
Update Solid DB to enable Suspense

fezproof:
🎯 Changes

The main change in this PR is that tanstack solid db now has Suspense support by default. This is more inline with the rest of the solid ecosystem and tanstack query.

```ts
// Use Suspense boundaries
const todosQuery = useLiveQuery((q) =>
  q.from({ todos: todoCollection })
)

return (
  <Suspense fallback={<div>Loading...</div>}>
    <For each={todosQuery()}>
      {(todo) => <li key={todo.id}>{todo.text}</li>}
    </For>
  </Suspense>
)
```

Updated the API for solid js to be more inline with solid js createResource. This is a breaking change to the API, as status, collection, and the various loading states are now getters instead of Accessors. This actually kinda simplifies the API, especially for collections.

```ts
// Access collection methods directly
const existingQuery = useLiveQuery(() => existingCollection)

// Use collection for mutations
const handleToggle = (id) => {
  existingQuery.collection.update(id, draft => { draft.completed = !draft.completed })
}
```

### DIFF
--- a/packages/solid-db/tests/useLiveQuery.test.tsx
+++ b/packages/solid-db/tests/useLiveQuery.test.tsx
@@ -1877,18 +1877,18 @@ describe(`Query Collections`, () => {
 
       function TestComponent() {
         const query = useLiveQuery((q) =>
-          q
-            .from({ persons: collection })
-            .select(({ persons }) => ({
-              id: persons.id,
-              name: persons.name,
-            })),
+          q.from({ persons: collection }).select(({ persons }) => ({
+            id: persons.id,
+            name: persons.name,
+          })),
         )
 
         return (
           <ul data-testid="list">
             <For each={query()}>
-              {(person) => <li data-testid={`person-${person.id}`}>{person.name}</li>}
+              {(person) => (
+                <li data-testid={`person-${person.id}`}>{person.name}</li>
+              )}
             </For>
           </ul>
         )
@@ -1928,12 +1928,10 @@ describe(`Query Collections`, () => {
 
       function TestComponent() {
         const query = useLiveQuery((q) =>
-          q
-            .from({ persons: collection })
-            .select(({ persons }) => ({
-              id: persons.id,
-              name: persons.name,
-            })),
+          q.from({ persons: collection }).select(({ persons }) => ({
+            id: persons.id,
+            name: persons.name,
+          })),
         )
 
         return (


KyleAMathews:
@MAST1999 what do you think?

KyleAMathews:
Also relevant — working on this https://github.com/TanStack/db/pull/842/files#diff-6bb6eea8d2771d74392090da8774873368fa5dca9cb915e442adc3a05e3b9b89

packages/solid-db/src/useLiveQuery.ts
```diff
@@ -250,79 +261,110 @@ export function useLiveQuery(
...
-              switch (change.type) {
-                case `insert`:
-                case `update`:
-                  state.set(change.key, change.value)
-                  break
-                case `delete`:
-                  state.delete(change.key)
-                  break
-              }
+  createEffect(() => {
```
MAST1999: Can we use `createComputed` here?

The effect fires after render, so I think we should use create computed.

MAST1999:
Thanks for the work.

I've been meaning to fix this for some time.
Other than the comment I left the changes look good.

The only other thing that worries me is changing the returned interface. If someone is already using the old one they'll have a lot of work migrating to the new output.

fezproof:
I don’t see your comment @MAST1999 
I could add data back as an accessor to enable a more stable update, and mark it deprecated? Should make updating a bit smoother

MAST1999:
> I could add data back as an accessor to enable a more stable update, and mark it deprecated? Should make updating a bit smoother

Sounds good!

The comment is on code, I left a review.
Why can't we use `createComputed` instead of `createEffect`. The effect will run after the render and it might cause more renders afterward.

fezproof:
> I could add data back as an accessor to enable a more stable update, and mark it deprecated? Should make updating a bit smoother

This is done now
 
> The comment is on code, I left a review. 

No reviews have been made on this PR, perhaps you forgot to hit submit @MAST1999 

> Why can't we use `createComputed` instead of `createEffect`. The effect will run after the render and it might cause more renders afterward.

If we use createComputed, it will run on render which means it also runs in SSR. Currently tanstack db doesn't support SSR, but this could be a huge pain in the future. createEffect also is designed for this kind of subscription to events, as it does not block rendering. Solid isn't React, only the data that changes will be re-rendered, making this not really a problem.

KyleAMathews:
We are going to be adding support for SSR soon to react btw. The patterns should be easy I assume to port to solid.

MAST1999:
> No reviews have been made on this PR, perhaps you forgot to hit submit @MAST1999

Damn it, I don't use GitHub too much, my bad. I indeed missed it.

LGTM

MAST1999:
> We are going to be adding support for SSR soon to react btw. The patterns should be easy I assume to port to solid.

Yep, should be really easy.
Give me a heads up and I'll update the Solid adapter.

fezproof:
@MAST1999 any progress on this?

MAST1999:
Sorry @fezproof been busy with a few things.

Right now, I've been trying to run the Solidjs todo example based on your changes and the latest changes on main branch to verify things, but I can't get it to run.

I get this error:

```sh
TypeError: (intermediate value).routerEntry.getRouter is not a function
    at getRouter (/home/mast/workspace/temp/mast-db/node_modules/.pnpm/@tanstack+start-server-core@1.139.14/node_modules/@tanstack/start-server-core/src/createStartHandler.ts:108:57)
    at async handleServerRoutes (/home/mast/workspace/temp/mast-db/node_modules/.pnpm/@tanstack+start-server-core@1.139.14/node_modules/@tanstack/start-server-core/src/createStartHandler.ts:348:18)
    at async eval (/home/mast/workspace/temp/mast-db/node_modules/.pnpm/@tanstack+start-server-core@1.139.14/node_modules/@tanstack/start-server-core/src/createStartHandler.ts:223:34) {
  cause: TypeError: (intermediate value).routerEntry.getRouter is not a function
      at getRouter (/home/mast/workspace/temp/mast-db/node_modules/.pnpm/@tanstack+start-server-core@1.139.14/node_modules/@tanstack/start-server-core/src/createStartHandler.ts:108:57)
      at async handleServerRoutes (/home/mast/workspace/temp/mast-db/node_modules/.pnpm/@tanstack+start-server-core@1.139.14/node_modules/@tanstack/start-server-core/src/createStartHandler.ts:348:18)
      at async eval (/home/mast/workspace/temp/mast-db/node_modules/.pnpm/@tanstack+start-server-core@1.139.14/node_modules/@tanstack/start-server-core/src/createStartHandler.ts:223:34),
  status: 500,
  statusText: undefined,
  headers: undefined,
  data: undefined,
  body: undefined,
  unhandled: true
}
```

MAST1999:
I managed to fix the issue with tanstack start, it was updated, and it had some breaking changes, but I don't have access to your repo, so I can't push my changes there.

fezproof:
All good, I’ll pull main and resolve conflicts and get this sorted

fezproof:
Sorry for the delay @MAST1999 want to take a look again?

MAST1999:
> Sorry for the delay @MAST1999 want to take a look again?

Did you try to run the todo example for Solid to make sure it works, since it's the only example we have for it?

fezproof:
@MAST1999 updated the solid example to work again (it was pretty broken) + moved to new api. Works will with and without suspense

fezproof:
Would love to do a follow up PR to get that wikipedia demo that @KyleAMathews was showing off on twitter the other day with durable streams

MAST1999:
LGTM!

let's go 😁

fezproof:
@MAST1999 is there anything more that I need to do to get this merged? I think the workflows have to be run to make it valid

MAST1999:
Well, I don't think I have the privileges to do it.

Will need @KyleAMathews to take another look.

PR #832, opened 2025-11-17
Fix claude action permissions

kevin-dp:


kevin-dp:
@claude can you reply?

Issue #827, opened 2025-11-15
`on-demand` prunes queryFn predicate for stale items


TimFL:
- [x] I've validated the bug against the latest version of DB packages

**Describe the bug**
When the result of a live query collection is stale, it is still being used to fulfill other live queries with the stale data.

In my testing, I query for users using their `id` property (which is also the item key), then run multiple live queries with various `id` conditions. I noticed that if I have overlapping `id` predicates in them, the later live queries return the stale user entries from previous stale queries and prune the affected user entity `id` predicates from the currently running queryFn.

**To Reproduce**

1) run query 1 asking for a specific user by `id`

```ts
const { data, isLoading } = useLiveQuery((q) =>
	q
		.from({ user: userCollection })
		.where(({ user }) => eq(user.id, `123`))
		.findOne()
);
```
Result: the REST API is successfully queried for the user id `123`

2) wait for the query stale time to hit (e.g. for testing, specify `3_000` as the default staleTime for the collection

3) mount query 2 asking for other users by `id`

```ts
const { data, isLoading } = useLiveQuery((q) =>
	q
		.from({ user: userCollection })
		.where(({ user }) => inArray(user.id, [`123`, `456`, `789`]))
);
```
Result: the REST API is *only* queried for user ids `456` and `789`, the respective queryFn of the userCollection only ever gets a filter for these 2 ids (the id `123` is not included in the parsed result set). The live query, however, returns the 3 respective users (including id `123`), with the 2 new ids being fresh from the REST API and the previously queried id being from the cache.

This is the collection at hand:
```ts
export const userCollection = createCollection(
	queryCollectionOptions<DbSyncMetaRecord<UserSchema>>({
		id: 'users',
		queryClient,
		getKey: (record) => record.id,
		syncMode: 'on-demand',
		startSync: false,
		queryKey: (opts) => ['users', JSON.stringify(opts || {})],
		queryFn: async (ctx) => {
                        // get existing data or empty array
			const existingData = (queryClient.getQueryData(ctx.queryKey) || []) as DbSyncMetaRecord<UserSchema>[];
			if (!ctx.meta?.loadSubsetOptions) return existingData; // return existing data or empty array
			const parsed = parseLoadSubsetOptions(ctx.meta?.loadSubsetOptions);
			const ids: string[] = [];
                        // the following portion is me trying to debug where the missing id went
			for (let i = 0; i < parsed?.filters?.length; i++) {
				const { field, value } = parsed.filters[i];
[...]

PR #835, opened 2025-11-17
Fix: do not dedupe `loadSubset` calls in query collection

kevin-dp:
Fixes https://github.com/TanStack/db/issues/827 by temporarily disabling deduplication in the query collection. We do want to re-introduce it properly in the near future as described in https://github.com/TanStack/db/issues/836.

The deduplication logic of `loadSubset` calls in the query collection is currently causing problems because upon deduplicating a `loadSubset` call we do not track which other queries are the reason why we can deduplicate this one. As a result, when the original query (or queries) is/are GCed this deduplicated query doesn't know that and continues to load stale data from the local collection (and the local collection won't receive updates to these rows anymore).

Keeping track of which queries are deduplicated because of which other queries would be quite complex. So instead, we remove this optimization for now and will rely on the query keys to deduplicate queries.

### DIFF
--- a/packages/query-db-collection/tests/query.test.ts
+++ b/packages/query-db-collection/tests/query.test.ts
@@ -3424,7 +3424,7 @@ describe(`QueryCollection`, () => {
       expect(collection.has(`4`)).toBe(false)
     })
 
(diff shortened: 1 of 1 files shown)


KyleAMathews (approved):
Thanks for investigating! Let's get this merged/shipped to unbreak people.

PR #843, opened 2025-11-17
Deprecate magic return APIs from collection handlers

KyleAMathews:
## Summary

Deprecates "magic" mutation handler return values in favor of explicit utility calls. Instead of returning `{ txid }` or relying on auto-refetch, developers now explicitly call `collection.utils.awaitTxId()` or `collection.utils.refetch()` within their handlers.

**User impact:** Deprecation warnings appear when using old patterns. Existing code continues to work throughout the 0.x series, with removal planned for v1.0.

---

## Root Cause

The implicit return-value API created confusing "magic" behavior:

- **ElectricCollection**: Returning `{ txid }` would magically wait for sync - but this control flow was hidden from developers and hard to customize (timeouts, error handling)
- **QueryCollection**: Auto-refetch after every handler made it hard to optimize performance or batch operations

Both patterns violated the principle of explicit over implicit - developers couldn't easily see or control what happened after their mutation completed.

### 1. Runtime Deprecation Warnings

Added `warnOnce` utility that logs deprecation messages once per key (prevents console spam):

```typescript
warnOnce('electric-txid-return', '[TanStack DB] DEPRECATED: Returning { txid }...')
```

### 2. Type System Changes

Handler types now use a union that prefers `void` but accepts `unknown` for backward compatibility:

```typescript
onInsert?:
  | InsertMutationFn<T, TKey, TUtils, void>      // Preferred
  | InsertMutationFn<T, TKey, TUtils, unknown>   // @deprecated
```

### 3. Documentation Updates

All docs and examples updated to show the new explicit pattern:

```typescript
// Before (deprecated)
  const result = await api.create(item)
  return { txid: result.txid }  // Magic!
}

// After (explicit)
  const result = await api.create(item)
  await collection.utils.awaitTxId(result.txid)  // Clear intent
}
```

## Key Changes

| Change | Before | After |
|--------|--------|-------|
| Handler return type default | `Promise<any>` | `Promise<void>` |
| ElectricCollection sync | Return `{ txid }` | Call `awaitTxId(txid)` |
| QueryCollection refetch | Auto-refetch always | Call `refetch()` explicitly |
| `awaitTxId` default timeout | 5000ms | 15000ms |
| `awaitMatch` default timeout | 3000ms | 15000ms |

## Non-goals

- **Not removing deprecated behavior** - that's v1.0
- **Not changing runtime behavior** - deprecated patterns still work identically
- **Not adding new features** - purely preparing the migration path

## Trade-offs

**Why explicit over implicit?**
- Easier to understand code flow
- Better error handling (try/catch around explicit calls)
- More control over timeouts and retry logic
- Clearer intent in code reviews

**Why deprecate now vs later?**
- Gives users the full 0.x series to migrate
- Warnings help identify code that needs updating
- Documentation can focus on the new pattern

---

## Verification

```bash
# Run affected package tests
pnpm vitest run packages/electric-db-collection/tests/electric.test.ts
pnpm vitest run packages/query-db-collection/tests/query.test.ts
pnpm vitest run packages/db/tests/collection-lifecycle.test.ts

# Type check
pnpm tsc --noEmit -p packages/db/tsconfig.json
pnpm tsc --noEmit -p packages/electric-db-collection/tsconfig.json
pnpm tsc --noEmit -p packages/query-db-collection/tsconfig.json
```

## Files Changed

| File | Change |
|------|--------|
| `packages/db/src/types.ts` | Handler types default to `void`, union with `unknown` for deprecated returns |
| `packages/db/src/utils.ts` | Add `warnOnce` and `resetWarnings` utilities |
| `packages/db/src/index.ts` | Export `warnOnce` and `resetWarnings` |
| `packages/electric-db-collection/src/electric.ts` | Add deprecation warnings, increase timeouts, clarify JSDoc |
| `packages/query-db-collection/src/query.ts` | Add deprecation warnings for auto-refetch, extract helper |
| `docs/**` | Update all examples to use explicit sync pattern |
| `.changeset/deprecate-handler-return-values.md` | Changeset with migration guide |

---

## Checklist

## Release Impact

---

## Summary by CodeRabbit

- **Deprecations**
  - Mutation handler return values, including transaction IDs, are deprecated.
  - Electric Collection handlers should await `collection.utils.awaitTxId(...)` instead of returning transaction IDs.
  - Automatic Query Collection refetching is transitional; use `collection.utils.refetch()` explicitly.

- **Compatibility**
  - Legacy handler results remain supported during the transition, with one-time deprecation warnings.
  - `{ refetch: false }` can suppress automatic refetching.

- **Documentation**
  - Updated guides, examples, and API references with the recommended synchronization patterns.
  - Increased default synchronization wait timeouts to 15 seconds.


samwillis:
I'm in favour of this change, it makes everything much more explicit. My only hesitation is on the query collection where it becomes a little more verbose with the need to refetch the majority of the time... but that explicitness does ensure that the user is aware of what is actually happening.

on the whole, I think it's worth making this change.

docs/guides/collection-options-creator.md
```diff
@@ -360,23 +362,23 @@ export function myCollectionOptions<TItem extends object>(
...
-      const result = await config.onInsert!(params)
-      // Additional sync coordination logic
-      return result
-    } : undefined
+
+    // Pass through user-provided handlers
+    // Users handle sync coordination in their own handlers
+    onInsert: config.onInsert,
+    onUpdate: config.onUpdate,
+    onDelete: config.onDelete
```
tannerlinsley, line 393: This is proof that this is a cleaner approach. Well done.

packages/db/src/types.ts
```diff
@@ -486,7 +500,9 @@ export interface BaseCollectionConfig<
   /**
    * Optional asynchronous handler function called before an insert operation
    * @param params Object containing transaction and collection information
-   * @returns Promise resolving to any value
+   * @returns Promise that should resolve to void
+   * @deprecated Returning values from this handler is deprecated. Use collection utilities to refetch/sync.
```
kevin-dp: I believe that @deprecated is going to mark the `onInsert` function as deprecated which is not really the case. It's only a specific usage of it that will be deprecated.

To do it right i think we should have an overload for the old signature and an overload for the new signature. And then deprecate the old signature via docstring. Like in this artificial example:
```ts
// Deprecated overload (old usage)
 /** @deprecated This function no longer returns a value. */
export function configure(): ConfigObject;

// New overload (current usage)
export function configure(): void;

// Implementation
export function configure() {
  // ...
}
```

kevin-dp (approved):
LGTM. I left a comment about improving the `@deprecated` annotation in the function's docstring.

packages/db/src/types.ts
kevin-dp: @claude Address this comment.

.changeset/deprecate-handler-return-values.md
```diff
@@ -0,0 +1,51 @@
+---
+'@tanstack/db': major
```
KyleAMathews: @claude make these minor — we're < 1.0

KyleAMathews:
@claude throughly review all docs for any other docs that are using this now deprecated pattern & rewrite as needed to the new pattern

kevin-dp (approved):
Approved.

Issue #397, opened 2025-08-11
Bug in update/delete functions with localStorageCollection


pwalczyszyn:
When calling `update` or `delete` functions on local storage collection it always applies to last item in the collection (not the one with provided `id`).

This is how I configure the collection:
```typescript
export const measurementsCollection = createCollection(
  localStorageCollectionOptions({
    id: "measurements",
    storageKey: "app-measurements",
    getKey: (item) => item.id,
    schema: $Measurement,
  })
);
```

This is how I call update/delete fns:
```typescript
// update example
measurementsCollection.update(measurementId, draft => {
  draft.sync = true
})

// delete example
measurementsCollection.update(measurementId)
```

samwillis:
Hey, could you show me the full code where you are adding the items to the collection? I've not been able to reproduce this yet.

pwalczyszyn:
Sure, here it is:

```typescript
  // Mutations
  const measurementsMutation = useMutation<Measurement, Error, string>({
    mutationFn: async (projectId: string) => {
      // Creating new Measurement
      const measurement: Measurement = {
        id: uuid(),
        project_id: projectId,
        radar: (selectedRadar?.id || "radar1") as Measurement["radar"],
        damping: damping,
        status: "started",
        notes: null,
        start_time: new Date(),
        end_time: null,
        sync: true,
        coords: [
          [52.237174982588726, 21.093376139051927],
          [52.237174982588726, 21.093],
        ],
        created_at: new Date(),
        updated_at: null,
      };

      // Inserting new Measurement
      await measurementsCollection.insert(measurement).isPersisted.promise;

      // Returning new Measurement
      return measurement;
    },
    onSuccess: (data) => {
      // Invalidate and refetch
      // queryClient.invalidateQueries({ queryKey: ["projects", project.id] });
      navigate(`/projects/${project.id}/measurements/${data.id}`);
      handleClose();
    },
    onError: (error) => {
      // TODO: Show error message
      console.error("Error creating measurement:", error);
      handleClose();
    },
  });
```

discoverlance-com:
I am facing a similar error, updates work alright but when i try to delete the data, the data in localStorage (when you check the browser dev tools) stays the same. Even though, the UI updates optimistically after deleting the data. I have a video of it that I was trying to send here, but github does not allow it here.

This is the code and the deletion is quite simple:

```ts
// lib/db.ts
export const notesCollection = createCollection(
  localStorageCollectionOptions({
    schema: notesSchema,
[...]

PR #845, opened 2025-11-18
Fix localStorage collections for numeric IDs

KyleAMathews:
Fixes https://github.com/TanStack/db/issues/397

Fix localStorage collections failing to update/delete items with numeric IDs. Previously, operations would target the wrong item or fail entirely when using numeric IDs (e.g., `id: 1`, `id: 2`) after the page reloaded, due to a type mismatch between numeric keys in memory and stringified keys from localStorage. Numeric keys are now prefixed with `__number__` in storage to ensure consistent lookups.

### DIFF
--- a/packages/db/tests/local-storage.test.ts
+++ b/packages/db/tests/local-storage.test.ts
@@ -869,8 +869,8 @@ describe(`localStorage collection`, () => {
       const storedData = mockStorage.getItem(`todos`)
       expect(storedData).toBeDefined()
       const parsed = JSON.parse(storedData!)
-      expect(parsed[`tx-1`].data.title).toBe(`Manual Tx Insert`)
-      expect(parsed[`tx-2`].data.title).toBe(`Manual Tx Insert 2`)
+      expect(parsed[`s:tx-1`].data.title).toBe(`Manual Tx Insert`)
+      expect(parsed[`s:tx-2`].data.title).toBe(`Manual Tx Insert 2`)
 
       subscription.unsubscribe()
     })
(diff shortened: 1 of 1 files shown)


packages/db/src/local-storage.ts
```diff
@@ -415,7 +415,12 @@ export function localStorageCollectionOptions(
     // Add new items with version keys
     params.transaction.mutations.forEach((mutation) => {
       // Use the engine's pre-computed key for consistency
-      const key = mutation.key
+      // Prefix numeric keys to prevent collision with string keys
+      // e.g., numeric 1 => "__number__1", string "1" => "1"
```
kevin-dp: This is not a proper solution. A numeric key 1 can still collide with a string key that is literally `"__number__1"`. A proper solution requires a real encoding scheme that for each key encodes both the type and the stringified value, e.g. `"<type>_<stringified value>"`. That way the numeric key 1 and the string key `"number_1"` are different because they would be encoded as `"number_1"` and `"string_number_1"` respectively.

Also, we need to do this for all supported key types. @claude what do you think?

packages/db/src/local-storage.ts
```diff
@@ -480,7 +490,12 @@ export function localStorageCollectionOptions(
     // Remove items
     params.transaction.mutations.forEach((mutation) => {
       // Use the engine's pre-computed key for consistency
-      const key = mutation.key
+      // Prefix numeric keys to prevent collision with string keys
```
kevin-dp: Please don't duplicate this stringification logic everywhere but extract it into a helper function.

kevin-dp:
@claude your previous fixes seem fine to me but there are some failing unit tests now (cf. "PR / Test" workflow). Analyze the failing tests and see if they are outdated an need to be changed. If they are not outdated, then analyze why they are now failing and what is needed to fix them.

kevin-dp:
@claude There are still some failing tests.

KyleAMathews:
haha nice @kevin-dp and @claude — I didn't know this existed. I think Kevin you need to approve your work now as I can't approve it

Issue #819, opened 2025-11-14
syncMode `"on-demand"` fails to send `loadSubsetOptions` when `like` is used in `where` with live query


Winwardo:
Love the 0.5 release, very cool piece of work. 

- [x] I've validated the bug against the latest version of DB packages

**Describe the bug**
When using the new `"on-demand"` sync mode, a `ctx.meta.loadSubsetOptions` is sent over with various filters.
When using a simple filter such as `.where((w) => eq(w.collection.id, 2))`, this filter correctly appears:
```
[{
    "field": [ "id" ],
    "operator": "eq",
    "value": 2
}]
```

When a `like` or `ilike` is used, whether standalone or chained, the filters object is empty: `[]`
The live query filter is still able to run correctly, but the `queryFn` on the collection itself does not see the predicates. This makes passing a `like` or `ilike` to an external API impossible.

**To Reproduce**
https://codesandbox.io/p/sandbox/q39dmj

Comment and uncomment the `where` options on line 45, then refresh the page and look at the console output. Note how the `ctx.meta?.loadSubsetOptions.where` becomes null.

**Expected behavior**
All predicates should be supported.

**Additional context**
I don't know your codebase very well, but I tried following this through.
I suspect the issue is in the `DeduplicatedLoadSubset` class - possibly the other predicates like `eq` and `lt` etc are properly set up to be compared / serialized, whereas `like` and `ilike` are missing some crucial definitions.
I have not tested extensively with other predicate functions.

`https://github.dev/TanStack/db/blob/5fb76fee0a7689f5563acca94a8d7463ea1b1e8a/packages/query-db-collection/src/query.ts#L978`
`https://github.dev/TanStack/db/blob/5fb76fee0a7689f5563acca94a8d7463ea1b1e8a/packages/db/src/query/subset-dedupe.ts#L85`

Thanks

kevin-dp:
Hi @Winwardo,

https://github.com/TanStack/db/pull/851 should fix the issue. Could you give the preview package a try?

kevin-dp:
I downloaded your code sandbox and upgraded typescript to v4.7 (required by tanstack/db v0.5) and modified the db and react-db packages to point to the preview packages. I can confirm it works now.

PR #851, opened 2025-11-19
Always pass where clause to loadSubset

kevin-dp:
Fixes https://github.com/TanStack/db/issues/819

The root cause was the `convertToBasicExpression` function that would filter out operations we can't (or currently don't) index. However, filtering those operators out here is unnecessary because the `extractIndexableExpressions` function also does the filtering. Hence, i removed all filtering from `convertToBasicExpression`.

[I added a unit test that reproduces the issue](https://github.com/TanStack/db/actions/runs/19497547928/job/55803378683?pr=851) and the unit test passes with the fixes from this PR.


pnpm-lock.yaml
samwillis, line 1: Not sure why this is here, there are no package.json changes

PR #852, opened 2025-11-19
fix electric collection progressive mode

samwillis:
~~**Depends on https://github.com/electric-sql/electric/pull/3463, must have deps bumped to new electric client version when that is released.**~~ *done*

Fix progressive mode to use fetchSnapshot and atomic swap

Progressive mode was broken because `requestSnapshot()` injected snapshots into the stream in causally correct position, which didn't work properly with the `full` mode stream. This release fixes progressive mode by:

**Core Changes:**
- Use `fetchSnapshot()` during initial sync to fetch and apply snapshots immediately in sync transactions
- Buffer all stream messages during initial sync (renamed flag to `isBufferingInitialSync`)
- Perform atomic swap on first `up-to-date`: truncate snapshot data → apply buffered messages → mark ready
- Track txids/snapshots only after atomic swap (enables correct optimistic transaction confirmation)

**Test Infrastructure:**
- Added `ELECTRIC_TEST_HOOKS` symbol for test control (hidden from public API)
- Added `progressiveTestControl.releaseInitialSync()` to E2E test config for explicit transition control
- Created comprehensive progressive mode E2E test suite (8 tests):
  - Explicit snapshot phase and atomic swap validation
  - Txid tracking behavior (Electric-only)
  - Multiple concurrent snapshots with deduplication
  - Incremental updates after swap
  - Predicate handling and resilience tests

**Bug Fixes:**
- Fixed type errors in test files
- All 166 unit tests + 95 E2E tests passing

### DIFF
--- a/packages/electric-db-collection/src/electric.ts
+++ b/packages/electric-db-collection/src/electric.ts
@@ -280,6 +280,93 @@ function hasTxids<T extends Row<unknown>>(
 }
 
 /**
+ * Creates a deduplicated loadSubset handler for progressive/on-demand modes
+ * Returns null for eager mode, or a DeduplicatedLoadSubset instance for other modes.
+ * Handles fetching snapshots in progressive mode during buffering phase,
+ * and requesting snapshots in on-demand mode
+ */
+function createLoadSubsetDedupe<T extends Row<unknown>>({
+  stream,
+  syncMode,
+  isBufferingInitialSync,
+  begin,
+  write,
+  commit,
+  collectionId,
+}: {
+  stream: ShapeStream<T>
+  syncMode: ElectricSyncMode
+  isBufferingInitialSync: () => boolean
+  begin: () => void
+  write: (mutation: {
+    type: `insert` | `update` | `delete`
+    value: T
+    metadata: Record<string, unknown>
+  }) => void
+  commit: () => void
+  collectionId?: string
+}): DeduplicatedLoadSubset | null {
+  // Eager mode doesn't need subset loading
+  if (syncMode === `eager`) {
+    return null
+  }
+
+  const loadSubset = async (opts: LoadSubsetOptions) => {
+    // In progressive mode, use fetchSnapshot during snapshot phase
+    if (isBufferingInitialSync()) {
+      // Progressive mode snapshot phase: fetch and apply immediately
+      const snapshotParams = compileSQL<T>(opts)
+      try {
+        const { data: rows } = await stream.fetchSnapshot(snapshotParams)
+
+        // Check again if we're still buffering - we might have received up-to-date
+        // and completed the atomic swap while waiting for the snapshot
+        if (!isBufferingInitialSync()) {
+          debug(
+            `${collectionId ? `[${collectionId}] ` : ``}Ignoring snapshot - sync completed while fetching`
+          )
+          return
+        }
+
+        // Apply snapshot data in a sync transaction (only if we have data)
+        if (rows.length > 0) {
+          begin()
+          for (const row of rows) {
+            write({
+              type: `insert`,
+              value: row.value,
+              metadata: {
+                ...row.headers,
+              },
+            })
+          }
+          commit()
+
+          debug(
+            `${collectionId ? `[${collectionId}] ` : ``}Applied snapshot with ${rows.length} rows`
+          )
+        }
+      } catch (error) {
+        debug(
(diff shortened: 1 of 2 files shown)


kevin-dp:
Mostly note to self during review:
> Perform atomic swap on first up-to-date: truncate snapshot data → apply buffered messages → mark ready

--> Need to check that we also cancel in-flight fetchSnapshot requests and/or ignore fetchSnapshot requests that resolve after the swap.

packages/electric-db-collection/src/electric.ts
```diff
@@ -669,6 +694,26 @@ function createElectricSync<T extends Row<unknown>>(
     sync: (params: Parameters<SyncConfig<T>[`sync`]>[0]) => {
       const { begin, write, commit, markReady, truncate, collection } = params
 
+      // Wrap markReady to wait for test hook in progressive mode
+      let progressiveReadyGate: Promise<void> | null = null
+      const wrappedMarkReady = () => {
+        // Only create gate if we're in buffering phase (first up-to-date)
+        if (
+          isBufferingInitialSync &&
```
kevin-dp: I don't like that we are closing over the `isBufferingInitialSync` which is actually define way down. I would rather explicitly pass it as an argument.

packages/electric-db-collection/src/electric.ts
```diff
@@ -742,12 +791,48 @@ function createElectricSync<T extends Row<unknown>>(
           ? null
           : new DeduplicatedLoadSubset({
               loadSubset: async (opts: LoadSubsetOptions) => {
-                // In progressive mode, stop requesting snapshots once full sync is complete
-                if (syncMode === `progressive` && hasReceivedUpToDate) {
-                  return
+                // In progressive mode, use fetchSnapshot during snapshot phase
+                if (syncMode === `progressive`) {
+                  if (hasReceivedUpToDate) {
```
kevin-dp: aren't `hasReceivedUpToDate` and `isBufferingInitialSync` equivalent when the `syncMode` is `'progressive'`?

packages/electric-db-collection/src/electric.ts
```diff
@@ -742,12 +791,48 @@ function createElectricSync<T extends Row<unknown>>(
...
+                    // Full sync complete, no need to load more
+                    return
+                  }
+                  // Snapshot phase: fetch and apply immediately
+                  const snapshotParams = compileSQL<T>(opts)
+                  try {
+                    const snapshot = await stream.fetchSnapshot(snapshotParams)
```
kevin-dp: nit: since we don't use `snapshot` let's immediately destructure it `const { data: rows } = ...`

packages/electric-db-collection/src/electric.ts
```diff
@@ -742,12 +791,48 @@ function createElectricSync<T extends Row<unknown>>(
...
-                // In progressive mode, stop requesting snapshots once full sync is complete
-                if (syncMode === `progressive` && hasReceivedUpToDate) {
-                  return
+                // In progressive mode, use fetchSnapshot during snapshot phase
+                if (syncMode === `progressive`) {
+                  if (hasReceivedUpToDate) {
+                    // Full sync complete, no need to load more
+                    return
+                  }
+                  // Snapshot phase: fetch and apply immediately
```
kevin-dp: I would extract the logic that handles the entire snapshot phase (L800-L830) into a separate function. This sync function is already massive and inlining it here only makes it worse. Small and clean functions will improve readability a lot here.

packages/electric-db-collection/src/electric.ts
kevin-dp: I think there is a bug here. When we call `fetchSnapshot` we still didn't receive the up-to-date. But, when this call resolves it could be that we received the up-to-date message in the meantime. We should then ignore this snapshot.

**We should have a unit test for this corner case.**

packages/electric-db-collection/src/electric.ts
kevin-dp: Essentially, i think `isBufferingInitialSync` should be a function that evaluates `syncMode === 'progressive' && !hasReceivedUpToDate`

kevin-dp (dismissed):
This is great work, i think we're almost there but i believe there is a bug with `fetchSnapshot` requests that resolve after the up-to-date message. For the rest i only have minor comments/nits.

samwillis:
@kevin-dp I've addressed your review. We should still come back to this file and tidy it up more.

Issue #804, opened 2025-11-13
`TransactionWithMutations` is missing operation based type when used with array methods


godiagonal:
- [x] I've validated the bug against the latest version of DB packages

**Describe the bug**
Some type information gets lost in `TransactionWithMutations['mutations']` when used with array methods such as `.map()`. 

**To Reproduce**
```ts
const schema = z.object({
  id: z.string(),
  title: z.string(),
  completed: z.boolean(),
});

const todosCollection = createCollection(
  electricCollectionOptions({
    id: 'todos',
    schema,
    getKey: (item) => item.id,
    shapeOptions: {
      url: '/api/todos',
      params: { table: 'todos' },
    },
    onDelete: async ({ transaction }) => {
      await Promise.all(
        transaction.mutations.map((mutation) =>
          // ❌ `original` is typed as `{} | { id: string; title: string; completed: boolean; }`
          api.todos.delete(mutation.original.id),
        ),
      );

      // ✅ `original` is typed as `{ id: string; title: string; completed: boolean; }`
      await api.todos.delete(transaction.mutations[0].original.id);

      // ✅ `original` is typed as `{ id: string; title: string; completed: boolean; }`
      await api.todos.delete(transaction.mutations[1]!.original.id);
    },
  }),
);
```

**Expected behavior**
The type of `transaction.mutations.map((mutation) => mutation.original)[0]` should match `transaction.mutations[0].original`.

**Additional context**
Omitting `mutations` from `Transaction<T>` in `TransactionWithMutations` does the trick:

```ts
export type TransactionWithMutations<T extends object = Record<string, unknown>, TOperation extends OperationType = OperationType> = Omit<Transaction<T>, 'mutations'> & {
    mutations: NonEmptyArray<PendingMutation<T, TOperation>>;
};
```

godiagonal:
My workaround currently:

```ts
await Promise.all(
  transaction.mutations.map((mutation: (typeof transaction.mutations)[0]) =>
    api.todos.delete(mutation.original.id),
  ),
);

```

kevin-dp:
Hi @godiagonal,

Thanks for the detailed report and reproducible example. I added a unit test that reproduces this type error. Looking into a fix now. I'll keep you posted when it's solved.

kevin-dp:
Thanks for this tip:
> Omitting mutations from Transaction<T> in TransactionWithMutations does the trick:

[...]

PR #854, opened 2025-11-19
Fix type issue with mutations in transaction

kevin-dp:
Fixes https://github.com/TanStack/db/issues/804


KyleAMathews:
Oh nice! The widening of the type here was frustrating me back in the day.

Issue #802, opened 2025-11-12
TypeScript error accessing ctx.meta.loadSubsetOptions in queryFn


GraxMonzo:
When accessing `ctx.meta?.loadSubsetOptions` inside a queryFn, an error occurs because meta is currently typed as `Record<string, unknown> | undefined`.

```tsx
import { parseLoadSubsetOptions, queryCollectionOptions } from '@tanstack/query-db-collection';
import { createCollection } from '@tanstack/react-db';

export const imageTypesCollection = createCollection(
  queryCollectionOptions({
    queryClient,
    schema: imageTypeSchema,
    queryKey: ['image-types'],
    queryFn: (ctx) => {
// Argument of type 'unknown' is not assignable to parameter of type '{ where?: BasicExpression<boolean> | undefined; orderBy?: OrderBy | undefined; limit?: number | undefined; } | null | undefined'.ts
      const params = parseLoadSubsetOptions(ctx.meta?.loadSubsetOptions);

      return api.images.getAllTypes({ pageSize: 1_000_000 }, { signal: ctx.signal }).then((res) => res.data);
    },
    getKey: (item) => item.name,
    syncMode: 'on-demand',
  }),
);
```

@tanstack/query-db-collection v1.0.0
@tanstack/react-db v0.1.44
react v19.2
typescript v5.9.3

PR #857, opened 2025-11-19
Modify type of query function's meta property to be more precise

kevin-dp:
Fixes https://github.com/TanStack/db/issues/802

By default a Tanstack Query function's `meta` property is typed as `Record<string, unknown> | undefined`. However, the Query Collection actually extends the meta property with a `loadSubsetOptions` property but this is not reflected in the type of `meta`.

This PR augments the `@tanstack/query-core` module such that the `meta` property is typed as `Record<string, unknown> & { loadSubsetOptions: LoadSubsetOptions }`. I find it a bit odd that we have to resort to module augmentation but [this seems to be the recommended approach](https://github.com/TanStack/query/discussions/2772#discussioncomment-7566892).

I added a type test that [reproduces the problem](https://github.com/TanStack/db/actions/runs/19506023770/job/55832379552?pr=857).

Note: the type of the meta property still allows it to be `undefined` even though we know that won't be the case. I didn't find an elegant way to solve this.

### DIFF
--- a/packages/query-db-collection/tests/query.test.ts
+++ b/packages/query-db-collection/tests/query.test.ts
@@ -3673,9 +3673,7 @@ describe(`QueryCollection`, () => {
       ]
 
       const queryFn = vi.fn((ctx: QueryFunctionContext) => {
-        const loadSubsetOptions = ctx.meta?.loadSubsetOptions as
-          | LoadSubsetOptions
-          | undefined
+        const loadSubsetOptions = ctx.meta?.loadSubsetOptions
         // Filter items based on the where clause if present
         if (loadSubsetOptions?.where) {
           // Simple mock filtering - in real use, you'd use parseLoadSubsetOptions


kevin-dp:
> @kevin-dp this may be the recommended approach, but like you I don't really like it.
> 
> Could we just modify the type here:
> 
> https://github.com/TanStack/db/blob/15270281f488a7187bf461d534635bef0296a50b/packages/query-db-collection/src/query.ts#L72-L76
> 
> such that it includes our metadata?

@samwillis i had given this a try before but i abandoned it early, so i tried again now with a bit more perseverance. I got close but i don't think this  is the right approach: https://github.com/TanStack/db/commit/d895a9854ac22e0a2171526a1a643f742a817c14

Essentially, it works on our side but it breaks types when interacting with Tanstack Query features. For example:
```ts
const queryFn = (
  context: QueryCollectionFunctionContext // <-- this is our new context type
) => {
  // ...
}
await queryClient.prefetchQuery({ queryKey, queryFn }) // expects queryFn to take a QueryFunctionContext
```
The call to `prefetchQuery` yields a type error because it expects the `queryFn` to have this type: `(context: QueryFunctionContext) => Promise<TData>`. However, with our approach of typing our query function differently, our function is typed as `(context: QueryCollectionFunctionContext<TQueryKey>) => Promise<...>`. This is incompatible with Tanstack's expected type because tanstack expect the function to be able to handle any `QueryFunctionContext` (i.e. any `Record<string, unknown> | undefined` value) but our function can only handle `Record<string, unknown> & { loadSubsetOptions: LoadSubsetOptions }` values. So our function is too narrow and the type checker complains.

If what i'm saying is not completely clear you can checkout the commit i linked and run the tests in the query package. You will see the error i'm describing above.

samwillis (approved):
Great work, and thanks for digging into this issue. Let's go for this module interface.

Just needs a changeset.

PR #859, opened 2025-11-19
Fix TypeScript type resolution for QueryCollectionConfig when using queryCollectionOptions without a schema.

KyleAMathews:
Previously, the `QueryCollectionConfig` interface extended `BaseCollectionConfig`, but TypeScript failed to resolve inherited properties like `getKey`, `onInsert`, `onUpdate`, etc. when the interface contained a conditional type for the `queryFn` property. This caused type errors when trying to use `queryCollectionOptions` without a schema.

See this discussion https://discord.com/channels/719702312431386674/1438802082382483496/1438802082382483496


KyleAMathews:
Oh yeah I think we already discussed this in the discord thread and decided it wasn't needed — I forgot to close it apparently. Thanks for checking again!

PR #869, opened 2025-11-20
Make ctx.meta.loadSubsetOptions type safe

KyleAMathews:
Reported here: https://discord.com/channels/719702312431386674/1441107002779631737/1441107002779631737

The module augmentation for ctx.meta.loadSubsetOptions is now guaranteed to load automatically when importing from @tanstack/query-db-collection. Previously, users needed to explicitly import QueryCollectionMeta or use @ts-ignore to pass ctx.meta?.loadSubsetOptions to parseLoadSubsetOptions.

Additionally, QueryCollectionMeta is now an interface (instead of a type alias), enabling users to safely extend meta with custom properties via declaration merging:

```typescript
declare module "@tanstack/query-db-collection" {
  interface QueryCollectionMeta {
    myCustomProperty: string
  }
}
```

### DIFF
--- a/packages/query-db-collection/src/global.ts
+++ b/packages/query-db-collection/src/global.ts
@@ -2,8 +2,9 @@
  * Global type augmentation for @tanstack/query-core
  *
  * This file ensures the module augmentation is always loaded when the package is imported.
- * The index.ts file includes a triple-slash reference to this file, which guarantees
- * TypeScript processes it whenever anyone imports from @tanstack/query-db-collection.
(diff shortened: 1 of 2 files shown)


utopyin:
nice, it fixed the issue on my end! 
image

Thank you :)

samwillis:
@kevin-dp can you pick up the review on this as it build on the previous fix you made.

KyleAMathews:
@kevin-dp yeah my first fix didn't work and then I pushed a commit over the weekend and forgot to update the description. Fixed now.

PR #870, opened 2025-11-20
fix(query-db-collection): implement reference counting for QueryObserver lifecycle

KyleAMathews:
## Summary
Fixes data loss on component remount by implementing reference counting infrastructure for QueryObserver lifecycle management.

## The Problem (vs main)
When live query subscriptions unsubscribed, there was no tracking of which rows were still needed by other active queries. This caused:
- Data loss during component remounts
- Issues when multiple live queries shared identical predicates
- No way to properly clean up rows when last subscriber left

## The Solution
Implemented reference counting to track QueryObserver usage:

1. Pass same predicates to `unloadSubset` that were passed to `loadSubset`
2. Use them to compute the queryKey (via `generateQueryKeyFromOptions`)
3. Use existing machinery (`queryToRows` map) to find rows that query loaded
4. Decrement the ref count
5. GC rows where count reaches 0 (no longer referenced by any active query)

### New Infrastructure (packages/db)
- Added `UnloadSubsetFn` type to sync config
- Collections now call `unloadSubset` when subscriptions clean up
- Subscriptions track all `loadSubset` calls to unload them symmetrically

### Core Implementation (packages/query-db-collection)
- Reference counting map tracks live queries per QueryObserver
- `generateQueryKeyFromOptions()` ensures loadSubset/unloadSubset use identical keys
- Refcount increments when queries load/reuse observers
- Refcount decrements via `unloadSubset` when subscriptions clean up
- Row-level GC when refcount reaches 0
- Preserves TanStack Query's cache (no `removeQueries` or `observer.destroy`)

## Impact
- ✅ Navigation back to previously loaded pages shows cached data immediately
- ✅ No unnecessary refetches during quick remounts (< gcTime)
- ✅ Multiple live queries with identical predicates correctly share QueryObservers
- ✅ Proper row-level cleanup when last subscriber leaves
- ✅ TanStack Query's cache lifecycle (gcTime, staleTime) is fully respected
- ✅ No data leakage from in-flight requests when unsubscribing

## Test Coverage
Added comprehensive tests for QueryObserver lifecycle:
- Reference counting with duplicate subset loads
- Refcount reset after TanStack Query GC and reload
- Mount/unmount/remount without breaking cache
- No data leakage when unsubscribing during in-flight load

All 20 tests in "Query Garbage Collection" suite pass.

### DIFF
--- a/packages/db/src/collection/subscription.ts
+++ b/packages/db/src/collection/subscription.ts
@@ -48,7 +48,10 @@ export class CollectionSubscription
   // While `snapshotSent` is false we filter out all changes from subscription to the collection.
   private snapshotSent = false
 
-  // Track all loadSubset calls made by this subscription so we can unload them on cleanup
+  /**
+   * Track all loadSubset calls made by this subscription so we can unload them on cleanup.
+   * We store the exact LoadSubsetOptions we passed to loadSubset to ensure symmetric unload.
+   */
   private loadedSubsets: Array<LoadSubsetOptions> = []
 
   // Keep track of the keys we've sent (needed for join and orderBy optimizations)
@@ -197,14 +200,14 @@ export class CollectionSubscription
 
     // Request the sync layer to load more data
     // don't await it, we will load the data into the collection when it comes in
-    const loadOptions = {
+    const loadOptions: LoadSubsetOptions = {
       where: stateOpts.where,
       subscription: this,
     }
     const syncResult = this.collection._sync.loadSubset(loadOptions)
 
     // Track this loadSubset call so we can unload it later
-    this.loadedSubsets.push({ where: stateOpts.where })
+    this.loadedSubsets.push(loadOptions)
 
     const trackLoadSubsetPromise = opts?.trackLoadSubsetPromise ?? true
     if (trackLoadSubsetPromise) {
@@ -341,7 +344,7 @@ export class CollectionSubscription
 
     // Request the sync layer to load more data
     // don't await it, we will load the data into the collection when it comes in
-    const loadOptions1 = {
+    const loadOptions1: LoadSubsetOptions = {
       where: whereWithValueFilter,
       limit,
       orderBy,
@@ -350,7 +353,7 @@ export class CollectionSubscription
     const syncResult = this.collection._sync.loadSubset(loadOptions1)
 
     // Track this loadSubset call
-    this.loadedSubsets.push({ where: whereWithValueFilter, limit, orderBy })
+    this.loadedSubsets.push(loadOptions1)
 
     // Make parallel loadSubset calls for values equal to minValue and values greater than minValue
     const promises: Array<Promise<void>> = []
@@ -360,14 +363,14 @@ export class CollectionSubscription
       const { expression } = orderBy[0]!
       const exactValueFilter = eq(expression, new Value(minValue))
 
-      const loadOptions2 = {
+      const loadOptions2: LoadSubsetOptions = {
         where: exactValueFilter,
         subscription: this,
       }
       const equalValueResult = this.collection._sync.loadSubset(loadOptions2)
 
       // Track this loadSubset call
-      this.loadedSubsets.push({ where: exactValueFilter })
+      this.loadedSubsets.push(loadOptions2)
 
       if (equalValueResult instanceof Promise) {
         promises.push(equalValueResult)
@@ -434,11 +437,9 @@ export class CollectionSubscription
 
   unsubscribe() {
     // Unload all subsets that this subscription loaded
-    for (const subset of this.loadedSubsets) {
-      this.collection._sync.unloadSubset({
-        ...subset,
-        subscription: this,
-      })
+    // We pass the exact same LoadSubsetOptions we used for loadSubset
+    for (const options of this.loadedSubsets) {
+      this.collection._sync.unloadSubset(options)
     }
     this.loadedSubsets = []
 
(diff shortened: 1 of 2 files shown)


samwillis (approved):
LGTM :shipit:

samwillis:
Although having said that, there isn't a test. May be a good idea to add one?

samwillis (approved):
After the refactor to introduce `unloadSubset` my ✅ stands.

This works well for the queryCollection where there is a way to go from predicates to queryKey - we should also explore how we can use it to GC rows on an electric collection.

I suspect a similar snapshotKey  that is uses to ref count loaded rows.

I wander if the key generation, and then row refcounting can become a genric we can reuse for both (and other) collections.

(obviously once CI is green)

packages/db/src/collection/subscription.ts
```diff
@@ -47,6 +48,12 @@ export class CollectionSubscription
   // While `snapshotSent` is false we filter out all changes from subscription to the collection.
   private snapshotSent = false
 
+  /**
+   * Track all loadSubset calls made by this subscription so we can unload them on cleanup.
+   * We store the exact LoadSubsetOptions we passed to loadSubset to ensure symmetric unload.
+   */
+  private loadedSubsets: Array<LoadSubsetOptions> = []
```
kevin-dp, line 55: I'm not a fan of storing the `LoadSubsetOptions` for all `loadSubset` calls that we make because:
- i think most collections do some form of hashing/computing an internal key from these options and store that key internally. Both `loadSubset` and `unloadSubset` then perform this hashing.
- some collections don't use them in `unloadSubset` (e.g. Electric) which means wasted memory

I believe a more traditional approach is to let the callee (optionally) return an identifier that it can use to trace requests (in this case `loadSubset` calls). So i'd rather have `loadSubset` return an opaque identifier and the subscription only storing those identifiers. In `unloadSubset` we then pass that identifier and it is up to the collection to decide how to handle it. E.g. the query collection would return the query key, and then in `unloadSubset` it would get that query key and immediately proceed to decrementing the corresponding ref counts without having to recompute the query key from the options. For the Electric collection, since we don't GC (yet) its `loadSubset` could always return `null` meaning we only store 1 `null` value even if the subscription makes a lot of `loadSubset` calls.

And if a collection's `unloadSubset` would really need to know which options were passed initially, it can always track that internally: generate the opaque identifier and then internally map that identifier to the options.

packages/db/src/collection/sync.ts
```diff
@@ -43,6 +43,8 @@ export class CollectionSyncManager<
   public syncLoadSubsetFn:
     | ((options: LoadSubsetOptions) => true | Promise<void>)
     | null = null
+  public syncUnloadSubsetFn: ((options: LoadSubsetOptions) => void) | null =
```
kevin-dp, line 46: why have it as `| null` instead of:
```ts
public syncUnloadSubsetFn?: (options: LoadSubsetOptions) => void
```
An optional property is easier to handle than one that can be `null`. For example on L215, you would not have to do `?? null`.

packages/query-db-collection/src/query.ts
```diff
@@ -651,29 +668,44 @@ export function queryCollectionOptions(
...
-      opts: LoadSubsetOptions = {},
-      queryFunction: typeof queryFn = queryFn
-    ): true | Promise<void> => {
-      // Push the predicates down to the queryKey and queryFn
-      let key: QueryKey
+    /**
+     * Generate a consistent query key from LoadSubsetOptions.
+     * CRITICAL: Must use identical logic in both createQueryFromOpts and unloadSubset
+     * so that refcount increment/decrement operations target the same hashedQueryKey.
+     * Inconsistent keys would cause refcount leaks and prevent proper cleanup.
```
kevin-dp, line 675: You don't have that problem if you use identifiers to relate `loadSubset` and `unloadSubset` calls

kevin-dp:
Based on the PR description, it seems that there were some problems/bugs with the existing ref counting implementation. However, this PR introduces a new ref counting mechanism while also keeping the old one but removing critical parts that break the old ref counting implementation. I don't think that introducing this new ref counting mechanism is the way to solve the problem. I'd rather keep the old ref counting system and fix the bugs it has.

The old ref counting system worked like this in a nutshell:
- Keeps track of which rows are referenced by which queries (based on query key)
- Keeps track of which queries (based on query key) reference which rows
- When a live query is GCed, the subscription is unsubscribed and we tell ts/query to GC all the query keys belonging to this subscription.
- When a ts/query is GCed we drop the ref count of all the rows it referenced
- If the ref count of a row hits zero we GC it

Iiuc, the bug you're describing in this PR is that several live queries may call `loadSubset` with the same predicate so they map to the same query key. But then when one of them is GCed, it would GC the ts/query even though the other one is still referencing it. I think the only thing we need to do to fix the existing implementation is to keep some additional state which for each query key keeps track which subscriptions are relying on that query, i.e. `const querySubscribers: Map<QueryKey, Set<SubscriberId>>`. Then, when modify the current flow like this (changes are in bolt):
- Keeps track of which rows are referenced by which queries (based on query key)
- Keeps track of which queries (based on query key) reference which rows
- **Keeps track of which queries are referenced by which subscribers: `const querySubscribers: Map<QueryKey, Set<SubscriberId>>`**
- When a live query is GCed, the subscription is unsubscribed and **for all query keys used by this subscription we remove the subscription from `querySubscribers`. When a `querySubscribers` set for a certain query key becomes empty we tell ts/query to GC that query key**.
- When a ts/query is GCed we drop the ref count of all the rows it referenced
- If the ref count of a row hits zero we GC it

PR #875, opened 2025-11-21
Throw a clear error when anything other than an object with an collection is passed to "from"

### LINKED ISSUE #873: "Only one source is allowed in the from clause" but not realy my code
**Describe the bug**
I`m trying a query but not success

**To Reproduce**

import { createCollection, localStorageCollectionOptions } from '@tanstack/react-db';
 
const TICKETS_COLLECTION_STORAGE_KEY = 'tickets-collection';

export const getTicketKey = ticket => ticket?.id;
 
export const ticketsCollection = createCollection(
    localStorageCollectionOptions({
        id: 'conversations',
        storageKey: TICKETS_COLLECTION_STORAGE_KEY,
        getKey: getTicketKey,
    }),
);

in my component

const ticketsQuery = useLiveQuery(q => q.from("conversations").orderBy("updatedAt", "desc"));

 but error 
 
**QueryBuilderError: Only one source is allowed in the from clause**

KyleAMathews:
Fixes #873

The issue was that users were getting a confusing error message when passing a string (or other invalid type) to the .from() method instead of an object with a collection.

For example:
```javascript
// Incorrect usage
q.from("conversations")  // String instead of object

// Correct usage
q.from({ conversations: conversationsCollection })
```

When a string was passed, Object.keys("conversations") would return an array of character indices ['0', '1', ...], which has length > 1, triggering the "Only one source is allowed in the from clause" error. This was misleading because the real issue was passing a string instead of an object.

Changes:
- Added validation to check if the source is a plain object before checking its key count
- Improved error message to explicitly state the expected format with an example: .from({ alias: collection })
- Added comprehensive tests for various invalid input types (string, null, array, undefined)

The new error message is:
"Invalid source for from clause: Expected an object with a single key-value pair like { alias: collection }. For example: .from({ todos: todosCollection }). Got: string \"conversations\""

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/db/src/query/builder/index.ts
+++ b/packages/db/src/query/builder/index.ts
@@ -10,9 +10,9 @@ import {
 } from "../ir.js"
 import {
   InvalidSourceError,
+  InvalidSourceTypeError,
   JoinConditionMustBeEqualityError,
   OnlyOneSourceAllowedError,
-  QueryBuilderError,
   QueryMustHaveFromClauseError,
   SubQueryMustHaveFromClauseError,
 } from "../../errors.js"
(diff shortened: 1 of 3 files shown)


packages/db/src/query/builder/index.ts
```diff
@@ -60,13 +61,49 @@ export class BaseQueryBuilder<TContext extends Context = Context> {
...
+    // We use try-catch to handle null/undefined gracefully
+    let keys: Array<string>
+    try {
+      keys = Object.keys(source)
+    } catch {
+      throw new QueryBuilderError(
+        `Invalid source for ${context}: Expected an object with a single key-value pair like { alias: collection }. ` +
+          // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+          `For example: .from({ todos: todosCollection }). Got: ${source === null ? `null` : `undefined`}`
+      )
```
samwillis: Can we move this error into the error.ts and then use it in all four places rather than duplicate the strings?

samwillis (approved):
Once small change, but other than that great addition. Approving for once the errors have been combined and moved

PR #879, opened 2025-11-21
Fixed query collection bug where optimistic state leaked into syncedData when using writeInsert inside onInsert handlers

### LINKED ISSUE #814: Query Collection "Incremental Updates" example never updates with server-sent data.

samwillis:
Fixes bug where optimistic state Leakes into syncedData when using writeInsert inside onInsert handlers.

Previously, when syncing server-generated fields (like IDs or timestamps) using writeInsert within an onInsert handler, the QueryClient cache was updated with combined visible state (including optimistic changes), which triggered the query observer to write optimistic values back to syncedData.

Now the cache is correctly updated with only server-confirmed state, ensuring syncedData maintains separation from optimistic state.

Fork of the reproduction using preview package from this PR showing it fixed: https://codesandbox.io/p/sandbox/tanstack-db-bug-forked-nqtkvp?file=%2Fsrc%2FApp.tsx

fixes #814
replaces #817

### DIFF
--- a/.changeset/fix-optimistic-state-in-synced-data.md
+++ b/.changeset/fix-optimistic-state-in-synced-data.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/query-db-collection": patch
+---
+
+Fixed bug where optimistic state leaked into syncedData when using writeInsert inside onInsert handlers. Previously, when syncing server-generated fields (like IDs or timestamps) using writeInsert within an onInsert handler, the QueryClient cache was updated with combined visible state (including optimistic changes), which triggered the query observer to write optimistic values back to syncedData. Now the cache is correctly updated with only server-confirmed state, ensuring syncedData maintains separation from optimistic state.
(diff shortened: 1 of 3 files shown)


KyleAMathews (approved):
So many caches to mess up! 😆 

:shipit:

PR #891, opened 2025-11-22
Add linearlarge example app for syncModes

KyleAMathews:
This example app is for demoing how `on-demand` syncMode allows DB to load data against very large datasets (millions of rows).

WIP still.

### DIFF
--- a/examples/react/linearlarge/package.json
+++ b/examples/react/linearlarge/package.json
@@ -25,6 +25,7 @@
     "@firefox-devtools/react-contextmenu": "^5.2.3",
     "@tailwindcss/vite": "^4.1.16",
     "@tanstack/electric-db-collection": "^0.1.7",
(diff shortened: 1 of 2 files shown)


examples/react/linearlarge/src/lib/queries.ts
```diff
@@ -0,0 +1,307 @@
...
+
+// Factory function for issue by ID query
+const createIssueByIdQuery = (
+  issueId: string,
+  issuesCollection: Collection<Issue>
+) => {
+  const query = createLiveQueryCollection((q) =>
+    q
+      .from({ issue: issuesCollection })
+      .where(({ issue }) => eq(issue.id, issueId))
```
TimFL, line 56: Is there a reason this is a plain query result and not returned with `findOne` if we only ever expect one issue to be returned?

PR #898, opened 2025-11-24
fix for schedular bug with lazy left-join sources

### LINKED ISSUE #813: `Scheduler detected unresolved dependencies`  `v0.5.0`

When trying to run an optimistic action on an electric collection, I get this error.

```Uncaught Error: Scheduler detected unresolved dependencies for context 1efbf3e1-41e3-4201-9f42-da1f38ec6ae9.```

This only seems to happen when mutating the draft object.

**To Reproduce**

```tsx

const someCollection = createCollection(
  electricCollectionOptions({
   ...
  })
)

const testAction = createOptimisticAction<string>({
  onMutate: id => {
   someCollection.update(id, draft => {
      draft.name = 'new name here';
    });
  },
  mutationFn: async (id, _params) => {
    return { txid: 0 };
  }
});

...
testAction("1efbf3e1-41e3-4201-9f42-da1f38ec6ae9")
```

Disabling optimistic update prevents the error:
```tsx
 someCollection.update(id, { optimistic: false }, draft => {
      draft.name = 'new name here';
    });
```

However this obviously defeats the point!

samwillis:
Fixes #813

Failed test run showing reproduction: https://github.com/TanStack/db/actions/runs/19632437525/job/56215175357?pr=898

----

## Summary

Fix scheduler deadlocks from lazy left-join sources (e.g., the nested left join in the “should handle optimistic mutations with nested left joins without scheduler errors” test) by only blocking on dependencies that are actually enqueued or have pending work in the same transaction. Non-enqueued lazy deps are treated as satisfied, preventing unresolved dependency errors.

## Background: How the scheduler and live queries interact

- The scheduler is a simple per-transaction queue. Each job is enqueued with optional dependencies. A job runs only after all its listed dependencies have run in that same context. If a dependency is listed but no such job is enqueued, the scheduler cannot make progress and throws `Scheduler detected unresolved dependencies for context ...`.
- Live query builders schedule their own jobs per transaction. Dependencies are used so parent queries run before dependent queries when both are scheduled.
- Lazy join sources (e.g., right side of a left join) may not enqueue work in a transaction; they often load data on-demand inside another builder’s run. If a consumer adds such a lazy source as a hard dependency but the lazy source never enqueues a job, the scheduler waits forever.

## Original issue

- Deadlock with lazy left joins (modeled by “should handle optimistic mutations with nested left joins without scheduler errors”): a consumer live query declared the lazy right side of a left join as a dependency, but that lazy source didn’t enqueue work in the transaction. The scheduler blocked waiting for a non-existent job and threw the unresolved-dependency error.
- Stale-read scenario (second new test): when a lazy source does have pending work but runs after its consumer, the consumer can read stale data if it isn’t ordered behind the lazy source. The test exercises this by keeping the transaction open.

## What changed (code)

- `packages/db/src/scheduler.ts`
  - Added a `PendingAwareJob` type and `isPendingAwareJob` guard to remove `any` casting when checking for pending work.
  - Dependency resolution now treats a dependency as blocking only if:
    - The dep job is enqueued and not yet completed, or
    - The dep reports a pending run in the same context via `hasPendingGraphRun`.
    - Otherwise, the dependency is considered satisfied. This avoids deadlocks on lazy sources that never schedule work while still ordering against deps that actually have work.
  - Provides `hasPendingGraphRun` so dependents can detect when a builder has work pending in a transaction (used by the scheduler dependency logic).
- Test adjustment (no API change)
  - The stale-read test uses `autoCommit: false` to keep the transaction open during the race, ensuring the scheduler ordering is exercised without dropping optimistic state prematurely.

## Why this fixes the problem

- Deadlocks avoided: lazy-only dependencies that aren’t enqueued no longer block the scheduler; the dependency edge is treated as satisfied if there is no job or pending work.
- Ordering preserved when needed: if a lazy source is actually scheduled or has pending work, the dependency still blocks, so the consumer won’t outrun it.
- Type safety: dependency checks no longer use duck typing; the scheduler knows how to ask a job if it has pending work.

## Testing

- `pnpm vitest packages/db/tests/query/scheduler.test.ts -t "live query scheduler"` (all 10 tests now pass, covering the deadlock scenario and the stale-read race).

## Notes

- No changes to public API.
- Changeset: `.changeset/soft-lazy-deps.md` (patch for `@tanstack/db`).

### DIFF
--- a/packages/db/src/scheduler.ts
+++ b/packages/db/src/scheduler.ts
@@ -29,12 +29,11 @@ interface PendingAwareJob {
   hasPendingGraphRun: (contextId: SchedulerContextId) => boolean
 }
 
-function isPendingAwareJob(dep: unknown): dep is PendingAwareJob {
+function isPendingAwareJob(dep: any): dep is PendingAwareJob {
   return (
     typeof dep === `object` &&
     dep !== null &&
-    typeof (dep as { hasPendingGraphRun?: unknown }).hasPendingGraphRun ===
-      `function`
+    typeof dep.hasPendingGraphRun === `function`
   )
 }
 


packages/db/src/scheduler.ts
```diff
@@ -25,6 +25,19 @@ interface SchedulerContextState {
...
 
+interface PendingAwareJob {
+  hasPendingGraphRun: (contextId: SchedulerContextId) => boolean
+}
+
+function isPendingAwareJob(dep: unknown): dep is PendingAwareJob {
+  return (
+    typeof dep === `object` &&
+    dep !== null &&
+    typeof (dep as { hasPendingGraphRun?: unknown }).hasPendingGraphRun ===
```
kevin-dp: It makes more sense to type this function's `dep` argument as `any` such that you don't need this typecast here.

PR #899, opened 2025-11-24
Also use compare small Uint8Array by content logic for joins

KyleAMathews:
Export the helper function from db-ivm for use in the main db package.


packages/db-ivm/tests/operators/join.test.ts
```diff
@@ -12,6 +12,7 @@ import {
 describe(`Operators`, () => {
   describe(`Join operation`, () => {
     testJoin()
+    testUint8ArrayKeyJoin()
```
samwillis, line 15: If we decide to keep this implementation, can we incorporate this into the main testJoin function rather than adding as a new one with just this single test?

packages/db-ivm/src/indexes.ts
```diff
@@ -151,6 +152,7 @@ export class Index<TKey, TValue, TPrefix = any> {
    */
   #inner: IndexMap<TKey, TValue, TPrefix>
   #consolidatedMultiplicity: Map<TKey, number> = new Map() // sum of multiplicities per key
+  #originalKeys: Map<TKey, TKey> = new Map() // map from normalized keys to original keys
```
samwillis, line 155: I'm a little nervous of the overhead of adding this additional map and the lookups. We are making joins slower for all cases in order to support an edge case.

samwillis (commented):
I think we should consider carefully if this is a change we want to make. The performance of joins is important, and currently this implementation makes all joins use a new (slightly) slower code path with the normalisation and tracking of original keys - thats not needed for 99% of joins that use string or numeric keys.

I'm honestly not sure that supporting uint8 arrays as join keys is a good idea or not. We can decide to just push this back on the user and have them normalise a uuid to a string when syncing data - I do wander if that is a better option that adding complexity.

KyleAMathews:
Closing as superseded by merged PR #1834, which fixes binary join-key equality through the shared equality identity path. The approach from this PR remains credited in #1834.

PR #903, opened 2025-11-25
Don't evict query keys from cache

kevin-dp:
When a live query is GCed its subscription is unsubscribed and the Query collection then immediately tells tanstack/query to remove the corresponding query keys. However, this means that the query data is immediately GCed. So if a user quickly browses to another page and comes back the data has to load again because it was immediately GCed on onmount. We should not evict it immediately but simply delete the tanstack/query observer such that tanstack/query will GC the data after the cache time has elapsed.

https://github.com/TanStack/db/pull/870 fixes this but it also introduces a couple of other changes that may not be needed. I removed the commits from that PR to see if we can fix this problem more surgically.


kevin-dp:
Closed in favor of https://github.com/TanStack/db/pull/905

PR #905, opened 2025-11-25
Revert commit and don't evict query keys from cache

kevin-dp:
When a live query is GCed its subscription is unsubscribed and the Query collection then immediately tells tanstack/query to remove the corresponding query keys. However, this means that the query data is immediately GCed. So if a user quickly browses to another page and comes back the data has to load again because it was immediately GCed on onmount. We should not evict it immediately but simply delete the tanstack/query observer such that tanstack/query will GC the data after the cache time has elapsed.

https://github.com/TanStack/db/pull/870 fixes this but it also introduces a couple of other changes that aren't needed. I reverted that commit.

I added a unit test that reproduces the problem: https://github.com/TanStack/db/actions/runs/19669831433/job/56335642878?pr=905

### DIFF
--- a/packages/query-db-collection/tests/query.test.ts
+++ b/packages/query-db-collection/tests/query.test.ts
@@ -969,7 +969,6 @@ describe(`QueryCollection`, () => {
       const cancelQueriesSpy = vi
         .spyOn(queryClient, `cancelQueries`)
         .mockResolvedValue()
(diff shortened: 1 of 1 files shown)


kevin-dp:
Closing this PR as the e2e tests revealed that we can have loadSubset calls with the same options leading to duplicate  subscribers so we need to ref count them which is essentially what https://github.com/TanStack/db/pull/870 did.

PR #910, opened 2025-11-26
 Fix change tracking for array items accessed via iteration methods (find, forEach, for...of, etc.)

KyleAMathews:
Previously, modifications to array items retrieved via iteration methods were not tracked by the change proxy because these methods returned raw array elements instead of proxied versions. This caused `getChanges()` to return an empty object, which in turn caused `createOptimisticAction`'s `mutationFn` to never be called when using patterns like:

```ts
collection.update(id, (draft) => {
  const item = draft.items.find(x => x.id === targetId)
  if (item) {
    item.value = newValue // This change was not tracked!
  }
})
```

The fix adds proxy handling for array iteration methods similar to how Map/Set iteration is already handled, ensuring that callbacks receive proxied elements and returned elements are properly proxied.

### DIFF
--- a/.changeset/fix-array-iteration-proxy.md
+++ b/.changeset/fix-array-iteration-proxy.md
@@ -16,3 +16,5 @@ collection.update(id, (draft) => {
 ```
 
 The fix adds proxy handling for array iteration methods similar to how Map/Set iteration is already handled, ensuring that callbacks receive proxied elements and returned elements are properly proxied.
+
+Also refactors proxy.ts for improved readability by extracting helper functions and hoisting constants to module scope.


packages/db/src/proxy.ts
```diff
@@ -411,6 +445,155 @@ export function createChangeProxy<
                 return result
               }
             }
+
+            // For Array methods that iterate with callbacks and may return elements
+            // These need to pass proxied elements to callbacks and return proxied results
+            if (CALLBACK_ITERATION_METHODS.has(methodName)) {
```
samwillis: The parent `get` method that this branch is part of has got *very* big. With this change it's a 450 line function. Can we move this branch out into a helper function for handling iteration.

I think we should also consider further breaking up this function.

packages/db/tests/proxy.test.ts
```diff
@@ -1653,5 +1653,165 @@ describe(`Proxy Library`, () => {
       })
       expect(obj.date).toEqual(originalDate)
     })
+
+    // Test that array iteration methods return proxied elements for change tracking
```
samwillis: Maybe we should group these tests in a `describe`

packages/db/src/proxy.ts
KyleAMathews: Yeah good time for a refactor yeah — Claude did it pretty and the get method is now ~100 lines of code

PR #913, opened 2025-11-26
Fix Date Precision Mismatch in Cursor-Based Pagination

samwillis:
Failed test run before the fix: https://github.com/TanStack/db/actions/runs/19701959124/job/56440153416?pr=913

### Summary

This PR fixes a bug where cursor-based pagination could stall or skip rows when ordering by Date columns, due to precision differences between JavaScript (milliseconds) and backends like PostgreSQL (microseconds).

### The Problem

JavaScript `Date` objects have **millisecond precision**, but databases like PostgreSQL store timestamps with **microsecond precision**. This creates a mismatch:

```
PostgreSQL stores:
- 2024-01-15T10:30:45.123000Z (row 1)
- 2024-01-15T10:30:45.123100Z (row 2)  
- 2024-01-15T10:30:45.123200Z (row 3)

JavaScript sees all as:
- 2024-01-15T10:30:45.123Z
```

When paginating, TanStack DB loads "all rows with the same cursor value" to avoid skipping duplicates. The previous code used exact equality (`eq`), which only matched rows with **exactly** `123.000000μs`, missing rows at other microsecond offsets.

### The Fix

For `Date` values, we now query for the **1ms range** that the Date represents:

```typescript
and(
  gte(expression, new Value(minValue)),           // >= 123.000000μs
  lt(expression, new Value(minValuePlus1ms))      // <  124.000000μs
)
```

### Changes

- **`packages/db/src/collection/subscription.ts`**: Added special handling for `Date` values to use range queries instead of exact equality
- **`packages/db/tests/query/order-by.test.ts`**: Added test that verifies the correct query structure is generated

### DIFF
--- a/.changeset/fix-date-precision-pagination.md
+++ b/.changeset/fix-date-precision-pagination.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+Fix pagination with Date orderBy values when backend has higher precision than JavaScript's millisecond precision. When loading duplicate values during cursor-based pagination, Date values now use a 1ms range query (`gte`/`lt`) instead of exact equality (`eq`) to correctly match all rows that fall within the same millisecond, even if the backend (e.g., PostgreSQL) stores them with microsecond precision.


packages/db/tests/query/order-by.test.ts
```diff
@@ -2923,3 +2923,157 @@ describe(`OrderBy with duplicate values`, () => {
 
   createOrderByBugTests(`eager`)
 })
+
+describe(`OrderBy with Date values and precision differences`, () => {
```
samwillis, line 2927: This test checks that we push down the range predicates rather than the `eq` - I considered making it test the effect, but that ends up testing the mock backend not the implementation.

Arguably we could add a e2e for electric on this.

PR #918, opened 2025-11-26
Automatically start sync when there's any mutations

KyleAMathews:
Fixed `SyncNotInitializedError` being thrown when calling write operations (`writeUpsert`, `writeInsert`, etc.) or mutations (`insert`, `update`, `delete`) on collections before sync is started. Previously, these operations required `startSync: true` to be explicitly set or `preload()` to be called first. Now, sync is automatically started when any write operation or mutation is called on an idle collection, enabling these operations to work immediately without explicit initialization.

Fixes the issue where users with on-demand collections could not manually insert related records using writeUpsert.


packages/db/src/collection/index.ts
```diff
@@ -210,9 +210,17 @@ export class CollectionImpl<
...
+  private _utils: Record<string, Fn> = {}
+
+  public get utils(): Record<string, Fn> {
+    this._lifecycle.validateCollectionUsable(`utils`)
+    return this._utils
+  }
+
+  public set utils(value: Record<string, Fn>) {
+    this._utils = value
+  }
```
samwillis, line 286: This is a very strange way to implement this. It seems to be a hack that means that when you access the collection `utils` **at all** it will call `_lifecycle.validateCollectionUsable` and trigger sync to start?

packages/query-db-collection/tests/query.test.ts
```diff
@@ -1622,6 +1622,74 @@ describe(`QueryCollection`, () => {
...
+      const newItem: TestItem = { id: `1`, name: `Item 1`, value: 10 }
+      collection.utils.writeInsert(newItem)
+
+      // After write, collection should be ready (sync was auto-started)
+      expect(collection.status).toBe(`ready`)
+      expect(collection.size).toBe(1)
+      expect(collection.get(`1`)).toEqual(newItem)
+
+      // Test writeUpsert (the specific operation from the bug report)
+      collection.utils.writeUpsert({ id: `2`, name: `Item 2`, value: 20 })
```
samwillis, line 1834: I thought the issue was with any mutation, not just the direct writes?

samwillis (changes requested):
This doesn't fix the issue. It only works on query collections with direct writes.

packages/db/src/collection/index.ts
KyleAMathews: yeah — it seems like a reasonable approach to me — the other way was to have query collection manually call startSync but the idea here is that if a collection is active in any way to start sync. Which covers a lot of stuff.

packages/query-db-collection/tests/query.test.ts
KyleAMathews: Yeah we handle regular mutations here https://github.com/TanStack/db/pull/918/files#diff-03176e533aea8affa9d789787bfdf48e6a7ce75817f44712dd9c1fe8cf9bd2eb

packages/db/src/collection/index.ts
samwillis: ~~But this doesn't fix `collection.insert` or using a custom mutation (are they also broken?)~~, and will trigger sync to start for any collection implementation that places anything it want on the utils, some of which could very much be intended not to start sync.

I really don't think this is the right approach.

samwillis: say a sync engine added a `utils.status`, accessing that would trigger sync to start.

mwalkersigma:
Just ran into this issue myself, trying to create an optimistic action using a collection that already has CRUD handlers. Any updates on this ?

flybayer:
so the current workaround is to always set `startSync: true` on `on-demand` collections?

PR #926, opened 2025-11-27
fix(db) loadSubset when orderby has multiple columns

samwillis:
**stacked on #958**

Enhanced multi-column orderBy support with lazy loading and composite cursor optimization.

**Changes:**
- Create index on first orderBy column even for multi-column orderBy queries, enabling lazy loading with first-column ordering
- Pass multi-column orderBy to loadSubset with precise composite cursors (e.g., `or(gt(col1, v1), and(eq(col1, v1), gt(col2, v2)))`) for backend optimization
- Use wide bounds (first column only) for local index operations to ensure no rows are missed
- Use precise composite cursor for sync layer loadSubset to minimize data transfer

**Benefits:**
- Multi-column orderBy queries with limit now support lazy loading (previously disabled)
- Sync implementations (like Electric) can optimize queries using composite indexes on the backend
- Local collection uses first-column index efficiently while backend gets precise cursor

Prior to this change composite order by was broken as shown in the first commit CI run: https://github.com/TanStack/db/actions/runs/19741724803/job/56567001027

### DIFF
--- a/packages/db/src/collection/subscription.ts
+++ b/packages/db/src/collection/subscription.ts
@@ -1,8 +1,8 @@
-import { ensureIndexForExpression } from "../indexes/auto-index.js"
-import { and, eq, gte, lt } from "../query/builder/functions.js"
-import { Value } from "../query/ir.js"
-import { EventEmitter } from "../event-emitter.js"
-import { buildCursor } from "../utils/cursor.js"
+import { ensureIndexForExpression } from '../indexes/auto-index.js'
+import { and, eq, gte, lt } from '../query/builder/functions.js'
+import { Value } from '../query/ir.js'
+import { EventEmitter } from '../event-emitter.js'
+import { buildCursor } from '../utils/cursor.js'
 import {
   createFilterFunctionFromExpression,
   createFilteredCallback,
@@ -324,7 +324,7 @@ export class CollectionSubscription
         const keysGreaterThanMin = index.take(
           limit - keys.length,
           minValueForIndex,
-          filterFn
+          filterFn,
         )
         keys.push(...keysGreaterThanMin)
       } else {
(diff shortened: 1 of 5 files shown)


packages/db/src/collection/subscription.ts
```diff
@@ -38,6 +43,75 @@ type CollectionSubscriptionOptions = {
...
+ * Builds a composite cursor expression for multi-column orderBy.
+ * For [col1 ASC, col2 DESC] with values [v1, v2], produces:
+ *   or(
+ *     gt(col1, v1),                         // col1 > v1
+ *     and(eq(col1, v1), lt(col2, v2))       // col1 = v1 AND col2 < v2 (DESC)
+ *   )
+ *
+ * This creates a precise cursor that works with composite indexes on the backend.
+ */
+function buildCompositeCursor(
```
kevin-dp: This utility function would be better in a utility file.

kevin-dp: We also need to unit test this utility function. I'm sure Cursor will be very good at spitting out all kinds of corner cases.

packages/db/src/collection/subscription.ts
```diff
@@ -233,17 +310,22 @@ export class CollectionSubscription
...
+   *
+   * Note 1: it may load more rows than the provided LIMIT because it loads all values equal to the first cursor value + limit values greater.
    *         This is needed to ensure that it does not accidentally skip duplicate values when the limit falls in the middle of some duplicated values.
    * Note 2: it does not send keys that have already been sent before.
    */
   requestLimitedSnapshot({
     orderBy,
     limit,
-    minValue,
+    minValues,
```
kevin-dp, line 260: Shall we rename `minValues` to `cursor`? You did this in the docstring but didn't rename the actual argument.

packages/db/src/collection/subscription.ts
kevin-dp: nit: i think `buildCursor` would be a better name, whether it's a composite cursor or not doesn't really matter.

kevin-dp (approved):
Great PR again! Addressing the problems in a concise and understandable way.

PR #932, opened 2025-11-28
Fix bigint serialization in pg-serializer

KyleAMathews:
- Add support for serializing BigInt values to strings
- Fix error message to handle non-JSON-serializable values (like BigInt) which was causing a secondary "Do not know how to serialize a BigInt" error from JSON.stringify instead of the intended error message
- Add comprehensive unit tests for pg-serializer with 100% coverage

## ✅ Checklist

## 🚀 Release Impact


samwillis:
@KyleAMathews looks good but perhaps we should add an e2e test to validate that this works when using a bigint as a predicate that's pushed to electric. I'm 99% sure this will work.

KyleAMathews:
Ah good call, added an e2e test

PR #933, opened 2025-11-28
Fix handling of frozen objects in proxy

KyleAMathews:
Use the unfrozen internal copy as the Proxy target instead of the original frozen object. This fixes a TypeError that occurred when trying to modify properties on a proxy wrapping a frozen object (such as state from libraries that freeze their data).

Added comprehensive tests for frozen object handling including deeply frozen nested objects and array change tracking.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/db/tests/proxy.test.ts
+++ b/packages/db/tests/proxy.test.ts
@@ -599,6 +599,7 @@ describe(`Proxy Library`, () => {
 
       // Modify and delete
       proxy.name = `Jane`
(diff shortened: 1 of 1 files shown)


KyleAMathews:
Yeah exactly. We were already making a clone and should have been using it. 

And yeah good call, also added support for .seal/.preventExtensions.

This is a good test of the improved models — it went in circles on this ~6 months ago when I first tried implementing this but Claude 4.5 fixed it immediately this go around.

PR #942, opened 2025-12-01
Tagged rows and support for move outs in Electric DB collection

kevin-dp:
Builds on top of Electric's ts-client support for tagging rows and move out events: https://github.com/electric-sql/electric/pull/3497

This PR extends tanstack DB such that it handles tagged rows and move out events. A tagged row is removed from the Electric collection when its tag set becomes empty. Note that rows only have tags when the shape they belong to has one or more subqueries.

### DIFF
--- a/.changeset/witty-animals-agree.md
+++ b/.changeset/witty-animals-agree.md
@@ -1,5 +1,5 @@
 ---
-"@tanstack/electric-db-collection": patch
+'@tanstack/electric-db-collection': patch
 ---
 
 Support tagged rows and move out events in Electric collection.


packages/electric-db-collection/src/electric.ts
```diff
@@ -1049,16 +1335,46 @@ function createElectricSync<T extends Row<unknown>>(
...
             for (const bufferedMsg of bufferedMessages) {
               if (isChangeMessage(bufferedMsg)) {
-                write({
-                  type: bufferedMsg.headers.operation,
-                  value: bufferedMsg.value,
-                  metadata: {
-                    ...bufferedMsg.headers,
-                  },
-                })
+                // Process tags for buffered messages
```
samwillis: almost exactly the same code as lines 1244-1277 (amused they are exactly 100 lines apart!)

samwillis (approved):
Looks good to me. Let's get the e2e tests in and have @icehaunter cast his eyes over it.

packages/electric-db-collection/src/electric.ts
```diff
@@ -806,6 +823,236 @@ function createElectricSync<T extends Row<unknown>>(
   // Store for the relation schema information
   const relationSchema = new Store<string | undefined>(undefined)
 
+  const tagCache = new Map<MoveTag, ParsedMoveTag>()
+
+  // Parses a tag string into a MoveTag.
+  // It memoizes the result parsed tag such that future calls
+  // for the same tag string return the same MoveTag array.
+  const parseTag = (tag: MoveTag): ParsedMoveTag => {
+    if (tagCache.has(tag)) {
```
icehaunter: Nit: I'd prefer one lookup with get & checking that against undefined

packages/electric-db-collection/src/electric.ts
```diff
@@ -806,6 +823,236 @@ function createElectricSync<T extends Row<unknown>>(
...
+    rowTagSet: Set<MoveTag>
+  ): void => {
+    if (tagLength === undefined) {
+      return
+    }
+
+    for (const tag of removedTags) {
+      const parsedTag = parseTag(tag)
+      const currentTagLength = getTagLength(parsedTag)
+      if (currentTagLength === tagLength) {
```
icehaunter: What's the point of this check? I think server guarantees this

packages/electric-db-collection/src/electric.ts
```diff
@@ -994,14 +1241,41 @@ function createElectricSync<T extends Row<unknown>>(
...
+
+              const rowId = collection.getKeyFromItem(message.value)
+              const rowTagSet = () =>
+                processTagsForChangeMessage(tags, removedTags, rowId)
+
+              // Check if row should be deleted (empty tag set)
+              // but only if the message includes tags
+              // because shapes without subqueries don't contain tags
+              // so we should keep those around
+              if (hasTags && rowTagSet().size === 0) {
```
icehaunter: I'm not sure this branch is possible on a change message, because an update that would have removed all tags from a row means it no longer matches the where clause and thus would be converted to a delete on the server

kevin-dp:
I addressed all feedback and added e2e tests. ~~All e2e tests involving move-outs are failing because of a bug in Electric. We should check that they pass when the bug is fixed.~~

PR #943, opened 2025-12-01
Find whereExpression examples in subscribeChanges

KyleAMathews:
Add a `where` callback option to `SubscribeChangesOptions` that allows filtering changes using query builder functions (eq, gt, and, etc.) with a row proxy, instead of requiring manual IR construction with PropRef.

Before:
```ts
import { eq, PropRef } from "@tanstack/db"
collection.subscribeChanges(callback, {
  whereExpression: eq(new PropRef(["status"]), "active")
})
```

After:
```ts
import { eq } from "@tanstack/db"
collection.subscribeChanges(callback, {
  where: (row) => eq(row.status, "active")
})
```

Changes:
- Add generic `where` callback to `SubscribeChangesOptions<T>`
- Compile `where` callback to `whereExpression` in CollectionChangesManager
- Update JSDoc examples to show correct usage
- Add tests for single and multiple condition filtering

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/db/src/collection/changes.ts
+++ b/packages/db/src/collection/changes.ts
@@ -110,15 +110,16 @@ export class CollectionChangesManager<
       )
     }
 
-    let whereExpression = options.whereExpression
-    if (options.where) {
+    const { where, ...opts } = options
+    let whereExpression = opts.whereExpression
+    if (where) {
       const proxy = createSingleRowRefProxy<TOutput>()
-      const result = options.where(proxy)
+      const result = where(proxy)
       whereExpression = toExpression(result)
     }
 
     const subscription = new CollectionSubscription(this.collection, callback, {
-      includeInitialState: options.includeInitialState,
+      ...opts,
       whereExpression,
       onUnsubscribe: () => {
         this.removeSubscriber()


packages/db/src/collection/changes.ts
```diff
@@ -94,13 +98,28 @@ export class CollectionChangesManager<
...
+    let whereExpression = options.whereExpression
+    if (options.where) {
+      const proxy = createSingleRowRefProxy<TOutput>()
+      const result = options.where(proxy)
+      whereExpression = toExpression(result)
+    }
+
     const subscription = new CollectionSubscription(this.collection, callback, {
-      ...options,
+      includeInitialState: options.includeInitialState,
```
kevin-dp: I'd rather keep `...options` because spreading is more future proof in case we add new fields to the options. But i understand we want to spread all except `where`. Perhaps:
```ts
const { where, ...opts } = options
const subscription = new CollectionSubscription(this.collection, callback, {
  ...opts,
  whereExpression,
  ...
```

kevin-dp (approved):
Nice addition. I left just one nit.

kevin-dp:
I rebased this branch and CI is green. I'll have Claude fix my nit and then it should be ready to go.

packages/db/src/collection/changes.ts
kevin-dp: @claude please apply the fix described in my previous comment.

PR #950, opened 2025-12-02
Make indexing opt-in

KyleAMathews:


packages/db/src/indexes/auto-index.ts
```diff
@@ -14,6 +14,19 @@ function shouldAutoIndex(collection: CollectionImpl<any, any, any, any, any>) {
     return false
   }
 
+  // Check if defaultIndexType is set on the collection
+  if (!collection.config.defaultIndexType) {
+    if (isDevModeEnabled()) {
+      console.warn(
+        `[TanStack DB] Auto-indexing is enabled but no defaultIndexType is set. ` +
```
kevin-dp: nit: this warning is misleading because it says auto-indexing is enabled but actually since we return false here it is disabled. We should make explicit that we are disabling it because no defaultIndexType is set.

I think it would make more sense to still keep it enabled and just fall back to a default index type in case the user does not provide one?

packages/db/src/indexes/basic-index.ts
```diff
@@ -0,0 +1,439 @@
...
+ * Basic index using Map + sorted Array.
+ *
+ * - Map for O(1) equality lookups
+ * - Sorted Array for O(log n) range queries via binary search
+ * - O(n) updates to maintain sort order
+ *
+ * Simpler and smaller than BTreeIndex, good for read-heavy workloads.
+ * Use BTreeIndex for write-heavy workloads with large collections.
+ */
+export class BasicIndex<
```
kevin-dp: Suggestion: For many users it may not be clear which index they should pick based on the names `BasicIndex` and `BTreeIndex`. Perhaps we should rename them to clarify this: `BasicIndex` -> `ReadOptimizedIndex` and `BTreeIndex` -> `WriteOptimizedIndex`. The docstrings can then explain like they do now that `ReadOptimizedIndex` is backed by a Map + sorted Array and its performance implications and that `WriteOptimizedIndex` is backed by a B+ tree.

packages/db/src/indexes/basic-index.ts
```diff
@@ -0,0 +1,439 @@
...
+      this.compareOptions = options!.compareOptions
+    }
+  }
+
+  protected initialize(_options?: BasicIndexOptions): void {}
+
+  /**
+   * Binary search to find insertion point in sorted array
+   */
+  private findInsertionIndex(value: any): number {
```
kevin-dp: Let's remove this method. We already have a `findInsertPosition` helper functon in `utils/array-utils.ts`.
That helper function expects the array to be an array of tuples though so perhaps we should modify that helper to be over a basic array of values. And we can define a variant of it on top for tupled arrays (such that the parts of the code that were using it before don't need to be changed, just need to use the tupled variant).

packages/db/src/indexes/basic-index.ts
```diff
@@ -0,0 +1,439 @@
...
+
+    if (this.valueMap.has(normalizedValue)) {
+      const keySet = this.valueMap.get(normalizedValue)!
+      keySet.delete(key)
+
+      if (keySet.size === 0) {
+        // No more keys for this value, remove from map and sorted array
+        this.valueMap.delete(normalizedValue)
+
+        const idx = this.findInsertionIndex(normalizedValue)
```
kevin-dp: This findInsertionIndex + the if test below is the logic that is needed to delete a value from a sorted array. As mentioned in a previous comment `findInsertionIndex` should be replaced by the already existing `findInsertPosition`. And we should add an additional helper function `deleteInSortedArray` to that `array-utils.ts` file that does exactly this logic. And then here just use that `deleteInSortedArray` function.

packages/db/src/indexes/read-optimized-index.ts
```diff
@@ -0,0 +1,439 @@
...
+   */
+  update(key: TKey, oldItem: any, newItem: any): void {
+    this.remove(key, oldItem)
+    this.add(key, newItem)
+  }
+
+  /**
+   * Builds the index from a collection of entries
+   */
+  build(entries: Iterable<[TKey, any]>): void {
```
kevin-dp, line 153: I was expecting `build` to do something like this:
```ts
entries.map(([key, value]) => this.add(key, value))
```
Because right now it duplicates the `add` logic.

packages/db/src/indexes/basic-index.ts
```diff
@@ -0,0 +1,439 @@
...
+  get keyCount(): number {
+    return this.indexedKeys.size
+  }
+
+  /**
+   * Performs an equality lookup - O(1)
+   */
+  equalityLookup(value: any): Set<TKey> {
+    const normalizedValue = normalizeValue(value)
+    return new Set(this.valueMap.get(normalizedValue) ?? [])
```
kevin-dp: nit: `this.valueMap.get(normalizedValue)` already returns a set so let's do:
```ts
return this.valueMap.get(normalizedValue) ?? new Set()
```

packages/db/src/indexes/read-optimized-index.ts
```diff
@@ -0,0 +1,439 @@
...
+    }
+
+    return result
+  }
+
+  /**
+   * Performs a reversed range query
+   */
+  rangeQueryReversed(options: RangeQueryOptions = {}): Set<TKey> {
+    // For BasicIndex, reversed is the same result set, just different iteration order
```
kevin-dp: Is this true?
`rangeQuery` looks up the index for `from` and the index for `to` and then iterates from `from` to `to` with this for loop:
```ts
for (let i = startIdx; i < endIdx; i++)
```
So it is explicitly expecting `startIdx` to be smaller than `endIdx` and so going left to right by incrementing the index.

However, when doing a reversed range query we go from right to left, e.g. from 10 to 5. So shouldn't we either swap the `from` and `to` here or modify `rangeQuery` to find the insertions indexes of both from and to and go from the smallest to the biggest index? i.e. swap `startIdx` and `endIdx` if `endIdx < startIdx`.

This is how the B+ tree index does it which seems to me that we also need to do this here:
```ts
rangeQueryReversed(options: RangeQueryOptions = {}): Set<TKey> {
  const { from, to, fromInclusive = true, toInclusive = true } = options
  return this.rangeQuery({
    from: to ?? this.orderedEntries.maxKey(),
    to: from ?? this.orderedEntries.minKey(),
    fromInclusive: toInclusive,
    toInclusive: fromInclusive,
  })
}
```

packages/db/src/indexes/basic-index.ts
```diff
@@ -0,0 +1,439 @@
...
+    return this.sortedValues.map((value) => [
+      value,
+      this.valueMap.get(value) ?? new Set(),
+    ])
+  }
+
+  get orderedEntriesArrayReversed(): Array<[any, Set<TKey>]> {
+    return [...this.sortedValues]
+      .reverse()
+      .map((value) => [value, this.valueMap.get(value) ?? new Set()])
```
kevin-dp: This code iterates 3 times through the sorted values. We can improve it by a single pass with a for loop or with `reduce`.

packages/db/src/indexes/index-options.ts
```diff
@@ -1,42 +1,22 @@
...
-          name?: string,
-          options?: infer O
-        ) => any
-        ? O
-        : never
-      : never
+  /** Index type to use (e.g., BasicIndex, BTreeIndex) */
+  indexType?: TIndexType
+  /** Options passed to the index constructor */
+  options?: TIndexType extends new (
```
kevin-dp, line 14: If we're going to pass both the index type (i.e. constructor) and the arguments for the constructor. Isn't that equivalent but a bit more verbose than to just pass an index creation function: `() => BaseIndex`.

packages/db/src/indexes/index-registry.ts
```diff
@@ -0,0 +1,170 @@
+/**
+ * Index Dev Mode - Helps developers identify when indexes would improve performance
+ *
+ * Dev mode suggestions are ON by default in non-production builds.
+ */
+
+// Dev mode detection settings - ON by default in non-production
+let devModeConfig: IndexDevModeConfig = {
+  enabled: true,
+  collectionSizeThreshold: 1000,
```
kevin-dp, line 10: This threshold seems arbitrary. Going through an non-indexed collection of 1000 rows should be pretty fast. I guess modern machines can easily go to perhaps 1M rows before it really becomes slow and needs indexes. Time-based suggestions (`slowQueryThresholdMs`) makes much more sense imo.

kevin-dp (changes requested):
I left some comments that i would like to see addressed. Nothing major. There are no unit tests for the new `BasicIndex`. We should add unit tests.

packages/db/src/indexes/basic-index.ts
kevin-dp: @claude Address this comment

packages/db/src/indexes/read-optimized-index.ts
kevin-dp: @claude address this comment

packages/db/src/indexes/basic-index.ts
kevin-dp: @claude address this comment

packages/db/src/indexes/index-options.ts
kevin-dp: Nevermind, i had wrongly interpreted this code. You can ignore this comment.

kevin-dp (approved):
I addressed my own review.

kevin-dp:
Closing in favor of https://github.com/TanStack/db/pull/1353

PR #951, opened 2025-12-02
Fix serializing null/undefined when generating subset queries

KyleAMathews:
When comparison operators (eq, gt, lt, etc.) were used with null/undefined values, the SQL compiler would generate placeholders ($1, $2) in the WHERE clause but skip adding the params to the dictionary because serialize() returns empty string for null/undefined.

This resulted in invalid queries being sent to Electric like:
  subset__where="name" = $1
  subset__params={}

The fix:
- For eq(col, null): Transform to "col IS NULL" syntax
- For other comparisons (gt, lt, gte, lte, like, ilike): Throw a clear error since null comparisons don't make semantic sense in SQL

Added comprehensive tests for the sql-compiler including null handling.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/.changeset/fix-select-write-operations.md
+++ b/.changeset/fix-select-write-operations.md
@@ -1,0 +1,9 @@
+---
+'@tanstack/query-db-collection': patch
+---
+
+Fix writeInsert/writeUpsert throwing error when collection uses select option
+
+When a Query Collection was configured with a `select` option to extract items from a wrapped API response (e.g., `{ data: [...], meta: {...} }`), calling `writeInsert()` or `writeUpsert()` would corrupt the query cache and trigger the error: "select() must return an array of objects".
+
+The fix routes cache updates through a new `updateCacheData` function that preserves the wrapper structure by using the `select` function to identify which property contains the items array (via reference equality), then updates only that property while keeping metadata intact.
--- a/examples/react/paced-mutations-demo/package.json
+++ b/examples/react/paced-mutations-demo/package.json
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "@tanstack/db": "^0.5.11",
-    "@tanstack/react-db": "^0.1.55",
+    "@tanstack/react-db": "^0.1.56",
     "mitt": "^3.0.1",
     "react": "^19.2.1",
     "react-dom": "^19.2.1"
--- a/examples/react/todo/package.json
+++ b/examples/react/todo/package.json
@@ -5,8 +5,8 @@
   "dependencies": {
     "@tanstack/electric-db-collection": "^0.2.12",
     "@tanstack/query-core": "^5.90.12",
-    "@tanstack/query-db-collection": "^1.0.6",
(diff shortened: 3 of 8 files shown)


docs/design/eq-null-handling.md
```diff
@@ -0,0 +1,105 @@
+# Handling Null Values in Comparison Operators
```
kevin-dp: Is this file intended to be merged into the codebase or is this cursor/claude writing an md file to tell us what it did but not actually intended to be merged?

.changeset/fix-electric-null-params.md
```diff
@@ -0,0 +1,12 @@
...
+"@tanstack/electric-db-collection": patch
+---
+
+Fix invalid Electric proxy queries with missing params for null/undefined values
+
+When comparison operators were used with null/undefined values, the SQL compiler would generate placeholders ($1, $2) in the WHERE clause but skip adding the params to the dictionary. This resulted in invalid queries being sent to Electric.
+
+Now:
+
+- `eq(col, null)` and `eq(col, undefined)` transform to `"col" IS NULL` syntax
```
kevin-dp: Where is this transformation happening? I don't see it in the changes?

kevin-dp (approved):

docs/design/eq-null-handling.md
KyleAMathews: yeah telling us what it did — I wanted a more detailed doc for you to look at

kevin-dp: Ok, let's remove it before we merge.

.changeset/fix-electric-null-params.md
```diff
@@ -0,0 +1,9 @@
+---
+'@tanstack/electric-db-collection': patch
+---
+
+Fix invalid Electric proxy queries with missing params for null/undefined values
+
+When comparison operators were used with null/undefined values, the SQL compiler would generate placeholders ($1, $2) in the WHERE clause but skip adding the params to the dictionary. This resulted in invalid queries being sent to Electric.
+
+Now all comparison operators (eq, gt, lt, gte, lte, like, ilike) throw a clear error when used with null/undefined values, since comparisons with NULL always evaluate to UNKNOWN in SQL. Users should use `isNull()` or `isUndefined()` to check for null values instead.
```
samwillis, line 9: This validation is happening within the sql compiler in the electric collection, and so is only applied when a query is pushed down to electric. Is that intended? Would it be better to make this a global validation on all collections? (I'm not sure at this point)

KyleAMathews: hmm yeah I dunno — this seems more like a you question haha — do we want this validation or not? For electric we need it because it breaks SQL compilation. So yeah, we could start here too and then add it later as the global validation, maybe.

samwillis: Yep, let's start with this as is, and then think about global validation.

PR #957, opened 2025-12-03
fix(db-ivm): use row keys for stable ORDER BY tie-breaking

samwillis:
## Summary

- Replace hash-based object ID tie-breaking with direct key comparison for deterministic ordering when ORDER BY values are equal
- Simplify internal data structures by removing the `TaggedValue` abstraction
- Remove `globalObjectIdGenerator` dependency from the topK operators

## Problem

When multiple rows have equal ORDER BY values, the previous implementation used `globalObjectIdGenerator.getId(key)` to generate a numeric tag for tie-breaking. This had several issues:

1. **Hash collisions**: The simple hash function could produce the same value for different keys
2. **Unstable for objects**: Object references get auto-incrementing IDs, which aren't stable across page reloads
3. **Unnecessary complexity**: The `TaggedValue<K, V> = [K, V, Tag]` 3-tuple added overhead

## Solution

Use the row key directly for tie-breaking:

```typescript
// Before (hash-based, potentially unstable):
const tieBreakerA = getTag(a)  // globalObjectIdGenerator.getId(key)
const tieBreakerB = getTag(b)
return tieBreakerA - tieBreakerB

// After (stable, deterministic):
if (aKey === bKey) return 0
if (aKey < bKey) return -1
return 1
```

This works because:
- Row keys are always `string | number` (from collection's `getKey`)
- Each key appears at most once in the topK (enforced by multiplicity tracking)
- Direct comparison is deterministic and stable across sessions

## Changes

- `topKWithFractionalIndex.ts`: 
  - Replace `TaggedValue<K, V>` with `[K, T]` tuple
  - Use direct key comparison for tie-breaking
  - Remove `globalObjectIdGenerator` import
  - Remove `tagValue`, `getKey`, `getVal`, `getTag` helpers
- `topKWithFractionalIndexBTree.ts`: Update type signatures to use `[K, T]`

## Test plan

- [x] All existing `db-ivm` tests pass (267 tests)
- [x] All existing `db` package tests pass (1718 tests)
- [x] ORDER BY tests specifically verified

### DIFF
--- a/packages/db-ivm/src/operators/topKWithFractionalIndex.ts
+++ b/packages/db-ivm/src/operators/topKWithFractionalIndex.ts
@@ -238,10 +238,7 @@ class TopKArray<V> implements TopK<V> {
 export class TopKWithFractionalIndexOperator<
   K extends string | number,
   T,
-> extends UnaryOperator<
-  [K, T],
-  [K, IndexedValue<T>]
-> {
+> extends UnaryOperator<[K, T], [K, IndexedValue<T>]> {
   #index: Map<K, number> = new Map() // maps keys to their multiplicity
 
   /**
@@ -261,7 +258,11 @@ export class TopKWithFractionalIndexOperator<
     super(id, inputA, output)
     const limit = options.limit ?? Infinity
     const offset = options.offset ?? 0
-    this.#topK = this.createTopK(offset, limit, createKeyedComparator(comparator))
+    this.#topK = this.createTopK(
+      offset,
+      limit,
+      createKeyedComparator(comparator),
+    )
     options.setSizeCallback?.(() => this.#topK.size)
     options.setWindowFn?.(this.moveTopK.bind(this))
   }
--- a/.changeset/stable-order-tiebreaker.md
+++ b/.changeset/stable-order-tiebreaker.md
@@ -1,5 +1,5 @@
 ---
-"@tanstack/db-ivm": patch
+'@tanstack/db-ivm': patch
 ---
 
 Use row keys for stable tie-breaking in ORDER BY operations instead of hash-based object IDs.


packages/db-ivm/src/operators/topKWithFractionalIndex.ts
```diff
@@ -261,30 +258,28 @@ export class TopKWithFractionalIndexOperator<K, T> extends UnaryOperator<
     super(id, inputA, output)
     const limit = options.limit ?? Infinity
     const offset = options.offset ?? 0
-    const compareTaggedValues = (
-      a: TaggedValue<K, T>,
-      b: TaggedValue<K, T>
-    ) => {
+    const compareKeyedValues = ([aKey, aVal]: [K, T], [bKey, bVal]: [K, T]) => {
```
kevin-dp: nit: I would extract this to the module-level instead of nesting this definition in the constructor.

packages/db-ivm/src/operators/topKWithFractionalIndex.ts
```diff
@@ -261,30 +258,28 @@ export class TopKWithFractionalIndexOperator<K, T> extends UnaryOperator<
...
       if (valueComparison !== 0) {
         return valueComparison
       }
-      // If the values are equal, compare on the tag (object identity)
-      const tieBreakerA = getTag(a)
-      const tieBreakerB = getTag(b)
-      return tieBreakerA - tieBreakerB
+      // If the values are equal, use the row key as tie-breaker
+      // This provides stable, deterministic ordering since keys are string | number
+      if (aKey === bKey) return 0
```
kevin-dp: This tie-breaking doesn't work if `aKey` and `bKey` have different types, e.g. `'a' < 5` and `5 < 'a'` and `5 == 'a'` all evaluate to false.
We need to keep into account their type:
```ts
if (typeof aKey == typeof bKey) {
  // current comparison
} else if (typeof aKey === 'number') {
  return -1
} else {
  // it's a string
  return 1
}
```

kevin-dp: There is a correct implementation in `comparison.ts` in the db package in your follow up PR:
```ts
/**
 * Compares two keys (string | number) for deterministic ordering.
 * Strings come before numbers, then sorted within type.
 */
export function compareKeys<TKey extends string | number>(
  a: TKey,
  b: TKey
): number {
  // Same type: compare directly
  if (typeof a === typeof b) {
    if (a < b) return -1
    if (a > b) return 1
    return 0
  }
  // Different types: strings come before numbers
  return typeof a === `string` ? -1 : 1
}
```

samwillis: I've used the one from the followup PR, and placed it in `utils.ts` with it exported from the package. Will change the followup to use the same implementation so we only have one copy.

PR #958, opened 2025-12-03
fix(db): ensure deterministic iteration order for collections and indexes

samwillis:
## Summary

This PR ensures deterministic iteration order for collections and indexes when multiple items have equal sort values. It builds on #957 which added stable tie-breaking for ORDER BY operations in the query engine.

**Changes:**

- **SortedMap**: Added key-based tie-breaking when values compare as equal, and optimized to skip value comparison entirely when no comparator is provided
- **BTreeIndex**: Keys within the same indexed value are now returned in deterministic sorted order, with fast paths for empty/single-key sets
- **CollectionStateManager**: Collections now always use `SortedMap` for `syncedData`, ensuring deterministic iteration even without a custom `compare` function
- **Shared utility**: Extracted `compareKeys` to `utils/comparison.ts` for reuse across modules

## Problem

When multiple rows share the same indexed value (e.g., same priority), their iteration order was previously non-deterministic:
- `SortedMap` returned any position among equal values
- `BTreeIndex.valueMap` used `Set` which iterates in insertion order
- Collections without a `compare` function used `Map` with insertion-order iteration

This caused issues with live queries using `orderBy` + `limit` where window operations (`setWindow`) could produce inconsistent results.

## Solution

All three components now use key-based tie-breaking to ensure deterministic ordering:
- When values compare as equal, keys are compared using `compareKeys(a, b)` 
- Strings sort before numbers, then lexicographically/numerically within type

## Test plan

- [x] Added comprehensive test suite (`deterministic-ordering.test.ts`) covering:
  - SortedMap ordering with equal values
  - BTreeIndex `take()`/`takeReversed()` with equal indexed values  
  - Collection iteration with and without `compare` function
  - `currentStateAsChanges` with `orderBy` and `limit`
- [x] All 1747 existing tests pass

**Note:** This PR is stacked on #957 and should be merged after it.

### DIFF
--- a/.changeset/deterministic-collection-ordering.md
+++ b/.changeset/deterministic-collection-ordering.md
@@ -1,5 +1,5 @@
 ---
-"@tanstack/db": patch
+'@tanstack/db': patch
 ---
 
 Ensure deterministic iteration order for collections and indexes.
--- a/packages/db/src/SortedMap.ts
+++ b/packages/db/src/SortedMap.ts
@@ -1,4 +1,4 @@
-import { compareKeys } from "@tanstack/db-ivm"
+import { compareKeys } from '@tanstack/db-ivm'
 
 /**
  * A Map implementation that keeps its entries sorted based on a comparator function
--- a/packages/db/tests/deterministic-ordering.test.ts
+++ b/packages/db/tests/deterministic-ordering.test.ts
@@ -1,9 +1,9 @@
-import { describe, expect, it } from "vitest"
-import { SortedMap } from "../src/SortedMap"
-import { BTreeIndex } from "../src/indexes/btree-index"
-import { createCollection } from "../src/collection/index.js"
-import { PropRef } from "../src/query/ir"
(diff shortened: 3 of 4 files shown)


packages/db/src/indexes/btree-index.ts
```diff
@@ -271,14 +276,25 @@ export class BTreeIndex<
     while ((pair = nextPair(key)) !== undefined && result.length < n) {
       key = pair[0]
       const keys = this.valueMap.get(key)
-      if (keys) {
-        const it = keys.values()
-        let ks: TKey | undefined
-        while (result.length < n && (ks = it.next().value)) {
+      if (keys && keys.size > 0) {
+        // Fast path: single key doesn't need sorting
+        if (keys.size === 1) {
```
kevin-dp: nit: i think this is an over optimization. Sorting a singleton array is fine.

kevin-dp (approved):
Perfect! Love how focused this PR is with small surgical fixes.

Issue #892, opened 2025-11-22
offset isn't passed to loadSubsetOptions


KyleAMathews:
- [ ] I've validated the bug against the latest version of DB packages

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

grunklejp:
Can confirm, the current behavior seems to just increase the limit. This makes direct pagination pretty painful if you are fetching page /?page=400&items=200

this would result in fetching 80,000 rows instead of just 200

KyleAMathews:
This appears fixed in current `main`.

`LoadSubsetOptions` now includes `offset` in `packages/db/src/types.ts`, and `packages/db/src/collection/subscription.ts` tracks/passes offset for limited snapshots so offset-based backends can page without only increasing the limit. `packages/db/src/query/live/collection-subscriber.ts` also has the ordered path that advances based on loaded row counts.

The likely merged fix is part of #960 / `b3b19400`, which added separate cursor expressions and offset support. I also found a side/non-main commit named `89397383 fix(db): pass offset to loadSubset for proper pagination support`; that exact commit is not an ancestor of current `main`, but the behavior appears incorporated by the #960-era code.

Closing as fixed; if `offset` is still missing in a current release, please share a fresh reproduction.

PR #960, opened 2025-12-03
feat(db,electric,query): separate cursor expressions for flexible pagination

samwillis:
## Summary

This PR enhances `LoadSubsetOptions` with separate cursor expressions and offset support, allowing sync layers to choose between cursor-based or offset-based pagination strategies.

**API Changes:**
- Add `CursorExpressions` type with `whereFrom`, `whereCurrent`, and optional `lastKey` properties
- `LoadSubsetOptions.where` no longer includes cursor expressions - these are now passed separately via `cursor`
- Add `offset` to `LoadSubsetOptions` for offset-based pagination support
- Electric sync layer now makes two parallel `requestSnapshot` calls when cursor is present:
  - One for `whereCurrent` (all ties at boundary, no limit)
  - One for `whereFrom` (rows after cursor, with limit)
- Query collection serialization now includes `offset` for query key generation
- Add `truncate` event to collections, emitted when synced data is truncated (e.g., after `must-refetch`)

**Bug Fixes:**
- Fixed `setWindow` pagination: cursor expressions are now correctly built when paging through multi-column orderBy results
- Fixed offset tracking: `loadNextItems` now passes the correct window offset to prevent incorrect deduplication
- `CollectionSubscriber` now listens for `truncate` events to reset cursor tracking state, preventing stale cursor data after `must-refetch`

## Benefits

- Sync layers can choose between cursor-based or offset-based pagination strategies
- Electric can efficiently handle tie-breaking with two targeted requests
- Better separation of concerns between filtering (`where`) and pagination (`cursor`/`offset`)
- `setWindow` correctly triggers backend loading for subsequent pages in multi-column orderBy queries
- Cursor state is properly reset after truncation

## Stacked PR

This PR is stacked on #926 (`samwillis/order-by-improvments`)

## Test plan

- [x] All unit tests pass for @tanstack/db, @tanstack/electric-db-collection, @tanstack/query-db-collection
- [x] E2E tests pass for electric-db-collection and query-db-collection
- [x] Updated test expectations for reduced requestSnapshot call counts (cursor-based pagination is more efficient)
- [x] Multi-column orderBy pagination tests pass with `setWindow`

### DIFF
--- a/packages/db-collection-e2e/src/suites/pagination.suite.ts
+++ b/packages/db-collection-e2e/src/suites/pagination.suite.ts
@@ -415,8 +415,6 @@ export function createPaginationTestSuite(
             .limit(10),
         )
 
-        console.log(`[QUERY DEBUG] query:`, query)
-
         await query.preload()
         await waitForQueryData(query, { minSize: 10 })
 
@@ -435,14 +433,10 @@ export function createPaginationTestSuite(
           }
         }
 
-        console.log(`[QUERY DEBUG] setting window`)
-
         // Move to second page using setWindow
         // IMPORTANT: setWindow returns a Promise when loading is required,
         // or `true` if data is already available. We verify loading occurs.
         const setWindowResult = query.utils.setWindow({ offset: 10, limit: 10 })
-
-        console.log(`[QUERY DEBUG] setWindowResult:`, setWindowResult)
 
         // In on-demand mode, moving to offset 10 should trigger loading
         // since only the first 10 records were initially loaded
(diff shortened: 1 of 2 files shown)


packages/db-collection-e2e/src/suites/pagination.suite.ts
```diff
@@ -415,6 +415,8 @@ export function createPaginationTestSuite(
             .limit(10),
         )
 
+        console.log(`[QUERY DEBUG] query:`, query)
```
KyleAMathews: logging

KyleAMathews (approved):
Nice! Excited to play around with all the improvements in saaslarge next week!

"when cursor is present" — this is used a number of places and it doesn't seem clear when/how this would happen? Is this just for infinite scroll type scenarios?

kevin-dp:
Fixes https://github.com/TanStack/db/issues/892

PR #962, opened 2025-12-03
Fix useLiveQuery loading status in on-demand sync

KyleAMathews:
For live queries using on-demand sync mode, the collection was being marked as ready immediately when source collections were ready, even though the initial loadSubset hadn't completed yet. This meant `isLoading` was false while data was still being fetched.

This fix ensures that:
1. Live queries with on-demand sources track the initial loadSubset promise and show `isLoading: true` until it completes
2. The collection status remains `loading` until the first data load finishes
3. Subsequent loadSubset calls (pagination/windowing) do NOT affect the ready status - only the first load matters

Changes:
- Add `hasMarkedReady` and `hasSetupLoadingListener` flags to track initial ready state in CollectionConfigBuilder
- Modify `updateLiveQueryStatus()` to wait for first loadSubset to complete before calling `markReady()`
- Update `subscribeToMatchingChanges()` in CollectionSubscriber to track the initial loadSubset promise for on-demand sources
- Add comprehensive tests for the new behavior

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/query-db-collection/e2e/query.e2e.test.ts
+++ b/packages/query-db-collection/e2e/query.e2e.test.ts
@@ -204,10 +204,12 @@ describe(`Query Collection E2E Tests`, () => {
       // out before cleanup) causes "[Live Query Error] Source collection was manually
       // cleaned up" warnings. Final cleanup happens in teardown (afterAll).
       //
-      // We do cancel pending queries between tests to prevent "operation was canceled"
-      // errors from accumulating observers.
+      // Remove inactive queries between tests while preserving active subscriptions.
+      // This prevents accumulation of stale query state without disrupting live queries.
       afterEach: async () => {
-        await queryClient.cancelQueries()
+        queryClient.removeQueries({
+          predicate: (query) => query.getObserversCount() === 0,
+        })
       },
       teardown: async () => {
         await Promise.all([


packages/db/src/query/live/collection-config-builder.ts
```diff
@@ -788,15 +800,42 @@ export class CollectionConfigBuilder<
...
+    // If the live query is currently loading a subset (e.g., initial on-demand load),
+    // wait for it to complete before marking ready. This ensures that for on-demand
+    // sync mode, the live query isn't marked ready until the first data is loaded.
+    // We only wait for the FIRST loadSubset - subsequent loads (pagination/windowing)
+    // should not affect the ready status.
+    if (this.liveQueryCollection?.isLoadingSubset) {
+      // Set up a one-time listener if we haven't already
+      if (!this.hasSetupLoadingListener) {
+        this.hasSetupLoadingListener = true
+        const unsubscribe = this.liveQueryCollection.on(
```
samwillis: We should lift up the saved unsubscribe to save it at the top of the collection config builder level, then call it (if it exists) in the collection cleanup function. This is what we do with `unsubscribeFromSchedulerClears` - we should do the same as that.

packages/db/tests/query/live-query-collection.test.ts
samwillis, line 1: Could we add a test that checks this behaviour for an on-demand collection with a join to another on-demand collection. Check that it only modes to ready when both are not loading subsets.

KyleAMathews:
Closing in favor of #1081 which addresses the same issue with a more complete solution (also fixes useLiveSuspenseQuery and addresses the race condition more fundamentally).

PR #970, opened 2025-12-04
fix(react-db): useLiveInfiniteQuery deletion not working with descending order on partial page

### LINKED ISSUE #872: useLiveInfiniteQuery is not updated after deletions on descending order when page is undersized

**Describe the bug**
When using useLiveInfiniteQuery with orderBy(..., 'desc'), deleting an item from the current page does not update the live result if the page has fewer rows than pageSize. The item disappears in storage but remains in the hook’s data until something else changes.

**To Reproduce**
 1. Build a descending infinite query (e.g. messages ordered by id desc, pageSize N).
 2. Ensure the current page has < pageSize items (e.g. last page).
 3. Delete one item.
 4. Observe the hook still returns the deleted item; data never refills.

I created an example in stackblitz so you can understand the issue easier: [https://stackblitz.com/edit/vitejs-vite-zhiz8bwg](https://stackblitz.com/edit/vitejs-vite-zhiz8bwg?file=src%2Fapp.tsx)

**Expected behavior**
After a delete, the hook should fetch the next item in order and the deleted row should disappear.

**Code**
If you have a live infinite query like this and the response has less than 20 items, deletes do not updated the result. If you change the order to `asc` it works.

```tsx
const messageListQuery = useLiveInfiniteQuery(
    (q) =>
      q
        .from({ messages: messagesCollection })
        .orderBy(({ messages }) => messages.id, 'desc'),
    {
      pageSize: 20,
      getNextPageParam: (lastPage, allPages) =>
        lastPage.length === 20 ? allPages.length : undefined,
    },
    []
  );
```

samwillis:
## Summary

Fixes #872

This PR fixes a bug in `useLiveInfiniteQuery` where deleting an item from a partial page (fewer items than `pageSize`) with descending order would not update the live result.

Reproduction of bug: https://stackblitz.com/edit/vitejs-vite-shmvuoh8?file=src%2Fapp.tsx
With the fix: https://stackblitz.com/edit/vitejs-vite-sqrhrspk?file=src%2Fapp.tsx

### Root Cause

The bug was in `requestLimitedSnapshot` in `subscription.ts`. The `biggestObservedValue` variable was incorrectly set to the full row object instead of the indexed value.

This caused the BTree comparison in `index.take()` to fail because comparing a row object with indexed values produces NaN, causing the same data to be loaded multiple times.

### The Fix

Create a value extractor from the orderBy expression to extract the actual indexed value from each row.

### Test Failure (Before Fix)

https://github.com/TanStack/db/actions/runs/19940582123/job/57176990721

## Test plan

- [x] Added reproduction tests
- [x] Verified fix works


hakanshehu:
Hi, thank you for the fix. I was just wondering is there any blocker for this to be merged? When can we expect the fix to be released? Thank you for the work!

PR #971, opened 2025-12-05
Allow access ShapeStream for electric collections

KyleAMathews:
Add `shapeStream` getter to ElectricCollectionUtils that allows users to access the underlying ShapeStream instance from an electric collection. This enables access to ShapeStream properties like the shape handle & offset & subscribing to the raw stream events.

### DIFF
--- a/.changeset/expose-shape-stream.md
+++ b/.changeset/expose-shape-stream.md
@@ -1,5 +1,5 @@
 ---
-"@tanstack/electric-db-collection": minor
+'@tanstack/electric-db-collection': minor
 ---
 
 feat(electric-db-collection): expose underlying ShapeStream via shapeStream getter


samwillis:
This is great, but one thing to note is that the shape stream can change to a new instance during the lifecycle of the collection. We should consider making the collections event emitter extendable so that collections can extend it with new events. Then the electric collection could emit a "shapeStreamChange" event.

we likely don't need this to use a Store either.

I'm happy to pick up extending the event emitter.

KyleAMathews:
oh interesting – yeah maybe just exposing the raw change events & getters for the shapestream metadata is better yeah than a fragile handle on a shapestream that might get changed

KyleAMathews:
Feel free to pickup!

PR #978, opened 2025-12-06
ci: sync changes from other projects

lachlancollins:
## 🎯 Changes

- Disable eslint error in rxdb-db-collection
- Disable vitest error in electric-db-collection
- Update gitignore
- Update repository links
- Remove unnecessary overrides
- Reorder package.json fields

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/electric-db-collection/tests/electric.test-d.ts
+++ b/packages/electric-db-collection/tests/electric.test-d.ts
@@ -155,7 +155,6 @@ describe(`Electric collection type resolution tests`, () => {
     // This causes a constraint error instead of a type mismatch error.
     // Instead, we test via type assignment which will show a proper type error if the types don't match.
     // Currently this shows that todosCollection.utils is typed as UtilsRecord, not ElectricCollectionUtils<TodoType>
-    // @ts-expect-error
     const testTodosUtils: ElectricCollectionUtils<TodoType> =
       todosCollection.utils
 


package.json
```diff
@@ -59,30 +60,5 @@
     "*.{ts,tsx}": [
       "eslint --fix"
     ]
-  },
-  "workspaces": [
```
KyleAMathews, line 63: what happened here @lachlancollins ? Our workspaces and overrides are gone...

lachlancollins: The only workspace config needed is in `pnpm-workspace.yaml`. pnpm also handles all overrides for local packages via `linkWorkspacePackages: true` and `preferWorkspacePackages: true`

KyleAMathews: I added the explicit overrides because the examples weren't loading the workspace versions — we need package versions there so they're installable outside of the repo but inside the repo, they resolve the workspace version.

lachlancollins: Hmmm that shouldn't be an issue, the examples should still load workspace versions if configured correctly - let me check!

KyleAMathews: I tested it some more and apparently the behavior is the workspace version is pulled in when it matches. There's perhaps other cases where it wasn't getting pulled into which I ran into before but in any case, yeah I think we're fine.

PR #980, opened 2025-12-06
ci: more changes from other projects

lachlancollins:
## 🎯 Changes

Having some difficulty migrating to Nx, as seen in #979. This PR takes some of the work around it.

## ✅ Checklist

## 🚀 Release Impact


.husky/pre-commit
```diff
@@ -1,4 +0,0 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-
-pnpm lint-staged
```
KyleAMathews, line 4: @lachlancollins I'm confused what you're doing here — can you write an issue & let us review PRs before you start changing behaviors? Why did you remove our pre-commit hook?

.github/workflows/release.yml
```diff
@@ -40,33 +40,6 @@ jobs:
           title: 'ci: Version Packages'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
KyleAMathews, line 43: why did you remove this?

lachlancollins: NPM tokens are going away as of 9th Dec - I've switched it to publish with trusted publishing

.husky/pre-commit
lachlancollins: Husky's something we don't use in the other projects, and has caused issues for us when used in the past. Happy to reconsider though!

I'll try to introduce Nx again in a subsequent PR.

.github/workflows/release.yml
KyleAMathews: I meant our doc publishing workflow

.husky/pre-commit
KyleAMathews: We find Husky useful & it hasn't caused issues. Please keep your changes focused. I'll add it back. I like sharing config across projects but we don't need to insist on a monoculture.

.github/workflows/release.yml
lachlancollins: Docs are generated during the autofix workflow now, in https://github.com/TanStack/ai/blob/main/.github/workflows/autofix.yml

KyleAMathews: yeah, that's what we started with too but it makes for an awful PR authoring experience as you're constantly running doc generation + the PRs are full of tiny changes. I already put a PR there to change that https://github.com/TanStack/ai/pull/97

lachlancollins: Yeah thats a good point, sounds like a better idea 👍

Issue #934, opened 2025-12-01
eq(…, …) does not work with Date objects


bradleyayers:
- [x] I've validated the bug against the latest version of DB packages

**Describe the bug**

When using `eq(…, …)` with Date objects, it does a referential comparison rather than a `Date#getTime()` comparison. A work around is to use `eq(concat(date1, ''), concat(date2, ''))` but this rounds to the nearest second and discards milliseconds so it's not very good.

**Expected behavior**

I expected that there would be special support for Date objects, as there is already for `UInt8Array` and `Buffer` ([areValuesEqual](https://github.com/TanStack/db/blob/7aedf12996a67ef64010bca0d78d51c919dd384f/packages/db/src/utils/comparison.ts#L170)).

Date handling is already supported by `gt` / `get` / `lt` / `let` comparisons, and also in `normalizeValue` https://github.com/TanStack/db/blob/7aedf12996a67ef64010bca0d78d51c919dd384f/packages/db/src/utils/comparison.ts#L143

kevin-dp:
@bradleyayers Comparison of dates (including equality comparison) should be correct since we are normalizing the dates (into epoch timestamps) before we compare them. I added some unit tests that confirm this behavior and pointed to the relevant code snippets in https://github.com/TanStack/db/pull/992.

I'm closing this issue as everything seems to work correctly. Feel free to reopen this issue if there is a case that i missed.

bradleyayers:
@kevin-dp thanks for taking a look. I dug a bit deeper and found that my issue is because I'm using `eq(…)` in a join, and it seems like the normalisation logic isn't applied in the same way for joins. I had a look at the code and it's a bit more complex because it considers whether to create an index or not (e.g. based on whether it's a computed value, which I think might need to be the case for normalised values?) https://github.com/TanStack/db/blob/63aa8ef8b09960ce0f93e068d41b37fb0503a21a/packages/db/src/query/compiler/joins.ts#L231. This might explain why the `concat` hack I used works because `hasComputedJoinExpr` is probably true in that condition.

Also I don't have permission to re-open the issue in this repo. And apologies for not mentioning the join in the original post, I wrongly assumed `eq(…)` handling.

kevin-dp:
Hi @bradleyayers,
Makes sense, i'm reopening this issue. Would you mind providing a reproducible example or at least the query with join that isn't working? Or is it essentially any join where the join condition involves eq'ing dates?

hugiex:
[...]

PR #992, opened 2025-12-09
Unit tests for data equality comparison

kevin-dp:
Added some unit tests to check whether `eq` works properly on Dates because according to https://github.com/TanStack/db/issues/934 dates are compared by reference instead of by value. However, the tests seem to indicate that the dates are being compared by value and thus are working as expected.

Looking at the code, we do normalize the values before comparing them:
https://github.com/TanStack/db/blob/7aedf12996a67ef64010bca0d78d51c919dd384f/packages/db/src/query/compiler/evaluators.ts#L165-L178

The normalization transforms dates into epoch numbers:
https://github.com/TanStack/db/blob/7aedf12996a67ef64010bca0d78d51c919dd384f/packages/db/src/utils/comparison.ts#L142-L145


packages/db/package.json
samwillis, line 1: Has this leaked in from another branch, or due to where you branched off?

samwillis (approved):
Tests look good, but unsure about the package.json and pnpm-lock change?

Approving for when you address that

packages/db/package.json
kevin-dp: It wouldn't compile for me. I assumed it is due to renovate having upgraded something and now requiring this package explicitly.

PR #995, opened 2025-12-09
fix: deduplicate joins keys when syncMode: 'on-demand'

nathanstitt:
When collections with syncMode: 'on-demand' are joined, the where clause sent to the server via loadSubset often can contain duplicate values in the 'in' condition. This occurs when the active side of the join has multiple rows with the same join key (e.g., multiple posts by the same user).

Before: userId IN (1, 1, 1, 2, 2)
After:  userId IN (1, 2)

I discovered this when developing a [pocketbase adapter](https://github.com/nathanstitt/pbtsdb/) where it was  crumbling under requests like:

image

I also see that DB isn't de-duplicating requests like:
```ts
useLiveQuery(q.from({ books: booksCollection }).where(({ books }) => inArray(books.id, [1,1,1,1,1])),
```
Will also send the where with duplicated values to the collection.  To me that's acceptable though because it's what the caller requested, but I can add deduplication to it if you'd like.

Or maybe this PR is off-base and the collection adapter should be handling this?  I did [add deduplication](https://github.com/nathanstitt/pbtsdb/commit/d18947451e3e458a0e5d08ef74dd8a9da58d0477) to my library but I did find it surprising and others may as well.

## ✅ Checklist

☝️ *this link is broken* db doesn't seem to have a CONTRIBUTING.md at all
☝️ there is not a package.json script for `test:pr`, but I have ran with `pnpm test`

## 🚀 Release Impact

I did not add a changelog because I'd consider this a simple bug fix not worth noting.  happy to do so if you'd like

### DIFF
--- a/packages/db/tests/query/compiler/join-key-deduplication.test.ts
+++ b/packages/db/tests/query/compiler/join-key-deduplication.test.ts
@@ -44,11 +44,26 @@ describe(`Join Key Deduplication`, () => {
       sync: {
         sync: ({ begin, write, commit, markReady }) => {
           begin()
-          write({ type: `insert`, value: { id: 1, userId: 1, title: `Post 1 by User 1` } })
-          write({ type: `insert`, value: { id: 2, userId: 1, title: `Post 2 by User 1` } })
-          write({ type: `insert`, value: { id: 3, userId: 1, title: `Post 3 by User 1` } })
-          write({ type: `insert`, value: { id: 4, userId: 2, title: `Post 1 by User 2` } })
-          write({ type: `insert`, value: { id: 5, userId: 2, title: `Post 2 by User 2` } })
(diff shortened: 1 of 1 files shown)


robertf224:
Relatedly I also noticed:
- it passes along undefined so collection implementations have to [filter those out](https://github.com/robertf224/foundry-tools/blob/0b8bbb11599c73ed5626f2051e1d939a19188bcb/packages/foundry-db/src/objects/convertLoadSubsetOptions.ts#L66)
- it will sometimes pass an empty array which should always return nothing

Think it prob makes sense to handle these upstream rather than pushing to collection implementations

goatrenterguy:
I am running into this issue as well. Makes using joins with `on-demand` almost useless without manually deduping.

PR #1000, opened 2025-12-10
Fix awaitMatch helper on collection inserts and export isChangeMessage

KyleAMathews:
Two issues fixed:

1. **isChangeMessage not exported**: The `isChangeMessage` and `isControlMessage` utilities were exported from electric.ts but not re-exported from the package's index.ts, making them unavailable to users despite documentation stating otherwise.

2. **awaitMatch race condition on inserts**: When `awaitMatch` was called after the Electric messages had already been processed (including up-to-date), it would timeout because:
   - The message buffer (`currentBatchMessages`) was cleared on up-to-date
   - Immediate matches found in the buffer still waited for another up-to-date to resolve

   Fixed by:
   - Moving buffer clearing to the START of batch processing (preserves messages until next batch)
   - Adding `batchCommitted` flag to track when a batch is committed
   - For immediate matches: resolve immediately only if batch is committed (consistent with awaitTxId)
   - For immediate matches during batch processing: wait for up-to-date (maintains commit semantics)

Fixes issue reported on Discord where inserts would timeout while updates worked.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/electric-db-collection/src/electric.ts
+++ b/packages/electric-db-collection/src/electric.ts
@@ -523,6 +523,10 @@ export function electricCollectionOptions<T extends Row<unknown>>(
   // Buffer messages since last up-to-date to handle race conditions
   const currentBatchMessages = new Store<Array<Message<any>>>([])
 
+  // Track whether the current batch has been committed (up-to-date received)
+  // This allows awaitMatch to resolve immediately for messages from committed batches
+  const batchCommitted = new Store<boolean>(false)
+
   /**
    * Helper function to remove multiple matches from the pendingMatches store
    */
@@ -560,6 +564,7 @@ export function electricCollectionOptions<T extends Row<unknown>>(
     syncMode: internalSyncMode,
     pendingMatches,
     currentBatchMessages,
+    batchCommitted,
     removePendingMatches,
     resolveMatchedPendingMatches,
     collectionId: config.id,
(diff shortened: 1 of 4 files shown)


docs/reference/electric-db-collection/functions/isChangeMessage.md
samwillis, line 1: Are these and the other changes in `docs/reference/` created by the build script or a Claude creation?

packages/electric-db-collection/src/electric.ts
```diff
@@ -476,6 +476,10 @@ export function electricCollectionOptions<T extends Row<unknown>>(
   // Buffer messages since last up-to-date to handle race conditions
   const currentBatchMessages = new Store<Array<Message<any>>>([])
 
+  // Track whether the current batch has been committed (up-to-date received)
+  // This allows awaitMatch to resolve immediately for messages from committed batches
+  const batchCommitted = new Store<boolean>(false)
```
samwillis, line 528: Nit: Not sure this needs to be a store as we are not reactively watching it? I guess Claude is following the pattern in the file.

(we don't really need store in here at all)

docs/reference/electric-db-collection/functions/isChangeMessage.md
KyleAMathews: This was the auto fix command that was running for a bit

PR #1001, opened 2025-12-11
feat(svelte-db): Add findOne() / SingleResult support to useLiveQuery

elohmeier:
Adds `SingleResult` support to `useLiveQuery`, matching the React implementation. When using `.findOne()`, data is typed as `T | undefined` instead of `Array<T>`.

Basically ports #440 to svelte-db.


samwillis (approved):
Thanks @elohmeier

PR #1004, opened 2025-12-11
Handle subset end message in Electric collection

kevin-dp:
Adds support for the subset-end message introduced in https://github.com/electric-sql/electric/pull/3582


samwillis (approved):
Perfect :shipit:

PR #1008, opened 2025-12-12
Handle invalid collection getKey return values

KyleAMathews:
Throws InvalidKeyError when getKey returns values other than string or number (e.g., null, objects, booleans). The validation is optimized to only require 1 typeof check on the happy path (string keys).

- Add InvalidKeyError class to errors.ts
- Update generateGlobalKey to validate key type before generating key
- Add tests for invalid key types (null, object, boolean)
- Add tests confirming valid keys (string, number, empty string, zero)

## ✅ Checklist

## 🚀 Release Impact


samwillis (approved):
Lgtm

PR #1014, opened 2025-12-13
consider isLoadingSubset in useLiveQuery isLoading state

nathanstitt:
Fix a bug where useLiveQuery would return isLoading=false while data was still undefined for on-demand sync mode. This would frequently cause flicker of 'not found' type messages.

For example,   I added a console.log of  isLoading, data for a component I'm working on:
```
{data: undefined, queryLoading: false}
{data: undefined, queryLoading: false}
{data: Array(0), queryLoading: true}
{data: Array(1), queryLoading: false}
```

The useLiveQuery hook now subscribes to loadingSubset:change events and factors isLoadingSubset into the isLoading computation.

I'm not 100% sure this is the best approach to fixing this, but it _seems_ like the best approach I could find given my relative lack of knowledge about the codebase.  Happy to rework with any suggestions anyone has to resolve this.

## ✅ Checklist

the below do not exist but I've followed the spirt of them :) 

## 🚀 Release Impact

### DIFF
--- a/packages/db/src/collection/changes.ts
+++ b/packages/db/src/collection/changes.ts
@@ -108,7 +108,9 @@ export class CollectionChangesManager<
     })
 
     if (options.includeInitialState) {
-      subscription.requestSnapshot({ trackLoadSubsetPromise: false })
+      // Track the loadSubset promise so isLoadingSubset reflects async data loading
+      // This ensures on-demand sync shows loading state until data is available
+      subscription.requestSnapshot()
     }
 
     // Add to batched listeners
--- a/packages/db/tests/query/live-query-collection.test.ts
+++ b/packages/db/tests/query/live-query-collection.test.ts
@@ -5,6 +5,7 @@ import {
   and,
   createLiveQueryCollection,
   eq,
+  gt,
   ilike,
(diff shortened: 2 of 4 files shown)


samwillis:
Thanks for looking into this. I think with this change `isLoading` would go back to `true` if the live query had to load data due to change on a source collection, such as loading from a joint collection for a new row. If the user is showing a full screen loading message that would show up when this happens.

I think what we need is two fold:

1. To expose the `isLoadingSubset` on the returned object from `useLiveQuery`
2. To track if the initial run of the query triggers a loadSubset and if so wait for that to resolve before setting loading status of the live query collection to `ready`. The useLiveInfiniteQuery hook does something similar when loading the next page, but we should do this for the `collection.status` not just the `useLiveQuery` hook.

@nathanstitt would you like to pick up making those changes? I'm happy to help

samwillis:
Ah! Kyle has an open PR for fixing this. Sorry @nathanstitt, I think we will use that one.

nathanstitt:
no prob at all, my knowledge of this was pretty shaky and wasn't sure this was the proper fix anyway.  glad it's been addressed tho, was most annoying to work around 😁

PR #1021, opened 2025-12-15
Prevent JavaScript in query selections

KyleAMathews:
## Summary

Adds warnings when users accidentally use JavaScript operators (`||`, `&&`, `??`, `?:`) in query callbacks. Warnings appear in development mode only; production has zero overhead.

**User-visible impact:** Developers get clear, actionable warnings when making this common mistake, instead of silently broken queries.

---

## Root Cause

Query callbacks like `.select()`, `.where()`, and `.having()` receive RefProxy objects that record property access paths. When users write:

```ts
.select(({ user }) => ({ name: user.name || 'default' }))
```

The `||` operator evaluates immediately at construction time. Since `user.name` is a RefProxy object (truthy), it always returns the RefProxy, ignoring `'default'`. The query silently produces unexpected results.

## Approach

Two-layer detection strategy:

1. **Static source analysis** (dev-only, warns) - Parse callback source via `.toString()` and detect operator patterns after stripping string literals and comments
2. **Runtime Symbol.toPrimitive trap** (always, throws) - Catch when RefProxy is coerced to a primitive (string concatenation, arithmetic)

The regex-based static analysis warns rather than throws because it can have false positives (e.g., operators in regex literals like `/a||b/`). The Symbol.toPrimitive traps have no false positives and throw errors.

## Key Invariants

- Optional chaining (`?.`) must NOT trigger false positives
- Operators inside string literals must NOT trigger warnings
- Static detection skipped entirely in production (`NODE_ENV === 'production'`)
- Symbol.toPrimitive traps work in all environments (no false positives)

## Non-goals

- **AST parsing** - Heavy dependency, slower, overkill for this use case
- **Blocking on false positives** - Regex can't perfectly parse JS; warning is appropriate
- **Comparison operators** (`===`, `>`, `<`) - Handled separately via return type validation

## Trade-offs

| Alternative | Why not |
|-------------|---------|
| Throw on all detections | False positives in regex literals would block valid code |
| Runtime-only detection | Would only catch primitive coercion, miss `\|\|`/`&&`/`??` which return objects |
| AST parsing | Heavy dependency for a dev-time warning |
| Production detection | Unnecessary overhead for something caught in dev |

---

## Verification

```bash
pnpm test  # All tests pass
```

## Files Changed

| File | Change |
|------|--------|
| `src/errors.ts` | Add `JavaScriptOperatorInQueryError` with helpful message and examples |
| `src/query/builder/ref-proxy.ts` | Add `Symbol.toPrimitive` trap to all proxy types, `checkCallbackForJsOperators()` with dev-only warning |
| `src/query/builder/index.ts` | Call `checkCallbackForJsOperators` in `select()`, `where()`, `having()` |
| `tests/query/builder/ref-proxy.test.ts` | Test all operators, string literal handling, production mode skip |
| `tests/query/scheduler.test.ts` | Fix existing test to use `coalesce()` instead of `??` |
| `tests/query/live-query-collection.test.ts` | Fix existing test to use `!` assertion instead of `??` |


KyleAMathews (commented):
## Code Review: PR #1021

**Overall Assessment: Approve** ✅

This is an excellent developer experience improvement. Like warning signs at the edge of a cliff, these checks prevent users from falling into a pit of silent query bugs. The dual detection strategy (static + runtime) provides comprehensive coverage.

### Summary

The PR adds detection for JavaScript operators (`||`, `&&`, `??`, `?:`) in query callbacks that would silently produce incorrect results due to eager evaluation at construction time.

### Strengths

1. **Two-layer detection strategy**:
   - **Static**: Parse callback source code and detect patterns
   - **Runtime**: `Symbol.toPrimitive` trap catches primitive coercion

2. **Excellent error messages**: The `JavaScriptOperatorInQueryError` provides:
   - Clear explanation of why it's wrong
   - Specific suggestions for alternatives
   - Before/after code examples

3. **Smart string/comment stripping**: The `stripStringsAndComments` function prevents false positives from operators inside string literals.

4. **Thorough test coverage**: Tests cover all operators, string literals, optional chaining, and the known regex limitation.

5. **Documentation of limitations**: The regex false positive is documented as a known limitation - this is the right call since regex literals in query callbacks are rare.

### Technical Notes

1. **ref-proxy.ts:283-287**: The ternary pattern `\?[^.?][^:]*:` correctly excludes optional chaining (`?.`) and nullish coalescing (`??`).

2. **Symbol.toPrimitive location**: Correctly added to all three proxy creation sites (root, alias, and nested property access).

3. **Test fixes**: Good catch on the existing tests using `??` - the fixes to use `coalesce()` and `!` assertion are appropriate.

### Minor Suggestions

1. **Consider orderBy**: Should `orderBy` also get the check? Users might try `({ users }) => users.name || 'zzz'` for null-last sorting.

2. **Error message enhancement**: Consider mentioning in the error that these operators are evaluated at *construction* time vs *execution* time - this is the key insight that helps users understand why.

3. **Performance consideration**: The `callback.toString()` approach has minimal overhead since it only runs once per query construction, not per row. Good trade-off.

### Question

For minified code (e.g., production builds), `callback.toString()` may return mangled code. The detection might not work, but the queries themselves still fail silently. Is this a concern, or is the assumption that developers catch these in development?

Great DX improvement that will save developers debugging time! 🛡️

packages/db/src/query/builder/ref-proxy.ts
```diff
@@ -303,3 +349,83 @@ export function isRefProxy(value: any): value is RefProxy {
...
+ * // This will log a warning in dev:
+ * checkCallbackForJsOperators(({users}) => users.data || [])
+ *
+ * // This is fine:
+ * checkCallbackForJsOperators(({users}) => users.data)
+ */
+export function checkCallbackForJsOperators<
+  T extends (...args: Array<any>) => any,
+>(callback: T): void {
+  if (process.env.NODE_ENV !== `production`) {
```
samwillis, line 376: I think this will throw in the browser as there is no `process` object. Some bundlers may strip this out, but I don't think we can depend on that?

something like:
```
const __DEV__ =
  typeof process !== "undefined" &&
  process.env &&
  process.env.NODE_ENV !== "production";

if (__DEV__) {
  // ...
}
```

samwillis (changes requested):
Other than above, I'm still a little hesitant about parsing with regex, there will be false positives where it warns about things that look like operators in strings / comments. We could extend `checkCallbackForJsOperators` to skip strings and comments when validating - I think its doable but would ad more code.

The other thing I think we should validate is the bundlers stripping out the validation code in a prod build. I had Cursor/Opus do some validation on anther PR to do with tree shaking, it did a good job of validating which bundlers worked and which didn't. I would ask it to look at validating that the code is removed in the prod build.

I think if we can be 100% sure that when DB is used as a lib this code is striped, then extending it a little to skip operators in string and comments would be a good idea.

sirgagahandino9-design (approved):

KyleAMathews:
Not going to do this — too much code & risk for too little gain.

PR #1023, opened 2025-12-15
Query collection select writeupsert error

samwillis:
## 🎯 Changes

Fixes a bug where `collection.utils.writeInsert()` or `collection.utils.writeUpsert()` would throw an error when the collection was configured with a `select` option.

The issue was that `setQueryData` in `performWriteOperations` would overwrite the query cache with a raw array of items, which conflicted with the `select` function's expectation of the original wrapped response format.

The fix involves:
- Adding a `hasSelect` flag to the `SyncContext`.
- Skipping `queryClient.setQueryData` in `performWriteOperations` when `hasSelect` is true, as the collection's internal state is already updated, and the cache format differs.
- Two new tests have been added to reproduce and verify the fix for `writeInsert` and `writeUpsert` with `select`.

## ✅ Checklist

## 🚀 Release Impact

---
<a href="https://cursor.com/background-agent?bcId=bc-49e1eb5c-32f7-47ba-bce1-ae9ae8a70407"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-cursor-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-cursor-light.svg">Open in Cursor</picture></a>&nbsp;<a href="https://cursor.com/agents?id=bc-49e1eb5c-32f7-47ba-bce1-ae9ae8a70407"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-web-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-web-light.svg">Open in Web</picture></a>

### DIFF
--- a/examples/react/paced-mutations-demo/package.json
+++ b/examples/react/paced-mutations-demo/package.json
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "@tanstack/db": "^0.5.11",
-    "@tanstack/react-db": "^0.1.55",
+    "@tanstack/react-db": "^0.1.56",
     "mitt": "^3.0.1",
     "react": "^19.2.1",
     "react-dom": "^19.2.1"
--- a/examples/react/todo/package.json
+++ b/examples/react/todo/package.json
@@ -5,8 +5,8 @@
   "dependencies": {
     "@tanstack/electric-db-collection": "^0.2.12",
     "@tanstack/query-core": "^5.90.12",
-    "@tanstack/query-db-collection": "^1.0.6",
-    "@tanstack/react-db": "^0.1.55",
+    "@tanstack/query-db-collection": "^1.0.7",
+    "@tanstack/react-db": "^0.1.56",
     "@tanstack/react-router": "^1.140.0",
     "@tanstack/react-start": "^1.140.0",
     "@tanstack/trailbase-db-collection": "^0.1.55",
--- a/examples/solid/todo/package.json
+++ b/examples/solid/todo/package.json
@@ -5,7 +5,7 @@
   "dependencies": {
     "@tanstack/electric-db-collection": "^0.2.12",
     "@tanstack/query-core": "^5.90.12",
-    "@tanstack/query-db-collection": "^1.0.6",
+    "@tanstack/query-db-collection": "^1.0.7",
     "@tanstack/solid-db": "^0.1.54",
     "@tanstack/solid-router": "^1.140.0",
     "@tanstack/solid-start": "^1.140.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -257,7 +257,7 @@ importers:
         specifier: ^0.5.11
         version: link:../../../packages/db
       '@tanstack/react-db':
-        specifier: ^0.1.55
+        specifier: ^0.1.56
         version: link:../../../packages/react-db
       mitt:
         specifier: ^3.0.1
@@ -433,10 +433,10 @@ importers:
         specifier: ^5.90.12
         version: 5.90.12
       '@tanstack/query-db-collection':
-        specifier: ^1.0.6
+        specifier: ^1.0.7
         version: link:../../../packages/query-db-collection
       '@tanstack/react-db':
-        specifier: ^0.1.55
+        specifier: ^0.1.56
         version: link:../../../packages/react-db
       '@tanstack/react-router':
         specifier: ^1.140.0
@@ -554,7 +554,7 @@ importers:
         specifier: ^5.90.12
         version: 5.90.12
       '@tanstack/query-db-collection':
-        specifier: ^1.0.6
+        specifier: ^1.0.7
         version: link:../../../packages/query-db-collection
       '@tanstack/solid-db':
         specifier: ^0.1.54


samwillis:
@cursoragent please don't inline updateCacheData in the context, lift it out somewhere to keep it tidy

samwillis:
@cursoragent please address this review

I like the suggestion of using select to get the array then traverse the object to find where it is, use that!

---

Here’s what PR TanStack/db#1023 is doing, how well it solves the problem, and what I’d tweak before merging.

What it fixes

The underlying bug is real and easy to reproduce: when a Query Collection is configured with a select that extracts items from a wrapped query result (e.g. { data: [...], meta: ... }), the direct write helpers (writeInsert, writeUpsert, etc.) were overwriting the TanStack Query cache with a raw array, so subsequent select executions would see the wrong shape and return undefined, producing the “select must return an array” error.  ￼

This also contradicts the docs’ promise that “Direct writes … also update the TanStack Query cache.”  ￼

What the PR changes

1) It routes cache updates through a new updateCacheData hook

SyncContext gets a new updateCacheData(items) function, and performWriteOperations calls that instead of directly doing queryClient.setQueryData(queryKey, updatedData).  ￼

That’s the right seam: the sync/write layer shouldn’t need to know how the query cache is shaped.

2) It implements updateCacheData inside queryCollectionOptions

The implementation is:
	•	If no select: cache is assumed to be the raw array → set it directly.
	•	If select exists: treat cache as a wrapper object, and attempt to preserve wrapper structure by:
	•	reading the old cached object
	•	finding the first top-level property that is an array
	•	shallow-copying the wrapper and replacing that property with the new items array  ￼

This is directionally correct: preserve wrapper + metadata; only swap the item array.

3) It adds tests that cover the regression and the intended behavior

You added:
	•	“should not log the select error” for writeInsert and writeUpsert under select  ￼
	•	“should keep the wrapped cache format preserved” across insert/update/delete (metadata stays intact; data mutates)  ￼

Those tests are doing the job: they fail on the old behavior and validate the key invariant (wrapper shape survives).

What I like
	•	Correct abstraction boundary: write ops call ctx.updateCacheData, and the Query Collection layer decides how to update the cache.  ￼
	•	Behavior matches docs: select is explicitly described as “extract array items when they’re wrapped with metadata” in the Query Collection docs, so preserving wrapper structure is exactly what users expect.  ￼
	•	The cache-preservation test is the right “hard” assertion (it’s better than purely checking for absence of console noise).  ￼

Things I would change (these are the sharp edges)

1) The “first array property wins” heuristic is risky

Right now, with select, you scan Object.keys(oldData) and replace the first array-valued property.  ￼

That works for { data: [...] }, but can go wrong for wrappers like:

{ errors: [], data: [...], meta: {...} }

If errors comes first, you’d overwrite errors and leave data untouched.

A more robust approach that still stays generic:
	•	Prefer common keys if present (data, items, results)
	•	Or, even better: use the select function to identify the property when it’s a pure extractor.

A practical trick:
	•	compute const selected = select(oldData)
	•	if selected is an array, find a top-level property prop where oldData[prop] === selected (reference equality)
	•	replace that prop

That handles “multiple arrays” wrappers cleanly without guessing.

2) In select mode, the fallback return items can reintroduce the bug

In the select branch you have several “give up” paths that return the raw array (items) (e.g., oldData missing, oldData not object, no array prop found).  ￼

But returning the raw array is exactly what caused the select mismatch in the first place.

If you can’t safely preserve wrapper shape, the safest behavior is usually:
	•	don’t corrupt the cache → return oldData unchanged (or undefined if it was undefined)
	•	optionally log a warning explaining you couldn’t locate the array field to patch

Yes, that means the cache might not reflect direct writes in that edge case, but it avoids breaking select and spamming the console.

3) Types could be tightened (minor, but low effort)

updateCacheData is typed as (items: Array<any>) in query.ts  ￼ despite SyncContext being generic (Array<TRow>) in manual-sync.ts.  ￼

If queryCollectionOptions already knows the row type (it should), it’d be worth threading that through so updateCacheData becomes (items: TRow[]).

Also: const key = queryKey as unknown as Array<unknown> is a bit of a smell. If queryKey is already a TanStack Query QueryKey, you shouldn’t need to cast it into a mutable array type.

4) Public API surface: SyncContext change might be a TS-breaking change

If SyncContext is exported and used externally, adding a required updateCacheData can break downstream TypeScript builds.  ￼

If that type is public, consider making it optional:

updateCacheData?: (items: Array<TRow>) => void

…and then in performWriteOperations do:
	•	if updateCacheData exists → use it
	•	else → fallback to old queryClient.setQueryData(...)

That keeps compatibility while fixing the issue for Query Collections (which will pass it).

5) Changeset / release bookkeeping

The PR looks like a user-facing bugfix in a published package; the PR itself shows the changeset-bot complaining there’s no changeset.  ￼
If this repo expects changesets for patches, you probably want a patch changeset for @tanstack/query-db-collection.

Suggested “best of both worlds” cache patching logic

If you want one concrete direction that’s still generic and avoids false positives:
	1.	If oldData is a plain object:
	•	if Array.isArray(oldData.data) → update data
	•	else if Array.isArray(oldData.items) → update items
	•	else if you can compute selected = select(oldData) and it’s an array:
	•	find a key where oldData[key] === selected and update that key
	•	else fallback to “first array prop”
	2.	If none of that worked: return oldData unchanged (don’t set raw array)

That would preserve the “it works for typical wrappers” behavior while not regressing into the original bug on weird wrappers.

KyleAMathews (approved):
:shipit:

PR #1025, opened 2025-12-16
fix(db): re-request subsets after truncate for on-demand sync mode

ibarakaiev:
When a must-refetch (409) occurs in on-demand sync mode, the collection receives a truncate which clears all data and resets the loadSubset deduplication state. However, subscriptions were not re-requesting their previously loaded subsets, leaving the collection empty.

This fix adds a truncate event listener to CollectionSubscription that:
1. Resets pagination/snapshot tracking state (but NOT sentKeys)
2. Re-requests all previously loaded subsets

Intentionally keeping sentKeys intact because the truncate event is emitted BEFORE delete events are sent to subscribers. If we cleared sentKeys, delete events would be filtered by filterAndFlipChanges.

### DIFF
--- a/packages/db/src/collection/subscription.ts
+++ b/packages/db/src/collection/subscription.ts
@@ -79,6 +79,9 @@ export class CollectionSubscription
   private _status: SubscriptionStatus = `ready`
   private pendingLoadSubsetPromises: Set<Promise<void>> = new Set()
 
+  // Cleanup function for truncate event listener
+  private truncateCleanup: (() => void) | undefined
+
   public get status(): SubscriptionStatus {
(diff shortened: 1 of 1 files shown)


ibarakaiev:
Note: the current implementation still causes a flicker, i.e. when 409 is received, data is emptied then once fetching completes it is repopulated, which is noticeable in the UI. Not sure what's the best way around that.

samwillis:
Hey @ibarakaiev, this was a really great start. I ended up forking it to add buffering on the live query subscriptions after a truncate such that there is no flash of no content at all. #1043 replaces this.

PR #1042, opened 2025-12-17
Move identifier quoting to Electric client

### LINKED ISSUE #1032: `columnMapper` encoding not applied to column names in subset queries
- [x] I've validated the bug against the latest version

## Describe the bug

When using `columnMapper: snakeCamelMapper()` in `shapeOptions`, column names in subset queries (used by `on-demand` and `progressive` sync modes) are sent to the server in camelCase instead of being converted to snake_case.

This affects:
- **WHERE clauses** - e.g., `subset__where: "workflowId" = $1`
- **ORDER BY clauses** - e.g., `subset__order_by: "createdAt" DESC`
- **Function arguments** - e.g., `upper("firstName")`

This causes queries to fail because the database columns are snake_case.

## To Reproduce

1. Create a collection with `columnMapper: snakeCamelMapper()` and `syncMode: 'on-demand'` or `'progressive'`
2. Query the collection with a filter or ordering on a camelCase property (e.g., `workflowId`, `createdAt`)
3. Observe the network request - the `subset__where` and `subset__order_by` parameters contain camelCase column names instead of snake_case

## Root cause

In `packages/electric-db-collection/src/sql-compiler.ts`:

```typescript
function quoteIdentifier(name: string): string {
  return `"${name}"`  // Quotes camelCase directly without encoding
}

// Used here:
case `ref`:
  return quoteIdentifier(exp.path[0]!)  // e.g., returns "workflowId" not "workflow_id"
```

The `compileSQL` function doesn't have access to the `columnMapper`, so it can't encode column names before quoting them.

Electric's `encodeWhereClause` *does* get called on the compiled SQL, but it intentionally preserves double-quoted identifiers (correct PostgreSQL semantics), so the camelCase names pass through unchanged.

## Additional context

- Electric client version: 1.3.0
- TanStack DB version: latest

KyleAMathews:
Remove quoteIdentifier function from sql-compiler.ts so that Electric's encodeWhereClause can see the original unquoted identifiers and properly apply columnMapper transformations (e.g., camelCase → snake_case) before quoting.

This fixes an issue where columnMapper encoding was not applied to subset queries because double-quoted identifiers were preserved by Electric's encodeWhereClause per PostgreSQL semantics.

Fixes #1032

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/electric-db-collection/src/sql-compiler.ts
+++ b/packages/electric-db-collection/src/sql-compiler.ts
@@ -75,9 +75,7 @@ function serializeExpression(
 /**
  * Serializes IR OrderBy clauses to structured format.
  */
-function serializeOrderBy(
-  orderBy: IR.OrderBy,
-): Array<SerializedOrderByClause> {
+function serializeOrderBy(orderBy: IR.OrderBy): Array<SerializedOrderByClause> {
   return orderBy.map((clause: IR.OrderByClause) => {
     const { expression, compareOptions } = clause
     if (expression.type !== `ref`) {
@@ -100,7 +98,9 @@ function serializeOrderBy(
   })
 }
 
-export function compileSQL<T>(options: LoadSubsetOptions): ExtendedSubsetParams {
+export function compileSQL<T>(
+  options: LoadSubsetOptions,
+): ExtendedSubsetParams {
   const { where, orderBy, limit } = options
 
   const params: Array<T> = []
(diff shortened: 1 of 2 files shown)


samwillis (changes requested):
Good in principal, but the e2e tests are failing and need investigating + changeset

KyleAMathews:
closing in favor of https://github.com/electric-sql/electric/pull/3629

Issue #921, opened 2025-11-27
Offline transactions applied twice on reload


oskarrough:
I'm trying to make use of the offline-transactions module together with a tanstack db query collection and think I found a bug.

When using @tanstack/offline-transactions, the `loadAndReplayTransactions()` method is called twice during initialization, causing persisted offline transactions to be executed multiple times (duplicating data).

## To reproduce

  1. Create an offline executor with a mutation function
  2. Go offline
  3. Create a transaction (e.g., insert a record)
  4. Close the tab while still offline
  5. Reopen the tab (now online)
  6. The mutation executes 2+ times, creating duplicate records

Note, if I don't refresh the tab, and only switch from offline to online, it works! No duplicated data.

## Expected behavior

loadAndReplayTransactions() should only be called once during initialization.
OR, there is some deduping going on?

## Possible fix

Not being overly confident with how this all plays together, I believe the issue is that `loadAndReplayTransactions()` is called in both `initialize()` and in `setupEventListeners()` inside `OfflineExecutor.ts`.

My AI robot overlord suggests to add a guard, but not sure that's the cleanest way to deal with this.

```js
  // In initialize():
  this._initialLoadDone = false;
  this.setupEventListeners();
  if (isLeader) {
    await this.loadAndReplayTransactions();
    this._initialLoadDone = true;
  }

  // In setupEventListeners():
  if (isLeader && this._initialLoadDone) {
    this.loadAndReplayTransactions();
  }
```

**Desktop:**

- OS: OSX
- Browser Chrome
- Version 142
"@tanstack/offline-transactions": "^1.0.0",
"@tanstack/query-db-collection": "^1.0.5",
"@tanstack/query-persist-client-core": "^5.91.10",
"@tanstack/svelte-db": "^0.1.49",
"@tanstack/svelte-query": "^6.0.9",

tobacode:
Yup, I have the same problem and posted a similar message on Discord a few hours ago.
https://discord.com/channels/719702312431386674/1451279903642222724/1451279903642222724

tobacode:
Should be fixed with https://github.com/TanStack/db/pull/1046

PR #1046, opened 2025-12-18
Fix a race condition in WebLocksLeader where transactions could be replayed twice on page load

KyleAMathews:
Fix a race condition in WebLocksLeader where transactions could be replayed twice on page load. The issue occurred because:

1. requestLeadership() returned true immediately when lock was available
2. But notifyLeadershipChange(true) was called asynchronously when the fire-and-forget navigator.locks.request() actually acquired the lock
3. This caused loadAndReplayTransactions() to be called twice - once manually and once via the callback

The fix sets isLeaderState = true synchronously when we know the lock is available, so the async notifyLeadershipChange(true) doesn't trigger the callback again (the guard checks isLeaderState !== isLeader).

Reported by Toba on Discord.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/.changeset/brave-foxes-dance.md
+++ b/.changeset/brave-foxes-dance.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/offline-transactions': patch
+---
+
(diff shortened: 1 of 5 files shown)


kevin-dp:
Fixes https://github.com/TanStack/db/issues/921

PR #1058, opened 2025-12-20
feat: add writeUpsert util to localstorage collection

LuggaPugga:
## 🎯 Changes

This PR adds a `writeUpsert` utility to the localstorage collection, inspired by the upsert utility in the query collection
Feel free to let me know if you'd like any changes

## ✅ Checklist

- [X] I have tested this code locally with `pnpm test`.

## 🚀 Release Impact

- [X] This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).


samwillis:
Hey @LuggaPugga, could I ask what the motivation is for this api? I'm not 100% sure how this is different from just calling `.write()` on a local stage collection, both write to the local storage and update the synced state, but by slight different mechanisms.

LuggaPugga:
> Hey @LuggaPugga, could I ask what the motivation is for this api? I'm not 100% sure how this is different from just calling `.write()` on a local stage collection, both write to the local storage and update the synced state, but by slight different mechanisms.

I have a JSON object and I want to update one of its values without needing to include the previous data in the `.write()`
And update won't work when the object isn't there yet.

PR #1068, opened 2025-12-23
feat: handle delete-insert in transaction merging

marbemac:
## 🎯 Changes

So related to the whole longer lived editing session concept, running into delete and the insert issue on pending mutation. Basically if user deletes something in our editor, we tx.mutate and delete the record. Then if the user cmd+z to undo in our editor, we reverse the operation and insert the record that was just deleted (on same tx, it has not been committed yet). 

This pattern results in the Uncaught Error: Unhandled mutation combination: delete-insert error.

This PR adds handling for delete-insert pattern.

## ✅ Checklist

## 🚀 Release Impact


marbemac:
Tests failing with:

```
error Dependency @tanstack/query-db-collection has multiple versions defined in the workspace. multiple-dependency-versions

error Dependency @tanstack/react-db has multiple versions defined in the workspace. multiple-dependency-versions
```

But only in CI - works on my local machine. Let me know if there's something I should do to resolve them!

samwillis:
Hey @marbemac 

This looks ok to me, but @KyleAMathews should cast his eyes over it as he spent some time working through the logic around handing these overlapping mutations in a transaction.

CI does a soft merge on main when it runs, and it was main that was broken (due to a new check being added)... CI is now green.

marbemac:
hi team, just following up here to see if this is still something ya'll might consider?

oschade:
I ran into the same issue today.

transaction.ts says:

```
  * Note: (delete, update) and (delete, insert) should never occur as the collection 
  * layer prevents operations on deleted items within the same transaction.
```

But I can confirm: it is possible to delete first and insert afterwards in the same tx, but then you get this error "Unhandled mutation combination: delete-insert"

KyleAMathews:
Closing as superseded by merged PR #1835, which fixes same-key delete-then-insert reduction with broader authoritative, repeated, and concurrent coverage while preserving the original contributor credit.

PR #1069, opened 2025-12-23
fix(electric-db-collection): prevent orphan transactions after must-refetch in progressive mode

ibarakaiev:
When a `must-refetch` message is received in progressive mode, it starts a transaction with `begin()` and calls `truncate()`. This resets `hasReceivedUpToDate` to `false`, causing `isBufferingInitialSync()` to return `true`.

The bug: subsequent messages after must-refetch were being buffered instead of written to the existing transaction. When `up-to-date` was received, the atomic swap code would create a NEW transaction, leaving the first transaction (from must-refetch) uncommitted forever. This "orphan transaction" caused the collection to become corrupted with undefined values.

The fix: Add `&& !transactionStarted` checks to 5 places so that when a transaction is already started (from must-refetch), messages are written directly to it instead of being buffered for atomic swap.

## 🚀 Release Impact


samwillis (approved):
This is great. I have added tests and validated that they fail on main and pass on this branch.

PR #1070, opened 2025-12-23
fix: handle overlapping subset queries returning same row with different values

ibarakaiev:
When multiple subset queries return the same row (e.g., different WHERE clauses that both match the same record), the server sends `insert` operations for each response. If there's a race condition and the row's data changed between requests (e.g., timestamp field updated), this caused a DuplicateKeySyncError because TanStack DB's sync layer throws when inserting an existing key with a different value.

This fix tracks synced keys in the Electric adapter and converts subsequent `insert` operations to `update` for keys that have already been synced. The tracked keys are cleared on truncate/must-refetch to stay in sync with the collection state.

Note: this might be against the original assumptions of the codebase, so perhaps there's a more reliable way to fix this instead of converting `insert` to `update`. Making this PR as a starting point.

## 🚀 Release Impact


samwillis (approved):
This is great @ibarakaiev - I've added some tests that validate that the fix workes. Going to merge

PR #1076, opened 2025-12-29
fix: ensure sync is started for on-demand collections in join queries

ibarakaiev:
When using join queries with collections configured with `syncMode: 'on-demand'`, the source collections were never having their sync initialized. This caused join queries to return empty results because:

1. On-demand collections use `offset: 'now'` and only receive changes from when sync starts, not historical data
2. The lazy join loader calls `requestSnapshot()` to fetch matching rows
3. Without `startSyncImmediate()`, the collection stays idle and `requestSnapshot()` has nothing to work with

The fix adds a `collection.startSyncImmediate()` call in `subscribeToAllCollections()` before subscribing, ensuring all source collections in a query (including those used in joins) have their sync properly initialized.

## 🚀 Release Impact

### DIFF
--- a/.changeset/fix-join-on-demand-sync.md
+++ b/.changeset/fix-join-on-demand-sync.md
@@ -1,5 +1,5 @@
 ---
-"@tanstack/db": patch
+'@tanstack/db': patch
 ---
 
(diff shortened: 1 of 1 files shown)


ibarakaiev:
Closing as it appears to be more involved than a one-line fix, opening an issue instead #1077.

PR #1081, opened 2025-12-30
fix(db): prevent on-demand live query from being marked ready before data loads

KyleAMathews:
## Summary

Live queries in on-demand sync mode would return `isReady=true` with empty data, or release suspense prematurely. Users see a flash of empty state before data appears.

## Root Cause

Race condition in `CollectionSubscriber`: the `status:change` listener was registered *after* the snapshot was triggered. If `loadSubset` resolved quickly, the `loadingSubset` → `ready` transition was missed, so `trackLoadPromise` was never called on the live query collection.

## Approach

**1. Deferred snapshot triggering** (core fix)
```
1. Create subscription with deferSnapshot: true
2. Register status listener  
3. THEN trigger snapshot
```

**2. Ready state gating** (defense in depth)
```typescript
if (this.allCollectionsReady() && !this.liveQueryCollection?.isLoadingSubset) {
  markReady()
}
```

### Key invariant
The status listener must be registered before any async work starts. This is now enforced by the `deferSnapshot` option.

### Non-goals
- No changes to `useLiveSuspenseQuery` hook. The fix at the collection level is sufficient—suspense naturally holds because `status !== 'ready'` while loading.

### Trade-offs
- Adds `deferSnapshot` to public API surface, but it's an internal option not meant for external use
- Slightly more complex subscription flow, but eliminates a class of timing bugs

## Verification

```bash
pnpm --filter "@tanstack/db" --filter "@tanstack/react-db" test
```

New test: `should not mark live query ready while isLoadingSubset is true`

## Files changed
- `packages/db/src/collection/changes.ts` - Add `deferSnapshot` option
- `packages/db/src/query/live/collection-config-builder.ts` - Ready gating + `loadingSubset:change` listener  
- `packages/db/src/query/live/collection-subscriber.ts` - Defer snapshot triggering
- `packages/db/src/types.ts` - Add `deferSnapshot` type

### DIFF
--- a/packages/db/src/query/live/collection-subscriber.ts
+++ b/packages/db/src/query/live/collection-subscriber.ts
@@ -5,7 +5,10 @@ import {
 } from '../compiler/expressions.js'
 import type { MultiSetArray, RootStreamBuilder } from '@tanstack/db-ivm'
 import type { Collection } from '../../collection/index.js'
-import type { ChangeMessage, SubscriptionStatusChangeEvent } from '../../types.js'
+import type {
+  ChangeMessage,
+  SubscriptionStatusChangeEvent,
+} from '../../types.js'
 import type { Context, GetResult } from '../builder/types.js'
 import type { BasicExpression } from '../ir.js'
 import type { OrderByOptimizationInfo } from '../compiler/order-by.js'


marbemac:
Doesn't seem to resolve the issue unfortunately.

Also, I'm not sure this is specific to suspense? For example if I have the below with regular useLiveQuery, sometimes `isReady=true` without any data, and sometimes after there is data.

```ts
const { data, isReady, state, } = useLiveQuery(
  q =>
    q
      .from({ w: workflowRunsCollection })
      .where(({ w }) => eq(w.id, 'some_id'))
      .findOne(),
  [],
);

console.log('run', { haveData: !!data, state, isReady });
```

I've stripped our app down to this one query, and reload the page, and see similar symptom to the suspense version (albeit using isReady as the boundary versus suspense):

Screenshot 2025-12-30 at 5 48 14 PM

marbemac:
Out of curiosity, I tried the latest from here and unfortunately the problem persists. So I had Claude add a bunch of debug logs to spots that might be relevant, then reloaded our app, and fed the logs back into claude for diagnosis. Here are the resulting findings from Claude in case helpful (including the debug logs from the actual reproduction):

---

# Debug Analysis: `isReady=true` before data is loaded

## Issue Summary

In on-demand sync mode with Electric, `useLiveQuery` returns `isReady=true` before the query data has been written to the live query collection. This causes:
- `useLiveQuery` to return `isReady=true` with empty `data`
- `useLiveSuspenseQuery` to release suspense prematurely

## Reproduction

```tsx
function Debug() {
  const { data, isReady, state } = useLiveQuery(
    q =>
      q
        .from({ w: workflowRunsCollection })
        .where(({ w }) => eq(w.id, 'wrun_eco6todnlmxfvg'))
        .findOne(),
    [],
  );

  console.log('run', { haveData: !!data, state, isReady });

  return <div>Debug</div>;
}
```

**Console output on page load:**
```
run {haveData: false, state: Map(0), isReady: false}
run {haveData: false, state: Map(0), isReady: true}   ← BUG: isReady=true but no data
run {haveData: true, state: Map(1), isReady: true}
```

## Debug Logs

Added console.logs to trace the flow. Full output:

```
[DEBUG allCollectionsReady] id=live-query-1
  statuses: [{id: 'workflowRuns', status: 'loading', isReady: false}]

[DEBUG updateLiveQueryStatus] id=live-query-1
  {allCollectionsReady: false, isLoadingSubset: false, currentStatus: 'loading', willMarkReady: false}

[DEBUG trackLoadPromise] id=workflowRuns
  {loadingStarting: true, pendingCount: 1}

[DEBUG CollectionSubscriber initial status check] alias=w
  {subscriptionStatus: 'ready'}                        ← Already 'ready'!

[DEBUG allCollectionsReady] id=live-query-1
  statuses: [{id: 'workflowRuns', status: 'loading', isReady: false}]

[DEBUG updateLiveQueryStatus] id=live-query-1
  {allCollectionsReady: false, isLoadingSubset: false, currentStatus: 'loading', willMarkReady: false}

run {haveData: false, state: Map(0), isReady: false}   ← Correct initial state

[DEBUG markReady] id=workflowRuns
  {currentStatus: 'loading'}                           ← Source collection becomes ready

[DEBUG allCollectionsReady] id=live-query-1
  statuses: [{id: 'workflowRuns', status: 'ready', isReady: true}]

[DEBUG updateLiveQueryStatus] id=live-query-1
  {allCollectionsReady: true, isLoadingSubset: false, currentStatus: 'loading', willMarkReady: true}
                                                       ↑ isLoadingSubset=false, so it marks ready!

[DEBUG markReady] id=live-query-1
  {currentStatus: 'loading'}                           ← Live query marked ready prematurely

run {haveData: false, state: Map(0), isReady: true}    ← BUG! isReady=true but no data

[DEBUG markReady] id=workflowRuns
  {currentStatus: 'ready'}

run {haveData: true, state: Map(1), isReady: true}     ← Data arrives later

[DEBUG trackLoadPromise.finally] id=workflowRuns
  {loadingEnding: true, pendingCount: 0}               ← Loading finishes after ready was set
```

## Root Cause Analysis

The fix in this branch added a check for `isLoadingSubset` on the live query collection:

```ts
// collection-config-builder.ts
if (this.allCollectionsReady() && !this.liveQueryCollection?.isLoadingSubset) {
  markReady()
}
```

However, `isLoadingSubset` is always `false` on the live query collection because:

1. **`trackLoadPromise` is called on the source collection (`workflowRuns`), not the live query collection (`live-query-1`)**

2. **The Electric subscription status is already `'ready'`** when `CollectionSubscriber` checks it at line 120-124:
   ```ts
   // collection-subscriber.ts
   if (subscription.status === `loadingSubset`) {
     trackLoadPromise()  // Never called because status is 'ready'
   }
   ```

3. Therefore, `trackLoadPromise()` is never called on the live query collection

4. The live query's `isLoadingSubset` remains `false`

5. When the source collection becomes ready, the condition `allCollectionsReady() && !isLoadingSubset` is satisfied, and the live query is marked ready before its data is populated

## Timeline

```
1. trackLoadPromise() called on workflowRuns (source) - starts loading
2. Electric subscription.status = 'ready' immediately
3. CollectionSubscriber checks status, sees 'ready', doesn't track load on live query
4. workflowRuns.markReady() - source collection ready
5. updateLiveQueryStatus() checks:
   - allCollectionsReady() = true (workflowRuns is ready)
   - isLoadingSubset = false (nothing tracked on live query)
   → Marks live-query-1 ready!
6. React renders: isReady=true, data=empty  ← BUG
7. Data is written to live query collection
8. React renders: isReady=true, data=populated
9. trackLoadPromise.finally() on workflowRuns - loading done
```

## Potential Fix Directions

1. **Track loading on live query collection**: When the source collection starts loading subset data, also track that loading on the live query collection that depends on it

2. **Investigate Electric subscription status**: Why is `subscription.status` already `'ready'` when data hasn't been loaded yet? This might be a bug in the Electric integration layer or Electric itself

3. **Use source collection's isLoadingSubset**: Instead of (or in addition to) checking the live query's `isLoadingSubset`, check if any source collection has `isLoadingSubset=true`

## Files Involved

- `packages/db/src/query/live/collection-config-builder.ts` - `updateLiveQueryStatus()` and `allCollectionsReady()`
- `packages/db/src/query/live/collection-subscriber.ts` - Electric subscription status tracking
- `packages/db/src/collection/sync.ts` - `trackLoadPromise()` and `isLoadingSubset`
- `packages/db/src/collection/lifecycle.ts` - `markReady()`

marbemac:
That changes since my last comment fixed it! There are several adjustments in this PR, unsure which one(s) are the actual fixes, but something in the last couple of commits finally resolved it. Might want to figure out a test that fails before the last couple of commits, and succeeds with the changes from the last couple of commits.

kevin-dp:
@KyleAMathews I find the new flow a bit cumbersome and error-prone since the caller has to remember to defer the snapshot, register the listener, then trigger the snapshot. I'd prefer if we replace the `deferSnapshot` option by a `onStatusChange` option and let `subscribeChanges` do the magic.

I asked Cursor what it thinks about this and it agreed and came up with this more detailed plan:

Analyzing the PR and evaluating the alternative. Searching for `deferSnapshot` and reviewing the relevant code.

## Proposed Solution Analysis

Passing the status listener as an option to `subscribeChanges` is better because:

1. Atomic registration: the listener is registered before any snapshot is requested, avoiding missed events.
2. Simpler API: no need for `deferSnapshot` and manual snapshot triggering.
3. Type safety: the listener can be part of the options interface.
4. Less error-prone: no manual sequencing of create → register → trigger.

## Recommended Implementation

Here's how the API could look:

```typescript
export interface SubscribeChangesOptions {
  /** Whether to include the current state as initial changes */
  includeInitialState?: boolean
  /** Pre-compiled expression for filtering changes */
  whereExpression?: BasicExpression<boolean>
  /** Listener for subscription status changes - registered before snapshot is requested */
  onStatusChange?: (event: SubscriptionStatusChangeEvent) => void
}
```

Then in `subscribeChanges`:

```typescript
public subscribeChanges(
  callback: (changes: Array<ChangeMessage<TOutput>>) => void,
  options: SubscribeChangesOptions = {}
): CollectionSubscription {
  this.addSubscriber()

  const subscription = new CollectionSubscription(this.collection, callback, {
    ...options,
    onUnsubscribe: () => {
      this.removeSubscriber()
      this.changeSubscriptions.delete(subscription)
    },
  })

  // Register status listener BEFORE requesting snapshot to avoid race condition
  if (options.onStatusChange) {
    subscription.on(`status:change`, options.onStatusChange)
  }

  if (options.includeInitialState) {
    subscription.requestSnapshot({ trackLoadSubsetPromise: false })
  }

  this.changeSubscriptions.add(subscription)
  return subscription
}
```

This ensures:
- The listener is registered before any status changes occur
- No need for `deferSnapshot`
- Cleaner, more intuitive API
- Backward compatible (the option is optional)

This approach is better than introducing `deferSnapshot` because it solves the root cause (listener registration timing) rather than requiring manual coordination.

KyleAMathews:
@kevin-dp implemented

kevin-dp (approved):

serco-chen:
@KyleAMathews This PR does **not** fix the issue where `isReady` is still marked as `true` prematurely in certain cases.

For example:

When the `syncMode` of `usersCollection` is set to **on-demand**:

```ts
const { data: user, status, isLoading } = useLiveQuery((q) =>
  q
    .from({ users: usersCollection })
    .where(({ users }) => eq(users.id, userId))
    .findOne()
, [userId])
```

`status` is `isReady` from the very beginning, and `isLoading` is `false`.

`findOne()` only sets `singleResult` and does not apply a `limit` (`packages/db/src/query/builder/index.ts:685`), so no `OrderByOptimizationInfo` is generated (`packages/db/src/query/compiler/order-by.ts:76`). As a result, `subscribeToOrderedChanges` is not used.

PR #1082, opened 2025-12-30
Fix QueryCompilationError with undefined expression type

KyleAMathews:
…perator usage

When users accidentally use JavaScript's comparison operators (===, !==, <, >, etc.) in where() callbacks instead of query builder functions (eq, gt, etc.), the query builder now throws a helpful InvalidWhereExpressionError with clear guidance.

Previously, this mistake would result in a confusing "Unknown expression type: undefined" error at query compilation time. Now users get immediate feedback with an example of the correct syntax.

Example of the mistake this catches:
  ❌ .where(({ user }) => user.id === 'abc')
  ✅ .where(({ user }) => eq(user.id, 'abc'))

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/db/src/query/builder/index.ts
+++ b/packages/db/src/query/builder/index.ts
@@ -366,15 +366,7 @@ export class BaseQueryBuilder<TContext extends Context = Context> {
     // This catches common mistakes like using JavaScript comparison operators (===, !==, etc.)
     // which return boolean primitives instead of expression objects
     if (!isExpressionLike(expression)) {
-      const valueType =
-        expression === null
-          ? `null`
-          : expression === undefined
-            ? `undefined`
-            : typeof expression === `object`
-              ? `object`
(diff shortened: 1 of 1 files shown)


packages/db/src/query/builder/index.ts
```diff
@@ -402,6 +418,21 @@ export class BaseQueryBuilder<TContext extends Context = Context> {
     const refProxy = createRefProxy(aliases) as RefsForContext<TContext>
     const expression = callback(refProxy)
 
+    // Validate that the callback returned a valid expression
```
kevin-dp, line 413: This code is an exact copy of L365-378. Let's extract this to a helper function that we can then use in both places. Probably we want to turn the valueType constant into a function.

kevin-dp:
:shipit:

PR #1089, opened 2026-01-05
merge: Claude/remove redundant unit tests

MentalGear:
## ✅ Checklist

- [ ] I have followed the steps in the [Contributing guide](https://github.com/TanStack/db/blob/main/CONTRIBUTING.md).
- [ ] I have tested this code locally with `pnpm test:pr`.

## 🚀 Release Impact

- [ ] This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).
- [ ] This change is docs/CI/dev-only (no release).

PR #1090, opened 2026-01-05
feat: trailbase syncMode support & e2e test suite

MentalGear:
## 🎯 Changes

trailbase-db-collection

- feat: **syncMode** support
- e2e tests integration: using official **packages/db-collection-e2e** (inspired by electric)

## Test

`cd db/packages/trailbase-db-collection && pnpm test:e2e`

## Current Status

** Of the official 118 tests test suite: **

- 114/118 tests pass ✅ 

However, of those **4 failing tests**:

**2 tests works individually:**
- should load subsequent pages correctly with multi-column orderBy
- should handle multi-column orderBy with mixed directions

**1 is flaky (sometimes works)**
- should handle insert appearing in matching queries

**always fails**
- should use setWindow with mixed direction multi-column orderBy

---

## Request for Help

I'm a bit at my wit's end here.

It might be an orderBy issue on how trailbase internally server-side does it's ordering vs the tanstack/db internal client, or it might be because of how trailbase's sync (instead of electric streaming) is a bit slower and the cleanUp/start takes longer.

In any case, I would appreciate someone more knowledgable having a look at the failing tests so we can finally bring the essential syncMode (partial replication) to Trailbase Collections as well!

-------

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/trailbase-db-collection/e2e/global-setup.ts
+++ b/packages/trailbase-db-collection/e2e/global-setup.ts
@@ -1,8 +1,8 @@
-import {  execSync, spawn } from 'node:child_process'
+import { execSync, spawn } from 'node:child_process'
 import { cpSync, existsSync, mkdirSync, rmSync } from 'node:fs'
 import { dirname, resolve as resolvePath } from 'node:path'
 import { fileURLToPath } from 'node:url'
-import type {ChildProcess} from 'node:child_process';
+import type { ChildProcess } from 'node:child_process'
 import type { GlobalSetupContext } from 'vitest/node'
 
 const __dirname = dirname(fileURLToPath(import.meta.url))
@@ -99,7 +99,14 @@ function startBinaryServer(): ChildProcess {
 
   const proc = spawn(
     binaryPath,
-    ['--data-dir', tempDataDir, 'run', '--address', `0.0.0.0:${TRAILBASE_PORT}`, '--dev'],
+    [
+      '--data-dir',
+      tempDataDir,
+      'run',
+      '--address',
+      `0.0.0.0:${TRAILBASE_PORT}`,
+      '--dev',
+    ],
     {
       stdio: ['ignore', 'pipe', 'pipe'],
       cwd: PACKAGE_DIR,
--- a/packages/trailbase-db-collection/docker/docker-compose.yml
+++ b/packages/trailbase-db-collection/docker/docker-compose.yml
@@ -6,9 +6,9 @@ services:
       context: .
       dockerfile: Dockerfile
     ports:
-      - "4000:4000"
+      - '4000:4000'
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:4000/api/healthz"]
+      test: ['CMD', 'curl', '-f', 'http://localhost:4000/api/healthz']
       interval: 5s
       timeout: 5s
       retries: 5
--- a/packages/trailbase-db-collection/e2e/trailbase.e2e.test.ts
+++ b/packages/trailbase-db-collection/e2e/trailbase.e2e.test.ts
@@ -36,7 +36,6 @@ declare module 'vitest' {
   }
 }
 
-
 /**
  * Decode base64-encoded BLOB UUID to standard UUID string format
  * TrailBase stores UUIDs as BLOBs and returns them as base64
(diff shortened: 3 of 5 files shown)


samwillis:
@ignatz would be hand if you could take a look at this too.

ignatz:
(copy & paste from #1089 for legibility)

@samwillis Ack

@MentalGear could you clarify the relationship between https://github.com/TanStack/db/pull/1090 and https://github.com/TanStack/db/pull/1089, I looked at the history:
Screenshot From 2026-01-07 09-59-14

And it seems like they are close siblings with both removing "redundant unit tests", maybe you didn't want to open two PRs? For now I will ignore https://github.com/TanStack/db/pull/1089 and comment on https://github.com/TanStack/db/pull/1090 only.

ignatz:
@MentalGear firstly, I love the addition of e2e tests, electric has them and it's just a great insurance to have :heart: 

Now, with you having failing and flaky tests, I can have a look but what would really help (w/o wanting to be a pain in your side) is to split up:

* The addition of e2e tests
* The addition of new/configurable sync modes

While I will continue to have a look, could we just add the new test harness first ensuring that what we already have is WAI and then stack on new functionality? :pray:

packages/trailbase-db-collection/e2e/global-setup.ts
```diff
@@ -0,0 +1,342 @@
...
+          `Please either:\n` +
+          `  1. Start TrailBase manually at ${TRAILBASE_URL}\n` +
+          `  2. Place the TrailBase binary at one of:\n` +
+          BINARY_PATHS.map((p) => `     - ${p}`).join('\n') +
+          `\n  3. Install Docker and run the tests again\n` +
+          `\nTo download TrailBase binary:\n` +
+          `  curl -sSL https://trailbase.io/install.sh | bash\n` +
+          `\nTo start TrailBase with Docker manually:\n` +
+          `  cd packages/trailbase-db-collection/docker\n` +
+          `  docker-compose up -d`,
```
ignatz, line 313: Recommending to use docker compose after docker isn't available (check just above), may be a bit misleading.

packages/trailbase-db-collection/e2e/global-setup.ts
```diff
@@ -0,0 +1,342 @@
...
+    console.log(`✓ TrailBase already running at ${TRAILBASE_URL}`)
+  } else {
+    // Try binary first (preferred for CI/local testing without Docker)
+    binaryPath = findBinaryPath()
+    if (binaryPath) {
+      console.log(`✓ TrailBase binary found at ${binaryPath}`)
+      serverProcess = startBinaryServer()
+      startedServer = true
+      usedMethod = 'binary'
+    } else if (isDockerAvailable()) {
```
ignatz, line 291: Why would we ever use anything but docker? If we want to test against a specific version we can achieve that with docker two, rather than checking in a binary. It will also be more portable (i.e. not just linux).

packages/trailbase-db-collection/e2e/global-setup.ts
```diff
@@ -0,0 +1,342 @@
...
+const PACKAGE_DIR = resolvePath(__dirname, '..')
+const REPO_ROOT = resolvePath(PACKAGE_DIR, '../..')
+const DOCKER_DIR = resolvePath(PACKAGE_DIR, 'docker')
+// Check multiple possible binary locations
+const BINARY_PATHS = [
+  resolvePath(PACKAGE_DIR, 'testing-bin-linux', 'trail'),
+  resolvePath(REPO_ROOT, 'packages/trailbase/test-linux-bin', 'trail'),
+]
+const CONTAINER_NAME = 'trailbase-e2e-test'
+const TRAILBASE_PORT = process.env.TRAILBASE_PORT ?? '4000'
```
ignatz, line 18: We should probably use a non-default port to avoid harder to debug issues, where a TB instance is running but not with the required configuration/schema.

MentalGear:
My apologies: You’re right that this should have been split into a separate E2E suite PR, followed by the syncMode addition. I added syncMode first and then added the full E2E coverage to properly validate the functionality.

Regarding the Docker/binary fallback: part of the development was done on a remote VM without Docker access, which necessitated that approach.

Thank you for taking the time to review this.

ignatz:
@MentalGear Just FYI, also have a pending change to add pagination support, which should fix some of the errors you observed: https://github.com/TanStack/db/pull/1101

Issue #1078, opened 2025-12-30
Unable to use aggregate fields from SELECT in ORDER BY


K-Mistele:
- [x] I've validated the bug against the latest version of DB packages

**Describe the bug**
When I am using `useLiveQuery` with aggregate functions in the `.select(({...}) => ({count: count(...), max: max(...)})` part of a query, I am unable to use these fields in a `.orderBy` or `.having` clause

**To Reproduce**

Perform a query involving an aggregate function with a `groupBy` or a `having`  or an `orderBy` clause, and observe that it is only possible to use columns on the table(s) referenced in `.from()`, joins and sub-queries - not columns defined in `.select()` and/or aggregates. 

```typescript
const sessionStats = q
    .from({ sessions: sessionsCollection })
    .where(({ sessions }) => eq(sessions.archived, false))
    .groupBy(({ sessions }) => sessions.task_id)
    .select(({ sessions}) => ({
    	taskId: sessions.task_id,
    	latestActivity: max(sessions.last_activity_at),
    	sessionCount: count(sessions.id),
    }))
    .orderBy(({sessions}) => /* Problem here */)

```

Observe that while it is valid to order by `sessions.<column>` based on the types of the `orderBy`, it is NOT possible to order by the fields of `latestActivity` or `sessionCount` defined by the aggregate functions of `max()` and `count()`, or by other fields defined in the `select` query - the types indicate it is only possible to filter based on the properties of the rows of the table(s)/joins/subqueries used in the query.

Similarly, it is impossible to filter with `.having()` based on `.select()`-defined fields and aggregates - it is only possible to use columns of the table(s) used in the `.from()` clause, subqueries and joins. 

**Expected behavior**
I should be able to use `.orderBy()` and `.having()` based on the fields and aggregate function-defined fields from the `.select()` clause

**Screenshots**
n/a

**Desktop (please complete the following information):**

- OS: MacOS
- Browser: n/a
- Version: n/a

**Smartphone (please complete the following information):**
n/a

**Additional context**
n/a

thruflo:
I've come across the same limitation when trying to orderBy aggregated values from the select. One workaround is to build a first query with a select and then pass that as the `from: ...` value to a subsequent query.

This works and avoids materialization of the interim results.

E.g.:

```tsx
const messageCollection = createLiveQueryCollection({
  query: (q) => {
    const baseQuery = q
      .from({ chunk: chunksCollection })
      .groupBy(({ chunk }) => chunk.messageId)
[...]

PR #1094, opened 2026-01-06
Support references to aggregated fields from SELECT in ORDERBY and HAVING

kevin-dp:
Fixes https://github.com/TanStack/db/issues/1078

This PR introduces a new `$selected` namespace that is present on `orderBy`, `having`, `fn.having` if the query contains a `select` clause.

-----------

# Bug Analysis: Unable to use aggregate fields from SELECT in ORDER BY

## Problem Summary

Users cannot reference SELECT fields (aggregate or non-aggregate) in ORDER BY or HAVING clauses. TypeScript types don't expose SELECT fields in the callback context, and the runtime doesn't transform references to SELECT aliases to use `__select_results`.

### Execution Order is Correct

The query execution pipeline runs SELECT before ORDER BY:

```typescript
// From compileQuery in index.ts:
// 1. Lines 219-255: SELECT is processed early and creates __select_results
// 2. Lines 257-281: GROUP BY is processed (updates __select_results with aggregates)
// 3. Lines 316-328: ORDER BY is processed AFTER GROUP BY
```

So `__select_results` is available when ORDER BY runs. Execution order is not the issue.

### The Real Problem: `replaceAggregatesByRefs` Function

The issue is in `replaceAggregatesByRefs` in `group-by.ts`. It only handles aggregate expressions (`agg` type), not regular `ref` expressions that match SELECT aliases.

**Current Implementation:**
```typescript
case `ref`: {
  // Non-aggregate refs are passed through unchanged (they reference table columns)
  return havingExpr as BasicExpression
}
```

When a `ref` in ORDER BY/HAVING matches a SELECT alias (e.g., `taskId`, `latestActivity`, `sessionCount`), it should be transformed to reference `__select_results[alias]` instead of a table column.

### What's Missing

When processing ORDER BY/HAVING expressions:
1. Aggregate expressions are handled — `replaceAggregatesByRefs` finds matching aggregates in SELECT and transforms them.
2. SELECT field references are not handled — `ref` expressions that match SELECT aliases are passed through unchanged, so they reference table columns that don't exist.

## Solution

Extend `replaceAggregatesByRefs` to also check if a `ref` matches a SELECT alias and transform it accordingly:

1. Extract the field path from the ref (removing the table alias if present).
2. Check if the field path matches any SELECT alias (supporting simple and nested aliases like `'taskId'` or `'meta.author'`).
3. If it matches: transform to `PropRef([resultAlias, ...aliasParts])` to reference `__select_results[alias]`.
4. If it doesn't match: pass through unchanged (treating it as a table column reference).

### Example Transformations

**Before (broken):**
```typescript
// Query
.select(({ sessions }) => ({
  taskId: sessions.customer_id,
  latestActivity: max(sessions.date),
  sessionCount: count(sessions.id),
}))
.orderBy(({ sessions }) => sessions.latestActivity)
// Creates: PropRef(['sessions', 'latestActivity'])
// Tries to access: row.sessions.latestActivity (doesn't exist!)
```

**After (fixed):**
```typescript
// Same query
.orderBy(({ sessions }) => sessions.latestActivity)
// Transforms to: PropRef(['__select_results', 'latestActivity'])
// Accesses: row.__select_results.latestActivity ✓
```

### How Nested Aliases Work

Nested SELECT aliases like `meta.author.name` are created in `addFromObject` in `select.ts`:
- When processing nested objects, it recursively builds a `prefixPath` array
- The final alias is created by joining: `[...prefixPath, key].join('.')`
- Example: `['meta', 'author', 'name']` → `'meta.author.name'`

The fix correctly:
- Splits aliases by dots: `alias.split('.')` → `['meta', 'author', 'name']`
- Compares against the field path from the ref expression
- Reconstructs the path: `PropRef([resultAlias, ...aliasParts])`

### Context-Specific Behavior

- ORDER BY: Uses `resultAlias = '__select_results'`, so transformed refs access `row.__select_results[alias]`
- HAVING: Uses `resultAlias = 'result'`, and the context is `{ result: row.__select_results }`, so transformed refs access `row.result[alias]`

## Testing

Unit tests were added to reproduce the bug:
- `BUG: orderBy cannot reference aggregate field from SELECT`
- `BUG: orderBy cannot reference non-aggregate field from SELECT`
- `BUG: having cannot reference aggregate field from SELECT`
- `BUG: having cannot reference non-aggregate field from SELECT`

These tests document the expected behavior and will pass once the fix is implemented.

### DIFF
--- a/.changeset/weak-colts-walk.md
+++ b/.changeset/weak-colts-walk.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/db': patch
+---
+
+Introduce $selected namespace for accessing fields from SELECT clause inside ORDER BY and HAVING clauses.


K-Mistele:
Thank you! Was waiting for this one :)

PR #1096, opened 2026-01-06
Add helper to extract the result type from a query `QueryResult`

KyleAMathews:
This allows users to extract the result type from a Query instance:

```typescript
import { Query, QueryResult } from '@tanstack/db'

const myQuery = new Query()
  .from({ users })
  .select(({ users }) => ({ name: users.name }))

// Clean and simple!
type MyQueryResult = QueryResult<typeof myQuery>
```

Fixes a gap in the public API where ExtractContext was defined but not re-exported from the main query module.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/.changeset/export-extract-context.md
+++ b/.changeset/export-extract-context.md
@@ -2,15 +2,17 @@
 '@tanstack/db': patch
 ---
 
-Export `ExtractContext` type to enable extracting query result types from Query instances.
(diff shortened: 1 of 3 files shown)


kevin-dp (approved):
I'll approve but i will advocate to expose a helper type that does both in one: `QueryResult<typeof myQuery>` similar to Zod's `z.infer<typeof mySchema>`

KyleAMathews:
Yeah good idea

PR #1099, opened 2026-01-07
Refactor useLiveQuery documentation for clarity

GwydionBr:
The documentation sections were repeating itself 8 times. I removed the 7 repetitions to show the sections only once.


KyleAMathews:
This file is autogenerated from the src so we don't edit it directly. I'm not sure why it'd generate duplicates but please feel free to investigate that!

PR #1114, opened 2026-01-09
fix(db): batch D2 output callbacks to prevent duplicate key errors in joins

KyleAMathews:
## Summary

Fixes `isReady()` returning `true` while `toArray()` returns `[]`. Also fixes a latent bug that was exposed by the timing fix.

---

## The Bug Report

> I have a case where `collection.toArray` is empty, but `collection.isReady()` is already true. Just to confirm, isReady should only be true once the query/collection got all query results for the first time, right?

```typescript
  q.from({ room: roomCollection }).where(({ room }) => eq(room.id, roomId))
);

await collection.toArrayWhenReady();
// isReady() === true, but toArray() === [] ← BUG
```

## Root Cause

`markReady()` was being called in `subscribeToAllCollections()` *before* the D2 graph had a chance to run and process data:

```typescript
// In subscribeToAllCollections():
syncState.subscribedToAllCollections = true
this.updateLiveQueryStatus(config)  // ← Called here, before graph.run()!
```

The graph only runs in `maybeRunGraph()`, which happens after subscriptions are set up. So `isReady` became `true` while the data pipeline was still empty.

## The Fix

Add a guard to `updateLiveQueryStatus()` so it only marks ready after subscriptions are set up AND the graph has had a chance to run:

```typescript
if (
  this.currentSyncState?.subscribedToAllCollections &&  // ← NEW: must have subscriptions
  this.allCollectionsReady() &&
  !this.liveQueryCollection?.isLoadingSubset
) {
  markReady()
}
```

And remove the premature call from `subscribeToAllCollections()`. The canonical place to mark ready is now after `graph.run()` in `maybeRunGraph()`.

---

## The Journey: Exposing a Latent Bug

When I made the timing fix, a test started failing: "should allow custom getKey with joins (1:1 relationships)"

```
Error: Cannot insert document with key '1' from sync because it already exists
```

**Why did this test "pass" before?**

The old code called `markReady()` before the graph ran, which resolved `preload()` early. The duplicate key error was thrown *after* the promise resolved, so it was swallowed. The test appeared to pass, but was actually broken.

**What was the actual bug?**

D2's incremental join produces multiple outputs during a single `graph.run()`:
1. First: partial join `[key=1, undefined]` with multiplicity +1
2. Second: full join `[key=1, data]` with multiplicity +1, AND delete of partial with multiplicity -1

Previously, each output committed immediately. The first commit added key "1" to syncedData, and the second output's insert failed because the key already existed.

**The fix:** Batch all outputs within a single `graph.run()` into one transaction. Changes for the same key are accumulated (+1, +1, -1 = +1) before committing.

---

## Key Invariants

1. **`markReady()` only after graph processes data** - the `subscribedToAllCollections` check ensures we don't mark ready before the graph runs
2. **All outputs within a `graph.run()` share one transaction** - prevents intermediate join states from causing duplicate key errors

## Verification

```bash
pnpm vitest run packages/db/tests/query/live-query-collection.test.ts
# All 46 tests pass

pnpm vitest run packages/db/tests
# 1716 tests pass
```

## Files Changed

| File | Change |
|------|--------|
| `collection-config-builder.ts` | Add `subscribedToAllCollections` guard to `updateLiveQueryStatus()`, remove premature call from `subscribeToAllCollections()`, batch outputs in `pendingChanges` Map |
| `types.ts` | Add `flushPendingChanges` to `SyncState` type |
| `sync.ts` | Minor refactor: extract `valuesEqual` variable |
| `live-query-collection.test.ts` | Fix test timing to properly await async operations |

### DIFF
--- a/packages/db/src/query/live/collection-config-builder.ts
+++ b/packages/db/src/query/live/collection-config-builder.ts
@@ -708,28 +708,7 @@ export class CollectionConfigBuilder<
 
         // Accumulate changes from this output callback into the pending changes map.
         // Changes for the same key are merged (inserts/deletes are added together).
-        for (const [[key, tupleData], multiplicity] of messages as Array<
-          [[unknown, [TResult, string | undefined]], number]
-        >) {
-          const [value, orderByIndex] = tupleData
-          const existing = pendingChanges.get(key) || {
-            deletes: 0,
-            inserts: 0,
-            value,
-            orderByIndex,
-          }
-          if (multiplicity < 0) {
-            existing.deletes += Math.abs(multiplicity)
-          } else if (multiplicity > 0) {
-            existing.inserts += multiplicity
-            // Update value to the latest version for this key
-            existing.value = value
-            if (orderByIndex !== undefined) {
-              existing.orderByIndex = orderByIndex
-            }
-          }
-          pendingChanges.set(key, existing)
-        }
+        messages.reduce(accumulateChanges<TResult>, pendingChanges)
       }),
     )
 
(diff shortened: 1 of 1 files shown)


packages/db/src/query/live/collection-config-builder.ts
```diff
@@ -687,22 +695,56 @@ export class CollectionConfigBuilder<
...
+
     pipeline.pipe(
       output((data) => {
         const messages = data.getInner()
         syncState.messagesCount += messages.length
 
-        begin()
-        messages
-          .reduce(
-            accumulateChanges<TResult>,
```
kevin-dp, line 698: Why are we removing the nice `accumulateChanges` helper function and inlining it?
I'd rather keep the helper function.

kevin-dp: @claude Address this comment

PR #1124, opened 2026-01-13
Fix offline mutations data loss on restart

KyleAMathews:
## Summary

Fix `mutation.changes` being lost when offline transactions are persisted and restored. After app restart, sync functions that rely on `mutation.changes` for partial updates would receive an empty object `{}`, causing server writes to fail with missing data.

## Root Cause

The `changes` field was never added to the serialization flow:

1. `SerializedMutation` interface was missing the `changes` field
2. `serializeMutation()` didn't serialize `changes`
3. `deserializeMutation()` hardcoded `changes: {}` with a misleading comment "Will be recalculated"

The comment was wrong—`changes` was never recalculated anywhere. This meant any mutation function using the common pattern `api.update(id, mutation.changes)` would send empty updates after restart.

## Approach

Add `changes` to the serialization round-trip, mirroring how `modified` and `original` are already handled:

```typescript
// Serialize
changes: this.serializeValue(mutation.changes),

// Deserialize (with legacy data fallback)
changes: data.changes !== undefined 
  ? this.deserializeValue(data.changes) 
  : {},
```

The legacy fallback handles data serialized before this fix—without it, `deserializeValue(undefined)` would return `undefined`, causing `TypeError` when downstream code accesses properties on `changes`.

## Key Invariants

- `mutation.changes` must survive the serialize → persist → deserialize cycle unchanged
- Legacy data (missing `changes` field) must deserialize without errors

## Non-goals

- Not adding explicit version migration—the fallback to `{}` matches the previous (broken) behavior, so existing data continues to work as before
- Not refactoring the broader silent-failure patterns in `OutboxManager`—that's pre-existing and out of scope

## Verification

```bash
cd packages/offline-transactions && pnpm test
```

All 26 tests pass. The existing "replays persisted transactions on startup" e2e test exercises this code path.

## Files Changed

| File | Change |
|------|--------|
| `types.ts` | Add `changes: any` to `SerializedMutation` interface |
| `TransactionSerializer.ts` | Serialize/deserialize `changes` field with legacy fallback |
| `TransactionSerializer.ts` | Code quality: safer `hasOwnProperty`, unused param prefixes |


kevin-dp:
> The legacy fallback handles data serialized before this fix—without it, deserializeValue(undefined) would return undefined, causing TypeError when downstream code accesses properties on changes.

Based on this explanation we can simplify this code:
```ts
changes: data.changes !== undefined 
  ? this.deserializeValue(data.changes) 
  : {},
```
to this:
```ts
changes: this.deserializeValue(data.changes) ?? {},
```

packages/offline-transactions/src/outbox/TransactionSerializer.ts
```diff
@@ -118,7 +118,7 @@ export class TransactionSerializer {
     if (typeof value === `object`) {
       const result: any = Array.isArray(value) ? [] : {}
       for (const key in value) {
-        if (value.hasOwnProperty(key)) {
+        if (Object.prototype.hasOwnProperty.call(value, key)) {
```
kevin-dp, line 112: Is this a hack such that it works if `value` is `null`? Wouldn't it be more readable if we did `if (value && value.hasOwnProperty(key))` ?

Same comment for line 143.

PR #1126, opened 2026-01-13
Fix subqueries app freezing issue

KyleAMathews:
## Summary

Fixes app freezing and browser lockups caused by 409 must-refetch response loops. When subqueries trigger rapid 409 responses, this creates a cascade that blocks the main thread and can freeze the entire application.

## Root Cause

Two bug reports described similar symptoms: app freezing, infinite `useLiveQuery` re-renders, and `awaitTxId` appearing to block forever. The root cause is a 409 must-refetch cascade:

```
409 must-refetch → truncate() → handleTruncate() → loadSubset()
                                                        ↓
                                    if loadSubset triggers 409 → REPEAT
```

Each cycle fires truncate events, triggers status changes, causes React re-renders, and with large datasets the main thread becomes completely blocked. The `awaitTxId` doesn't actually block—it's async—but the event loop starvation makes the timeout ineffective.

### 1. Cycle Detection in Subscription (packages/db/src/collection/subscription.ts)

Added truncate cycle detection to break infinite 409 loops:

```typescript
private lastTruncateTime = 0
private truncateCycleCount = 0
private static readonly TRUNCATE_CYCLE_WINDOW_MS = 2000
private static readonly MAX_TRUNCATES_PER_WINDOW = 5

private handleTruncate() {
  const now = Date.now()
  if (now - this.lastTruncateTime < TRUNCATE_CYCLE_WINDOW_MS) {
    this.truncateCycleCount++
    if (this.truncateCycleCount > MAX_TRUNCATES_PER_WINDOW) {
      console.warn('[TanStack DB] Detected rapid 409 must-refetch cycle...')
      // Reset state but DON'T re-request subsets (breaks the cycle)
      return
    }
  }
  // ...normal truncate handling
}
```

### 2. Reject Pending awaitTxId on 409 (packages/electric-db-collection/src/electric.ts)

When 409 occurs, the shape stream restarts from a different offset. The original txid may never arrive on the new stream, leaving `awaitTxId` hanging indefinitely. Now we reject pending promises immediately:

```typescript
if (isMustRefetchMessage(message)) {
  rejectPendingAwaitTxIds()  // New: reject with SyncInterruptedByRefetchError
  // ...existing 409 handling
}
```

Added new `SyncInterruptedByRefetchError` error type so callers can handle this gracefully.

## Key Invariants

1. **Cycle detection window (2s) must exceed typical truncate processing time** — ensures legitimate back-to-back truncates aren't falsely detected as cycles
2. **Threshold (5 truncates) must be low enough to catch loops before freeze** — but high enough to allow burst scenarios
3. **Pending awaitTxId promises must be rejected before truncate processing** — prevents hanging promises and allows callers to retry

## Non-goals

- **Async/chunked commit processing** — would help with large datasets but is a larger architectural change
- **Rate-limiting status change events** — could reduce re-render pressure but needs more investigation
- **Root cause fix in Electric server** — the 409 responses may be correct behavior for certain subquery configurations; this fix handles the client-side consequences

## Trade-offs

**Why cycle detection instead of preventing the 409s?**
- 409 must-refetch is expected behavior for certain subquery configurations (per Electric maintainers)
- The client must handle this gracefully regardless of server behavior
- Cycle detection is defensive—it only triggers when there's actually a problem

**Why reject awaitTxId instead of just letting it timeout?**
- The timeout is ineffective when the main thread is starved
- Explicit rejection lets callers handle the error and retry if appropriate
- Cleaner than waiting 5 seconds for something that will never arrive

## Verification

```bash
# Run the new tests
pnpm test packages/db/tests/collection-truncate.test.ts
pnpm test packages/electric-db-collection/tests/electric.test.ts

# Specific test cases for this fix:
# - "should detect and break rapid 409 must-refetch cycles to prevent app freezing"
# - "should reset truncate cycle counter after window expires"  
# - "should reject pending awaitTxId promises when must-refetch occurs"
# - "should reject multiple pending awaitTxId promises on must-refetch"
```

## Files Changed

| File | Changes |
|------|---------|
| `packages/db/src/collection/subscription.ts` | Added truncate cycle detection with configurable window/threshold |
| `packages/electric-db-collection/src/electric.ts` | Added `rejectPendingAwaitTxIds()` called on must-refetch, tracks pending awaitTxId calls |
| `packages/electric-db-collection/src/errors.ts` | Added `SyncInterruptedByRefetchError` for explicit rejection |
| `packages/db/tests/collection-truncate.test.ts` | Tests for cycle detection behavior |
| `packages/electric-db-collection/tests/electric.test.ts` | Tests for awaitTxId rejection on 409 |
| `INVESTIGATION-409-FREEZE.md` | Detailed investigation report |

---

### DIFF
--- a/packages/db/tests/collection-truncate.test.ts
+++ b/packages/db/tests/collection-truncate.test.ts
@@ -1150,28 +1150,29 @@ describe(`Collection truncate operations`, () => {
         | undefined
       let loadSubsetCallCount = 0
 
-      const collection = createCollection<{ id: number; value: string }, number>(
-        {
-          id: `cycle-detection-test`,
-          getKey: (item) => item.id,
-          startSync: true,
-          syncMode: `on-demand`,
-          sync: {
-            sync: (cfg) => {
-              syncOps = cfg
-              cfg.markReady()
+      const collection = createCollection<
+        { id: number; value: string },
+        number
+      >({
+        id: `cycle-detection-test`,
+        getKey: (item) => item.id,
+        startSync: true,
+        syncMode: `on-demand`,
+        sync: {
+          sync: (cfg) => {
+            syncOps = cfg
+            cfg.markReady()
 
-              return {
-                loadSubset: () => {
-                  loadSubsetCallCount++
-                  // Simulate async loadSubset
-                  return Promise.resolve()
-                },
-              }
-            },
+            return {
+              loadSubset: () => {
+                loadSubsetCallCount++
+                // Simulate async loadSubset
+                return Promise.resolve()
+              },
+            }
           },
         },
-      )
+      })
 
       await collection.stateWhenReady()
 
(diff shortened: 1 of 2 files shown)


KyleAMathews:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

PR #1127, opened 2026-01-13
Fix date field corruption after app restart

KyleAMathews:
## Summary

Fixes date field corruption that occurred after app restart. String values that happened to match ISO date format (e.g., `eventId: "2024-01-15T10:30:00.000Z"`) were being incorrectly converted to Date objects during deserialization, corrupting user data.

---

## Root Cause

The `JSON.parse` reviver in `deserialize()` used a regex to detect ISO date strings and eagerly converted ALL matches to Date objects:

```typescript
JSON.parse(data, (key, value) => {
  if (typeof value === 'string' && /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}/.test(value)) {
    return new Date(value)  // Corrupts string fields that look like dates!
  }
  return value
})
```

This ignored that the serializer already has an explicit marker system for Date objects:
- `serializeValue()` wraps actual Dates as `{ __type: 'Date', value: '...' }`
- `deserializeValue()` only converts values with that marker back to Dates

The reviver was converting things it shouldn't, bypassing the marker system entirely.

## Approach

1. **Remove the eager reviver** - Let `JSON.parse` return raw data
2. **Handle `createdAt` explicitly** - The only top-level Date field gets explicit conversion
3. **Trust the marker system** - Mutation data uses `{ __type: 'Date' }` markers, handled by existing `deserializeValue()`
4. **Fix the type** - `SerializedOfflineTransaction.createdAt` is now `string` (matches JSON-parsed reality)
5. **Add validation** - Throw on Invalid Dates rather than silently corrupting data

## Key Invariants

1. **String values are never converted to Dates** unless explicitly wrapped with `{ __type: 'Date' }`
2. **Actual Date objects round-trip correctly** via the marker system
3. **Invalid dates fail fast** with clear error messages, not silent corruption

## Non-goals

- Did not add date validation to `OutboxManager`'s catch blocks (separate concern)
- Did not add nominal typing for ISO strings (adds complexity without clear benefit)

## Trade-offs

**Alternative considered:** Keep the reviver but add an exclusion list for fields that shouldn't be converted.

**Why rejected:** Fragile - requires maintaining a list of every string field that might contain ISO-formatted values. The marker system already solves this correctly.

---

## Verification

```bash
cd packages/offline-transactions && pnpm test
```

The new test file covers:
- Plain ISO strings preserved as strings (the bug scenario)
- Actual Date objects restored via marker system
- Mixed Date objects and ISO strings handled correctly
- Nested ISO string values not corrupted
- Top-level `createdAt` correctly restored as Date

## Files Changed

| File | Changes |
|------|---------|
| `TransactionSerializer.ts` | Remove reviver, add explicit `createdAt` conversion, add date validation |
| `types.ts` | `SerializedOfflineTransaction.createdAt`: `Date` → `string` |
| `TransactionSerializer.test.ts` | New test file with 5 tests covering date handling |

### DIFF
--- a/.changeset/fix-date-corruption.md
+++ b/.changeset/fix-date-corruption.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/offline-transactions': patch
+---
+
+Fix date field corruption after app restart. String values matching ISO date format were incorrectly converted to Date objects during deserialization, corrupting user data. Now only explicit Date markers are converted, preserving string values intact.


KyleAMathews:
### Code review

Found 1 issue:

1. Missing changeset file. This PR modifies published package code (`@tanstack/offline-transactions`) to fix a bug, but no changeset file was included. The project uses changesets for versioning and the PR template requires: "This change affects published code, and I have generated a changeset."

https://github.com/TanStack/db/blob/b475edd0ef6fe09cec693ae420f2830da5f8c8e6/.github/pull_request_template.md#L5-L6

Run `pnpm changeset` to generate a changeset file describing this patch fix.

PR #1128, opened 2026-01-13
Pass in join information into queryFn for on-demand syncMode

KyleAMathews:
## Summary

Passes join information to `queryFn` in on-demand mode, enabling server-side joins before pagination. This fixes inconsistent page sizes when queries combine pagination with filters on joined collections.

---

### Root Cause

When using `queryCollection` with joins and pagination (e.g., `limit: 10`), the pagination was applied **before** join filters:

1. `queryFn` received `{ limit: 10, where: <main collection filters only> }`
2. Backend returned 10 rows from main collection
3. Client-side join with other collections
4. Join filters applied client-side
5. **Result:** Only 6 rows match → inconsistent page size

The join information was lost during compilation, so `queryFn` had no way to construct a server-side query that joins and filters before paginating.

### Approach

Extract join information during query optimization and thread it through the compilation pipeline to `LoadSubsetOptions`:

```typescript
// Now available in queryFn
context.meta.loadSubsetOptions.joins = [
  {
    collectionId: 'accounts',
    alias: 'account',
    type: 'inner',
    localKey: { type: 'ref', path: ['task', 'account_id'] },
    foreignKey: { type: 'ref', path: ['account', 'id'] },
    where: /* filters on account collection */,
    orderBy: /* order by expressions referencing account */,
  }
]
```

### Key Invariants

1. **Join info is keyed by main source alias** - A query `FROM tasks JOIN accounts` produces `{ tasks: [accountJoinInfo] }`
2. **WHERE clauses are partitioned** - Filters referencing `account.*` go into `joinInfo.where`, not `loadSubsetOptions.where`
3. **Key direction is normalized** - `localKey` always references main collection, `foreignKey` always references joined collection (even if IR has them swapped)
4. **Subqueries are traversed** - `getInnermostCollectionId()` recursively finds the actual collection ID through QueryRef chains

### Non-goals

- **No automatic server-side query generation** - We provide the information; `queryFn` implementations decide how to use it
- **No changes to client-side join execution** - Existing IVM-based joins continue to work unchanged
- **No breaking changes** - `joins` is optional; existing `queryFn` implementations work without modification

### Trade-offs

**Why extract at optimizer level?**
The optimizer already analyzes WHERE clauses by source alias (`sourceWhereClauses`). Extracting join info here reuses that analysis and keeps join-related logic centralized.

**Why include `where` and `orderBy` in JoinInfo?**
Server-side queries need the complete picture. Including these avoids multiple round-trips through the compilation result to reconstruct which filters apply to which joins.

---

## Verification

```bash
# Run optimizer tests (includes new join info extraction tests)
pnpm vitest --run optimizer.test.ts

# Run query-db-collection tests (includes serialization)
cd packages/query-db-collection && pnpm vitest --run
```

## Files Changed

| File | Change |
|------|--------|
| `packages/db/src/types.ts` | Add `JoinInfo` type, extend `LoadSubsetOptions` with `joins` |
| `packages/db/src/query/optimizer.ts` | Add `extractJoinInfo()`, include `joinInfoBySource` in optimization result |
| `packages/db/src/query/compiler/index.ts` | Pass `joinInfoBySource` through `CompilationResult` |
| `packages/db/src/query/live/collection-config-builder.ts` | Store `joinInfoBySourceCache` |
| `packages/db/src/query/live/collection-subscriber.ts` | Add `getJoinInfoForAlias()`, pass to subscriptions |
| `packages/db/src/collection/subscription.ts` | Include `joins` in `loadSubset` calls |
| `packages/query-db-collection/src/serialization.ts` | Serialize joins in query keys with validation |
| `packages/db/tests/query/optimizer.test.ts` | Add 8 tests for join info extraction |


KyleAMathews:
Closing as fixed by a different current architecture. The joined-window pagination oracle now returns the correct limited result and verifies that the source is loaded without a pre-join limit, so the reported underfill no longer reproduces and the proposed join-metadata API is not required. Thanks @KyleAMathews; this PR remains provenance for the invariant.

PR #1130, opened 2026-01-14
Fix syncedData not updating on manual writes in mutationFn

KyleAMathews:
## Summary

Manual write operations (`writeUpsert`, `writeInsert`, `writeUpdate`, `writeDelete`) now correctly update `syncedData` when called from within mutation handlers that include async operations. Previously, `syncedData` would appear stale until the next sync cycle.

---

## Root Cause

The bug occurs in this sequence:

1. User calls `collection.update()` which creates a transaction
2. The `onUpdate` handler fires and makes an async API call (`await fetch(...)`)
3. During the `await`, the transaction transitions to `persisting` state
4. Handler calls `writeUpsert()` to update `syncedData` with server response
5. **Bug**: `commitPendingTransactions()` sees `hasPersistingTransaction=true` and skips processing the sync transaction

The guard `if (!hasPersistingTransaction || hasTruncateSync)` exists to prevent sync operations from overwriting in-flight optimistic updates. But manual writes are different—they're intentionally updating `syncedData` to match server state, and should be processed immediately.

## Approach

Added an `immediate` flag to sync transactions that bypasses the persisting-transaction check:

```typescript
// sync.ts - begin() now accepts options
begin: (options?: { immediate?: boolean }) => {
  this.state.pendingSyncedTransactions.push({
    committed: false,
    operations: [],
    deletedKeys: new Set(),
    immediate: options?.immediate,  // new
  })
}

// manual-sync.ts - write operations use immediate: true
ctx.begin({ immediate: true })
```

The condition becomes:
```typescript
if (!hasPersistingTransaction || hasTruncateSync || hasImmediateSync) {
  // Process sync transactions
}
```

## Key Invariants

1. Manual writes (`writeUpsert`, etc.) must update `syncedData` synchronously, regardless of transaction state
2. When an immediate (or truncate) transaction triggers processing, **all** committed sync transactions are processed together to preserve causal ordering. This prevents earlier non-immediate transactions from being applied later and overwriting newer state.
3. Without an immediate/truncate transaction, sync operations wait for persisting transactions to complete

## Non-goals

- Did not change the general sync flow behavior—only manual writes get the `immediate` flag
- Did not expose `immediate` in the public API beyond the existing write utilities

## Trade-offs

**Alternative considered**: Process only the immediate transaction, leaving other committed sync transactions queued.

Rejected because this would break ordering: if you have [syncA, syncB(immediate)] and only process syncB, then syncA would apply later and could overwrite syncB's changes with stale data. Processing all committed transactions together maintains the queue's causal ordering.

**Alternative considered**: Process all sync transactions immediately regardless of persisting state.

Rejected because this could cause race conditions where a slow sync response overwrites optimistic state that's still being persisted. The `immediate` flag is surgical—it only triggers when there's an intentional manual write.

---

## Verification

```bash
pnpm test -- packages/query-db-collection/tests/query.test.ts
```

The new test `should update syncedData immediately when writeUpsert is called after async API in onUpdate handler` reproduces the exact bug scenario.

## Files Changed

| File | Changes |
|------|---------|
| `packages/db/src/types.ts` | Added `immediate` option to `begin()` signature with JSDoc |
| `packages/db/src/collection/sync.ts` | Pass `immediate` option through to pending transaction |
| `packages/db/src/collection/state.ts` | Track `hasImmediateSync` flag, process when true; added comment explaining ordering semantics |
| `packages/query-db-collection/src/manual-sync.ts` | Use `begin({ immediate: true })` for write operations |
| `packages/query-db-collection/tests/query.test.ts` | Regression test for the async-onUpdate scenario |

---

### DIFF
--- a/packages/db/src/collection/state.ts
+++ b/packages/db/src/collection/state.ts
@@ -448,10 +448,10 @@ export class CollectionStateManager<
       (acc, t) => {
         if (t.committed) {
           acc.committedSyncedTransactions.push(t)
-          if (t.truncate === true) {
+          if (t.truncate) {
             acc.hasTruncateSync = true
           }
-          if (t.immediate === true) {
+          if (t.immediate) {
             acc.hasImmediateSync = true
           }
         } else {


tobacode:
I tested the PR and it solves my issue. I no longer have this "off-by-one" cache problem.

packages/db/src/collection/state.ts
```diff
@@ -437,13 +443,17 @@ export class CollectionStateManager<
...
       hasTruncateSync,
+      hasImmediateSync,
     } = this.pendingSyncedTransactions.reduce(
       (acc, t) => {
         if (t.committed) {
           acc.committedSyncedTransactions.push(t)
           if (t.truncate === true) {
             acc.hasTruncateSync = true
           }
+          if (t.immediate === true) {
```
kevin-dp: @claude Since `t.immediate` is of type `boolean` this can be shortened to `if (t.immediate) {`. Similarly, fix the same thing 3 lines above for the `t.truncate` check.

PR #1139, opened 2026-01-15
Let Claude reproduce and fix new issues

kevin-dp:
Let's see how this goes


KyleAMathews:
My main concern is this runs automatically — I think a quick human-in-the-loop for ok-ing the agent starting would be good as a precaution against abuse. But until that happens... it's probably not worth setting it up. Ideal I think would be a discord bot that pulls in an issue and does a check and asks us for approval to start working on it so very easy/fast. But that can be a future follow-up.

KyleAMathews (approved):

kevin-dp:
@KyleAMathews i want to strike a good balance here. If it requires too much manual input then issues will keep accumulating. If we require no interaction at all, we are indeed at risk of abuse. I'll go for ok-ing the agent via a comment by a maintainer.

PR #1141, opened 2026-01-15
Compile query in DB w/ the columnMapper

KyleAMathews:
## Summary

Fixes SQL compilation for on-demand sync to properly encode column names using the configured `columnMapper`. Previously, queries with camelCase property names would generate incorrect SQL when `snakeCamelMapper` was configured. Also adds developer warnings when on-demand collections are passed directly to `useLiveQuery` without a query builder.

---

## Root Cause

When using `snakeCamelMapper` (which converts between camelCase JS properties and snake_case DB columns), the SQL compiler was not applying the encoding transformation. For example:

```typescript
// User writes a query using camelCase (JS convention):
q.from({ p: programs }).where(({ p }) => p.programTemplateId === someId)

// SQL was incorrectly generated as:
// WHERE "programTemplateId" = $1  ❌ (column doesn't exist in Postgres)

// Should be:
// WHERE "program_template_id" = $1  ✅
```

The `columnMapper.encode` function existed and was used during data sync, but was never passed to the SQL compilation step.

## Approach

**SQL Compiler Enhancement:**
- Added optional `encodeColumnName` parameter to `compileSQL()` 
- Threading the encoder through all expression compilation functions (`compileBasicExpression`, `compileOrderBy`, `compileFunction`)
- Applied encoding in `quoteIdentifier()` before quoting column names

**Integration:**
- `createLoadSubsetDedupe()` now receives `shapeOptions.columnMapper?.encode` 
- All `compileSQL()` calls pass the compile options when encoding is available

**Developer Experience:**
- Added console warnings in React, Vue, and Svelte `useLiveQuery` hooks when an on-demand collection is passed directly (which loads no data since there's no predicate)

## Key Invariants

1. Column names in generated SQL must match the actual database column names
2. The `columnMapper` transformation must be applied consistently: during sync (already worked) AND during query compilation (this fix)
3. When no `columnMapper` is configured, column names pass through unchanged

## Non-goals

- Did not change how `columnMapper` works for data sync (already correct)
- Did not add runtime validation that column names exist in the schema
- Did not make the on-demand + direct collection usage an error (just a warning)

## Trade-offs

**Warning vs Error for on-demand misuse:** Chose a warning because some developers might intentionally pass an on-demand collection and rely on external mechanisms to load data. An error would be breaking; a warning educates without blocking.

---

## Verification

```bash
pnpm test --filter @tanstack/electric-db-collection
```

The new tests in `sql-compiler.test.ts` verify:
- camelCase → snake_case encoding in WHERE clauses
- Encoding in ORDER BY clauses  
- Encoding in IS NULL / IS NOT NULL expressions
- Encoding in nested AND/OR expressions
- Encoding in LIKE expressions
- No transformation when encoder is absent

## Files Changed

| File | Change |
|------|--------|
| `packages/electric-db-collection/src/sql-compiler.ts` | Added `ColumnEncoder` type and `CompileSQLOptions` interface; threaded `encodeColumnName` through all compilation functions |
| `packages/electric-db-collection/src/electric.ts` | Passed `shapeOptions.columnMapper?.encode` to `createLoadSubsetDedupe()` and through to `compileSQL()` |
| `packages/electric-db-collection/tests/sql-compiler.test.ts` | Added 9 tests for column name encoding scenarios |
| `packages/react-db/src/useLiveQuery.ts` | Added warning when on-demand collection passed directly |
| `packages/vue-db/src/useLiveQuery.ts` | Added warning when on-demand collection passed directly |
| `packages/svelte-db/src/useLiveQuery.svelte.ts` | Added warning when on-demand collection passed directly |

### DIFF
--- a/packages/electric-db-collection/src/electric.ts
+++ b/packages/electric-db-collection/src/electric.ts
@@ -13,7 +13,7 @@ import {
   TimeoutWaitingForMatchError,
   TimeoutWaitingForTxIdError,
 } from './errors'
-import {  compileSQL } from './sql-compiler'
+import { compileSQL } from './sql-compiler'
 import {
   addTagToIndex,
   findRowsMatchingPattern,
@@ -22,7 +22,7 @@ import {
   removeTagFromIndex,
   tagMatchesPattern,
 } from './tag-index'
-import type {ColumnEncoder} from './sql-compiler';
+import type { ColumnEncoder } from './sql-compiler'
 import type {
   MoveOutPattern,
   MoveTag,
@@ -437,7 +437,10 @@ function createLoadSubsetDedupe<T extends Row<unknown>>({
           orderBy,
           // No limit - get all ties
         }
-        const whereCurrentParams = compileSQL<T>(whereCurrentOpts, compileOptions)
+        const whereCurrentParams = compileSQL<T>(
+          whereCurrentOpts,
+          compileOptions,
+        )
         promises.push(stream.requestSnapshot(whereCurrentParams))
 
         debug(
--- a/packages/react-db/src/useLiveQuery.ts
+++ b/packages/react-db/src/useLiveQuery.ts
@@ -353,7 +353,9 @@ export function useLiveQuery(
       // Warn when passing a collection directly with on-demand sync mode
       // In on-demand mode, data is only loaded when queries with predicates request it
       // Passing the collection directly doesn't provide any predicates, so no data loads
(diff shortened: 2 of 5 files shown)


ButlerBorst:
Hey hey! Long time follower first time commenter. 

I reported this bug 1/15 after noticing it the day prior. I've installed this PR locally and can confirm the fix works perfectly. Thanks for all y'all are doing. Keep up the good work

PR #1146, opened 2026-01-17
Fix sql-compiler to handle empty strings like `eq(column, '')`

### LINKED ISSUE #1147: [Bug] Empty String & NULL/undefined values are silently omitted from query electric db collection query params
### Summary

When a where clause contains an empty string, null, or undefined value, the parameter is silently omitted from the params record, producing a malformed query.

### Reproduction

```typescript
useLiveQuery((q) => 
  q.from({ sessions }).where(({ sessions }) => eq(sessions.host_id, ''))
)
```

Produces: `WHERE "host_id" = $1` with `params: {}` — the placeholder has no corresponding value.

### Root Cause

**1. Serializer returns empty string for null/undefined and passes through empty strings**
[pg-serializer.ts:15-22](https://github.com/TanStack/db/blob/main/packages/electric-db-collection/src/pg-serializer.ts#L15-L22)
```typescript
if (value === null || value === undefined) {
  return ``
}
if (typeof value === `string`) {
  return value  // empty string passes through as ''
}
```

**2. Empty strings are silently omitted from params record**
[sql-compiler.ts:43-45](https://github.com/TanStack/db/blob/main/packages/electric-db-collection/src/sql-compiler.ts#L43-L45)
```typescript
if (serialized !== ``) {
  acc[`${index + 1}`] = serialized
}
```

Note: null/undefined in comparison operators like `eq()` will throw an error earlier ([sql-compiler.ts:145-159](https://github.com/TanStack/db/blob/main/packages/electric-db-collection/src/sql-compiler.ts#L145-L159)), but empty strings bypass this check entirely.

### Suggested Fix

Either:
- Include empty strings in the params record (valid use case to query for empty string values)
- Throw an error when a param would be omitted, rather than silently producing a broken query

KyleAMathews:
## Summary

Fixes empty string values being incorrectly treated as NULL in SQL queries. Queries like `eq(status, '')` now correctly generate `WHERE status = ''` instead of silently omitting the parameter.

---

### Root Cause

The sql-compiler's param filtering logic was checking if the *serialized* value was an empty string (`serialized !== ''`) to decide whether to include a param. This conflated two distinct cases:
1. A value that serializes to empty string because it IS an empty string (legitimate)
2. A null/undefined value that also happens to serialize to empty string (should be omitted)

This caused legitimate empty string queries to break silently — the param would be omitted from the params record, but the `$N` placeholder would still appear in the SQL, leading to incorrect query behavior.

### Approach

Changed the filter condition from checking the serialized output to checking the original input value:

```typescript
// Before: checked serialized output (broken for empty strings)
if (serialized !== ``) {
  acc[`${index + 1}`] = serialized
}

// After: checks original value (empty strings pass through)
if (param != null) {
  acc[`${index + 1}`] = serialized
}
```

### Key Invariants

1. `null` and `undefined` params MUST be omitted (they indicate optional/unset values)
2. Empty strings MUST be included (they're valid query values)
3. The number of params in the record must match the number of `$N` placeholders in the SQL

### Non-goals

- Did not change how null/undefined values are handled in the WHERE clause generation itself (that's a separate concern)
- Did not add support for explicit `IS NULL` queries (different feature)

### Trade-offs

Alternative considered: Could have made `serialize()` return `null` for null/undefined and then filter on that. Rejected because it would change the serialize function's contract and add complexity for a simple fix.

---

## Verification

```bash
pnpm vitest --run tests/sql-compiler.test.ts
```

The tests verify:
- Single `eq` with empty string includes the param
- Compound `AND` clause with empty string includes all params
- Exact reproduction case from issue #1147 (`host_id = ''`)
- Multiple empty strings get correct param indices

### Files Changed

| File | Change |
|------|--------|
| `packages/electric-db-collection/src/sql-compiler.ts` | Changed param filtering to check original value instead of serialized output |
| `packages/electric-db-collection/tests/sql-compiler.test.ts` | Added 4 test cases for empty string equality queries |

---

Fixes #1147

### DIFF
--- a/.changeset/fix-empty-string-params.md
+++ b/.changeset/fix-empty-string-params.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/electric-db-collection': patch
+---
+
+Fix empty string values being incorrectly omitted from SQL query params. Queries like `eq(column, '')` now correctly include the empty string parameter instead of producing a malformed query with a missing `$1` value.


K-Mistele:
related #1146

thruflo (approved):

PR #1151, opened 2026-01-18
Export and document subtract, multiply, divide math functions

KyleAMathews:
## Summary

Add `subtract`, `multiply`, and `divide` math functions to enable computed columns in queries. This unlocks ranking algorithms and dynamic calculations directly in `select` and `orderBy` clauses without pre-computing values.

## Approach

Extended the existing math function pattern established by `add()`. Each function:
- Takes two numeric operands (field refs, literals, or nested expressions)
- Returns `BasicExpression<number>` for `add`/`subtract`/`multiply` because nullish operands are coalesced to `0`; `divide` returns `BasicExpression<number | null>` because divide-by-zero returns `null`
- Compiles to the corresponding IR node for execution

```ts
// Functions follow the same pattern as add()
export function subtract<T1, T2>(left: T1, right: T2): BasicExpression<number> {
  return new Func(`subtract`, [toExpression(left), toExpression(right)])
}
```

**Key design choice:** Functions can be nested for complex calculations:
```ts
// (salary * 1.1) - 500
subtract(multiply(employees.salary, 1.1), 500)
```

### Key Invariants

1. **Type safety preserved** — Return types match runtime semantics: `add`/`subtract`/`multiply` coalesce nullish operands to `0`, while `divide` can return `null` for divide-by-zero
2. **Composability** — Functions nest arbitrarily without special handling
3. **Consistent pattern** — All four arithmetic functions share identical structure with `add()`

### Non-goals

- **Operator syntax** — No `a + b` syntax; explicit function calls keep the IR clean

## Verification

```bash
pnpm test --filter @tanstack/db
```

Tests cover:
- Each function individually (`subtract`, `multiply`, `divide`)
- Nested composition (e.g., `subtract(multiply(...), ...)`)
- Usage in `orderBy` clauses

## Files Changed

| File | Change |
|------|--------|
| `packages/db/src/query/builder/functions.ts` | Add `subtract`, `multiply`, `divide` functions |
| `packages/db/src/query/index.ts` | Export new functions |
| `packages/db/tests/query/builder/functions.test.ts` | Add tests for new functions |
| `docs/guides/live-queries.md` | Document functions + computed orderBy example |
| `.changeset/add-math-functions.md` | Changeset for release |

---

## Release Notes

* **New Features**
  * Added math expression helpers (`subtract`, `multiply`, `divide`) for computed `select` and `orderBy` expressions.
  * `divide(a, b)` now returns `null` when dividing by zero.
* **Documentation**
  * Updated the Expression Functions Reference with new math helper examples.
  * Added guidance that computed expressions in `orderBy` combined with `limit()` disable a lazy-loading optimization.
* **Tests**
  * Added type and runtime coverage for `subtract`, `multiply`, and `divide`, including composed expressions and `orderBy` behavior.


kevin-dp (changes requested):
There are 2 type related issues.

## Issue 1: Unnecessary type widening on `null`/`undefined` operands

The PR description mentions:
> Nullability propagates correctly (if either operand can be null/undefined, result can be too)

At the type level, an operand that is `null` or `undefined` indeed widens the result type:
```ts
subtract(x: number, y: number) => BasicExpression<number>
subtract(x: number, y: number | null) => BasicExpression<number | null>
subtract(x: number, y: number | undefined) => BasicExpression<number | undefined>
```

But at the runtime level (`evaluators.ts`) it coalesces `null`/`undefined` away:
```ts
return (a ?? 0) - (b ?? 0)
```
So while the type widening is correct (typewise), it is misleading and imprecise since at runtime it will never return `null`.

## Issue 2: Type of divide is incorrect

At runtime `divide` returns `null` on divide-by-zero but the type gives non-null operands a non-null result:
```ts
divide(order.total: number, order.itemCount: number) => BasicExpression<number>
```
So according to this type `divide(10, 0)` should return a number but it returns `null`.

kevin-dp (approved):

PR #1169, opened 2026-01-20
Restore offline transactions to optimistic store upon restart the app

KyleAMathews:
## Summary

Fixes offline transactions not restoring optimistic state when the page is refreshed while offline. Users now see their pending changes immediately on page load, providing a seamless offline UX.

---

## Root Cause

When the `OfflineExecutor` loaded pending transactions from storage on startup, it only scheduled them for execution—it never re-applied the mutations to the collection's state manager. The optimistic state lived only in memory and was lost on page refresh.

## Approach

Created "restoration transactions" that shadow the persisted offline transactions:

```typescript
// TransactionExecutor.restoreOptimisticState()
const restorationTx = createTransaction({
  id: offlineTx.id,
  autoCommit: false,
  mutationFn: async () => {},
})
restorationTx.applyMutations(offlineTx.mutations)
mutation.collection._state.transactions.set(restorationTx.id, restorationTx)
mutation.collection._state.recomputeOptimisticState(true)
```

When pending transactions are loaded from storage, we:
1. Create a restoration transaction with the same ID
2. Apply the persisted mutations to it
3. Register it with each affected collection's state manager
4. Clean up when the real offline transaction completes or fails

## Key Invariants

1. **Restoration transactions never commit** - they exist only to hold mutations for optimistic display
2. **Same ID as offline transaction** - ensures proper cleanup when the real transaction resolves
3. **Cleanup on completion** - whether success or failure, restoration transactions are removed so sync provides authoritative data
4. **Key extraction during deserialization** - mutations need their `key` populated for optimistic state to work

## Non-goals

- **Re-executing mutations on restore** - we only restore the visual state, not re-run the mutation logic
- **Handling conflicts** - conflict resolution happens when the actual sync completes, not during restoration

## Trade-offs

**Alternative considered**: Storing the optimistic state separately from the transaction data.
**Why this approach**: Reusing the existing transaction/mutation infrastructure is simpler and leverages the collection's built-in optimistic state management. No new storage schema or state reconciliation logic needed.

---

## Verification

```bash
pnpm test:pr
```

The new test `should restore optimistic state to collection on startup` verifies the complete flow:
1. Create transaction with optimistic mutation
2. Dispose executor (simulate page refresh)
3. Create new executor with same storage
4. Assert optimistic state is visible before transaction completes

## Files Changed

| File | Description |
|------|-------------|
| `OfflineExecutor.ts` | Added `restorationTransactions` map, `registerRestorationTransaction()`, `cleanupRestorationTransaction()`, and `waitForInit()` helper |
| `TransactionExecutor.ts` | Added `restoreOptimisticState()` that creates restoration transactions on load |
| `TransactionSerializer.ts` | Extract `key` from modified data during deserialization |
| `offline-e2e.test.ts` | New test covering the page refresh scenario |
| `harness.ts` | Minor test harness updates |

### DIFF
--- a/packages/offline-transactions/src/executor/TransactionExecutor.ts
+++ b/packages/offline-transactions/src/executor/TransactionExecutor.ts
@@ -269,6 +269,13 @@ export class TransactionExecutor {
           mutationFn: async () => {},
         })
 
+        // Prevent unhandled promise rejection when cleanup calls rollback()
+        // We don't care about this promise - it's just for holding mutations
+        restorationTx.isPersisted.promise.catch(() => {
+          // Intentionally ignored - restoration transactions are cleaned up
+          // via cleanupRestorationTransaction, not through normal commit flow
+        })
+
         restorationTx.applyMutations(offlineTx.mutations)
 
         // Register with each affected collection's state manager


TomasGonzalez:
I reported this bug and just tested this PR build locally. After refreshing while offline, the optimistic state is now restored correctly and pending changes show up immediately on page load.
Looks good on my end ✅ Thanks for the quick fix!

samwillis:
Review from GPT5.2:

-----

Here’s my review of PR #1169 (“Restore offline transactions to optimistic store upon restart the app”). Overall: the direction is solid and the tests you added are the right shape. The core idea—rehydrating pending offline mutations into a “restoration transaction” so collections can rebuild optimistic state on startup—makes sense. ([[GitHub](https://github.com/TanStack/db/pull/1169/commits/aad8aea19ce52678986b69873a92025e1b6c7be3)][1])

## What looks good

* **Repro + coverage**: The new e2e repro captures the real UX failure mode (refresh while offline → UI loses optimistic state until sync eventually happens). ([[GitHub](https://github.com/TanStack/db/pull/1169/commits/72f94f9d31597601423db97a22ca5b4e60a774d3)][2])
* **Restoration mechanism**: Creating an internal transaction (never committed) and registering it in each touched collection’s optimistic transaction map is a pragmatic way to “plug back into” the existing optimistic-state machinery. ([[GitHub](https://github.com/TanStack/db/pull/1169/commits/aad8aea19ce52678986b69873a92025e1b6c7be3)][1])
* **Cleanup & rollback paths**: Tracking restoration transactions and explicitly removing/rolling them back on completion or permanent failure is the right lifecycle model, and the added rollback test is a strong guardrail. ([[GitHub](https://github.com/TanStack/db/pull/1169/commits/aad8aea19ce52678986b69873a92025e1b6c7be3)][1])
* **Defensive hardening**: The try/catch isolation and null checks around potentially-corrupt deserialized mutations are sensible. ([[GitHub](https://github.com/TanStack/db/pull/1169/files)][3])

## Must-fix / likely bugs

### 1) `executeAll()` is called twice during startup

In `OfflineExecutor.loadAndReplayTransactions()` you `await this.executor.executeAll()` and then immediately call `this.executor.executeAll().catch(...)` again, while the comment says “don’t await”. That’s almost certainly unintended and risks double-scheduling / double-processing. ([[GitHub](https://github.com/TanStack/db/pull/1169/commits/aad8aea19ce52678986b69873a92025e1b6c7be3)][1])

**Suggestion:** pick one:

* If you truly want non-blocking init: `await loadPendingTransactions(); executor.executeAll().catch(...)` (no awaited `executeAll`)
* If you want blocking init: `await loadPendingTransactions(); await executor.executeAll()` and remove the second call + comment

(Your new `waitForInit()` is a nice hook either way.) ([[GitHub](https://github.com/TanStack/db/pull/1169/commits/aad8aea19ce52678986b69873a92025e1b6c7be3)][1])

### 2) `TransactionSerializer.deserializePendingMutation()` returns duplicate fields + likely wrong `key` for deletes

In the returned object literal, you currently set:

* `modified` twice (`modified: this.deserializeValue(...)` and also `modified`), and
* `key` twice (`key: null` and then `key`) — the earlier one is dead and confusing. ([[GitHub](https://github.com/TanStack/db/pull/1169/files)][3])

More importantly: `key` is computed only from `modified`. For `delete` mutations, `modified` is usually null and `original` holds the record—so deletes may *not* restore optimistic state correctly after refresh.

**Suggestion (safer + cleaner):**

* Deserialize once (`modified`, `original`)
* Compute key from `modified ?? original` (or based on `type`, if you prefer)
* Return each field exactly once

This is pretty load-bearing because restoration relies on `key` lookups to rebuild optimistic state. ([[GitHub](https://github.com/TanStack/db/pull/1169/files)][3])

## Nice-to-haves / design notes

* **Private `_state` poking**: You’re reaching into `collection._state.transactions` and calling `_state.recomputeOptimisticState(...)`. It works, but it’s a sharp edge. If you can introduce a tiny internal helper on Collection/StateManager (even non-public) like `registerOptimisticTransaction(tx)` / `unregisterOptimisticTransaction(txId, opts)` you’d centralize invariants and reduce “knowledge leakage.” ([[GitHub](https://github.com/TanStack/db/pull/1169/commits/aad8aea19ce52678986b69873a92025e1b6c7be3)][1])
* **`waitForLeader()` loop**: In the test harness you call `waitForInit()` inside the polling loop. It’s fine, but you could simplify to: wait for leader → `await waitForInit()` once. (Current version is still correct if `waitForInit` resolves quickly.) ([[GitHub](https://github.com/TanStack/db/blob/1eda746745e9836a9305da4613c6bf3a1fe194f9/packages/offline-transactions/src/OfflineExecutor.ts)][4])
* **Test naming/comment**: The original repro test still reads like it’s expected to fail (“BUG: currently does not”). If it now passes in this PR, it’d be good to update the wording so it doesn’t confuse future readers. ([[GitHub](https://github.com/TanStack/db/pull/1169/commits/72f94f9d31597601423db97a22ca5b4e60a774d3)][2])

## One question to sanity-check (not blocking, but important)

Do offline “real” transactions use the **same transaction id** as the persisted offline transaction id (the one you’re using for `restorationTx.id`)? It *seems* like you’re intentionally matching ids so the optimistic subsystem behaves consistently across refresh—which is good—but if there’s any path where a second live Transaction with the same id is created while the restoration tx is still registered, you could get subtle clashes. The cleanup logic helps, but it’s worth confirming the invariants. ([[GitHub](https://github.com/TanStack/db/pull/1169/commits/aad8aea19ce52678986b69873a92025e1b6c7be3)][1])

[1]: https://github.com/TanStack/db/pull/1169/commits/aad8aea19ce52678986b69873a92025e1b6c7be3 "Restore offline transactions to optimistic store upon restart the app by KyleAMathews · Pull Request #1169 · TanStack/db · GitHub"
[2]: https://github.com/TanStack/db/pull/1169/commits/72f94f9d31597601423db97a22ca5b4e60a774d3 "Restore offline transactions to optimistic store upon restart the app by KyleAMathews · Pull Request #1169 · TanStack/db · GitHub"
[3]: https://github.com/TanStack/db/pull/1169/files "Restore offline transactions to optimistic store upon restart the app by KyleAMathews · Pull Request #1169 · TanStack/db · GitHub"
[4]: https://github.com/TanStack/db/blob/1eda746745e9836a9305da4613c6bf3a1fe194f9/packages/offline-transactions/src/OfflineExecutor.ts "db/packages/offline-transactions/src/OfflineExecutor.ts at 1eda746745e9836a9305da4613c6bf3a1fe194f9 · TanStack/db · GitHub"

PR #1173, opened 2026-01-22
Add safety limits and diagnostics for ORDER BY + LIMIT infinite loops

KyleAMathews:
## Summary

Adds iteration safety limits to prevent app freezes from excessive iterations in ORDER BY + LIMIT queries. When limits are hit, warnings are logged with diagnostic info but **queries continue normally** with available data.

**User impact**: Apps no longer freeze. If iteration limits are hit, users get their data (possibly incomplete TopK) and a console warning with diagnostic info.

## Root Cause

ORDER BY + LIMIT queries can enter excessive iteration cycles when:
1. TopK operator requests `limit` items (e.g., 10)
2. WHERE clause filters out most data (only 2 items match)
3. TopK keeps requesting more data that doesn't exist locally
4. Each iteration triggers more graph processing

The exact trigger conditions in production are still unclear - we observed debugger pausing in WHERE clause evaluation (`and`, `gte` cases) during `requestLimitedSnapshot`.

## Approach

**Graceful circuit breakers**: Add iteration limits that break out of loops and continue with available data:

| Location | Limit | Purpose |
|----------|-------|---------|
| `D2.run()` | 100,000 | Prevents infinite dataflow cycles |
| `maybeRunGraph` | 10,000 | Prevents graph/loading feedback loops |
| `requestLimitedSnapshot` | 10,000 | Prevents index iteration issues |

**Key insight**: By the time we hit the limit, we likely have all available data - we just couldn't fill the TopK completely because WHERE filtered out most items. So we log a warning and continue rather than erroring.

**Iteration breakdown tracking**: When limits are hit, warnings show WHERE the loop got stuck:
```
Iteration breakdown (where the loop spent time):
    1-5: operators with work = [Filter,TopK]
    6-100000: operators with work = [TopK]
```

This makes it clear that the loop got stuck requesting more data for TopK after filtering completed.

**Additional diagnostic info logged**:
```
- Live query ID
- Source collection IDs  
- Run count
- OrderBy optimization info (limit, offset, dataNeeded)
- Cursor position and iteration counts
- Which D2 operators have pending work
```

## Non-goals

- **Root cause fix**: We're shipping a graceful band-aid. The diagnostic info will help identify the actual cause when warnings occur in production.
- **Performance optimization**: The limits are generous enough to never trigger for legitimate workloads.

## Trade-offs

| Option | Pros | Cons |
|--------|------|------|
| **Error state** (rejected) | Clear signal something is wrong | Breaks app, requires refresh |
| **Graceful recovery** (chosen) | App keeps working, data is shown | Users might not notice warnings |

## Verification

```bash
# Run the safety limit tests
pnpm vitest run packages/db/tests/infinite-loop-prevention.test.ts

# Run NaN handling tests (related edge case exploration)
pnpm vitest run packages/db/tests/nan-comparator-infinite-loop.test.ts
```

### DIFF
--- a/packages/db/src/collection/subscription.ts
+++ b/packages/db/src/collection/subscription.ts
@@ -400,19 +400,22 @@ export class CollectionSubscription
   }
 
   /**
-   * Sends a snapshot that fulfills the `where` clause and all rows are bigger or equal to the cursor.
-   * Requires a range index to be set with `setOrderByIndex` prior to calling this method.
-   * It uses that range index to load the items in the order of the index.
+   * Loads rows from the local collection in sorted order using the BTree index.
+   *
+   * Uses the BTree index (set via `setOrderByIndex`) to iterate through the local
+   * collection's data in sorted order, starting from the cursor position (`minValues`).
+   * Also triggers an async `loadSubset` call to fetch more data from the sync layer
+   * (e.g., Electric backend) if needed.
    *
    * For multi-column orderBy:
-   * - Uses first value from `minValues` for LOCAL index operations (wide bounds, ensures no missed rows)
-   * - Uses all `minValues` to build a precise composite cursor for SYNC layer loadSubset
+   * - Uses first value from `minValues` for BTree index operations (wide bounds, ensures no missed rows)
+   * - Uses all `minValues` to build a precise composite cursor for sync layer loadSubset
    *
-   * Note 1: it may load more rows than the provided LIMIT because it loads all values equal to the first cursor value + limit values greater.
-   *         This is needed to ensure that it does not accidentally skip duplicate values when the limit falls in the middle of some duplicated values.
-   * Note 2: it does not send keys that have already been sent before.
+   * Note 1: May load more rows than `limit` because it includes all rows equal to the
+   *         cursor value. This prevents skipping duplicates when limit falls mid-duplicates.
+   * Note 2: Skips keys that have already been sent to prevent duplicates.
    *
-   * @returns true if local data was found and sent, false if the local index was exhausted
+   * @returns true if local data was found, false if no more matching rows exist locally
    */
   requestLimitedSnapshot({
     orderBy,
@@ -635,12 +638,8 @@ export class CollectionSubscription
     this.loadedSubsets.push(loadOptions)
     this.trackLoadSubsetPromise(syncResult)
 
-    // Return whether local data was found and iteration completed normally.
-    // Return false if:
-    // - No local data was found (index exhausted)
-    // - Iteration limit was hit (abnormal exit)
-    // Either case signals that the caller should stop trying to load more.
-    // The async loadSubset may still return data later.
+    // Return true if we found and sent local data, false otherwise.
+    // The async loadSubset call may still fetch data from the backend.
     return changes.length > 0 && !hitIterationLimit
   }
 


packages/db-ivm/src/d2.ts
```diff
@@ -57,7 +58,30 @@ export class D2 implements ID2 {
...
+      MAX_RUN_ITERATIONS,
+      (state) => `operators with work = [${state}]`,
+    )
+
     while (this.pendingWork()) {
+      const operatorsWithWork = this.#operators
+        .filter((op) => op.hasPendingWork())
+        .map((op) => op.constructor.name)
+        .sort()
+        .join(`,`)
```
samwillis: I'm not too keen on this inside the while loop, it's stringifing a bunch of stuff on each iteration even if it's not shown. Ideally anything we pass to `trackAndCheckLimit` is lazily stringified so we are minimising the work.
At the minimum lift it out of the loop.

packages/db/src/query/live/collection-subscriber.ts
```diff
@@ -222,8 +222,9 @@ export class CollectionSubscriber<
     const sendChangesInRange = (
       changes: Iterable<ChangeMessage<any, string | number>>,
     ) => {
+      const changesArray = Array.isArray(changes) ? changes : [...changes]
```
samwillis: I don't understand why it's making this change here - splitUpdates accepts an iterator?  I think it's a redundant change thats doing a wasteful allocation.

kevin-dp:
I would definitely make this opt-in or enable only in debug mode because i'm worried about 2 things:
- performance hit of doing additional tracking in D2 pipeline
- limits being hit in a healthy application

For the limits being hit in a healthy application what i mean is that an application could do a `LIMIT 1` to get the single best result. But the query may have an additional WHERE clause that filters out most of the data. The table may be big (e.g. 2M rows) and the first result that passes the filter may be at the end of the table. So in order to get to that data, topK needs to make many load requests (more than our artificial limits). So it will hit the limits first and the user will think something's wrong with their app, which isn't the case.

To avoid these kinds of situations i would make it opt-in. Only if you encounter this bug that we are currently unable to reproduce would you enable these limits in an attempt to get tracking information that helps us identifying the root cause.

**I'm also worried about this debugging code remaining in the codebase forever.** I'd rather ask the users that are currently facing this bug to use the preview package from this PR to help find the root cause, rather than merging this into our main codebase. If we do that, then no need to make this opt-in or enabled only in debug mode.

packages/db/src/query/live/collection-subscriber.ts
```diff
@@ -320,7 +321,11 @@ export class CollectionSubscriber<
       return
     }
 
-    const trackedChanges = this.trackSentValues(changes, orderByInfo.comparator)
+    const changesArray = Array.isArray(changes) ? changes : [...changes]
```
samwillis: Also same as above, trackSentValues accepts an iterator, this is redundant.

packages/db/src/query/live/collection-subscriber.ts
```diff
@@ -342,10 +347,14 @@ export class CollectionSubscriber<
 
   // Loads the next `n` items from the collection
   // starting from the biggest item it has sent
-  private loadNextItems(n: number, subscription: CollectionSubscription) {
+  // Returns true if local data was found, false if the local index is exhausted
+  private loadNextItems(
+    n: number,
+    subscription: CollectionSubscription,
+  ): boolean {
```
samwillis: This returning a boolean doesn't seem to be used anywhere, can we keep the same signature for now?

packages/db/src/collection/subscription.ts
```diff
@@ -410,13 +411,15 @@ export class CollectionSubscription
...
+   *
+   * @returns true if local data was found and sent, false if the local index was exhausted
    */
   requestLimitedSnapshot({
     orderBy,
     limit,
     minValues,
     offset,
-  }: RequestLimitedSnapshotOptions) {
+  }: RequestLimitedSnapshotOptions): boolean {
```
samwillis, line 425: This return values is passed through by `loadNextItems` but then not used anywhere.

packages/db/src/collection/subscription.ts
```diff
@@ -556,20 +588,20 @@ export class CollectionSubscription
...
+        if (cursorMinValue instanceof Date) {
+          const minValuePlus1ms = new Date(cursorMinValue.getTime() + 1)
           whereCurrentCursor = and(
-            gte(expression, new Value(minValue)),
+            gte(expression, new Value(cursorMinValue)),
             lt(expression, new Value(minValuePlus1ms)),
           )
         } else {
-          whereCurrentCursor = eq(expression, new Value(minValue))
+          whereCurrentCursor = eq(expression, new Value(cursorMinValue))
```
samwillis, line 612: Renames for no reasion

packages/db/src/collection/subscription.ts
```diff
@@ -502,8 +505,38 @@ export class CollectionSubscription
...
+    type SnapshotState = { valuesNeeded: number; keysInBatch: number }
+    const tracker = createIterationTracker<SnapshotState>(
+      MAX_SNAPSHOT_ITERATIONS,
+      (state) =>
+        `valuesNeeded=${state.valuesNeeded}, keysInBatch=${state.keysInBatch}`,
+    )
+    let hitIterationLimit = false
+
     while (valuesNeeded() > 0 && !collectionExhausted()) {
-      const insertedKeys = new Set<string | number>() // Track keys we add to `changes` in this iteration
```
samwillis, line 506: It's interesting that it's spotted this is unused and removed it.

@kevin-dp insertedKeys was introduced in one of your PRs, can you remember what it was for?

I am wandering a little bit if this could be a bug, if it was intended to be used to track sent data, but we are not doing so?

packages/db/src/query/live/collection-subscriber.ts
kevin-dp: Normally the boolean indicates whether it's done loading more items or not. IIRC if we hit the end of the data then we return true indicating we're done loading items (even though the topK may not be completely filled). This is essential to know that we need to stop requesting for more data.

I'm confused about the comment it added about loading this locally. This could be loaded from the backend as well.

packages/db/src/collection/subscription.ts
kevin-dp: I traced back the original PR that introduced this variable. Keys were added to this set but the set was never used. I think it must have been remnants of a previous approach i had tried to fix the bug and then i ended up switching to a different approach in this variable went unnoticed. (This is the PR that introduced it: https://github.com/TanStack/db/pull/713)

PR #1175, opened 2026-01-22
Improve awaitTxId timeout error with diagnostic info

KyleAMathews:
## Summary

Enhanced `TimeoutWaitingForTxIdError` to provide actionable debugging information when `awaitTxId` times out, showing which txids arrived during the timeout and linking to documentation about the common cause.

**User impact**: Developers debugging timeout issues can now see whether transactions are being received at all, making it much easier to diagnose whether the issue is network-related or a `pg_current_xact_id()` placement problem.

---

## Root Cause (of the DX problem)

When `awaitTxId` times out, users previously got a generic "Timeout waiting for txId: X" message with no context about what went wrong. The most common cause—calling `pg_current_xact_id()` outside the transaction that performs the mutation—was not surfaced.

## Approach

1. **Track txids during timeout**: Capture a snapshot of seen txids at the start of the wait, then record any new txids that arrive during the timeout period
2. **Enhanced error message**: Include the list of received txids (or "no txids received") plus a hint about the common cause with a link to documentation

**Key code path** (`electric.ts:639-665`):
```typescript
const txidsAtStart = new Set(seenTxids.state)
const txidsReceivedDuringTimeout: Array<Txid> = []

// In timeout handler:
reject(new TimeoutWaitingForTxIdError(txId, config.id, txidsReceivedDuringTimeout))

// In subscription:
for (const seenTxid of seenTxids.state) {
  if (!txidsAtStart.has(seenTxid) && !txidsReceivedDuringTimeout.includes(seenTxid)) {
    txidsReceivedDuringTimeout.push(seenTxid)
  }
}
```

## Key Invariants

- The `receivedTxids` parameter is optional for backward compatibility
- Tracking only starts when entering the Promise (not on immediate resolution)
- Duplicate txids are not added to the tracking array

## Non-goals

- Not changing timeout duration or retry behavior
- Not adding automatic retry logic

---

## Verification

```bash
pnpm test
```

All 404 tests pass.

## Files Changed

| File | Change |
|------|--------|
| `packages/electric-db-collection/src/electric.ts` | Track txids received during timeout, pass to error |
| `packages/electric-db-collection/src/errors.ts` | Add `receivedTxids` param, format debugging info, add hint with docs link |

---

## Checklist

- [x] Tested locally with `pnpm test`
- [x] Generated changeset

### DIFF
--- a/.changeset/improve-txid-timeout-error.md
+++ b/.changeset/improve-txid-timeout-error.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/electric-db-collection': patch
+---
+
+Improve awaitTxId timeout error with debugging info showing which txids were received during the timeout period, plus a hint about the common cause (calling pg_current_xact_id outside the mutation transaction).
--- a/packages/electric-db-collection/src/electric.ts
+++ b/packages/electric-db-collection/src/electric.ts
@@ -637,13 +637,33 @@ export function electricCollectionOptions<T extends Row<unknown>>(
     if (hasSnapshot) return true
 
     return new Promise((resolve, reject) => {
+      // Track txids at start to know which ones arrived during timeout
+      const txidsAtStart = new Set(seenTxids.state)
+      const txidsReceivedDuringTimeout: Array<Txid> = []
+
       const timeoutId = setTimeout(() => {
         unsubscribeSeenTxids()
         unsubscribeSeenSnapshots()
-        reject(new TimeoutWaitingForTxIdError(txId, config.id))
+        reject(
+          new TimeoutWaitingForTxIdError(
+            txId,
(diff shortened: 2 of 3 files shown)


KyleAMathews:
Closing as not planned. This is a diagnostic-only enhancement rather than a correctness fix, and the implementation targets the older pre-ElectricLifecycle state design. The current timeout plus existing documentation is adequate; adding waiter-local tracking and additional runtime/error machinery is not justified without evidence of a recurring support problem.\n\nIf a concrete current-version debugging case shows that the existing timeout information is insufficient, we can reconsider a smaller targeted improvement. Thanks to the original implementation and documentation work captured here.

PR #1183, opened 2026-01-26
Fix: Enable $selected access in orderBy after fn.select

KyleAMathews:
## Summary

Fixed `$selected` namespace availability in `orderBy()`, `having()`, and `fn.having()` clauses when using functional select (`fn.select()`). Users can now order and filter by computed/derived fields from functional selects.

## Root Cause

The query builder only checked for `this.query.select` when deciding whether to expose the `$selected` namespace in `orderBy` and `having` callbacks. It did not account for `this.query.fnSelect` (functional select), so users couldn't reference `$selected` fields after using `fn.select()`.

## Approach

Added `|| this.query.fnSelect` to the condition in both `orderBy()` and `having()` methods:

```typescript
const refProxy = (
  this.query.select || this.query.fnSelect
    ? createRefProxyWithSelected(aliases)
    : createRefProxy(aliases)
) as RefsForContext<TContext>
```

## Key Invariants

- `$selected` is available in `orderBy`, `having`, and `fn.having` when either `.select()` or `fn.select()` has been called
- Table refs remain accessible alongside `$selected` (both namespaces work together)
- Ordering/filtering by original table fields still works after `fn.select()`

## Limitations

- `fn.select` + `groupBy` combination is not fully supported (returns only group keys, not computed fields)
- For aggregates with groupBy, continue using regular `.select()` with `.having()` or `.fn.having()`

## Verification

```bash
pnpm test:pr
pnpm vitest run packages/db/tests/query/functional-variants.test.ts
```

## Files Changed

| File | Change |
|------|--------|
| `packages/db/src/query/builder/index.ts` | Added `fnSelect` check to `orderBy` and `having` methods |
| `packages/db/tests/query/functional-variants.test.ts` | Added 7 new runtime tests for `fn.select` with `orderBy` and `fn.having` |
| `packages/db/tests/query/functional-variants.test-d.ts` | Added 2 type tests for `$selected` type inference |

---
https://claude.ai/code/session_01EPRCG2K8348FMNWzjPiacC

### DIFF
--- a/packages/db/tests/query/functional-variants.test.ts
+++ b/packages/db/tests/query/functional-variants.test.ts
@@ -653,7 +653,11 @@ describe(`Functional Variants Query`, () => {
 
       // Only users with salary > 60k: Alice (75k), Charlie (85k), Dave (65k)
       expect(results).toHaveLength(3)
-      expect(results.map((r) => r.name).sort()).toEqual([`Alice`, `Charlie`, `Dave`])
+      expect(results.map((r) => r.name).sort()).toEqual([
+        `Alice`,
+        `Charlie`,
+        `Dave`,
+      ])
     })
 
     test(`should allow orderBy with both table refs and $selected`, () => {


kevin-dp:
@KyleAMathews The `$selected` namespace should also be available on `having` and `fn.having` clauses. You can point Claude to the original PR https://github.com/TanStack/db/pull/1094 such that it has the necessary context to implement this fix also for `having` and `fn.having` (together with the necessary unit tests). It can copy paste the unit tests from the original PR and replace `select` by `fn.select`.

KyleAMathews:
Thanks for the feedback @kevin-dp!

I've added `$selected` support for `having` and `fn.having` when using `fn.select`. The fix follows the same pattern as `orderBy` - checking for `fnSelect` when deciding whether to create the ref proxy with `$selected`.

**What works now:**
- `fn.select` → `orderBy` with `$selected` ✅
- `fn.select` → `fn.having` with `$selected` (no groupBy) ✅
- `fn.select` → `having` with `$selected` (needs groupBy for regular `having`)

**Limitation found:**
When I tried to add tests that combine `fn.select` + `groupBy` + `having`/`fn.having` (as you suggested - replacing `select` with `fn.select` in the existing tests), I discovered that `fn.select` + `groupBy` doesn't work correctly. The query returns only `{ __key_0: ... }` instead of the computed fields.

This happens because:
1. `fn.select` runs before `groupBy`, creating `$selected` per-row
2. `processGroupBy()` doesn't know about `fnSelect` (only receives `selectClause`)
3. Without a `selectClause`, it falls back to returning only the group keys

Supporting `fn.select` + `groupBy` would require changes to the query compiler to either:
- Pass `fnSelect` to `processGroupBy()` and preserve the first row's `$selected` values
- Or extract aggregate expressions from `fnSelect` (not possible since it's a JS function, not an expression tree)

For now, users should continue using regular `.select()` for aggregates with groupBy, and can use `fn.select` → `fn.having` for non-grouped filtering use cases.

Let me know if you'd like me to pursue the `fn.select` + `groupBy` feature as a follow-up!

KyleAMathews:
Posted issue to https://github.com/TanStack/db/issues/1189

Issue #1186, opened 2026-01-27
Infinite loop in `BTreeIndex.takeInternal` when using orderBy and limit with `undefined` indexed values


1bod:
- [x] I've validated the bug against the latest version of DB packages

**Describe the bug**
When a collection contains items where an indexed field (used in `orderBy`) evaluates to `undefined`, and the query uses both `.orderBy()` and `.limit()`, the `BTreeIndex.takeInternal` method enters an infinite loop. This happens because `nextPair(undefined)` returns `[undefined, undefined]` (an array) instead of `undefined`, so the while loop condition `pair !== undefined` is always true.
**To Reproduce**
https://stackblitz.com/edit/vitejs-vite-xcu9tcef?file=src%2FApp.vue

```vue
<script setup lang="ts">
import {
  createCollection,
  eq,
  useLiveQuery,
  localStorageCollectionOptions,
} from '@tanstack/vue-db';

const collection = createCollection<{
  id: number;
  assignedTo?: { id: number };
}>(
  localStorageCollectionOptions({
    id: 'tasks',
    storageKey: 'app-tasks',
    getKey: (item) => item.id,
  })
);

// Add an item with undefined assignedTo
collection.insert({ id: 1, assignedTo: undefined });

// Query with a filter on assignedTo.id
const { data } = useLiveQuery((q) =>
  q
    .from({ item: collection })
    .where(({ item }) => eq(item.assignedTo.id, 35))
    .orderBy(({ item }) => item.taskDate, 'desc')
    .limit(1)
);

// Insert another item - this triggers the infinite loop
collection.insert({ id: 2, assignedTo: { id: 35 } });

</script>

<template>
  <div>
    {{ data }}
  </div>
</template>

```

Debug information:

When pausing in the debugger, execution is stuck in `btree-index.ts` at `takeInternal`:

Variables at freeze:

`pair` = `[undefined, undefined]`
`key` = `undefined`
`keys` = Set with one empty string
`result` = `[]`
`n` = 1

**Expected behavior**
The `takeInternal` method should handle `undefined` keys gracefully.

kevin-dp:
This problem is inherited from the B+-tree library we use.

Our `BTreeIndex.take` method relies on the underlying BTree's `nextHigherPair` method to start iterating from that key (exclusive). But that `nextHigherPair` method interprets `undefined` as start from the beginning which means it doesn't support starting from a key that is `undefined`:
```ts
[...]

PR #1191, opened 2026-01-27
test: Add regression tests for BTreeIndex infinite loop with undefined keys

KyleAMathews:
## 🎯 Changes

This PR adds comprehensive regression tests for GitHub issue #1186, which describes an infinite loop in `BTreeIndex.takeInternal` when using `orderBy()` and `limit()` on collections containing items with undefined indexed values.

The test suite includes:
- Basic tests for `take()` and `takeReversed()` with undefined indexed values
- Tests mixing undefined and defined indexed values
- Tests with multiple items sharing the same undefined key
- Full end-to-end tests reproducing the exact scenario from issue #1186 using live queries with `orderBy` and `limit`
- Tests combining `where`, `orderBy`, and `limit` with undefined values

All tests include a 5-second timeout to catch infinite loops during test execution.

**Root Cause (documented in test comments):**
When `nextHigherPair(undefined)` is called, it returns the minimum pair from the B-tree. If that minimum key is itself `undefined`, the method returns `[undefined, undefined]`. After processing, the code sets `key = pair[0]` (which is `undefined`) and calls `nextHigherPair(undefined)` again, returning the same pair and creating an infinite loop.

## ✅ Checklist

## 🚀 Release Impact

**Note:** This is a test-only addition that documents the bug for future reference and validation once the underlying issue is fixed.

https://claude.ai/code/session_01RKBKXMoKVe1hSXGy3VVEFo


KyleAMathews:
Closing as superseded by merged PR #1198. The merged runtime fix includes a focused undefined-key BTree regression file and end-to-end ordering coverage, so this earlier standalone test PR is redundant.

PR #1192, opened 2026-01-27
fix(db): track loadSubset promise for on-demand live queries

KyleAMathews:
## Summary

Fixes `isReady` tracking for on-demand live queries. This PR addresses two related issues:

1. **Non-ordered queries**: Were incorrectly marked as ready before data finished loading
2. **Ordered queries with paging**: Could hang indefinitely when multiple live queries share a source collection or when paging with `setWindow`

Also fixes `preload()` promises hanging when cleanup occurs before the collection becomes ready.

### Issue 1: Non-ordered queries (original fix)

For non-ordered live queries, `subscribeChanges` is called with `includeInitialState: true`, which internally calls `requestSnapshot({ trackLoadSubsetPromise: false })`. This disables the subscription's status tracking—the subscription never transitions to `loadingSubset` status, so the live query's `isReady` stayed true during loading.

### Issue 2: Ordered paging hang (this session)

`requestLimitedSnapshot` used `limitedSnapshotRowCount` to compute offset for offset-based backends. That counter only advanced when the **local index** produced rows. When `loadSubset` resolved asynchronously (cached query observer) without local rows at request time, the counter stayed at 0. Subsequent `loadNextItems` calls kept requesting offset 0, re-writing the same rows and scheduling `loadMore` again, leading to an infinite loop.

**History**: Introduced in commit b3b194000 (2025-12-12) when limitedSnapshotRowCount/offset pagination was added; the counter never advanced for async loadSubset results. The hang has been possible since then in on-demand ordered/limited queries (especially with multiple live queries or window moves).

### For non-ordered queries

The fix couldn't simply change `trackLoadSubsetPromise` to `true` because that breaks truncate handling (must-refetch scenarios). The truncate buffering logic depends on the existing status behavior.

Instead, the fix tracks loading at the collection level (`collection-subscriber.ts`): After subscribing, checks if the source collection's `isLoadingSubset` changed. If so, listens for the `loadingSubset:change` event and tracks a promise on the live query collection for `isReady`.

### For ordered paging

1. **Keep offset in sync** (`subscription.ts`): Sync `limitedSnapshotRowCount` with `sentKeys.size` after callbacks—advances the offset even when async loadSubset results arrive without local index rows

2. **Deduplicate by request parameters** (`collection-subscriber.ts`): Track `lastLoadRequestKey` (serialized cursor + window) to prevent repeated identical requests while still allowing window moves (different offset/limit)

3. **Track seen keys** (`collection-subscriber.ts`): Use `seenKeys` to detect genuinely new rows (as opposed to repeated updates for the same keys) and reset the dedup flag when new keys arrive

4. **Pass orderBy/limit to snapshot requests**: Ensures each live query with different parameters gets its own cache entry in on-demand source collections

### Cleanup handling

- Calls `onFirstReady` callbacks during cleanup (`lifecycle.ts`) so `preload()` promises resolve instead of hanging
- Cleans up event listeners on early unsubscribe to prevent memory leaks

## Key Invariants

1. `isReady` must return `false` while `loadSubset` is in progress
2. `preload()` must always resolve (either when ready OR during cleanup)
3. Truncate handling must continue to work correctly (buffering events during must-refetch)
4. Event listeners must be cleaned up on early unsubscribe to prevent memory leaks
5. Multiple live queries sharing a source collection must independently track their loading state
6. Window moves (`setWindow`) must trigger new data fetches even if cursor hasn't advanced

## Regression Tests Added

- `live-query-collection`: advances offset after async loadSubset when initial window is empty
- `live-query-collection`: window moves across identical orderBy values request new offsets
- `e2e pagination`: multiple live queries with distinct windows + paging a second query

## Verification

```bash
pnpm --filter @tanstack/db exec vitest run
pnpm --filter @tanstack/query-db-collection exec vitest run
pnpm --filter @tanstack/query-db-collection exec vitest run --config vitest.e2e.config.ts
```

All tests pass:
- db: 1935 passed (3 skipped)
- query-db-collection: 187 passed
- e2e: 112 passed

## Files Changed

- `packages/db/src/collection/subscription.ts` - Sync offset counter with sentKeys, add onLoadSubsetResult/trackLoadSubsetPromise options
- `packages/db/src/query/live/collection-subscriber.ts` - Deduplicate by request parameters, track seen keys, direct load promise tracking
- `packages/db/src/query/live/collection-config-builder.ts` - Pass orderBy/limit to snapshot requests
- `packages/db/src/collection/lifecycle.ts` - Call onFirstReady callbacks during cleanup

### DIFF
--- a/packages/db/src/query/live/collection-subscriber.ts
+++ b/packages/db/src/query/live/collection-subscriber.ts
@@ -425,12 +425,12 @@ export class CollectionSubscriber<
       : undefined
 
     // Normalize to array format for minValues
-    const minValues =
-      extractedValues !== undefined
-        ? Array.isArray(extractedValues)
-          ? extractedValues
-          : [extractedValues]
-        : undefined
+    let minValues: Array<unknown> | undefined
+    if (extractedValues !== undefined) {
+      minValues = Array.isArray(extractedValues)
+        ? extractedValues
+        : [extractedValues]
+    }
 
     const loadRequestKey = this.getLoadRequestKey({
       minValues,
--- a/packages/db/src/collection/subscription.ts
+++ b/packages/db/src/collection/subscription.ts
@@ -683,13 +683,9 @@ export class CollectionSubscription
 
     for (const change of changes) {
       if (change.type === `delete`) {
-        // Remove deleted keys from sentKeys so future re-inserts are allowed
         this.sentKeys.delete(change.key)
       } else {
-        // For inserts and updates, track the key as sent
-        if (!this.sentKeys.has(change.key)) {
-          this.sentKeys.add(change.key)
-        }
+        this.sentKeys.add(change.key)
       }
     }
 


chicojasl:
Verified this locally. Working great! 👍

packages/db/src/query/live/collection-subscriber.ts
```diff
@@ -117,14 +138,25 @@ export class CollectionSubscriber<
     // Check current status after subscribing - if status is 'loadingSubset', track it.
     // The onStatusChange listener will catch the transition to 'ready'.
     if (subscription.status === `loadingSubset`) {
-      trackLoadPromise(subscription)
+      if (!this.subscriptionLoadingPromises.has(subscription)) {
```
kevin-dp: This code is duplicated on line 93. It used to call `trackLoadPromise`. We should extract this duplicate code into a function (if `trackLoadPromise` isn't the right one to use).

KyleAMathews: good call — working on adding a helper to make the state changes cleaner/less bug prone

PR #1193, opened 2026-01-28
test: Add regression test for OR queries with mixed indexed/non-indexed fields

KyleAMathews:
## 🎯 Changes

Bug reported in https://discord.com/channels/719702312431386674/1465855364502126662/1465855364502126662

Added a regression test that demonstrates a bug in OR query handling when mixing indexed and non-indexed fields. The test documents the issue where items matching non-indexed field conditions are silently dropped when combined with indexed field conditions in an OR query.

**Bug Description:**
When using `or(eq(element.id, itemId), eq(element.group_id, itemId))` where `id` has an index but `group_id` does not, only items matching the indexed condition are returned. Items matching the non-indexed condition are silently dropped.

**Expected Behavior:**
Items matching EITHER condition should be returned. When not all OR branches can be optimized with indexes, the query should fall back to a full scan to ensure all matching items are found.

**Test Details:**
- Creates a query: `age=25 OR name='Bob'`
- `age` field has an index, `name` field does not
- Expected results: Both Alice (matches indexed condition) and Bob (matches non-indexed condition)
- Current behavior: Only Alice is returned due to the bug

This test will fail until the underlying bug is fixed in the query optimization logic.

## ✅ Checklist

## 🚀 Release Impact

**Note:** This is a test-only change that documents an existing bug. No changeset is needed.

https://claude.ai/code/session_01R16FFtQbQr1VXjEVCYAeUk

### DIFF
--- a/packages/db/tests/collection-auto-index.test.ts
+++ b/packages/db/tests/collection-auto-index.test.ts
@@ -387,6 +387,56 @@ describe(`Collection Auto-Indexing`, () => {
     subscription.unsubscribe()
   })
 
+  it(`should create auto-indexes for fields inside OR expressions`, async () => {
+    // BUG: Auto-indexing only recurses into AND expressions, not OR expressions.
+    // This means fields used inside OR don't get auto-indexed, which combined with
+    // the optimizer bug (dropping non-indexed OR branches), causes silent data loss.
+    //
+    // This test demonstrates what SHOULD happen: auto-indexing should create indexes
+    // for all fields in OR expressions so queries return correct results.
+
+    const autoIndexCollection = createCollection<TestItem, string>({
+      getKey: (item) => item.id,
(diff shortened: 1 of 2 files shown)


th1nkful:
Just adding some extra detail. I tried correcting this by adding `elementsCollection.createIndex((element) => element.group_id);` right after the `createCollection` call, which at first appeared to resolve it. But continuing to work with it, it stopped returning results again. I think this could be related to Vite's HMR.

KyleAMathews:
Superseded by merged #1582, which fixed partial OR index soundness and landed equivalent mixed indexed/unindexed coverage. Current main is GREEN for mixed, all-indexed, and no-index OR paths. The separate suggestion to auto-index every OR branch is a performance policy, not required for correctness.

PR #1204, opened 2026-01-30
fix(react-db): include peek-ahead item in useLiveInfiniteQuery initial query

melvinhagberg:
The initial query was fetching exactly `pageSize` items, but the peek-ahead logic requires `pageSize + 1` to determine if more pages exist. This mismatch caused `hasNextPage` to incorrectly return `false` on initial load when using SQLite predicate push-down with `syncMode: "on-demand"`.

The issue occurred because:
1. Initial query loaded exactly `pageSize` items (e.g., 50)
2. `setWindow({ limit: pageSize + 1 })` tried to access the peek-ahead item
3. The item didn't exist yet, and lazy loading had timing issues
4. `hasNextPage` was incorrectly `false` even when more data existed

The fix aligns the initial query limit with what `setWindow` expects, fetching `pageSize + 1` items upfront. The extra item is only used internally for the `hasNextPage` check - the returned `data` is still sliced to `pageSize` items via the existing useMemo logic.

## ✅ Checklist

- [X] I have followed the steps in the [Contributing guide](https://github.com/TanStack/db/blob/main/CONTRIBUTING.md).
- [X] I have tested this code locally with `pnpm test:pr`.

## 🚀 Release Impact

### DIFF
--- a/.changeset/fix-infinite-query-peek-ahead.md
+++ b/.changeset/fix-infinite-query-peek-ahead.md
@@ -1,0 +1,7 @@
+---
+"@tanstack/react-db": patch
+---
+
+fix(react-db): include peek-ahead item in useLiveInfiniteQuery initial query
+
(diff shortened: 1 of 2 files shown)


KyleAMathews:
Should be fixed by https://github.com/TanStack/db/pull/1209

PR #1213, opened 2026-02-02
Virtual props implementation

samwillis:
## 🎯 Changes

This PR implements the virtual properties system as described in the RFC. This introduces computed, read-only metadata fields prefixed with `$` to every row, providing information about its sync status, origin, key, and source collection ID.

Key aspects of this implementation:
- **`$synced`**: Indicates whether a row reflects confirmed state from the backend (`true`) or has pending optimistic mutations (`false`).
- **`$origin`**: Specifies if the last confirmed change originated from the local client (`'local'`) or was received via sync (`'remote'`).
- **`$key`**: The primary identifier of the row.
- **`$collectionId`**: The ID of the collection the row originated from.
- **Pass-through semantics**: Virtual properties are preserved and passed through in live queries and nested collections using an "add-if-missing" pattern.
- **Queryability**: Virtual properties are available for use in query builder expressions (e.g., `where(({ item }) => eq(item.$synced, true))`).
- **Type Integration**: Updated `RefProxy` and `Ref` types to include virtual properties, ensuring type safety and discoverability.

This feature enhances data visibility and enables more sophisticated UI logic based on the state and origin of data.

## ✅ Checklist

## 🚀 Release Impact

---
<a href="https://cursor.com/background-agent?bcId=bc-763756c0-b317-42fd-bc3e-c085c81437f1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;<a href="https://cursor.com/agents?id=bc-763756c0-b317-42fd-bc3e-c085c81437f1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>

### DIFF
--- a/packages/db/src/query/compiler/group-by.ts
+++ b/packages/db/src/query/compiler/group-by.ts
@@ -233,7 +233,9 @@ export function processGroupBy(
           VIRTUAL_HAS_LOCAL_KEY
         ]
         resultRow.$synced = groupSynced ?? true
-        resultRow.$origin = (groupHasLocal ? `local` : `remote`) satisfies VirtualOrigin
+        resultRow.$origin = (
+          groupHasLocal ? `local` : `remote`
+        ) satisfies VirtualOrigin
         resultRow.$key = resultKey
         resultRow.$collectionId =
           aggregateCollectionId ?? resultRow.$collectionId
@@ -407,7 +409,9 @@ export function processGroupBy(
         VIRTUAL_HAS_LOCAL_KEY
       ]
       resultRow.$synced = groupSynced ?? true
-      resultRow.$origin = (groupHasLocal ? `local` : `remote`) satisfies VirtualOrigin
+      resultRow.$origin = (
+        groupHasLocal ? `local` : `remote`
+      ) satisfies VirtualOrigin
       resultRow.$key = finalKey
       resultRow.$collectionId = aggregateCollectionId ?? resultRow.$collectionId
       if (mainSource && correlationKey !== undefined) {
--- a/packages/db/tests/query/includes.test-d.ts
+++ b/packages/db/tests/query/includes.test-d.ts
@@ -168,12 +168,12 @@ describe(`includes subquery types`, () => {
       expectTypeOf(result.issues.toArray[0]!.$key).toEqualTypeOf<
         string | number
       >()
-      expectTypeOf(result.issues.toArray[0]!.$collectionId).toEqualTypeOf<
-        string
-      >()
-      expectTypeOf(result.issues.toArray[0]!.comments.toArray[0]!).toMatchTypeOf<
-        WithVirtualProps<{ id: number; body: string }>
-      >()
+      expectTypeOf(
+        result.issues.toArray[0]!.$collectionId,
+      ).toEqualTypeOf<string>()
+      expectTypeOf(
+        result.issues.toArray[0]!.comments.toArray[0]!,
+      ).toMatchTypeOf<WithVirtualProps<{ id: number; body: string }>>()
     })
   })
 
--- a/packages/db/tests/query/includes.test.ts
+++ b/packages/db/tests/query/includes.test.ts
@@ -656,9 +656,7 @@ describe(`includes subqueries`, () => {
 
       // Beta has 1 issue; limit(1) keeps it
       const beta = collection.get(2) as any
-      expect(plainRows(beta.issues)).toEqual([
-        { id: 20, title: `Bug in Beta` },
-      ])
+      expect(plainRows(beta.issues)).toEqual([{ id: 20, title: `Bug in Beta` }])
 
       // Gamma has 0 issues; limit(1) still empty
       const gamma = collection.get(3) as any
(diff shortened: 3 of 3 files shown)


packages/db/src/query/compiler/group-by.ts
```diff
@@ -80,11 +117,39 @@ export function processGroupBy(
...
+        return false
+      },
+    },
+  }
+
   // Handle empty GROUP BY (single-group aggregation)
   if (groupByClause.length === 0) {
     // For single-group aggregation, create a single group with all data
-    const aggregates: Record<string, any> = {}
+    const aggregates: Record<string, any> = { ...virtualAggregates }
```
kevin-dp: Do we really need to make a copy?

packages/db/src/query/compiler/group-by.ts
```diff
@@ -200,7 +277,7 @@ export function processGroupBy(
   }
 
   // Create aggregate functions for any aggregated columns in the SELECT clause
-  const aggregates: Record<string, any> = {}
+  const aggregates: Record<string, any> = { ...virtualAggregates }
```
kevin-dp: Same, really needed to copy?

samwillis:
@kevin-dp 

The aggregation concern is already addressed in the current implementation. We derive aggregate-level virtual props intentionally:
- `$synced` is `false` if any row in the group is unsynced
- `$origin` is `local` if any contributing row is `local`
- `$key` is a derived group key rather than a source-row key

On joins, I think the ambiguity is narrower than described. After a join, the virtual props remain attached per alias: `todo.$collectionId`, `list.$collectionId`. The place where ambiguity does exist is flattened or spread select output, where top-level merges can overwrite one source's virtual props with another's, but I would see that as a users concern when projecting in the `select`.

Removed the object copes you flagged.

KyleAMathews:
## Code Review: Virtual Props Implementation

Nice architecture overall — the "add-if-missing" enrichment pattern, WeakMap caching, and aggregate semantics (all-synced / any-local) are well-designed. A few issues worth addressing before merge:

---

### Bug: Truncate doesn't clear origin tracking maps

`packages/db/src/collection/state.ts` ~line 839

The truncate phase in `commitPendingTransactions` clears `syncedData`, `syncedMetadata`, and `syncedKeys`, but **not** `rowOrigins`, `pendingLocalChanges`, or `pendingLocalOrigins`. After truncate, if the server re-inserts the same keys, `getRowOrigin(key)` returns the stale value from `rowOrigins` instead of defaulting to `'remote'`.

The full `cleanup()` method at line 1257 does clear these maps, but truncate-during-sync doesn't.

```typescript
// After line 841 in the truncate block:
this.rowOrigins.clear()
this.pendingLocalChanges.clear()
this.pendingLocalOrigins.clear()
```

---

### Bug: `pendingLocalOrigins` not cleaned up on transaction failure

`packages/db/src/collection/mutations.ts` + `state.ts`

`markPendingLocalOrigins()` is called before `commit()`, but if the transaction fails, the `.catch(() => undefined)` on the commit promise doesn't clean up `pendingLocalOrigins`. The `recomputeOptimisticState` `failed` branch clears `pendingOptimistic*` maps but doesn't touch `pendingLocalOrigins` for non-optimistic mutations. Result: `$origin` stays `'local'` for affected keys indefinitely after a failed mutation.

Similarly, `scheduleTransactionCleanup`'s empty catch block (line 1198) keeps the transaction for reference but never cleans up origin tracking state.

---

### Minor: `previousValue` virtual props in optimistic path use post-change state

`packages/db/src/collection/state.ts` — `enrichChangeMessage` (line 215)

When enriching `previousValue`, it calls `enrichWithVirtualProps(change.previousValue, change.key)` which reads *current* `isRowSynced`/`getRowOrigin` — the post-mutation state. The sync commit path handles this correctly with `getPreviousVirtualProps`, but the optimistic recompute path at line 533 (`collectOptimisticChanges` → `emitEvents` → `enrichChangeMessage`) doesn't capture pre-change virtual props.

Impact is limited to `previousValue.$synced` / `previousValue.$origin` in change events during optimistic transitions.

---

### Cleanup: Dead `VirtualRefProps` type

`packages/db/src/virtual-props.ts:105` — `VirtualRefProps` is exported but never imported anywhere. The actual ref typing is handled by `VirtualPropsRef` in `types.ts` and `VirtualPropsRefProxy` in `ref-proxy.ts`. Should be removed.

---

### Suggestion: Derive types from single source

`WithoutVirtualProps` hardcodes the four property names as string literals rather than using `Omit<T, keyof VirtualRowProps>`. Similarly, `VirtualPropsRef` and `VirtualPropsRefProxy` duplicate the same four properties independently. If a fifth virtual prop is added, three files need manual updates. Consider deriving from `VirtualRowProps`:

```typescript
// WithoutVirtualProps
export type WithoutVirtualProps<T> = Omit<T, keyof VirtualRowProps>

// Ref types
type VirtualRefShape<TKey extends string | number = string | number> = {
  readonly [K in keyof VirtualRowProps<TKey>]: RefLeaf<VirtualRowProps<TKey>[K]>
}
```

---

### Suggestion: `hasVirtualProps` should check all 4 properties

`packages/db/src/virtual-props.ts:161` — Currently only checks for `$synced` and `$origin`, but the type guard asserts `VirtualRowProps` which also requires `$key` and `$collectionId`. Code relying on this guard could access `$key` on an object that doesn't have it.

PR #1214, opened 2026-02-02
fix(db): prevent cross-query data contamination in useLiveInfiniteQuery

melvinhagberg:
When using useLiveInfiniteQuery with syncMode: 'on-demand', navigating between different queries on the same collection causes the wrong data to appear on the first page.

This happens because a new subscription's first snapshot reads from the local index, which may contain data loaded by a different subscription's loadSubset call.

The fix tracks whether loadSubset has been called on a collection. When a new subscription's first snapshot runs:
- If loadSubset has never been called, local data is from initial sync (valid for all queries)
- If loadSubset has been called, local data may be contaminated, so we skip local reads and let loadSubset provide the correct data

## ✅ Checklist

- [X] I have followed the steps in the [Contributing guide](https://github.com/TanStack/db/blob/main/CONTRIBUTING.md).
- [X] I have tested this code locally with `pnpm test:pr`.

## 🚀 Release Impact

- [X] This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).

### DIFF
--- a/.changeset/fix-cross-query-contamination.md
+++ b/.changeset/fix-cross-query-contamination.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/db': patch
+---
+
+Fix cross-query data contamination in `useLiveInfiniteQuery` when navigating between different queries on the same collection with `syncMode: 'on-demand'`. The first page now correctly loads from `loadSubset` instead of potentially stale local index data.
--- a/packages/db/src/collection/subscription.ts
+++ b/packages/db/src/collection/subscription.ts
@@ -476,7 +476,20 @@ export class CollectionSubscription
     // For multi-column orderBy, we use the first column value for index operations (wide bounds)
     // This may load some duplicates but ensures we never miss any rows.
     let keys: Array<string | number> = []
-    if (hasMinValue) {
+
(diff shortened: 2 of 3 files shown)


samwillis:
Hey @melvinhagberg 

Just to check the intention of this fix. Am I correct that what you were experiencing was that the page would briefly show *valid* data for the query, but data that would not be the first page, and appear incorrect? So for example, after changing the orderBy, it would show the data that had already loaded (as its in the on-demand collection) until the "true" first page had loaded from the server?

I'm a little hesitant of the coarseness of this change (it's scoped to the source collection, and flicks on after the first loadSubset), and that it defaults to enabled.

I think we have a few options:

1. Something like this, but make it scoped to the useLiveInfiniteQuery call site. Essentially ensuring that it always waits for the promise from the first `loadSubset` to return before we loading state changes. We essentially then *always* go to the server.
2. Do 1, but make it opt in.
3. Make a much more intelligent tracking of the queries and determine if the first page is already covered by what we have loaded in the collection. I think we have some of the part need for this with the "loadSnapshot Deduper", but it's a more complex solve.

PR #1221, opened 2026-02-07
feat(db): implement createEffect reactive effects API

samwillis:
## Summary

- Implements `createEffect` and `useLiveQueryEffect` as specified in the reactive effects RFC
- `createEffect` attaches an effect handler to a query's delta stream, firing `enter`, `exit`, and `update` events as rows match/unmatch/change within the query result
- Builds a D2 differential dataflow pipeline directly (no intermediate collection materialisation), with full support for `where`, `join`, `select`, `orderBy` + `limit` queries
- Extracts shared utilities (`buildQueryFromConfig`, `extractCollectionAliases`, `sendChangesToInput`, `splitUpdates`, etc.) from `CollectionConfigBuilder` and `CollectionSubscriber` into `src/query/live/utils.ts` to reduce duplication
- Integrates with the transaction-scoped scheduler to coalesce multiple source changes within a single transaction into one graph run
- Supports `skipInitial`, `onError`, `batchHandler`, `handler`, and async `dispose()` with `AbortSignal` for cancelling in-flight work
- Auto-disposes when source collections enter error or cleaned-up state
- Adds `useLiveQueryEffect` React hook with proper mount/unmount lifecycle management

## Key design decisions

- **Direct D2 pipeline** rather than wrapping `createLiveQueryCollection` — avoids materialising results into a collection when only deltas are needed
- **Shared utilities** extracted to enable a future refactor where `CollectionConfigBuilder` could wrap `createEffect` to further reduce duplication
- **Source error handling** — listens for `status:change` on source collections and auto-disposes on error/cleanup (gap identified during review vs `CollectionConfigBuilder`)

## Test plan

- [x] 37 unit tests covering:
  - Basic delta events (enter, exit, update)
  - Filtered queries (where clause matching/unmatching)
  - `on` parameter variations (single type, array, `delta`)
  - `batchHandler` batching semantics
  - `skipInitial` suppression of initial data
  - Error handling (sync, async, console fallback)
  - Disposal (event suppression, in-flight await, AbortSignal, idempotency)
  - Auto-generated and custom IDs
  - `QueryBuilder` instance input
  - Join queries with multiple source collections
  - Select projections
  - Transaction coalescing (single batch) vs immediate execution (separate batches)
  - Truncate handling (re-insert, full clear, post-truncate data)
  - `orderBy` + `limit` (top-K window, insert displacement, delete backfill, desc, in-window updates)
  - Source error handling (auto-dispose on cleanup)
- [x] 3 React hook tests for `useLiveQueryEffect` (lifecycle, dep recreation, event reception)

### DIFF
--- a/packages/react-db/src/useLiveQueryEffect.ts
+++ b/packages/react-db/src/useLiveQueryEffect.ts
@@ -31,11 +31,11 @@ export function useLiveQueryEffect<
   TRow extends object = Record<string, unknown>,
   TKey extends string | number = string | number,
 >(config: EffectConfig<TRow, TKey>, deps: React.DependencyList = []): void {
-  const configRef = useRef(config)
+  const configRef = useRef<EffectConfig<TRow, TKey>>(config)
   configRef.current = config
 
   useEffect(() => {
-    const effect: Effect = createEffect({
+    const effect: Effect = createEffect<TRow, TKey>({
       id: config.id,
       query: config.query,
       skipInitial: config.skipInitial,


docs/guides/live-queries.md
```diff
@@ -1687,6 +1688,370 @@ const users = createLiveQueryCollection((q) =>
...
+
+```ts
+import { createEffect, eq } from '@tanstack/db'
+
+const effect = createEffect({
+  query: (q) =>
+    q
+      .from({ msg: messagesCollection })
+      .where(({ msg }) => eq(msg.role, 'user')),
+  on: 'enter',
```
kevin-dp: The current API of `createEffect` only allows you to register a single enter/update/exit handler. Sometimes we may want to register several ones. This would require storing the query in a variable and creating multiple effects. What if we could do that in a single effect definition?

```ts
createEffect({
  query: ...,
  onEnter: async (event) => ...,
  onUpdate: async (event) => ...,
  onDelete: async (event) => ...
})
```
I also like that now you don't have to provide `on` and `handler` props but directly the handler: `onEnter`/`onUpdate`/`onExit`.

docs/guides/live-queries.md
```diff
@@ -1687,6 +1688,370 @@ const users = createLiveQueryCollection((q) =>
...
+  handler: async (event) => {
+    console.log('New user message:', event.value)
+    await generateResponse(event.value)
+  },
+})
+
+// Later: stop the effect
+await effect.dispose()
```
kevin-dp: How come `dispose()` is async?

docs/guides/live-queries.md
kevin-dp: Another design option for the API that is more consistent with the collections API:
```ts
createEffect({
  query: (q) => ...
}).subscribeChanges(event => {
  const deltaType = event.tpe
  ...
})
```

samwillis: I really like:
```ts
createEffect({
  query: ...,
  onEnter: async (event) => ...,
  onUpdate: async (event) => ...,
  onDelete: async (event) => ...
})
```

I think I change it to that.

samwillis:
Implemented the API reshape @kevin-dp proposed in the comments.

This removes the old `on` + `handler` / `batchHandler` config shape and replaces it with named callbacks on `createEffect`:

- `onEnter`
- `onUpdate`
- `onExit`
- `onBatch`

The goal is to make the common case read more clearly while still preserving batch-oriented handling for transaction / graph-run coalescing.

A few details:
- `onEnter`, `onUpdate`, and `onExit` are per-event callbacks.
- `onBatch` receives the full unfiltered `DeltaEvent[]` for a graph run.
- Semantics are unchanged: these are still query-result deltas, so `exit` means "left the query result", not necessarily "source row was deleted".
- Existing disposal, abort-signal, error handling, and batch coalescing behavior are preserved.

I also updated the effect tests, the React wrapper/tests, and the docs to match the new API.

packages/db/src/query/effect.ts
```diff
@@ -982,17 +986,18 @@ class EffectPipelineRunner<TRow extends object, TKey extends string | number> {
...
-  if (on === `delta`) {
-    return new Set<DeltaType>([`enter`, `exit`, `update`])
-  }
-  if (Array.isArray(on)) {
-    return new Set<DeltaType>(on)
+function getHandlerForEvent<TRow extends object, TKey extends string | number>(
+  event: DeltaEvent<TRow, TKey>,
+  config: EffectConfig<TRow, TKey>,
+): EffectEventHandler<TRow, TKey> | undefined {
+  switch (event.type) {
```
kevin-dp, line 1015: nit: this switch is in fact a dictionary. Could just store it in a map from event type to handler and then do a lookup in the map.

PR #1225, opened 2026-02-09
feat(db): add type-level queryable field constraints

ajoslin:
Hey, love Tanstack! here's a quick PR allowing separation of queryable/sortable types from returned types.

## Summary

Problem: TanstackDB live query callbacks expose all model fields even when I want only allow a subset for filtering/sorting.

I return a lot of fields to the client from my API, using a query-backed collection with syncMode on demand. However, I don't want the client to be able to filter/sort by ALL of those fields, since not all of them can be efficiently queried on the backend.

The solution: add optional type-level queryable constraints so `where`/`orderBy` refs can be restricted while result rows remain full objects.

Non-invasive: additive typing only (`querySchema` + optional `queryable`/helper), defaults preserved, no runtime or planner behavior changes.

## Example

```ts
type EngagementDocument = { // Returned from API, fully readable
  _id: string
  status: `ACTIVE` | `INACTIVE`
  user_id: string
  employer_id: string
  created_at: string
  updated_at: string
  platform_fee_percent: number
}
```
```ts
   createLiveQueryCollection({
      queryable: {
        filterable: [ // platform_fee_percent is NOT queryable!
          `_id`,
          `status`,
          `user_id`,
          `employer_id`,
          `created_at`,
          `updated_at`,
        ] as const,
        sortable: [`created_at`, `updated_at`, `status`] as const,
      },
      query: (q) =>
        q
          .from({ engagement: engagementsCollection })
          .where(({ engagement }) => {
            // @ts-expect-error platform_fee_percent is not queryable
            return eq(engagement.platform_fee_percent, 0.15)
          }),
    })
```

### DIFF
--- a/packages/db/src/query/builder/index.ts
+++ b/packages/db/src/query/builder/index.ts
@@ -42,6 +42,9 @@ import type {
   MergeContextWithJoinType,
   OrderByCallback,
   OrderByOptions,
+  QueryRefsForContext,
+  QuerySchemaFromSource,
+  QueryWhereCallback,
   RefsForContext,
   ResultTypeFromSelect,
   SchemaFromSource,
@@ -138,6 +141,7 @@ export class BaseQueryBuilder<TContext extends Context = Context> {
   ): QueryBuilder<{
     baseSchema: SchemaFromSource<TSource>
     schema: SchemaFromSource<TSource>
+    querySchema: QuerySchemaFromSource<TSource>
     fromSourceName: keyof TSource & string
     hasJoins: false
   }> {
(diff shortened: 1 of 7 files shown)


samwillis:
Hey @ajoslin 

This looks interesting and I can see the real need for it.

Currently you are defining the queryable fields on the liveQueryCollection, not on the source collection. Would it be better to do it on the source collections so that they are not repeated?

KyleAMathews:
Thanks for documenting the use case and building the proposal. We are closing this PR for the pre-1.0 cleanup because separating returned fields from filterable/sortable capabilities needs a public provider/planner capability contract, not only compile-time restrictions. The current type-only design cannot guarantee that runtime pushdown and backend support match the declared fields. We can revisit the use case through a focused capability-design issue if it becomes a priority.

PR #1229, opened 2026-02-10
fix: normalize keys before join operation

hugiex:
## Description
To fix the issue https://github.com/TanStack/db/issues/934 

### Problem
- join conditions like `eq(left.date, right.date)` were using raw Date object as join keys
- the join index uses Map key identify => the diff Date instance with the same timestamp did not match

### Solution
- normalize join keys (Date -> timestamp)
- also tests created to cover the initial matching and live updates when timestamps changes

### DIFF
--- a/.changeset/fix-join-date-eq.md
+++ b/.changeset/fix-join-date-eq.md
@@ -2,4 +2,4 @@
 '@tanstack/db': patch
 ---
 
-Fix `eq()` with Date objects in join conditions by normalizing join keys via `normalizeValue` (#934)
+Fix `eq()` with Date objects in join conditions and `inArray()` with Date values in WHERE clauses by normalizing values via `normalizeValue` (#934)


kevin-dp:
Thanks for the fix! I pushed an additional commit to also normalize the value in the `in` evaluator (`evaluators.ts`).

**Why this is needed:** The join compiler has a lazy-loading optimization where the active side's join keys are passed into `inArray(ref, joinKeys)` to selectively load matching rows from the lazy side (instead of loading the entire collection). After your fix, those `joinKeys` are normalized (e.g. `Date` → timestamp number), but the `in` evaluator was comparing them against the raw field values using `Array.includes()` (strict equality). So `[1234567890].includes(new Date(1234567890))` would return `false`, causing the optimization to silently fall back to loading the full collection.

The fix normalizes the value in the `in` evaluator, consistent with how `eq` already normalizes both operands.

kevin-dp:
While reviewing, I found an additional related bug: `inArray` in WHERE clauses also doesn't work with Date objects. Reproduced in this test: https://github.com/TanStack/db/actions/runs/21949227368/job/63395185329

For example, `.where(({ item }) => inArray(item.createdAt, [new Date('2025-01-01')]))` returns no results because the `in` evaluator uses `Array.includes()` which relies on strict reference equality (`===`), so different `Date` instances with the same timestamp never match.

This is the same class of bug as the original join issue — just in a different code path. The fix (commit 3a23b306) normalizes both the value and the array elements in the `in` evaluator:

```ts
// Before
return array.includes(value)

// After
return array.some((item) => normalizeValue(item) === value)
```

This makes `in` consistent with `eq`, which already normalizes both operands. A test was added in the previous commit (f57873c7) to cover this case.

Issue #865, opened 2025-11-20
Persistence of synced data


samwillis:
A common request, and something we want to develop in time, is a persistence layer for TanStack DB. This would allow users to persist synced data onto the user’s device, providing faster startup and potentially offline support.

This issue is intended as an opening to the discussion: laying out the problem space and a few initial ideas. We will follow up with a more detailed plan once we’ve done some research and experiments.

*See also the older issue for the wider "offline support" #82*

## Requirements

1. A pluggable persistence API that allows multiple alternative backends, such as IndexedDB or OPFS in the browser, or SQLite in a React Native app. OPFS + SQLite-WASM is now a real option for high-performance browser persistence, but comes with a different set of constraints to IDB.
2. Support for larger-than-memory data, with paging in/out of the persistence layer depending on the queries being executed.
3. In the browser we need to safely handle multiple tabs/windows syncing to the same persistence layer.
4. Access to the persisted data should be possible while the app is offline.
5. Schema evolution - persisted stores need a version number and upgrade path. We should assume schemas (and internal encodings) will evolve over time and provide hooks to migrate, or to wipe/resync safely when migration isn’t possible.

## Relationship to offline transactions

We already have an early offline mutation persister in `@tanstack/offline-transactions`:
[https://github.com/TanStack/db/tree/main/packages/offline-transactions](https://github.com/TanStack/db/tree/main/packages/offline-transactions)

This is intentionally a separate problem to “persist synced data”. That said, once we do have persisted collections, queries will need to compose:

* state loaded from the persisted store **plus**
* any pending offline mutations from the offline transaction log

So persistence can’t assume “disk == canonical local view”. The in-memory view is: **persisted base + pending local delta**, and that delta may exist even if the main persistence layer is not enabled for a given collection. This distinction is important to keep in mind in the design.

## Issues with larger-than-memory data and paging

A core motivation for persistence is not only to provide offline support, but also to enable very large datasets that are persisted to a non-memory-backed store. We would then need to support paging in/out of the persistence layer depending on the queries being executed.

There is significant complexity in implementing this, and it needs to be carefully considered in the initial design. We likely need some form of indexing that enables fetching rows or pages that are needed. There are two strategies we need to consider:

1. **Row-level paging.**
   We can use the facilities of the underlying persistence layer to handle paging. Both IndexedDB and SQLite can support row-level filtering and pagination. We can use this to only load the rows that are needed.

[...]

PR #1230, opened 2026-02-10
WIP - Persistence

samwillis:
## 🎯 Changes

Implements https://github.com/TanStack/db/issues/865#issuecomment-3873495126 from #865

## ✅ Checklist

## 🚀 Release Impact

---
<p><a href="https://cursor.com/background-agent?bcId=bc-1751b2c1-b92e-480e-8b95-270f619234d9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;<a href="https://cursor.com/agents?id=bc-1751b2c1-b92e-480e-8b95-270f619234d9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a></p>

### DIFF
--- a/packages/db-electron-sqlite-persisted-collection/src/renderer.ts
+++ b/packages/db-electron-sqlite-persisted-collection/src/renderer.ts
@@ -251,12 +251,7 @@ function createResolvedRendererAdapter<
       latestSeq: number
       latestRowVersion: number
     }> => {
-      return executeRequest(
-        `getStreamPosition`,
-        collectionId,
-        {},
-        resolution,
-      )
+      return executeRequest(`getStreamPosition`, collectionId, {}, resolution)
     },
   }
 }


solarsoft0:
, not sure if this is the right place to share this, but I was able to get it working on my end.
but In local-only persisted mode, the stream position resets after restart, causing valid mutations to be skipped as duplicates and leading to silent data loss. I implemented a temporary workaround, and aside from this issue, I haven’t noticed any other problems so far. thank you @samwillis.

samwillis:
@solarsoft0 thats awesome!
I'm hoping to get back to this next week (or maybe the week after). Very happy to hear it's so close. 
Which platform/storage layer did you use?

solarsoft0:
> @solarsoft0 thats awesome! I'm hoping to get back to this next week (or maybe the week after). Very happy to hear it's so close. Which platform/storage layer did you use?

, I created one for Tauri, and I think my fix confirms it is not a storage layer issue.

kevin-dp:
> , not sure if this is the right place to share this, but I was able to get it working on my end. but In local-only persisted mode, the stream position resets after restart, causing valid mutations to be skipped as duplicates and leading to silent data loss. I implemented a temporary workaround, and aside from this issue, I haven’t noticed any other problems so far. thank you @samwillis.

Hi @solarsoft0, i can confirm this issue in local-only mode as i'm hitting the same problem. Working on a fix for it.

kevin-dp:
## Multi-tab sync bug fixes

Three bugs were identified and fixed that caused multi-tab sync to silently break. Together they caused follower tab mutations to disappear from the leader tab's in-memory collection, and eventually caused sync to stop working in both directions.

---

### Bug 1: Leadership state stuck after setup error — 51630874

**What:** `state.isLeader = true` was set *before* `getStreamPosition()` in `acquireLeadership`. If `getStreamPosition` threw (e.g. due to a `UNIQUE constraint failed: collection_registry.tombstone_table_name` from React StrictMode double-mounting), `isLeader` remained permanently stuck at `true` because the `finally` block that resets it was inside an inner `try/finally` that was never entered.

**Fix:** Wrap the entire lock callback body in a single `try/finally`. Set `state.isLeader = true` only *after* successful setup. The `finally` block always executes and resets `isLeader = false` + cleans up the heartbeat timer.

---

### Bug 2: Seq collision between leader direct path and coordinator — 533be706

**What:** The leader tab had two mutation paths: a "direct" path (write to SQLite and broadcast) and a coordinator RPC path. Only follower tabs used the RPC path — the leader bypassed the coordinator entirely.

This caused a **seq collision**: the leader's direct writes incremented the runtime's `localSeq` (e.g. to 1, 2, 3) but left the coordinator's `state.latestSeq` at 0. When a follower later sent an RPC, the coordinator assigned seq starting from 1 again, producing duplicate seq numbers. The leader then skipped these "already-seen" `tx:committed` messages via `txCommitted.seq <= this.latestSeq`, causing follower mutations to silently disappear.

**Fix:** Always route through `requestApplyLocalMutations` when available, regardless of leader/follower status. This keeps the coordinator's seq counter in sync with all writes. Also removes the stub `requestApplyLocalMutations` from `SingleProcessCoordinator` — it returned success without persisting, so single-process mode correctly falls back to the direct path.

---

### Bug 3: Leader ignoring coordinator-delivered tx:committed for follower mutations — 49720f63

**What:** The runtime's `onCoordinatorMessage` handler skipped ALL messages where `senderId` matched the coordinator's own node ID. But when the coordinator processes a follower's RPC in `handleApplyLocalMutations`, it delivers the resulting `tx:committed` to local subscribers using the coordinator's own `senderId`. This caused the leader's runtime to ignore follower mutations — they were written to SQLite but never applied to the leader's in-memory collection.

**Fix:** Allow `tx:committed` messages from self to pass through the filter. The seq dedup logic in `processCommittedTxUnsafe` prevents double-processing of the leader's own mutations: `observeStreamPosition` is called with the response's term/seq before the local delivery runs under the mutex, so the duplicate is detected via `txCommitted.seq <= this.latestSeq`. Other message types (heartbeats, resets) from self are still skipped.

kevin-dp:
Superseded by https://github.com/TanStack/db/pull/1358

Issue #1012, opened 2025-12-12
type errors when using ts project references / declaration


marbemac:
- [x] I've validated the bug against the latest version of DB packages

**Describe the bug**

Getting the dreaded `The inferred type of "xyz" cannot be named without a reference to ../node_modules/.pnpm/@tanstack+db@0.5.11_typescript@5.9.3/node_modules/@tanstack/db/dist/esm/query/builder/types. This is likely not portable. A type annotation is necessary.` error when trying to use query joins in a project that uses ts project references.

**To Reproduce**

1. Clone https://github.com/getdatanaut/tanstack-repro
2. `pnpm i`
3. Open `src/index.ts` and you should see the ts error.

Or just copy https://github.com/getdatanaut/tanstack-repro/blob/main/src/index.ts into a project that has `declaration: true` in the `tsconfig.json`.

marbemac:
Different package, but kind of relevant / similar - https://github.com/TanStack/query/issues/8453

kevin-dp:
Hi @marbemac,

Thank you for your reproduction which helped confirming the issue. I opened a PR that fixes it. I can confirm that it solves the problems from your reproduction.

PR #1231, opened 2026-02-11
fix: export types used in public API signatures for declaration emit compatibility

kevin-dp:
Fixes https://github.com/TanStack/db/issues/1012


samwillis (approved):
LGTM

Issue #1180, opened 2026-01-25
Collection utils are untyped for Electric, Trailbase, PowerSync, localOnly collections


pawelblaszczyk5:
- [x] I've validated the bug against the latest version of DB packages

**Describe the bug**

Hello 👋

While building app based on Electric I've spotted that `utils` on collection aren't typed correctly - they fall back to untyped `UtilsRecord`. Before just reporting it - I verified it for every built-in collection that has custom utils whether it works and found out that it doesn't for a few more collections.

It doesn't have utils (so it works correctly):

- RxDB collection

It works correctly for:

- Local Storage collection
- TanStack Query collection

Utils aren't typed correctly for:

- Local only collection
- Electric collection
- PowerSync collection
- Trailbase collection

I've created a reproduction including every single of the above collections (except of rxdb which doesn't have custom utils), so it's easy to verify:

https://github.com/pawelblaszczyk5/tanstack-db-missing-collection-utils/blob/main/src/test.ts

<img width="1416" height="999" alt="Image" src="https://github.com/user-attachments/assets/755b6af0-18a2-4b11-a77d-da383d55398b" />

As you can see on above screenshot, a few of these has generic `UtilsRecord` instead of correctly typed ones.

**To Reproduce**

1. Create Electric/Trailbase/PowerSync/localOnly collection
2. Try to access utils
3. Observe them being untyped (they're still there at runtime)

**Expected behavior**

Collection utils are correctly typed

**Screenshots**

Provided above

**Desktop (please complete the following information):**

- OS: N/A
- Browser N/A
- Version N/A

**Smartphone (please complete the following information):**

- Device: N/A
- OS: N/A
- Browser N/A
- Version N/A

**Additional context**

N/A

PR #1236, opened 2026-02-12
Improve type of collection utils for Trailbase, PowerSync, and localOnly collections

kevin-dp:
Fixes https://github.com/tanstack/db/issues/1180

### DIFF
--- a/.changeset/tough-dragons-fold.md
+++ b/.changeset/tough-dragons-fold.md
@@ -1,7 +1,8 @@
 ---
 '@tanstack/powersync-db-collection': patch
 '@tanstack/trailbase-db-collection': patch
+'@tanstack/electric-db-collection': patch
 '@tanstack/db': patch
 ---
 
-Make type of collection utils more precise for localOnly, PowerSync, and Trailbase collections
+Make type of collection utils more precise for localOnly, PowerSync, Trailbase, and Electric collections


pawelblaszczyk5:
Would there be a chance to fix Electric one as well here? It seems to be missing, thanks a lot in advance!

kevin-dp:
> Would there be a chance to fix Electric one as well here? It seems to be missing, thanks a lot in advance!

Based on [this type test](https://github.com/TanStack/db/blob/main/packages/electric-db-collection/tests/electric.test-d.ts#L103-L166) i believe it is typed correctly already. Lmk if we're missing something here.

pawelblaszczyk5:
> > Would there be a chance to fix Electric one as well here? It seems to be missing, thanks a lot in advance!
> 
> Based on [this type test](https://github.com/TanStack/db/blob/main/packages/electric-db-collection/tests/electric.test-d.ts#L103-L166) i believe it is typed correctly already. Lmk if we're missing something here.

I think I see the issue, in tests options aren't passed to `createCollection` in that case it indeed works. But after passing it through `createCollection` utils types are lost.

edit: note around 157. Line in the test you linked seems to describe the exact issue that happens there? 🤔

image

I've updated the reproduction from mine issue that you linked in this PR. 

https://github.com/pawelblaszczyk5/tanstack-db-missing-collection-utils/blob/main/src/test.ts

kevin-dp:
@pawelblaszczyk5 i had to fight a bit around how types are inferred for Electric. This seems to do the trick though. Could you give it a new try and confirm whether it's solved now?

pawelblaszczyk5:
I can confirm it fixes the issue both in my small reproduction and in my bigger project when I use the version from latest commit here. Thanks a lot!

PR #1238, opened 2026-02-12
Make loadSubsetOptions optional in QueryCollectionMeta

### LINKED ISSUE #1061: @tanstack/query-core interface pollution
When using Tanstack Query alongside DB, the extension of `Register` [here](https://github.com/TanStack/db/blob/5fd0f1e8907cf9e24a4fc9790c55be2ff4624375/packages/query-db-collection/src/global.ts#L30-L40) affects all Query uses, leading to widespread type failures due to missing `loadSubsetOptions`:

```

error TS2741: Property 'loadSubsetOptions' is missing in type '{ ...: string; }' but required in type 'QueryCollectionMeta'.

138     meta: { source: '...' },
        ~~~~

  ../../node_modules/@tanstack/query-db-collection/dist/esm/global.d.ts:18:5 - 'loadSubsetOptions' is declared here.
    18     loadSubsetOptions: LoadSubsetOptions;
```

 Extending the shared interface with a required DB-only field seems like the wrong solution.

kevin-dp:
## Summary
- Makes `loadSubsetOptions` optional in `QueryCollectionMeta` to prevent `@tanstack/query-core` `Register` interface pollution
- Updates type tests to use non-null assertion operator where needed

Fixes #1061

## Test plan
- [x] TypeScript type check passes (`tsc --noEmit`)
- [x] All 85 runtime tests pass
- [x] Type test file compiles without errors

### DIFF
--- a/.changeset/optional-loadsubset-meta.md
+++ b/.changeset/optional-loadsubset-meta.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/query-db-collection': patch
+---
+
+Make loadSubsetOptions optional in QueryCollectionMeta to fix query-core interface pollution (#1061)


samwillis (approved):
LGTM

PR #1252, opened 2026-02-16
fix(angular-db): resolve issue where query function was called twice inside of the computed collection object

emcdaniel1624:
the query function was being called twice in the computed.


emcdaniel1624:
Should I have opened an issue for this as well? I noticed it and it was a quick enough fix that I just skipped that step.

Issue #685, opened 2025-10-16
leftJoin + where isUndefined on right field filters out relation but not the row


john-rtr:
- [x] I've validated the bug against the latest version of DB packages

**Describe the bug**
When using a left join, applying a `where` predicate on a right-side field with `isUndefined(right?.payload)` removes the right side only (sets `right` to `undefined`) but still returns the left rows. The filter does not exclude the entire joined row.

**To Reproduce**

1. Create two collections:

```ts
it('left join with isUndefined filtering on right-side field', () => {
  const leftSchema = z.object({ id: z.string(), rightId: z.string().nullable() })
  const rightSchema = z.object({ id: z.string(), payload: z.string().nullish() })

  const left = createCollection(
    localOnlyCollectionOptions({
      id: 'left',
      getKey: (item: { id: string }) => item.id,
      schema: leftSchema,
      initialData: [
        { id: 'l1', rightId: 'r1' },
        { id: 'l2', rightId: 'r2' },
        { id: 'l3', rightId: 'r3' },
        { id: 'l4', rightId: null }
      ]
    })
  )

  const right = createCollection(
    localOnlyCollectionOptions({
      id: 'right',
      getKey: (item: { id: string }) => item.id,
      schema: rightSchema,
      initialData: [
        { id: 'r1', payload: 'ok' },
        { id: 'r2', payload: null },
        { id: 'r3', payload: undefined }
      ]
    })
  )

  const lq = createLiveQueryCollection({
    startSync: true,
    query: (q) =>
      q
        .from({ l: left })
        .leftJoin({ r: right }, ({ l, r }) => eq(l.rightId, r.id))
        .where(({ r }) => isUndefined(r?.payload))
        .select(({ l, r }) => ({ leftId: l.id, right: r }))
  })

  const data = lq.toArray

  expect(data.sort((a, b) => a.leftId.localeCompare(b.leftId))).toEqual([
    { leftId: 'l3', right: { id: 'r3', payload: undefined } },
    { leftId: 'l4', right: undefined }
  ])
})
```

Current results:

```ts
[
  { leftId: 'l1', right: undefined },
  { leftId: 'l2', right: undefined },
  { leftId: 'l3', right: { id: 'r3', payload: undefined } },
  { leftId: 'l4', right: undefined }
]
```

**Expected behavior**
[...]

PR #1254, opened 2026-02-17
fix(db): left join where filter on right-side field incorrectly includes unmatched rows

kevin-dp:
## Summary
- [Adds a failing test](https://github.com/TanStack/db/actions/runs/22096372284/job/63854417048?pr=1254) that reproduces a bug where `.where(({ r }) => isUndefined(r?.payload))` after a `leftJoin` returns all left rows with the right side stripped to `undefined`, instead of excluding rows whose right-side payload is defined.
- Fixes the query optimizer to stop pushing single-source WHERE clauses into the nullable side of outer join subqueries, which was changing join semantics (turning excluded rows into unmatched left-join rows that then pass the residual filter).

Fixes https://github.com/TanStack/db/issues/685

## Test plan
- [x] New unit test `where with isUndefined on right-side field should filter entire joined rows` passes
- [x] Existing join tests continue to pass
- [x] Verify no regressions in optimizer push-down for inner joins

### DIFF
--- a/packages/db/tests/query/optimizer.test.ts
+++ b/packages/db/tests/query/optimizer.test.ts
@@ -1653,10 +1653,7 @@ describe(`Query Optimizer`, () => {
       // pre-filter the right data, converting excluded matches into unmatched rows
       expect(optimizedQuery.where).toHaveLength(1)
       expect(optimizedQuery.where![0]).toEqual(
-        createEq(
-          createPropRef(`teamMember`, `user_id`),
-          createValue(100),
-        ),
+        createEq(createPropRef(`teamMember`, `user_id`), createValue(100)),
       )
 
       // The optimizer should NOT create a subquery for the nullable side
@@ -1701,10 +1698,7 @@ describe(`Query Optimizer`, () => {
       // It must NOT be pushed down to the nullable (left/from) side
       expect(optimizedQuery.where).toHaveLength(1)
       expect(optimizedQuery.where![0]).toEqual(
-        createEq(
-          createPropRef(`user`, `department_id`),
-          createValue(1),
-        ),
+        createEq(createPropRef(`user`, `department_id`), createValue(1)),
       )
 
       // The optimizer should NOT create a subquery for the nullable side (from)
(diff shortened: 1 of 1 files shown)


kevin-dp:
## Root cause

The query optimizer has two mechanisms that push single-source WHERE clauses closer to the data:

1. **Predicate pushdown** (`applyOptimizations`) — wraps a collection reference in a filtering subquery inside the IVM pipeline
2. **Source WHERE extraction** (`extractSourceWhereClauses`) — passes the clause to the collection subscriber for `loadSubset` index optimization

Both are correct for inner joins, but break outer join semantics when applied to the **nullable side** (right side of LEFT, left side of RIGHT, both for FULL).

### Why it breaks

Consider `.leftJoin({ r: right }, ...).where(({ r }) => isUndefined(r?.payload))`:

**Without pushdown** (correct): the join matches all right rows, then the WHERE filters the result:
- l1 ↔ r1 (payload='ok') → `isUndefined('ok')` → false → excluded ✓
- l2 ↔ r2 (payload=null) → `isUndefined(null)` → false → excluded ✓
- l3 ↔ r3 (payload=undefined) → `isUndefined(undefined)` → true → kept ✓
- l4 ↔ (no match) → `isUndefined(undefined)` → true → kept ✓

**With pushdown** (broken): the right side is pre-filtered to only r3, so the join can't find r1/r2:
- l1 ↔ (no match!) → `isUndefined(undefined)` → true → kept ✗
- l2 ↔ (no match!) → `isUndefined(undefined)` → true → kept ✗
- l3 ↔ r3 → `isUndefined(undefined)` → true → kept ✓
- l4 ↔ (no match) → `isUndefined(undefined)` → true → kept ✓

The optimizer keeps a residual WHERE for outer joins to handle exactly this, but it can't distinguish "genuinely unmatched" rows (l4) from "had a match but it was pre-filtered away" rows (l1, l2) — they look identical. When the WHERE condition happens to be true for unmatched rows (like `isUndefined`), the residual passes them through incorrectly.

## Fix

A new `getNullableJoinSources()` helper identifies which source aliases are on the nullable side of each outer join. Both `applyOptimizations` and `extractSourceWhereClauses` now exclude these sources, keeping their WHERE clauses as post-join filters on the main pipeline.

This is conservative — it also blocks pushdown for cases where the residual would work correctly (e.g. `eq(other.status, 'active')` where the residual evaluates to false for unmatched rows). But statically determining which expressions are safe is complex, and correctness is more important than the optimization.

Issue #896, opened 2025-11-24
Joining collections by small Uint8Array's checks by reference rather than content


Tomaszal:
- [x] I've validated the bug against the latest version of DB packages

**Describe the bug**
Collections joined by small Uint8Array comparisons are checked using reference, when they should be checked by content

**To Reproduce**
https://stackblitz.com/edit/vitejs-vite-jwhabpic

If data initialization on lines 10-12 is moved up to the root, references will be consistent across both collections, and join operation will work correctly. However, it should also work in the initial arrangement if Uint8Array's are compared by content

**Expected behavior**
Using the provided reproduction, "Joined by ID" and "Joined by name" should result in the same output

**Additional context**
Related to https://github.com/TanStack/db/pull/779

kevin-dp:
Closing this as won't-fix. We are considering joining on `Uint8Array` values out of scope due to the added complexity and the ambiguity around what size threshold should trigger content-based comparison vs. reference comparison.

We appreciate the detailed reproduction, but this is a known trade-off we've accepted. If your use case requires joining on binary data, we'd recommend converting the `Uint8Array` values to a comparable primitive (e.g., a hex string) before using them as join keys.

Tomaszal:
Understandable, although I wonder if #779 should be reverted then? Because the way it is right now is even more ambiguous, as the `eq` comparisons behave differently when used in `query` vs `join`

KyleAMathews:
Yes, I'll revert it

PR #1258, opened 2026-02-17
Remove Uint8Array content-based equality comparison

KyleAMathews:
## 🎯 Changes

This PR removes the special handling for Uint8Array and Buffer content-based equality comparisons. Previously, the codebase attempted to normalize small Uint8Arrays (≤128 bytes) to string representations to enable content-based equality checks, while using reference equality for larger arrays.

**Changes made:**

1. **Removed `normalizeValue` logic for Uint8Arrays/Buffers** - Deleted the code that converted small Uint8Arrays to string representations (`__u8__` prefixed strings) for Map key usage
2. **Removed `areValuesEqual` function** - Deleted the utility function that provided special Uint8Array comparison logic
3. **Simplified `eq` operator** - Changed from using `normalizeValue` and `areValuesEqual` to simple reference equality (`===`)
4. **Removed `hashUint8Array` function** - Deleted the content-based hashing for small Uint8Arrays; all Uint8Arrays/Buffers now use reference-based hashing
5. **Updated hash function** - All Uint8Arrays, Buffers, and Files now consistently use reference-based hashing via `cachedReferenceHash`
6. **Removed test cases** - Deleted integration tests for Uint8Array ID comparison and unit tests for Uint8Array equality handling

**Rationale:**

The previous approach of normalizing small Uint8Arrays to strings for content-based equality was complex and had limitations. Users working with Uint8Array IDs (like ULIDs) should use a derived string key in their data model instead of relying on implicit content-based comparison. This simplification:
- Reduces code complexity
- Improves performance by avoiding string allocations
- Makes equality semantics consistent (always reference-based)
- Aligns with how other binary types (Files) are handled

## ✅ Checklist

## 🚀 Release Impact

https://claude.ai/code/session_014bN4QUhGYQpuNAy1Ec42pC


kevin-dp (approved):
Approving but looks like you forgot to remove the "Uint8Arrays with same content have same hash" test and you should also add a changeset.

KyleAMathews:
Closing as superseded by the current equality architecture. Current main intentionally defines binary content equality across `eq`/`in`, hashing, grouping, routing, and demand; the public join path is the remaining inconsistent owner, with a deterministic RED matrix for distinct Uint8Array/Buffer copies in both auto-index modes. Reverting to reference equality would break that established contract and remove normative tests. We’ll use #896/#899/#1258 as credited design provenance for a fresh current-main join/equality/planning fix, rather than revive this rollback. Thanks for making the alternative policy and its cost concerns explicit.

Issue #1251, opened 2026-02-16
LEFT JOIN: select() callback receives truthy proxy objects for unmatched rows instead of null/undefined


marbemac:
- [x] I've validated the bug against the latest version of DB packages

**Describe the bug**

When using `leftJoin`, the `select()` callback receives proxy objects for unmatched (right-side) rows instead of `null` or `undefined`. These proxy objects are truthy, so conditional checks like `right ? { id: right.id } : null` always take the truthy branch, even when there was no matching row on the right side of the join.

This means there's no way to distinguish matched from unmatched rows inside `select()`. Accessing fields on these proxy objects (e.g. `right.name`) also returns truthy empty objects `{}` rather than `undefined`, so even `right?.id ? ... : null` doesn't work.

**To Reproduce**

```ts
import { createCollection, localOnlyCollectionOptions, createLiveQueryCollection, eq, Query } from '@tanstack/db';
import { z } from 'zod';

const users = createCollection(
  localOnlyCollectionOptions({
    id: 'users',
    getKey: (item: { id: string }) => item.id,
    schema: z.object({ id: z.string(), name: z.string() }),
    initialData: [
      { id: 'u1', name: 'Alice' },
      { id: 'u2', name: 'Bob' },
    ],
  }),
);

const profiles = createCollection(
  localOnlyCollectionOptions({
    id: 'profiles',
    getKey: (item: { id: string }) => item.id,
    schema: z.object({ id: z.string(), userId: z.string(), bio: z.string() }),
    initialData: [
      { id: 'p1', userId: 'u1', bio: 'Hello world' },
      // No profile for u2
    ],
  }),
);

const lq = createLiveQueryCollection({
  startSync: true,
  query: (q) =>
    q
      .from({ user: users })
      .leftJoin({ profile: profiles }, ({ user, profile }) => eq(user.id, profile.userId))
      .select(({ user, profile }) => ({
        name: user.name,
        // This conditional NEVER takes the null branch because `profile` is always a truthy proxy object
        bio: profile ? profile.bio : null,
      })),
});

const data = lq.toArray;
console.log(data);
// Actual:   [{ name: 'Alice', bio: 'Hello world' }, { name: 'Bob', bio: {} }]
// Expected: [{ name: 'Alice', bio: 'Hello world' }, { name: 'Bob', bio: null }]
```

[...]

PR #1262, opened 2026-02-18
fix(db): use Ref<T, Nullable> for left join select refs instead of Ref<T> | undefined

kevin-dp:
Improves types of LEFT, RIGHT, and FULL joined data. Closes https://github.com/TanStack/db/issues/1251.

## Summary

- Adds failing type tests that assert the desired behavior for left join refs in declarative `.select()` callbacks
- Currently, left-joined table refs are typed as `Ref<T> | undefined`, which forces optional chaining (`dept?.name`). Since these refs are always truthy proxy objects at build time, the `| undefined` is misleading and encourages runtime conditional patterns (`profile ? profile.bio : null`) that silently fail — the truthy branch is always taken
- The fix (in a follow-up) will introduce `Ref<T, Nullable>` where nullability is encoded as a type parameter rather than a union, so that:
  - Direct property access works without `?.` (honest — the proxy is always present)
  - Result types still correctly produce `T | undefined` for fields from left-joined tables
  - Users are guided toward `coalesce()` / `isNull()` for null handling in declarative queries

## Test plan

- [x] Type test for left, right, and full join select with direct property access (currently fails — expected)
- [x] Type test for inner join select with direct property access (passes — baseline)

### DIFF
--- a/.changeset/fix-nullable-join-select-ref-typing.md
+++ b/.changeset/fix-nullable-join-select-ref-typing.md
@@ -1,0 +1,7 @@
+---
+'@tanstack/db': patch
+---
+
+fix(db): use `Ref<T, Nullable>` brand instead of `Ref<T> | undefined` for nullable join refs in declarative select
+
+The declarative `select()` callback receives proxy objects that record property accesses. These proxies are always truthy at build time, but nullable join sides (left/right/full) were typed as `Ref<T> | undefined`, misleading users into using `?.` and `??` operators that have no effect at runtime. Nullable join refs are now typed as `Ref<T, true>`, which allows direct property access without optional chaining while correctly producing `T | undefined` in the result type.


kevin-dp:
## How the new typing works

### Problem

The declarative `select()` callback receives proxy objects that record property access paths at build time. These proxies are **always truthy** — they're JavaScript objects, never `undefined`. But nullable join sides (left/right/full) were typed as `Ref<T> | undefined`, which led users into traps like:

```ts
.leftJoin({ dept }, ...)
.select(({ user, dept }) => ({
  // dept is typed as Ref<Department> | undefined, but it's always a truthy proxy
  deptName: dept?.name,          // ?. has no effect at runtime
  safeName: dept?.name ?? 'N/A', // ?? never triggers at runtime
}))
```

### Solution

Nullable join refs are now typed as `Ref<T, true>` instead of `Ref<T> | undefined`. The second type parameter (`Nullable`) is a brand that propagates through all nested property accesses:

```ts
// RefLeaf uses a conditional brand — only present when Nullable is true
type RefLeaf<T, Nullable extends boolean = false> = {
  readonly [RefBrand]?: T
} & ([Nullable] extends [true] ? { readonly [NullableBrand]?: true } : {})
```

This means:
- **`Ref<T, false>`** (inner join / from table) — no `NullableBrand`, result extracts as `T`
- **`Ref<T, true>`** (nullable join side) — has `NullableBrand`, result extracts as `T | undefined`
- **Direct property access works** — `dept.name` compiles without `?.` since the proxy is never `undefined`
- **Result types are correct** — `dept.name` produces `string | undefined` in the query result via `IsNullableRef` detection

### Why not `Ref<T | undefined>` or `Ref<T> | undefined`?

- **`Ref<T | undefined>`** breaks because `RefLeaf<T>` uses intersection branding (`T & { [RefBrand]?: T }`), and `(T | undefined) & { brand }` collapses to `T & { brand }` — the `undefined` gets erased. It also breaks mapped type keys since `keyof (T | undefined)` = `never`.
- **`Ref<T> | undefined`** is the current approach and the root cause of the problem — it makes the proxy appear nullable when it's always truthy.

### Assignability

The conditional brand ensures `RefLeaf<T, false>` extends `RefLeaf<T, true>` (an absent optional property satisfies `true | undefined`). This preserves the `QueryBuilder<SpecificContext>` → `QueryBuilder<Context>` assignability needed for subqueries.

marbemac:
❤️  yes this looks great!

packages/db/tests/query/live-query-collection.test.ts
```diff
@@ -2324,9 +2325,9 @@ describe(`createLiveQueryCollection`, () => {
             .join({ users }, ({ comments: c, users: u }) => eq(c.userId, u.id))
             .select(({ comments: c, users: u }) => ({
               id: c.id,
-              userId: u?.id ?? c.userId,
+              userId: coalesce(u.id, c.userId),
```
samwillis, line 2328: When our own tests have this bug it makes it VERY clear we should change this!

PR #1263, opened 2026-02-18
fix(db): like not matching newline characters

ldirer:
## Summary

Fix `like`/`ilike` operators not matching newline characters with `%` or `_` wildcards. SQL LIKE semantics treat `%` as matching **any** sequence of characters (including newlines) and `_` as matching **any** single character (including newlines), but JavaScript's RegExp `.` does not match newline characters by default.

## Root Cause

The `evaluateLike` function converts SQL LIKE patterns to JavaScript regex (`%` → `.*`, `_` → `.`), but `.` in JS regex doesn't match line terminators (`\n`, `\r`, `\u2028`, `\u2029`) unless the `s` (dotAll) flag is set. This meant any field value containing a newline would fail to match patterns that should logically match it.

## Approach

Add the `s` (dotAll) flag to the `RegExp` constructor in both copies of `evaluateLike`:

```ts
// Before
const regex = new RegExp(`^${regexPattern}$`)

// After — dotAll flag so '.' matches newline characters (SQL LIKE semantics)
const regex = new RegExp(`^${regexPattern}$`, `s`)
```

The fix is applied in two locations:
- `packages/db/src/query/compiler/evaluators.ts` — the core runtime
- `packages/query-db-collection/e2e/query-filter.ts` — the e2e test helper (duplicated logic)

## Verification

```bash
pnpm vitest run packages/db/tests/query/compiler/evaluators.test.ts
```

All 180 evaluator tests pass (90 per copy), including 4 new tests covering `like` and `ilike` with `%` and `_` against newline characters.

## Files Changed

| File | Change |
|------|--------|
| `packages/db/src/query/compiler/evaluators.ts` | Add `s` (dotAll) flag to RegExp |
| `packages/query-db-collection/e2e/query-filter.ts` | Same fix in duplicated e2e helper |
| `packages/db/tests/query/compiler/evaluators.test.ts` | Add 4 unit tests for like/ilike newline matching |
| `packages/db-collection-e2e/src/fixtures/seed-data.ts` | Add seed user with newline in name |
| `packages/db-collection-e2e/src/suites/predicates.suite.ts` | Add e2e test for like with newline |
| `.changeset/nice-months-nail.md` | Patch changeset for @tanstack/db |

---

### DIFF
--- a/.changeset/nice-months-nail.md
+++ b/.changeset/nice-months-nail.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/db': patch
+---
+
+Fix like/ilike `%` and `_` not matching newline characters


KyleAMathews:
Thanks for the fix!

PR #1268, opened 2026-02-18
fix(react-db): make getNextPageParam optional in useLiveInfiniteQuery

### LINKED ISSUE #1241: cleanup: getNextPageParam is required but never used in useLiveInfiniteQuery
## Description

In `useLiveInfiniteQuery`, the `getNextPageParam` callback is a **required** field in `UseLiveInfiniteQueryConfig`:

```ts
export type UseLiveInfiniteQueryConfig<TContext extends Context> = {
  pageSize?: number
  initialPageParam?: number
  getNextPageParam: (
    lastPage: Array<InferResultType<TContext>[number]>,
    allPages: Array<Array<InferResultType<TContext>[number]>>,
    lastPageParam: number,
    allPageParams: Array<number>,
  ) => number | undefined
}
```

However, `getNextPageParam` is **never called** anywhere in the implementation. The `hasNextPage` value is determined entirely by the peek-ahead strategy (checking if `dataArray.length > totalItemsRequested`).

Every consumer is forced to write a function that has zero effect on behavior. The docstring examples even show `getNextPageParam` implementations, reinforcing the false impression that it controls pagination.

## Suggested Fix

Either:
1. **Remove `getNextPageParam`** from the config type since it's unused, or
2. **Make it optional** (`getNextPageParam?:`) as a no-op for API compatibility with TanStack Query conventions

Option 1 is a breaking change. Option 2 is non-breaking and still communicates that it's not required.

## Impact

Low — this is a DX papercut. Users write dead code and may be confused when changing `getNextPageParam` has no effect on pagination behavior.

sleitor:
## Summary

`getNextPageParam` is required in `UseLiveInfiniteQueryConfig` but is never called anywhere in the implementation. Pagination is determined entirely by the peek-ahead strategy (`dataArray.length > totalItemsRequested`).

This forces every consumer to write a callback that has zero effect on behavior.

### Change

Mark `getNextPageParam` as optional (`?`) with a `@deprecated` JSDoc tag explaining that pagination is handled internally.

This is a non-breaking change — existing code that passes `getNextPageParam` continues to work; code that omits it no longer gets a type error.

Fixes #1241

### DIFF
--- a/.changeset/fix-optional-getnextpageparam.md
+++ b/.changeset/fix-optional-getnextpageparam.md
@@ -1,5 +1,5 @@
 ---
-"@tanstack/react-db": patch
+'@tanstack/react-db': patch
 ---
 
 fix(react-db): make getNextPageParam optional in useLiveInfiniteQuery


kevin-dp:
Thanks for your contribution. I would be tempted to remove it altogether even though that's a breaking change. Another possibility is to mark it deprecated until v1 and then make the breaking change in v1. @samwillis @KyleAMathews what do you think?

KyleAMathews:
We can do the deprecated thing — let's just make an issue with things to remove in the first RC like this.

kevin-dp:
> We can do the deprecated thing — let's just make an issue with things to remove in the first RC like this.

Was thinking the same about making a list but we could also just do a search for @deprecated prior to v1 release. I believe we want to remove all @deprecated features in v1.

sleitor:
Happy to add a `@deprecated` JSDoc annotation to the field in this PR if you'd like — that way the fix ships now and the cleanup can be tracked in your v1 deprecation list. Just say the word and I'll update the PR.

PR #1269, opened 2026-02-18
fix(react-db): handle rejected/stale setWindow promises in useLiveInfiniteQuery

### LINKED ISSUE #1240: bug: useLiveInfiniteQuery Promise handling can leave isFetchingNextPage stuck
## Description

`useLiveInfiniteQuery` has several Promise handling issues in the `useEffect` that calls `setWindow()` (lines 236-248 of `useLiveInfiniteQuery.ts`):

### 1. No `.catch()` on `setWindow` Promise

```ts
if (result !== true) {
  setIsFetchingNextPage(true)
  result.then(() => {
    setIsFetchingNextPage(false)
  })
}
```

If the `setWindow` Promise rejects (e.g. async `loadSubset` throws), `isFetchingNextPage` stays `true` forever with no error surfaced to the user.

### 2. Missing `useEffect` cleanup for stale Promises

The effect doesn't return a cleanup function. If the effect re-runs while a previous `setWindow` Promise is still pending, the old `.then()` handler can fire and set `isFetchingNextPage(false)` prematurely while new async loading is in progress. This causes the loading state to flicker or become unreliable.

### 3. `setWindow` Promise can hang indefinitely

The Promise returned by `setWindow` has no rejection path or timeout. If an async `loadSubset` never resolves (network hang, etc.), the Promise hangs forever.

## Suggested Fix

```ts
useEffect(() => {
  let cancelled = false
  // ... existing validation ...

  const result = utils.setWindow({ offset: expectedOffset, limit: expectedLimit })

  if (result !== true) {
    setIsFetchingNextPage(true)
    result
      .then(() => {
        if (!cancelled) setIsFetchingNextPage(false)
      })
      .catch((error) => {
        if (!cancelled) {
          setIsFetchingNextPage(false)
          console.error(`useLiveInfiniteQuery: setWindow failed:`, error)
        }
      })
  } else {
    setIsFetchingNextPage(false)
  }

  return () => { cancelled = true }
}, [/* deps */])
```

## Impact

When using `useLiveInfiniteQuery` with an on-demand collection that has an async `loadSubset`, if the async operation fails, the user sees a permanent loading state with no error feedback.

sleitor:
## Summary

The `useEffect` that calls `utils.setWindow()` in `useLiveInfiniteQuery` has three issues:

1. **No `.catch()`** — if `setWindow` rejects (e.g. async `loadSubset` throws), `isFetchingNextPage` stays `true` forever with no error surfaced to the user.

2. **No cleanup for stale promises** — if the effect re-runs while a previous `setWindow` promise is pending, the old `.then()` handler fires and sets `isFetchingNextPage = false` prematurely while new loading is in progress.

3. **Unhandled promise rejection** — Node/browser logs an unhandled rejection warning.

### Fix

- Add `.catch()` that resets `isFetchingNextPage` and logs the error
- Add a `cancelled` flag with `useEffect` cleanup to ignore stale promise settlements

Fixes #1240

### DIFF
--- a/packages/react-db/src/useLiveInfiniteQuery.ts
+++ b/packages/react-db/src/useLiveInfiniteQuery.ts
@@ -256,7 +256,8 @@ export function useLiveInfiniteQuery<TContext extends Context>(
       setIsFetchingNextPage(true)
       result
         .catch((error: unknown) => {
-          if (!cancelled) console.error(`useLiveInfiniteQuery: setWindow failed:`, error)
+          if (!cancelled)
+            console.error(`useLiveInfiniteQuery: setWindow failed:`, error)
         })
         .finally(() => {
           if (!cancelled) setIsFetchingNextPage(false)


packages/react-db/src/useLiveInfiniteQuery.ts
```diff
@@ -233,19 +233,31 @@ export function useLiveInfiniteQuery<TContext extends Context>(
...
     })
 
     if (result !== true) {
       setIsFetchingNextPage(true)
-      result.then(() => {
-        setIsFetchingNextPage(false)
-      })
+      result
+        .then(() => {
+          if (!cancelled) setIsFetchingNextPage(false)
```
kevin-dp: Since both the `then` and `catch` branches are doing something similar we can improve this by using a `finally` block:
```ts
result
  .catch((error: unknown) => {
    if (!cancelled) console.error(`useLiveInfiniteQuery: setWindow failed:`, error)
  })
  .finally(() => {
    if (!cancelled) setIsFetchingNextPage(false)
  })
```

sleitor: Good catch — done in the latest commit. Much cleaner with `finally`.

kevin-dp (approved):

PR #1270, opened 2026-02-19
feat(electric-db-collection): DNF/active_conditions support

kevin-dp:
## Summary

- Adds support for the new Electric server wire protocol ([electric-sql/electric#3791](https://github.com/electric-sql/electric/pull/3791)) which enables arbitrary boolean WHERE clauses with OR/NOT and multiple subqueries
- Changes tag delimiter from `|` to `/` and replaces `_` wildcards with empty segments (`NON_PARTICIPATING` positions)
- Adds `active_conditions` header support: shapes with subquery dependencies use DNF evaluation (row visible if ANY disjunct has ALL its positions satisfied), simple shapes retain existing empty-tag-set deletion behavior
- Derives `disjunct_positions` once per shape (global) rather than per-row, since the DNF structure is fixed by the WHERE clause

## Test plan

- [x] Updated all existing tag tests to use new `/` delimiter and empty-segment syntax
- [x] Added 7 new test cases covering DNF scenarios: partial deactivation, full deactivation, multi-disjunct visibility, active_conditions overwrite on re-send, simple shapes without active_conditions, and mixed rows
- [x] Full test suite passes (419 tests, 0 failures, no type errors)

### DIFF
--- a/packages/react-db/package.json
+++ b/packages/react-db/package.json
@@ -54,7 +54,7 @@
     "react": ">=16.8.0"
   },
   "devDependencies": {
-    "@electric-sql/client": "^1.5.13",
+    "@electric-sql/client": "^1.5.14",
     "@testing-library/react": "^16.3.2",
     "@types/react": "^19.2.13",
     "@types/react-dom": "^19.2.3",
--- a/packages/solid-db/package.json
+++ b/packages/solid-db/package.json
@@ -49,7 +49,7 @@
     "solid-js": ">=1.9.0"
   },
   "devDependencies": {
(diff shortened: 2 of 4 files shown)


kevin-dp:
> [!IMPORTANT]
> **Preview package dependency**: This PR currently uses a preview build of `@electric-sql/client` from https://github.com/electric-sql/electric/pull/4043:
> ```
> "@electric-sql/client": "https://pkg.pr.new/@electric-sql/client@4043"
> ```
> Before merging this PR, replace the preview URL with the published version (e.g. `"^1.6.0"` or whatever version includes the move-in event support).

Issue #720, opened 2025-10-25
Aggregate inside expression (e.g. `coalesce(count(...), 0)`) throws `QueryCompilationError: Unknown expression type: agg`


evanheckert:
### Summary
Wrapping an aggregate in another expression (e.g., `coalesce(count(item.property), 0)`) throws during compilation. Using the plain aggregate (`count(item.property)`) succeeds. Oddly, the wrapped form *does* work inside a live query, so behavior is inconsistent.

### Repro

Building the subquery outside of the hook threw the error:
```ts
const vehicleDispatchCountSubquery = new Query()
  .from({ vehicleDispatch: vehicleDispatchCollection })
  .groupBy(({ vehicleDispatch }) => vehicleDispatch.dispatchId)
  .select(({ vehicleDispatch }) => ({
    dispatchId: vehicleDispatch.dispatchId,
    vehicleDispatchCount: coalesce(count(vehicleDispatch.id), 0),
  }))
```
**However**, just using `count(vehicleDispatch.id)` without `coalesce` no longer throws the error.

Alternatively, within a `useLiveQuery()` nesting works without error:
```ts
export const useDispatchSidebarData = (date: string) =>
  useLiveQuery(
    q => {
      const vdCountSubquery = q
        .from({ vehicleDispatch: vehicleDispatchCollection })
        .groupBy(({ vehicleDispatch }) => vehicleDispatch.dispatchId)
        .select(({ vehicleDispatch }) => ({
          dispatchId: vehicleDispatch.dispatchId,
          vehicleDispatchCount: coalesce(count(vehicleDispatch.id)),
        }))
        .distinct()
      return q
        .from({ dispatch: dispatchCollection })
        .where(({ dispatch }) => eq(dispatch.date, date))
        .leftJoin({ vdCount: vdCountSubquery }, ({ dispatch, vdCount }) => eq(dispatch.id, vdCount.dispatchId))
        .select(({ dispatch, vdCount }) => ({
          ...dispatch,
          vehicleDispatchCount: vdCount?.vehicleDispatchCount,
        }))
    },
    [date]
  )
```

#### Error
```
QueryCompilationError: Unknown expression type: agg
    at compileExpressionInternal (…/chunk-TT3G3ME3.js:703:13)
    at …/chunk-TT3G3ME3.js:751:14
    at Array.map (<anonymous>)
    at compileFunction (…/chunk-TT3G3ME3.js:750:34)
```

### Environment

[...]

PR #1274, opened 2026-02-19
fix(db): support aggregates nested inside expressions (#720)

kevin-dp:
## Summary

Fixes https://github.com/TanStack/db/issues/720.

Aggregates wrapped inside other expressions (e.g. `coalesce(count(...), 0)`) threw `QueryCompilationError: Unknown expression type: agg` because the compiler only recognized top-level aggregates in SELECT, not aggregates nested inside function expressions.

### Approach

HAVING clauses already solve a similar problem — `replaceAggregatesByRefs` walks the HAVING expression tree, matches aggregates against SELECT entries, and replaces them with refs to computed values. But for SELECT, the aggregates are *buried inside* other expressions rather than existing as top-level entries, so there's nothing to match against.

The fix adds an extract-register-replace pass: before groupBy processing, nested aggregates are extracted from wrapper expressions and registered under synthetic aliases (e.g. `__agg_0`). The aggregates in the original expression are replaced with `PropRef` references to these aliases. After groupBy computes the aggregate values, a second pass evaluates the wrapper expressions against the populated results.

### Changes

- **`select.ts`**: Defer expressions containing nested aggregates to `processGroupBy` (same as plain aggregates)
- **`group-by.ts`**: Add `containsAggregate` / `extractAndReplaceAggregates` helpers; update both single-group and multi-group paths to extract, register, and evaluate wrapped aggregates
- **`index.ts`**: Detect nested aggregates for implicit single-group aggregation

### Reproduction tests

- 5 tests in `@tanstack/db` (`packages/db/tests/query/group-by.test.ts`): `coalesce(count(...))`, `coalesce(sum(...))`, `add(count(...), count(...))`, mixed plain + wrapped, subquery join source
- 3 tests in `@tanstack/react-db` (`packages/react-db/tests/useLiveQuery.test.tsx`): `coalesce(count(...))`, `coalesce(sum(...))`, subquery join source — confirming the bug also affects `useLiveQuery`

## Test plan

- [x] All 8 new tests pass (previously threw `Unknown expression type: agg`)
- [x] All existing tests pass (94 group-by, 32 useLiveQuery)
- [x] TypeScript builds cleanly (`tsc --noEmit`)

### DIFF
--- a/packages/db/tests/query/group-by.test.ts
+++ b/packages/db/tests/query/group-by.test.ts
@@ -1786,18 +1786,18 @@ function createGroupByTests(autoIndex: `off` | `eager`): void {
         })
 
         expect(customerSummary.size).toBe(3)
-
-        const customer1 = customerSummary.get(1)
-        expect(customer1?.customer_id).toBe(1)
-        expect(customer1?.order_count).toBe(3)
-
-        const customer2 = customerSummary.get(2)
-        expect(customer2?.customer_id).toBe(2)
-        expect(customer2?.order_count).toBe(2)
-
-        const customer3 = customerSummary.get(3)
-        expect(customer3?.customer_id).toBe(3)
-        expect(customer3?.order_count).toBe(2)
+        expect(customerSummary.get(1)).toEqual({
+          customer_id: 1,
+          order_count: 3,
+        })
+        expect(customerSummary.get(2)).toEqual({
+          customer_id: 2,
+          order_count: 2,
+        })
+        expect(customerSummary.get(3)).toEqual({
+          customer_id: 3,
+          order_count: 2,
+        })
       })
 
       test(`coalesce wrapping sum returns the sum value or the fallback`, () => {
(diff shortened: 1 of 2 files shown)


packages/db/src/query/compiler/group-by.ts
```diff
@@ -234,11 +269,18 @@ export function processGroupBy(
...
+        // Add synthetic aggregate values so wrapped expressions can reference them
+        for (const key of Object.keys(aggregatedRow)) {
+          if (key.startsWith(`__agg_`)) {
+            finalResults[key] = aggregatedRow[key]
+          }
+        }
+        // Second pass: evaluate wrapped-aggregate expressions
+        for (const [alias, evaluator] of Object.entries(wrappedAggExprs)) {
+          finalResults[alias] = evaluator({ $selected: finalResults })
+        }
```
samwillis: Nit: This is exactly the same code as on 134

kevin-dp:
@samwillis Good catch on all three points. Fixed in the latest commits:

1. **Synthetic keys leaking onto result rows** — Added a cleanup pass in `evaluateWrappedAggregates` that deletes all `__agg_*` keys from `finalResults` after the wrapped expressions have been evaluated. They're only needed temporarily during that evaluation step.

2. **Tests not validating exact row shape** — Replaced per-field assertions (`expect(row?.field).toBe(...)`) with `toEqual` checks on the full row object. This ensures no extra keys (like `__agg_0`) are present on projected results.

3. **Duplicated code between single-group and multi-group paths** — Extracted the synthetic-value copy, evaluate, and cleanup logic into a shared `evaluateWrappedAggregates` helper used by both paths.

packages/db/src/query/compiler/group-by.ts
kevin-dp: Fixed — extracted into a shared `evaluateWrappedAggregates` helper.

PR #1276, opened 2026-02-19
Fix select() to handle direct ref returns as spreads

KyleAMathews:
## 🎯 Changes

Fixed a common pitfall in the `select()` query builder where returning a ref directly (e.g., `select(({ users }) => users)`) would produce ref/proxy metadata instead of actual row data.

The fix detects when a callback returns a `RefProxy` directly and automatically treats it as if the user had spread it (e.g., `select(({ users }) => ({ ...users }))`). This is accomplished by:

1. Adding an `isRefProxy()` check in the `select()` method
2. Converting direct ref returns into a sentinel-based spread object that the existing `buildNestedSelect()` logic already handles correctly
3. This preserves all existing behavior for explicit spreads while fixing the implicit case

## ✅ Checklist

## 🚀 Release Impact

---

**Test Coverage:** Added three comprehensive test cases covering:
- Basic direct ref returns projecting full rows
- Nested object field preservation with direct ref returns
- Live update handling with direct ref returns

https://claude.ai/code/session_01KAGFqbqFxY2YYV4uTVTBa8

### DIFF
--- a/packages/db/tests/query/select-spread.test.ts
+++ b/packages/db/tests/query/select-spread.test.ts
@@ -280,9 +280,7 @@ describe(`select spreads (runtime)`, () => {
 
   it(`returning a ref directly (without spread) projects the full row`, async () => {
     const collection = createLiveQueryCollection((q) =>
-      q
-        .from({ message: messagesCollection })
-        .select(({ message }) => message),
+      q.from({ message: messagesCollection }).select(({ message }) => message),
     )
     await collection.preload()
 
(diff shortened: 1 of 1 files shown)


samwillis:
Other than that, looking good. Needs a changeset

samwillis:
Closing this in favor of the work that landed in #1384.

We did incorporate the useful part of this PR, but in a narrower form.

What we kept:
- support for returning a top-level alias ref directly, e.g. `select(({ users }) => users)`, so it behaves like the spread form and projects the full row instead of ref metadata

What we intentionally did not keep:
- the broader “any direct `RefProxy` return becomes a spread” behavior

The reason is that we needed to land scalar child selects for `toArray(...)` / `concat(toArray(...))`, and that requires leaf refs like:

```ts
select(({ c }) => c.text)
```

to remain genuine scalar selections, not be rewritten as spread-like object projections.

So the behavior now is:
- top-level alias/object ref returns are treated as spread-equivalent
- scalar leaf refs are left alone

That gives us the safe/fixed case this PR was targeting, without conflicting with the scalar-select work in [#1384](https://github.com/TanStack/db/pull/1384) and the follow-up boundary tightening there.

Because of that, this PR is effectively superseded by the narrower implementation on that branch, so I’m closing this one.

Issue #1279, opened 2026-02-20
broken typescript types: tanstack/query `queryOptions` are not compatible with `queryCollectionOptions`


valerii15298:
- [x] I've validated the bug against the latest version of DB packages

**Describe the bug**
Result of `queryOptions` is not compatible with `queryCollectionOptions`

# To Reproduce
Code to reproduce the behavior:
```ts
import { queryCollectionOptions } from "@tanstack/query-db-collection";
import { createCollection } from "@tanstack/react-db";
import { QueryClient, queryOptions } from "@tanstack/react-query";

const queryClient = new QueryClient();

const queryKey = ["numbers"];

const queryFn = () =>
  Promise.resolve([
    { id: 1, value: "one" },
    { id: 2, value: "two" },
  ]);

const opts = queryOptions({ queryKey, queryFn });

const collection = createCollection(
  queryCollectionOptions({
    ...opts,
    // uncomment these to make it work
    // queryKey,
    // enabled: true,
    // queryFn: (ctx) => Promise.resolve(opts.queryFn!(ctx)),
    queryClient,
    getKey(item) {
      return item.id;
    },
  }),
);

```
**Expected behavior**
There should be no typescript errors with the code above.

## Motivation
A lot of tools already provide options using `queryOptions` utility and would be great if it was compatible with tanstack/db.
I specifically use `oRPC` and need to do these workarounds in order to make things works :(

Latest dependencies on time of this issue creation:
```
dependencies:
@tanstack/query-db-collection 1.0.24
@tanstack/react-db 0.1.71
@tanstack/react-query 5.90.21
```

# Additional Context and debugging
While testing here is additional context when testing each key compatibility, basically only commenting one line from the next working code which uses workaround:
```ts
import { queryCollectionOptions } from "@tanstack/query-db-collection";
import { createCollection } from "@tanstack/react-db";
import { QueryClient, queryOptions } from "@tanstack/react-query";

const queryClient = new QueryClient();

const queryKey = ["numbers"];

const queryFn = () =>
  Promise.resolve([
    { id: 1, value: "one" },
    { id: 2, value: "two" },
  ]);

const opts = queryOptions({ queryKey, queryFn });

const collection = createCollection(
  queryCollectionOptions({
    ...opts,
    queryKey,
    enabled: true,
    queryFn: (ctx) => Promise.resolve(opts.queryFn!(ctx)),
    queryClient,
    getKey(item) {
      return item.id;
    },
  }),
);
```

[...]

PR #1287, opened 2026-02-23
fix(query-db-collection): align queryOptions interop types

hugiex:
## Summary
To fix this issue: https://github.com/TanStack/db/issues/1279

This PR makes `queryOptions(...)` interop first-class in `@tanstack/query-db-collection`. You can now spread `...queryOptions(...)` into `queryCollectionOptions(...)` without re-declaring `queryKey`, re-shaping `enabled`, or wrapping `queryFn`.

## 🎯 Changes
1. Aligned query collection types with TanStack Query core contracts:
- `queryFn` now accepts `T | Promise<T>` (not Promise-only).
- `enabled` now aligns with `QueryObserverOptions['enabled']` (boolean or callback).
- `staleTime`, `refetchInterval`, `retry`, and `retryDelay` now follow `QueryObserverOptions` typing with the correct query data/key generics.
- Added typed overloads for `DataTag` query keys returned by `queryOptions(...)`.
2. Kept runtime safety strict and unchanged:
- `queryFn` is still required at runtime.
- wrapped responses still require `select` to extract item arrays.
3. Added coverage and docs:
- Added interop tests (including sync `queryFn` return support).
- Added docs section showing `queryOptions(...)` spread usage.
- Intentionally did not add `SkipToken` support (query collections require an executable fetch function).

## Why
The issue is a type-level mismatch between `query-db-collection` and TanStack Query option shapes. This removes boilerplate for existing Query users while preserving query-collection safety guarantees.

Reference: [original TanStack Query types file](https://github.com/TanStack/query/blob/main/packages/query-core/src/types.ts)

## ✅ Checklist

Note: there are some failed test files but I didn't changes them 

## 🚀 Release Impact

### DIFF
--- a/packages/query-db-collection/src/query.ts
+++ b/packages/query-db-collection/src/query.ts
@@ -77,7 +77,7 @@ type QueryOptionsInteropConfig<
   `queryFn` | `queryKey`
 > & {
   queryKey: TaggedQueryKey<TQueryKey, TQueryData, TError>
-  queryFn?: (
+  queryFn: (
     context: QueryFunctionContext<
       TaggedQueryKey<TQueryKey, TQueryData, TError>
     >,
@@ -617,15 +617,10 @@ export function queryCollectionOptions<
 }
 
 export function queryCollectionOptions(
-  config: Omit<
-    QueryCollectionConfig<
-      Record<string, unknown>,
-      (context: QueryFunctionContext<any>) => any
-    >,
-    `queryFn`
-  > & {
-    queryFn?: (context: QueryFunctionContext<any>) => any
-  },
+  config: QueryCollectionConfig<
+    Record<string, unknown>,
+    (context: QueryFunctionContext<any>) => any
+  >,
 ): CollectionConfig<
   Record<string, unknown>,
   string | number,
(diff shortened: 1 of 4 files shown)


packages/query-db-collection/src/query.ts
```diff
@@ -47,6 +48,41 @@ type InferSchemaInput<T> = T extends StandardSchemaV1
...
+    TError,
+    TaggedQueryKey<TQueryKey, TQueryData, TError>,
+    TKey,
+    TSchema,
+    TQueryData
+  >,
+  `queryFn` | `queryKey`
+> & {
+  queryKey: TaggedQueryKey<TQueryKey, TQueryData, TError>
+  queryFn?: (
```
kevin-dp: The types should correspond to runtime expectations. The query function is required at runtime so this should be reflected by the type, even if that is incompatible with TanStack query options.

hugiex: make sense, I've made the changes

kevin-dp:
Hi @hugiex, thanks for your contribution. I think we can simplify it a bit as we don't need the entire interop overloads. These simplified typings seem to work fine: https://github.com/TanStack/db/pull/1289. Would appreciate your thoughts about it.

hugiex:
Thanks for the suggestion! I'm totally good with simplifying it if this already covers what we need.

Happy to stick with the minimal approach as long as it fits the repo/team direction. I'll go ahead and close this PR 👍

PR #1294, opened 2026-02-25
feat: Subqueries in SELECT for hierarchical data (includes)

### LINKED ISSUE #288: Joins with a hierarchical projection (includes)
The data shown in a user interface if often hierarchical (projects -> issues -> comments)

The joins supported by Tanstack DB are SQL like joins where the resulting rows are flat, however users will often what to project to a hierarchical structure to render in the UI.

The proposed syntax is to use subqueries in the `select` clause:

```ts
const projects = new Query
  .from({ project: projectsCllctn })
  .where(....)
  .select(({ project }) => ({
    project,
    issues: new Query
      .from({ issue: issuesCllctn })
      .where(({ issue }) => 
        eq(issue.projectId, project.id)
      )
      .select(({ issue }) => ({
        issue,
        comments: new Query
          .from({ comment: commentsCllctn })
          .where(({ comment }) =>
            eq(comment.issueId, issue.id)
          )
      })
   })
```

kevin-dp:
## Summary

- Adds support for subqueries inside `.select()` that produce hierarchical results — each parent row gets a child Collection (e.g., projects with nested issues, issues with nested comments)
- Child queries are inner-joined with the parent pipeline so only children matching filtered parents flow through
- Supports ORDER BY and LIMIT/OFFSET on child queries (uses grouped ORDER BY so limits are per-parent)
- Nested includes work recursively (projects → issues → comments)

Closes #288

## Test plan

- [x] Basic includes: parent rows have child Collections with correct items
- [x] Reactivity: adding/removing children updates child Collections without touching parents
- [x] Parent remove + re-add: child Collection resets correctly
- [x] Inner join filtering: children only shown for parents matching WHERE
- [x] Nested includes: two levels deep (projects → issues → comments)
- [x] Ordered child queries: child Collections respect ORDER BY
- [x] Ordered + LIMIT: limit applied per parent, not globally; insertions displace correctly
- [x] All 1815 existing tests pass (no regressions)

### DIFF
--- a/.changeset/includes-subqueries.md
+++ b/.changeset/includes-subqueries.md
@@ -1,5 +1,5 @@
 ---
-'@tanstack/db': minor
+'@tanstack/db': patch
 ---
 
 feat: support for subqueries for including hierarchical data in live queries


packages/db/src/query/builder/index.ts
```diff
@@ -861,11 +862,126 @@ function buildNestedSelect(obj: any): any {
...
+
+  if (childQuery.where) {
+    for (let i = 0; i < childQuery.where.length; i++) {
+      const where = childQuery.where[i]!
+      const expr =
+        typeof where === `object` && `expression` in where
+          ? where.expression
+          : where
+
+      // Look for eq(a, b) where one side references parent and other references child
```
samwillis: I believe this is finding the first expression that references both sides, this is correct. We should consider what something like this does:

```ts
  id: p.id,
  name: p.name,
  issues: q
    .from({ i: issues })
    .where(({ i }) => and(eq(i.projectId, p.id)), eq(i.createdBy, p.createdBy))
    .select(({ i }) => ({
      id: i.id,
      title: i.title,
    })),
})),
)
```

I suspect it breaks at the moment, and so we may want to throw if there is more than one expression matching both sources.

I think it's possible to make this work though by pulling the parent project value temporarily into the child issue pipeline.

packages/db/src/query/live/collection-config-builder.ts
```diff
@@ -1132,13 +1265,583 @@ function extractCollectionAliases(query: QueryIR): Map<string, Set<string>> {
...
+  const resultKeys = new WeakMap<object, unknown>()
+  const orderByIndices = hasOrderBy ? new WeakMap<object, string>() : null
+  let syncMethods: SyncMethods<any> | null = null
+
+  const compare = orderByIndices
+    ? createOrderByComparator(orderByIndices)
+    : undefined
+
+  const collection = createCollection<any, string | number>({
+    id: `${parentId}-${fieldName}-${String(correlationKey)}`,
```
samwillis: tempted to prefix with `child-collection` or similar

packages/db/tests/query/includes.test.ts
```diff
@@ -0,0 +1,568 @@
...
+        value: { id: 1, name: `Alpha Reborn` },
+      })
+      projects.utils.commit()
+
+      const alpha = collection.get(1) as any
+      expect(alpha).toMatchObject({ id: 1, name: `Alpha Reborn` })
+      expect(childItems(alpha.issues)).toEqual([
+        { id: 10, title: `Bug in Alpha` },
+        { id: 11, title: `Feature for Alpha` },
+      ])
```
samwillis, line 265: 🥳

samwillis:
ChatGPT review:

----

Here’s my review of **TanStack/db PR #1294** (adds “includes” subqueries / nested child collections). ([[GitHub](https://github.com/TanStack/db/pull/1294)][1])

## What this PR is doing (as I understand it)

* **New “includes subquery” syntax**: returning a `QueryBuilder` from inside a parent `.select()` field now becomes an `IncludesSubquery` IR node, by detecting a correlating `eq(child.fk, parent.pk)` in the child query’s `where`. ([[GitHub](https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/src/query/builder/index.ts)][2])
* **Compiler support**: the compiler extracts those `IncludesSubquery` nodes, compiles the child query “per parent correlation key”, and plumbs an extra `correlationKey` through the result tuples so the output layer can route rows into the correct child collection. ([[GitHub](https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/src/query/compiler/index.ts)][3])
* **Output/runtime support**: live query builder wires child pipelines via `output()` callbacks, creates per-parent child `Collection`s, attaches them onto parent rows, and handles **nested includes** via a shared-buffer + routing-index approach. ([[GitHub](https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/src/query/live/collection-config-builder.ts)][4])
* **Tests**: good coverage for basic includes, reactivity (insert/delete), ordered children, per-parent `limit`, and 2-level nesting (projects → issues → comments). ([[GitHub](https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/tests/query/includes.test.ts)][5])
* **Related fix**: `containsAggregate` is now defensive against nested `Select` objects (important because includes introduces nested select-shaped objects). ([[GitHub](https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/src/query/compiler/group-by.ts)][6])

Overall: the API is *very* ergonomic, and the nested routing solution is clever.

---

## Things I like

* The user-facing API is dead simple and reads like a real ORM include. ([[GitHub](https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/tests/query/includes.test.ts)][5])
* The compiler approach (extract includes early, compile child with a parent-key stream, and then let the live layer attach real child collections) is a solid separation of concerns. ([[GitHub](https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/src/query/compiler/index.ts)][3])
* The tests hit the highest-risk areas: ordering + limit per parent + nested includes. ([[GitHub](https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/tests/query/includes.test.ts)][5])

---

## Key concerns / suggested changes

### 1) Alias collisions between parent and child queries (correctness bug risk)

`extractCorrelation()` decides “parent vs child” purely by membership in `parentAliases` / `childAliases`. If an alias name appears in **both** sets (e.g. user reuses `p` inside the child query), the correlation detection can mis-classify and/or silently do the wrong thing. ([[GitHub](https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/src/query/builder/index.ts)][2])

**Suggestion**

* Enforce **disjoint alias sets** for includes subqueries at build time:

  * If `childAliases` intersects `parentAliases`, throw a dedicated error (ideally the same family as `DuplicateAliasInSubqueryError` used elsewhere). ([[GitHub](https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/src/query/compiler/index.ts)][3])
* Also consider extending `validateQueryStructure` to walk `select` and validate nested `IncludesSubquery` nodes too (right now it looks focused on `from/join` QueryRefs). ([[GitHub](https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/src/query/compiler/index.ts)][3])

### 2) Correlation extraction only matches *top-level* `eq(ref, ref)`

Right now the includes correlation must be a direct `where(() => eq(child.fk, parent.pk))`. If someone writes:

```ts
.where(({ i }) => and(eq(i.projectId, p.id), eq(i.status, 'open')))
```

…correlation won’t be found (because it doesn’t traverse `and()` trees). That’s fine for v1, but it needs to be *explicit*.

**Suggestion**

* Either:

  1. document “correlation must be a top-level `where(eq(...))`”, **and** add a clearer error type/message; or
  2. traverse boolean expression trees (`and/or`) to find the first correlating `eq`.

The current thrown `Error(...)` is a bit “raw” for public API behavior. ([[GitHub](https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/src/query/builder/index.ts)][2])

### 3) Compiler mutates the query IR’s `select` in-place

`replaceIncludesInSelect(query.select, key)` replaces includes entries with `Val(null)` so `processSelect()` doesn’t see them. That’s convenient, but **in-place mutation** of a query object that may be cached/reused is a footgun (especially if you ever recompile without cache, or if other passes expect to see includes). ([[GitHub](https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/src/query/compiler/index.ts)][3])

**Suggestion**

* Treat IR as immutable here:

  * clone `select` into `selectWithoutIncludes`,
  * compile using the clone,
  * keep the original IR intact.

### 4) Parent “correlation key” changes on update aren’t handled cleanly

In `flushIncludesState`, Phase 1 runs on `changes.inserts > 0`, so it will also run for **updates** (which often show as delete+insert deltas internally). But Phase 5 only cleans up on “pure delete” (`deletes > 0 && inserts === 0`). ([[GitHub](https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/src/query/live/collection-config-builder.ts)][4])

If a parent row’s correlation value can change (even if rare), you can end up with:

* stale `correlationToParentKeys` membership under the old key,
* orphaned child collections / routing entries,
* parent row still pointing at the old child collection.

**Suggestion**

* Either explicitly state/validate that the correlation field must be stable (usually the parent primary key), **or**
* Track the last correlationKey per parent key and on update:

  * remove old mapping + routing entries when the correlationKey changes,
  * attach the new child collection.

Given the current code already supports “multiple parents per correlationKey”, it’s close—just missing the “move” case. ([[GitHub](https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/src/query/live/collection-config-builder.ts)][4])

### 5) Child collection IDs use `String(correlationKey)`

`id: ${parentId}-${fieldName}-${String(correlationKey)}`

* If `correlationKey` is an object/composite, `String()` becomes `"[object Object]"` → collisions.
* If it contains awkward chars, IDs get messy.

**Suggestion**

* If you want IDs to be stable + readable, consider `serializeValue(correlationKey)` (you’re already using `serializeValue` elsewhere) or a tiny hash of a stable serialization. ([[GitHub](https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/src/query/compiler/group-by.ts)][6])

### 6) Internal `__correlationKey` property naming

Compiler/runtime appears to rely on a magic `__correlationKey` on the source row for grouping/limit-per-parent logic. If user data can contain that field, there’s collision risk. ([[GitHub](https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/src/query/compiler/index.ts)][3])

**Suggestion**

* Use a `Symbol`, or a namespaced internal key that can’t realistically collide, or store correlation metadata outside user rows.

---

## Tests I’d add (small but high value)

1. **Alias collision**: parent uses alias `p`, child also uses alias `p` → assert it throws a clear error.
2. **Correlation inside `and()`**: demonstrate either supported traversal or a friendly “must be top-level eq” error.
3. **Parent update that changes correlation value** (even if discouraged): verify the old child collection detaches and the new one attaches (or verify it’s rejected).
4. **Multiple parents share same correlationKey** (since your reverse index supports it): ensure attach/update works for all parents in the set. ([[GitHub](https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/src/query/live/collection-config-builder.ts)][4])

---

## Bottom line

This is a strong feature with a very nice API and solid initial test coverage. The biggest things I’d address before merging are:

* **alias overlap validation** (likely correctness),
* **avoid mutating query IR** in compiler,
* **define behavior/constraints** for correlation extraction (top-level eq vs expression traversal),
* and **handle or forbid parent correlationKey changes**.

[1]: https://github.com/TanStack/db/pull/1294 "feat: Subqueries in SELECT for hierarchical data (includes) by kevin-dp · Pull Request #1294 · TanStack/db · GitHub"
[2]: https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/src/query/builder/index.ts "raw.githubusercontent.com"
[3]: https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/src/query/compiler/index.ts "raw.githubusercontent.com"
[4]: https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/src/query/live/collection-config-builder.ts "raw.githubusercontent.com"
[5]: https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/tests/query/includes.test.ts "raw.githubusercontent.com"
[6]: https://raw.githubusercontent.com/TanStack/db/15b98621076a4fb5031d330c7c984f56e63c5746/packages/db/src/query/compiler/group-by.ts "raw.githubusercontent.com"

packages/db/src/query/builder/index.ts
kevin-dp: Indeed, it breaks right now because the parent row is not in the child pipeline. I added support for this in this PR: https://github.com/TanStack/db/pull/1307

kevin-dp:
@samwillis response to codex' review:

> 1) Alias collisions between parent and child queries (correctness bug risk)

Valid concern but I don't think it's a real risk in practice. The child query is built via the builder API where the user explicitly declares aliases in `.from({ i: issues })`. If they reuse `p` as a child alias, it would shadow the parent's p in the closure scope — so p.id in the child's `.where()` would already refer to the child's p, not the parent's.

> 2) Correlation extraction only matches top-level eq(ref, ref)

Fixed in https://github.com/TanStack/db/pull/1307

> 3) Compiler mutates the query IR’s select in-place

This is a fair observation but the compiler already runs on the output of `optimizeQuery()`, which returns a new object. And the cache is keyed by the raw query, with queryMapping linking optimized → raw. So in practice the mutation happens on a fresh optimized copy, not the user's original IR.

Still a valid code hygiene concern, but it's out of scope for this PR. If it were to be fixed, it should be its own cleanup.

> 4) Parent “correlation key” changes on update aren’t handled cleanly

The correlation field is almost always the parent's primary key (e.g., `p.id` in `eq(i.projectId, p.id)`). PKs don't change by definition.

Could a user correlate on a non-PK field? Technically yes, but it would be semantically wrong — the correlation field determines how children are grouped to parents. If it's not stable, the entire grouping model is broken, not just the cleanup logic.

I'd lean toward the first suggestion: document/validate that the correlation field should be a stable key (which it naturally is in every real use case). The "move" case handling would add complexity for a scenario that doesn't really make sense to support.

> 5) Child collection IDs use `String(correlationKey)`

True — the correlation field doesn't have to be the PK (which are restricted to `string | number`). Someone could correlate on any field, and arbitrary field values could be objects, arrays, dates, etc. Even though it's unusual, `String()` would silently produce `"[object Object]"` and cause collisions.

Using `serializeValue` is a cheap one-line fix that makes it robust. Will do.

> 6) Internal `__correlationKey` property naming

This is fine. We have a couple of these reserved properties around the code. It's unlikely someone would use this name.


Regarding the additional tests:
1. alias collisions: isn't needed because as we explained this is handled by standard shadowing in TS.
2. correlation inside `and()`: these tests already exists in follow up PRs since we introduced support for this.
3. Parent update that changes correlation value: As discussed, the correlation field is practically always the PK which doesn't change. Testing this would be testing undefined behavior. I'd rather document the constraint than test around it.
4. Multiple parents sharing same correlationKey: This is a good one. It tests a real scenario (e.g., multiple
  projects with the same foreign key value). Added this one.

KyleAMathews:
## Code Review

### Bugs

**1. `__correlationKey` leaks into child results when child query omits `.select()`**
`compiler/index.ts:211` stamps `__correlationKey` onto the child row. When the child has no explicit `.select()`, lines 376-377 set `$selected = namespacedRow[mainSource]` — the raw row including `__correlationKey`. It passes straight through `unwrapValue` at line 479 and becomes a visible property on every child Collection item. All tests use `.select()` on child queries so this path is untested.

Fix: strip `__correlationKey` from `finalResults` when `parentKeyStream` is present (similar to how `__includesCorrelationKeys` is cleaned up at lines 1791-1796 of collection-config-builder.ts).

**2. `replaceIncludesInSelect` mutates the shared IR `select` object**
`compiler/index.ts:326` — the optimizer copies `select` by reference (`optimizer.ts:754`), so `replaceIncludesInSelect(query.select, key)` permanently replaces `IncludesSubquery` entries on both the optimized AND raw query. Today the compilation cache masks this (nobody re-reads `rawQuery.select`), but it violates the immutable-IR convention the rest of the codebase follows. A shallow clone before mutation would be cheap insurance:
```ts
const selectCopy = { ...query.select }
query = { ...query, select: selectCopy }
```

**3. Nested `IncludesSubquery` in nested select silently produces `null`**
`compiler/index.ts:877-887` — `extractIncludesFromSelect` only checks top-level select entries. But the builder's `buildNestedSelect` converts `BaseQueryBuilder` at any depth, so `.select(({p}) => ({ info: { issues: childQuery } }))` produces a nested `IncludesSubquery` that the compiler never extracts. The child pipeline is never compiled; the user gets `null` with no error. Either make extraction recursive or throw when a nested `IncludesSubquery` is detected.

### Test gaps

**4. No test for updating an existing child row** — Insert and delete are covered, but the update branch in `flushIncludesState` (collection-config-builder.ts:1714-1726) is never exercised. Updating children is the most common real-world mutation. A test that changes an issue's `title` and asserts the child collection reflects it would cover this.

**5. No test for the error on missing/invalid correlation** — `buildIncludesSubquery` (builder/index.ts:930-935) throws when no valid `eq()` correlation is found. This is the primary user-facing validation for the feature and has zero test coverage. Worth testing with: no WHERE at all, a WHERE without `eq()`, and an `eq()` referencing two child-side aliases.

**6. No test for multiple sibling includes on the same parent** — Every test uses one includes field. A parent with both `issues` and `milestones` would verify that multiple correlation stamps, child registries, and output callbacks operate independently.

### Minor

**7. Child collections not explicitly cleaned up on teardown** — When the parent sync ends (collection-config-builder.ts:636), `includesCache` is nulled but child Collections created with `startSync: true` are never explicitly stopped. Probably fine via GC in practice, but something to watch in long-running apps with frequent parent churn.

**8. Type system has no awareness of includes** — `SelectValue` doesn't include `QueryBuilder` and `ResultTypeFromSelect` has no branch for it. Users get zero autocomplete or type safety on child collection fields. Makes sense as a follow-up.

kevin-dp:
**Addressing review feedback from #1294 (comment)**

### Bugs fixed

**1. `__correlationKey` leaks into child results when child query omits `.select()`** — Fixed in f8327ed4. Strip `__correlationKey` from `finalResults` before returning when `parentKeyStream` is present.

**2. `replaceIncludesInSelect` mutates the shared IR `select` object** — Fixed in 0e0e3cb6. Shallow-clone `query.select` before mutating when includes entries are found, so the original IR is preserved.

**3. Nested `IncludesSubquery` in nested select silently produces `null`** — Fixed in 1ef6098b and 1ab02ba9. `extractIncludesFromSelect` now recursively checks nested objects and throws a clear error when a nested `IncludesSubquery` is detected. Also skips `__SPREAD_SENTINEL__` entries to avoid infinite recursion on RefProxy objects.

### Test gaps filled

**4. No test for updating an existing child row** — Added in 29c34f54. Exercises the update branch in `flushIncludesState`.

**5. No test for the error on missing/invalid correlation** — Added in 7f1066ab. Tests: no WHERE at all, WHERE without `eq()`, and `eq()` referencing two child-side aliases.

**6. No test for multiple sibling includes on the same parent** — Added in 7f1066ab. Parent with both `issues` and `milestones`, verifying independent child collections and independent reactivity.

### Minor items

**7. Child collections not explicitly cleaned up on teardown** — Agreed this is fine via GC for now, something to watch.

**8. Type system has no awareness of includes** — Fixed in 5db9ae0a and 50abd991. Added `QueryBuilder<any>` to `SelectValue` and a `ResultTypeFromSelect` branch mapping `QueryBuilder<TChildContext>` → `Collection<GetResult<TChildContext>>`. Covered by 4 new type tests.

.changeset/includes-subqueries.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tanstack/db': minor
```
samwillis:
```suggestion
'@tanstack/db': patch
```

the v0.6 will be triggered by the breaking changes with opt-in indexing and removal of the magic return values

PR #1295, opened 2026-02-25
feat: add toArray() for includes subqueries

kevin-dp:
## Summary
- Adds `toArray()` wrapper that converts child Collections to plain `Array<T>` on parent rows
- When children change, the parent row is re-emitted with a fresh array snapshot
- Supports ordering and limits on toArray subqueries

## Test plan
- [x] Basic toArray produces arrays, not Collections
- [x] Empty parents get `[]`
- [x] Adding/removing a child re-emits parent with updated array
- [x] Array respects ORDER BY
- [x] Ordered toArray with LIMIT applied per parent
- [x] Existing Collection-based includes still work unchanged
- [x] Full db test suite passes with no regressions

### DIFF
--- a/packages/db/src/query/live/collection-config-builder.ts
+++ b/packages/db/src/query/live/collection-config-builder.ts
@@ -1805,8 +1805,7 @@ function flushIncludesState(
         for (const parentKey of parentKeys) {
           const item = parentCollection.get(parentKey as any)
           if (item) {
-            const key =
-              parentSyncMethods.collection.getKeyFromItem(item)
+            const key = parentSyncMethods.collection.getKeyFromItem(item)
             // Capture previous value before in-place mutation
             const previousValue = { ...item }
             if (entry) {


samwillis:
Review from GPT5.4:

----

## Findings

1. `packages/db/src/query/builder/functions.ts`, `packages/db/src/query/builder/types.ts`

`toArray()` isn’t wired into the result type inference, so the new public API loses type safety even though the runtime behavior works.

Right now `toArray()` returns a non-generic `ToArrayWrapper`, and `ResultTypeFromSelect` never maps that wrapper to `Array<GetResult<...>>`. As a result, a projection like `issues: toArray(...)` won’t infer as `Issue[]` on the parent row, which is a user-visible regression for the feature this PR is adding.

I think this needs a dedicated type-level branch in the select result inference so the wrapper unwraps to the child query result array type.

2. `packages/db/tests/query/includes.test.ts`, `packages/db/tests/query/*.test-d.ts`

The runtime coverage here is good, but there’s no `.test-d.ts` coverage for the new `toArray()` API. That leaves the typing regression above completely unguarded.

A focused type test around something like `issues: toArray(q.from(...).select(...))` and an assertion that the parent result field is `Array<...>` would catch this immediately and help keep the public API stable.

samwillis:
This is looking very good, other than the type issue that GPT5.4 found above, the other thing we need to do is ensure there are tests that validate the reactivity model:

- non-`toArray` subscribe to the main result collection and validate that changes that result in a child change don't trigger a change event on the parent.
- `toArray` same, but validate that there *is* an event on the child collection.

I believe these tests will pass, but we should have them as they are a core design feature.

PR #1300, opened 2026-02-25
fix(offline-transactions): enforce strict FIFO when retries are pending

KyleAMathews:
## Summary

Enforce strict FIFO ordering in `KeyScheduler.getNextBatch` so that a transaction waiting for retry delay blocks all later transactions from executing. Previously, later transactions could jump the queue, causing dependent UPDATE-before-INSERT failures.

## Root Cause

`getNextBatch` used `.find()` to locate **any** pending transaction that was ready to run. If the oldest transaction had failed and was in a retry backoff window (`nextAttemptAt` in the future), `.find()` would skip past it and return a later transaction that was ready. This violated the FIFO contract — a newer transaction could execute before an older one that it depended on.

## Approach

Replace the `.find()` scan with a check on only `this.pendingTransactions[0]` (the oldest transaction). If it's not ready, return an empty batch. If it is ready, return it.

```typescript
// Before: find ANY ready transaction (could skip the first)
const readyTransaction = this.pendingTransactions.find((tx) =>
  this.isReadyToRun(tx),
)

// After: only check the FIRST transaction (strict FIFO)
const firstTransaction = this.pendingTransactions[0]!
if (!this.isReadyToRun(firstTransaction)) {
  return []
}
return [firstTransaction]
```

The `scheduleNextRetry` mechanism in `TransactionExecutor` ensures that when the first transaction's retry delay elapses, execution resumes automatically.

### Key Invariants

- The queue always processes in `createdAt` order — no transaction can execute before an older one
- When the head transaction is in retry backoff, the entire queue waits
- `scheduleNextRetry` provides the eventual recovery path after a break

### Non-goals

- No changes to the retry/backoff timing logic itself
- No changes to `TransactionExecutor` or the execution loop
- Pre-existing concerns (e.g., `markStarted`/`markFailed` not validating transaction identity) are out of scope

## Verification

```bash
pnpm --filter @tanstack/offline-transactions test -- KeyScheduler.test.ts
```

8 tests cover: FIFO blocking during retry, delay elapse, queue advancement after completion, `isRunning` guard, full retry-then-success lifecycle, out-of-order scheduling, `updateTransaction` sort preservation, and empty queue boundary.

## Files Changed

- **`packages/offline-transactions/src/executor/KeyScheduler.ts`** — Replace `.find()` scan with strict first-element check; use non-null assertion after length guard
- **`packages/offline-transactions/tests/KeyScheduler.test.ts`** — New test file with 8 tests covering the FIFO ordering contract

---

### DIFF
--- a/packages/offline-transactions/src/executor/KeyScheduler.ts
+++ b/packages/offline-transactions/src/executor/KeyScheduler.ts
@@ -22,15 +22,14 @@ export class KeyScheduler {
     )
   }
 
-  getNextBatch(_maxConcurrency: number): Array<OfflineTransaction> {
+  getNext(): OfflineTransaction | undefined {
     return withSyncSpan(
-      `scheduler.getNextBatch`,
+      `scheduler.getNext`,
       { pendingCount: this.pendingTransactions.length },
       (span) => {
-        // For sequential processing, we ignore maxConcurrency and only process one transaction at a time
         if (this.isRunning || this.pendingTransactions.length === 0) {
           span.setAttribute(`result`, `empty`)
-          return []
+          return undefined
         }
 
         const firstTransaction = this.pendingTransactions[0]!
@@ -38,12 +37,12 @@ export class KeyScheduler {
         if (!this.isReadyToRun(firstTransaction)) {
           span.setAttribute(`result`, `waiting_for_first`)
           span.setAttribute(`transaction.id`, firstTransaction.id)
-          return []
+          return undefined
         }
 
         span.setAttribute(`result`, `found`)
         span.setAttribute(`transaction.id`, firstTransaction.id)
-        return [firstTransaction]
+        return firstTransaction
       },
     )
   }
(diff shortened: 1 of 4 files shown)


packages/offline-transactions/src/executor/KeyScheduler.ts
```diff
@@ -33,19 +33,17 @@ export class KeyScheduler {
...
+        if (!this.isReadyToRun(firstTransaction)) {
+          span.setAttribute(`result`, `waiting_for_first`)
+          span.setAttribute(`transaction.id`, firstTransaction.id)
+          return []
         }
 
-        return readyTransaction ? [readyTransaction] : []
+        span.setAttribute(`result`, `found`)
+        span.setAttribute(`transaction.id`, firstTransaction.id)
+        return [firstTransaction]
```
kevin-dp: nit: `getNextBatch` has return type `Array<OfflineTransaction>` but it always returns 0 or 1 transaction. So `OfflineTransaction | undefined` or `OfflineTransaction | null` would be slightly more precise.

kevin-dp (approved):

PR #1301, opened 2026-02-25
fix(offline-transactions): infinite retries by default and online-status gating

KyleAMathews:
## Fix offline transaction retry behavior and add online-status gating

Transactions were being permanently dropped after 10 failed retries, and retry attempts were wasted while the device was offline. Both issues are now fixed — retries are infinite by default and gated on connectivity.

## Root Cause

Two separate issues in `@tanstack/offline-transactions`:

1. **`DefaultRetryPolicy` hardcoded `maxRetries = 10`** — After 10 transient failures (e.g., a brief server outage), transactions were treated as permanent failures, removed from the outbox, and lost.

2. **No online-status awareness in the execution loop** — The executor would attempt transactions regardless of connectivity, burning through retries against an unreachable server.

## Approach

**Infinite retries by default**: Changed `DefaultRetryPolicy` from `maxRetries = 10` to `maxRetries = Infinity`. Retriable errors (anything that isn't `NonRetriableError`, `AbortError`, or a 4xx status) will keep retrying with exponential backoff until the server responds.

**Online-status gating**: Added `isOnline()` checks at two points:
- `TransactionExecutor.runExecution()` — breaks out of the batch loop when offline
- `TransactionExecutor.scheduleNextRetry()` — skips scheduling retry timers when offline

When the device comes back online, the `OnlineDetector` subscription handler resets retry delays and triggers execution immediately.

**`OnlineDetector.isOnline()` is now required** on the interface. Both existing implementations (`WebOnlineDetector`, `ReactNativeOnlineDetector`) already provided it. Making it required eliminates `typeof` guards with dangerous `return true` fallbacks.

**`OfflineExecutor.notifyOnline()` removed** — this method was redundant with the `OnlineDetector` subscription. The detector is the single source of truth for connectivity. Users who need imperative control over online status should implement a custom `OnlineDetector`.

## Key Invariants

- Retriable errors never stop retrying (only `NonRetriableError` and specific status codes are permanent)
- No transaction execution is attempted while `isOnline()` returns false
- The `OnlineDetector` is the single source of truth for connectivity — no force flags or override mechanisms

## Non-goals

- Configurable `maxRetries`/`retryDelay`/`retryPolicy` — stripped from the PR to avoid unnecessary API surface. Can be added when there's demand.
- Retry timer suppression test coverage for online→offline→online transitions — the guard logic is simple and the offline→online path is tested.

## Trade-offs

**Breaking interface changes shipped as patch**: `OnlineDetector.isOnline()` is now required and `OfflineExecutor.notifyOnline()` is removed. This is technically a breaking change, but this package is pre-1.0 with very few users, making a patch appropriate.

## Verification

```bash
pnpm --filter @tanstack/offline-transactions test
```

## Files changed

- **src/retry/RetryPolicy.ts** — Default `maxRetries` changed from 10 to `Infinity`
- **src/executor/TransactionExecutor.ts** — Online gating in execution loop and retry scheduling; `TransactionSignaler` interface replaces `any` type
- **src/OfflineExecutor.ts** — Removed `notifyOnline()`; added `isOnline()` method; subscription handler triggers execution on connectivity change
- **src/connectivity/ReactNativeOnlineDetector.ts** — Added `isOnline()`, initial `NetInfo.fetch()`, extracted `toConnectivityState()`
- **src/types.ts** — `OnlineDetector.isOnline()` now required; added `TransactionSignaler` interface
- **tests/offline-e2e.test.ts** — Regression test for >10 retries; offline execution gating test
- **tests/harness.ts** — Simplified config passthrough with spread
- **README.md** — Updated API reference, removed speculative retry config sections

### DIFF
--- a/packages/offline-transactions/src/connectivity/ReactNativeOnlineDetector.ts
+++ b/packages/offline-transactions/src/connectivity/ReactNativeOnlineDetector.ts
@@ -120,6 +120,6 @@ export class ReactNativeOnlineDetector implements OnlineDetector {
     isConnected: boolean | null
     isInternetReachable: boolean | null
   }): boolean {
-    return state.isConnected === true && state.isInternetReachable !== false
+    return !!state.isConnected && state.isInternetReachable !== false
   }
 }


packages/offline-transactions/src/connectivity/ReactNativeOnlineDetector.ts
```diff
@@ -98,8 +107,19 @@ export class ReactNativeOnlineDetector implements OnlineDetector {
...
   dispose(): void {
     this.stopListening()
     this.listeners.clear()
   }
+
+  private toConnectivityState(state: {
+    isConnected: boolean | null
+    isInternetReachable: boolean | null
+  }): boolean {
+    return state.isConnected === true && state.isInternetReachable !== false
```
kevin-dp: could write `state.isConnected` instead of `=== true`.

KyleAMathews: Done, simplified to `!!state.isConnected`.

PR #1307, opened 2026-02-26
Support parent-referencing WHERE filters in includes child queries

kevin-dp:
## Summary
- Allow child queries in `includes` to have additional WHERE clauses that reference parent fields (e.g., `eq(i.createdBy, p.createdBy)`), beyond the single correlation `eq()`
- When no parent-referencing filters exist, behavior stays exactly as-is with zero overhead
- Parent-referencing WHEREs are detected in the builder, parent fields are projected into the key stream, and filters are re-injected into the child query where parent context is available

## Test plan
- [x] Basic parent-referencing filter: correlation `eq(i.projectId, p.id)` + filter `eq(i.createdBy, p.createdBy)` — only matching issues appear
- [x] Reactivity — parent field change: updating `p.createdBy` causes child results to update
- [x] Reactivity — child field change: updating `i.createdBy` causes issue to appear/disappear
- [x] Mixed filters: parent-referencing WHERE + pure-child WHERE both apply correctly
- [x] All existing includes tests pass (no regression)

### DIFF
--- a/packages/db/src/query/compiler/index.ts
+++ b/packages/db/src/query/compiler/index.ts
@@ -462,7 +462,6 @@ export function compileQuery(
       // Replace includes entry in select with a null placeholder
       replaceIncludesInSelect(query.select!, key)
     }
-
   }
 
   if (query.distinct && !query.fnSelect && !query.select) {


samwillis:
GPT5.4 found an interesting problem with this PR:

---

## Blocking Review Comment

I think this PR has a correctness issue in the new parent-referencing filter path.

The implementation now allows the child include query to depend on additional parent fields, but the includes runtime still appears to materialize and route child collections solely by the correlation key. That means the results become incorrect whenever:

1. multiple parent rows share the same correlation key, and
2. the extra parent-referencing filters would produce different child result sets for those parents

### Why this matters

Before this PR, includes effectively meant "one child collection per correlation value", which was fine because the child result only depended on that correlation itself.

With this PR, the child query can now depend on more parent state than just the correlation field, for example:

```ts
issues: q
  .from({ i: issues })
  .where(({ i }) => eq(i.projectGroupId, p.groupId))
  .where(({ i }) => eq(i.createdBy, p.createdBy))
```

At that point, the result is no longer a function of `groupId` alone. But the runtime still seems to cache/attach child collections by just that correlation key.

### Concrete failing case

Consider this data:

```ts
parents = [
  { id: 1, groupId: 1, createdBy: 'alice' },
  { id: 2, groupId: 1, createdBy: 'bob' },
]

children = [
  { id: 10, groupId: 1, createdBy: 'alice' },
]
```

Query:

```ts
q.from({ p: parents }).select(({ p }) => ({
  id: p.id,
  createdBy: p.createdBy,
  items: q
    .from({ c: children })
    .where(({ c }) => eq(c.groupId, p.groupId))
    .where(({ c }) => eq(c.createdBy, p.createdBy))
    .select(({ c }) => ({
      id: c.id,
      createdBy: c.createdBy,
    })),
}))
```

Expected result:

```ts
[
  {
    id: 1,
    createdBy: 'alice',
    items: [{ id: 10, createdBy: 'alice' }],
  },
  {
    id: 2,
    createdBy: 'bob',
    items: [],
  },
]
```

Actual result with the current implementation:

```ts
[
  {
    id: 1,
    createdBy: 'alice',
    items: [{ id: 10, createdBy: 'alice' }],
  },
  {
    id: 2,
    createdBy: 'bob',
    items: [{ id: 10, createdBy: 'alice' }],
  },
]
```

### Where I think the issue comes from

In `packages/db/src/query/compiler/index.ts`, the new code correctly projects parent context into the child pipeline and re-injects parent filters there. That means child rows are evaluated using parent-specific state.

But in `packages/db/src/query/live/collection-config-builder.ts`, includes output state still appears to maintain one child collection per `correlationKey`, with parents attached through `correlationToParentKeys`. So once two parents share the same correlation key, they also share the same child collection instance.

That was a valid design assumption before, but I don't think it is valid anymore once extra parent-referencing predicates are allowed.

### Important nuance

I do **not** think this is just "non-PK correlation is bad".

The real issue is: the correlation key no longer uniquely determines the child result set.

If all parents sharing a correlation key would also share the same values for every parent field used by the extra filters, then behavior is still fine. But in the general case the PR advertises, that is not guaranteed.

### Suggested direction

I think this needs one of these before merge:

1. change the includes runtime so child collections are keyed by a parent-specific identity that includes the projected parent filter context, not just the correlation key, or
2. explicitly reject parent-referencing filters unless the implementation can guarantee one parent-specific child collection per distinct parent filter context

At minimum, I think this needs a regression test covering duplicate correlation keys with differing parent filter values, because the current test suite only covers cases where the correlation key effectively identifies the parent result set.

### Requested test

Please add a test for:

- two parents with the same correlation key
- different value in a parent-referenced filter field
- one child matching only one of those parents
- verify the two parents receive different child collections
- verify reactivity still works when either the parent filter field or child field changes

I’d consider this blocking, because the PR introduces a new semantic capability, but the current routing model still assumes child results are only a function of the correlation key.

samwillis:
I think this suggestion above makes sense:

> change the includes runtime so child collections are keyed by a parent-specific identity that includes the projected parent filter context, not just the correlation key

samwillis:
@kevin-dp I had the same previous GPT5.4 session review the changes, this is its feedback:

---

## Findings
1. High: parent-specific routing is still lost for child queries with `orderBy` + `limit`/`offset`.

The unlimited case looks fixed, but the top-k path still groups includes by raw `__correlationKey`, not by the new composite routing identity. So if two parents share the same correlation key but differ on parent-referenced filters, their child rows are still collapsed together during grouped ordering.

```560:598:packages/db/src/query/compiler/index.ts
  // Process orderBy parameter if it exists
  if (query.orderBy && query.orderBy.length > 0) {
    // When in includes mode with limit/offset, use grouped ordering so that
    // the limit is applied per parent (per correlation key), not globally.
    const includesGroupKeyFn =
      parentKeyStream &&
      (query.limit !== undefined || query.offset !== undefined)
        ? (_key: unknown, row: unknown) =>
            (row as any)?.[mainSource]?.__correlationKey
        : undefined

    const orderedPipeline = processOrderBy(
      rawQuery,
      pipeline,
      query.orderBy,
      query.select || {},
      collections[mainCollectionId]!,
      optimizableOrderByCollections,
      setWindowFn,
      query.limit,
      query.offset,
      includesGroupKeyFn,
    )

    // Final step: extract the $selected and include orderBy index
    const resultPipeline: ResultStream = orderedPipeline.pipe(
      map(([key, [row, orderByIndex]]) => {
        // Extract the final results from $selected and include orderBy index
        const raw = (row as any).$selected
        const finalResults = unwrapValue(raw)
        // When in includes mode, embed the correlation key and parentContext
        if (parentKeyStream) {
          const correlationKey = (row as any)[mainSource]?.__correlationKey
          const parentContext = (row as any).__parentContext ?? null
          return [
            key,
            [finalResults, orderByIndex, correlationKey, parentContext],
          ] as any
```

The runtime now computes a composite routing key later:

```1584:1593:packages/db/src/query/live/collection-config-builder.ts
 * Computes a composite routing key from correlation key and parent context.
 * When parentContext is null (no parent filters), returns the raw correlationKey
 * for zero behavioral change on existing queries.
 */
function computeRoutingKey(
  correlationKey: unknown,
  parentContext: Record<string, any> | null,
): unknown {
  if (parentContext == null) return correlationKey
  return JSON.stringify([correlationKey, parentContext])
}
```

…but by then it is too late for `limit`/`offset`, because the grouped ordering window has already been computed per bare correlation key.

I reproduced this locally with:

- parents:
  - `{ id: 1, groupId: 1, createdBy: 'alice' }`
  - `{ id: 2, groupId: 1, createdBy: 'bob' }`
- children:
  - `{ id: 10, groupId: 1, createdBy: 'alice' }`
  - `{ id: 11, groupId: 1, createdBy: 'bob' }`

and query:

```ts
items: q
  .from({ c: children })
  .where(({ c }) => eq(c.groupId, p.groupId))
  .where(({ c }) => eq(c.createdBy, p.createdBy))
  .orderBy(({ c }) => c.id, 'asc')
  .limit(1)
```

Expected:

- parent 1 gets child 10
- parent 2 gets child 11

Actual:

- parent 1 gets child 10
- parent 2 gets `[]`

The existing includes limit tests only cover plain correlation and therefore still pass:

```375:452:packages/db/tests/query/includes.test.ts
  describe(`ordered child queries with limit`, () => {
    it(`limits child collection to N items per parent`, async () => {
          id: p.id,
          name: p.name,
          issues: q
            .from({ i: issues })
            .where(({ i }) => eq(i.projectId, p.id))
            .orderBy(({ i }) => i.title, `asc`)
            .limit(1)
            .select(({ i }) => ({
              id: i.id,
              title: i.title,
            })),
        })),
      )
      // ...
```

## Summary
The original shared-correlation bug appears addressed for unrestricted child queries, but I would still block on the `orderBy` + `limit`/`offset` case. The grouping key for top-k selection needs to switch from raw correlation key to the same composite routing identity used by the includes runtime, and the test suite needs one regression test that combines:

- duplicate correlation key across parents
- parent-referencing child filter
- child `orderBy`
- child `limit` or `offset`

I didn’t find another concrete regression beyond that in this pass.

samwillis:
## Findings
1. High: nested parent-referencing includes are still broken.

The top-level shared-correlation cases now look fixed, including the `orderBy + limit` regression. But I reproduced a broader failure one level deeper: when a nested include also uses parent-referencing filters, the grandchild collection comes back empty even in the simplest single-parent case.

Repro I ran locally:

- projects:
  - `{ id: 1, groupId: 1, createdBy: 'alice' }`
- issues:
  - `{ id: 10, groupId: 1, createdBy: 'alice', categoryId: 7 }`
- comments:
  - `{ id: 100, categoryId: 7, createdBy: 'alice', body: 'a' }`

Query:

```ts
  id: p.id,
  issues: q
    .from({ i: issues })
    .where(({ i }) => eq(i.groupId, p.groupId))
    .where(({ i }) => eq(i.createdBy, p.createdBy))
    .select(({ i }) => ({
      id: i.id,
      createdBy: i.createdBy,
      categoryId: i.categoryId,
      comments: q
        .from({ c: comments })
        .where(({ c }) => eq(c.categoryId, i.categoryId))
        .where(({ c }) => eq(c.createdBy, i.createdBy))
        .select(({ c }) => ({ id: c.id, createdBy: c.createdBy, body: c.body })),
    })),
}))
```

Expected:
- issue `10` should contain comment `100`

Actual:
- issue `10` had `comments: []`

That points to the nested routing path still being incomplete for parent-referencing filters.

The likely failure path is that nested buffers are keyed by the composite routing key, but the nested routing index is still populated and looked up using only `nestedCorrelationKey`, not the full routing identity:

```1513:1528:packages/db/src/query/live/collection-config-builder.ts
  for (const setup of state.nestedSetups) {
    for (const [, change] of childChanges) {
      if (change.inserts > 0) {
        // Read the nested correlation key from the INCLUDES_ROUTING stamp
        const nestedRouting =
          change.value[INCLUDES_ROUTING]?.[setup.compilationResult.fieldName]
        const nestedCorrelationKey = nestedRouting?.correlationKey

        if (nestedCorrelationKey != null) {
          state.nestedRoutingIndex!.set(nestedCorrelationKey, correlationKey)
          let reverseSet = state.nestedRoutingReverseIndex!.get(correlationKey)
          if (!reverseSet) {
            reverseSet = new Set()
            state.nestedRoutingReverseIndex!.set(correlationKey, reverseSet)
          }
          reverseSet.add(nestedCorrelationKey)
        }
```

while the nested shared buffer is already using `computeRoutingKey(correlationKey, parentContext)`:

```1355:1374:packages/db/src/query/live/collection-config-builder.ts
    entry.pipeline.pipe(
      output((data) => {
        const messages = data.getInner()
        syncState.messagesCount += messages.length

        for (const [[childKey, tupleData], multiplicity] of messages) {
          const [childResult, _orderByIndex, correlationKey, parentContext] =
            tupleData as unknown as [
              any,
              string | undefined,
              unknown,
              Record<string, any> | null,
            ]

          const routingKey = computeRoutingKey(correlationKey, parentContext)

          let byChild = buffer.get(routingKey)
          if (!byChild) {
            byChild = new Map()
            buffer.set(routingKey, byChild)
          }
```

And nested attachment still depends on reading routing metadata from the parent result value:

```483:497:packages/db/src/query/compiler/index.ts
  if (includesRoutingFns.length > 0) {
    pipeline = pipeline.pipe(
      map(([key, namespacedRow]: any) => {
        const routing: Record<
          string,
          { correlationKey: unknown; parentContext: Record<string, any> | null }
        > = {}
        for (const { fieldName, getRouting } of includesRoutingFns) {
          routing[fieldName] = getRouting(namespacedRow)
        }
        namespacedRow.$selected[INCLUDES_ROUTING] = routing
        return [key, namespacedRow]
      }),
    )
  }
```

```1667:1674:packages/db/src/query/live/collection-config-builder.ts
          const parentResult = changes.value
          // Extract routing info from INCLUDES_ROUTING symbol (set by compiler)
          const routing = parentResult[INCLUDES_ROUTING]?.[state.fieldName]
          const correlationKey = routing?.correlationKey
          const parentContext = routing?.parentContext ?? null
          const routingKey = computeRoutingKey(correlationKey, parentContext)
```

So I’d still block on this. The current added tests cover:
- top-level shared correlation key with parent-referenced filters
- top-level shared correlation key with `orderBy + limit`

But they still don’t cover:
- nested includes where the grandchild query references fields from the child include parent

## Notes
I reran `pnpm --filter @tanstack/db test -- packages/db/tests/query/includes.test.ts`; it passes. The issue above is not covered by the current test file, which is why it still slips through.

A good next regression test would be:
- one project
- one issue selected by a parent-referencing filter
- one comment selected by a second-level parent-referencing filter against the issue
- assert the comment appears
- then add the duplicate-correlation variant at the nested level once the base nested case works

kevin-dp:
## Addressed review findings

### Critical #1: Premature child collection deletion when parents share a routing key
**Fixed.** Two changes:
1. In Phase 5, moved `childRegistry.delete(routingKey)` inside the `parentKeys.size === 0` check so the child collection is only deleted when the last parent referencing it is removed.
2. Added a `reduce` operator on the `parentKeyStream` to clamp multiplicities to 1. This prevents the inner join from producing duplicate child entries (multiplicity > 1) when multiple parents share the same correlation key, which was causing incorrect deletions in the output callback when one parent was removed.

### Critical #2: Grouped ordering uses raw correlation key
**Already fixed** in a prior commit (cf0f28c5). `includesGroupKeyFn` now incorporates `__parentContext` into the group key.

### Important #3: `__correlationKey` leaks into user results with spread selects
**Fixed.** Added `delete finalResults.__correlationKey` and `delete finalResults.__parentContext` in both output paths (with and without orderBy), after extracting the routing values from the raw namespaced row.

### Important #4: `replaceIncludesInSelect` mutates shared select reference
Deferred — low practical impact since query IR is typically single-use.

All fixes have corresponding regression tests committed before the fix.

kevin-dp:
Re: https://github.com/TanStack/db/pull/1307#issuecomment-4032096319

## Addressed: nested parent-referencing includes routing

The nested routing index was using the raw `nestedCorrelationKey` while the nested shared buffers were already using `computeRoutingKey(correlationKey, parentContext)`. This mismatch caused grandchild collections to come back empty when nested includes used parent-referencing filters.

**Fixed** in commit dcc6b7cd by updating `updateRoutingIndex` to use `computeRoutingKey(nestedCorrelationKey, nestedParentContext)` in both the insert and delete paths, matching the composite key used by the nested buffers.

Also verified the fix works for 3+ levels of nesting (commit 9e5a6c4d adds a test for three levels of nested includes with parent-referencing filters at each level).

PR #1317, opened 2026-03-02
Docs for includes

kevin-dp:
Add an Includes section to `docs/guides/live-queries.md` covering:
- Basic includes with correlation conditions
- Additional filters including parent-referencing WHERE clauses
- Ordering and limiting per parent
- `toArray()` for plain array results
- Aggregates per parent
- Nested includes

### DIFF
--- a/packages/react-db/tests/useLiveQuery.test.tsx
+++ b/packages/react-db/tests/useLiveQuery.test.tsx
@@ -2481,4 +2481,112 @@ describe(`Query Collections`, () => {
       })
     })
   })
+
+  describe(`includes subqueries`, () => {
+    type Project = {
+      id: string
+      name: string
+    }
+
+    type ProjectIssue = {
+      id: string
+      title: string
+      projectId: string
+    }
+
+    const sampleProjects: Array<Project> = [
+      { id: `p1`, name: `Alpha` },
+      { id: `p2`, name: `Beta` },
+    ]
+
+    const sampleProjectIssues: Array<ProjectIssue> = [
+      { id: `i1`, title: `Bug in Alpha`, projectId: `p1` },
+      { id: `i2`, title: `Feature for Alpha`, projectId: `p1` },
+      { id: `i3`, title: `Bug in Beta`, projectId: `p2` },
+    ]
+
+    it(`should render includes results and reactively update child collections`, async () => {
+      const projectsCollection = createCollection(
+        mockSyncCollectionOptions<Project>({
+          id: `includes-react-projects`,
+          getKey: (p) => p.id,
+          initialData: sampleProjects,
+        }),
+      )
+
+      const issuesCollection = createCollection(
+        mockSyncCollectionOptions<ProjectIssue>({
+          id: `includes-react-issues`,
+          getKey: (i) => i.id,
+          initialData: sampleProjectIssues,
+        }),
+      )
+
+      // Parent hook: runs includes query that produces child Collections
+      const { result: parentResult } = renderHook(() =>
+        useLiveQuery((q) =>
+          q.from({ p: projectsCollection }).select(({ p }) => ({
+            id: p.id,
+            name: p.name,
+            issues: q
+              .from({ i: issuesCollection })
+              .where(({ i }) => eq(i.projectId, p.id))
+              .select(({ i }) => ({
+                id: i.id,
+                title: i.title,
+              })),
+          })),
+        ),
+      )
+
+      // Wait for parent to be ready
+      await waitFor(() => {
+        expect(parentResult.current.data).toHaveLength(2)
+      })
+
+      const alphaProject = parentResult.current.data.find(
+        (p: any) => p.id === `p1`,
+      )!
+      expect(alphaProject.name).toBe(`Alpha`)
+
+      // Child hook: subscribes to the child Collection from the parent row,
+      // simulating a subcomponent using useLiveQuery(project.issues)
+      const { result: childResult } = renderHook(() =>
+        useLiveQuery((alphaProject as any).issues),
+      )
+
+      await waitFor(() => {
+        expect(childResult.current.data).toHaveLength(2)
+      })
+
+      expect(childResult.current.data).toEqual(
+        expect.arrayContaining([
+          expect.objectContaining({ id: `i1`, title: `Bug in Alpha` }),
+          expect.objectContaining({ id: `i2`, title: `Feature for Alpha` }),
+        ]),
+      )
+
+      // Add a new issue to Alpha — the child hook should reactively update
+      act(() => {
+        issuesCollection.utils.begin()
+        issuesCollection.utils.write({
+          type: `insert`,
+          value: { id: `i4`, title: `New Alpha issue`, projectId: `p1` },
+        })
+        issuesCollection.utils.commit()
+      })
+
+      await waitFor(() => {
+        expect(childResult.current.data).toHaveLength(3)
+      })
+
+      expect(childResult.current.data).toEqual(
+        expect.arrayContaining([
+          expect.objectContaining({ id: `i1`, title: `Bug in Alpha` }),
+          expect.objectContaining({ id: `i2`, title: `Feature for Alpha` }),
+          expect.objectContaining({ id: `i4`, title: `New Alpha issue` }),
+        ]),
+      )
+    })
+  })
 })


samwillis:
GPT5.4s proposed fix:

---

## Proposal

Implement option 1 by introducing an explicit **include scope key** that represents the identity of the specific parent row instance an included child collection belongs to.

Today the includes runtime is keyed by `correlationKey`, which is only sufficient when the child result is purely a function of that correlation value. For parent-referencing filters, the correct identity is instead:

- the specific parent row instance
- at the specific nesting level
- with a stable key that can be routed through the compiled child pipeline

The cleanest fix is to make includes routing operate on a **parent-specific scope key**, while still using `correlationKey` only for the initial join between parent rows and child rows.

---

## High-Level Design

### Core idea

Split the two concerns that are currently conflated into `correlationKey`:

- `correlationKey`
  - Used only to join child rows to candidate parent rows.
  - Example: `issue.projectId = project.groupId`

- `includeScopeKey`
  - Used to identify which parent row instance owns a particular included child collection.
  - Example: `"project:1"` at the first include level, or `"project:1/issue:10"` at the next level down.

After the child query is joined against candidate parents and parent-dependent filters are evaluated, every surviving child row should carry the **include scope key** of the parent row it matched. From that point onward, all includes runtime state should be keyed by `includeScopeKey`, not by `correlationKey`.

---

## Why This Works

For this query:

```ts
items: q
  .from({ c: children })
  .where(({ c }) => eq(c.groupId, p.groupId))
  .where(({ c }) => eq(c.createdBy, p.createdBy))
```

two parents can share the same `groupId`, but they should not share the same included collection if `createdBy` differs.

With this proposal:

- both parents still participate in the join for `groupId = 1`
- child rows are evaluated separately against each parent context
- surviving child rows are tagged with the matching parent’s `includeScopeKey`
- the runtime creates one child `Collection` per `includeScopeKey`
- therefore each parent gets its own correct child collection

---

## Proposed Runtime Model

### New internal concept

Introduce:

- `type IncludeScopeKey = string`

This should be an internal, opaque, stable string.

### Scope key semantics

At each include level:

- top-level include scope key identifies the root parent row
- nested include scope key identifies the full ancestor path to the immediate parent row

Examples:

- top-level parent row `project.id = 1`
  - scope key: `"n:1"` or equivalent

- nested child row `issue.id = 10` under that project
  - scope key: `"n:1|n:10"` or equivalent

The exact encoding is not important as long as it is:

- deterministic
- collision-free for `string | number` row keys
- append-only
- cheap to compute

I would recommend a dedicated helper instead of ad-hoc string concat.

---

## Helper Utilities

Add small internal helpers, probably near `collection-config-builder.ts` or in a tiny shared util:

- `encodeIncludeScopeSegment(key: string | number): string`
- `appendIncludeScopeKey(parentScopeKey: IncludeScopeKey | undefined, rowKey: string | number): IncludeScopeKey`

Example encoding strategy:

- number key `1` -> `n:1`
- string key `abc` -> `s:3:abc`
- append with delimiter `|`

So:

- root project `1` -> `n:1`
- issue `10` under project `1` -> `n:1|n:10`

This avoids collisions like `"1|23"` vs `"12|3"`.

I would not use raw JSON stringification as the first choice unless you want the simplest possible implementation; an explicit segment encoder is cheaper and clearer.

---

## Compiler Changes

### Files

- `packages/db/src/query/compiler/index.ts`
- maybe `packages/db/src/query/ir.ts` only if you want to document the new metadata concept, but it is not strictly required

### Current behavior

The compiler currently:

1. builds `parentKeyStream` keyed by `correlationKey`
2. optionally includes projected parent context
3. joins child rows against `parentKeyStream`
4. tags child rows with `__correlationKey`
5. returns pipeline tuples containing `correlationKey`

### Proposed behavior

Keep step 1, but change the parent-side payload from:

- `null`
- or `parentContext`

to:

- `{ includeScopeKey, parentContext }`

#### Building `parentKeyStream`

When extracting includes from a parent pipeline, compute for each parent row:

- `correlationKey`
- `includeScopeKey`
- `parentContext`

`includeScopeKey` should be derived from the **current parent row identity**:

- top-level parent row:
  - based on the pipeline tuple `key`
- nested parent row:
  - based on the current row’s existing `__includeScopeKey` plus the tuple `key`

That means the compiler needs a helper like:

- if current row already has `__includeScopeKey`, append current row key to it
- otherwise create a fresh root scope key from current row key

#### After join

When a child row is joined to parent candidates, tag the child row with:

- `__includeScopeKey`
- optionally `__correlationKey` if still useful internally

The child pipeline should evaluate parent-referencing filters exactly as it does now, using projected parent context. The key difference is that **the surviving child rows must retain the matched parent’s include scope key**.

#### Final pipeline output

For includes mode, change emitted tuple shape from:

- `[result, orderByIndex, correlationKey]`

to:

- `[result, orderByIndex, includeScopeKey]`

This is the key contract change that lets the runtime stop grouping by correlation value.

#### `ORDER BY` / `LIMIT` / `OFFSET`

This is important.

Currently `includesGroupKeyFn` groups by `__correlationKey`. That becomes incorrect under parent-specific includes, because per-parent top-k behavior should be grouped by **scope key**, not correlation key.

Change grouped ordering to use:

- `row[mainSource].__includeScopeKey`

not

- `row[mainSource].__correlationKey`

Without this change, duplicate-correlation parents would still share a limit/offset window even if routing is otherwise fixed.

---

## Includes Runtime Changes

### File

### Current state model

`IncludesOutputState` currently uses correlation-based structures such as:

- `childRegistry: Map<correlationKey, ChildCollectionEntry>`
- `pendingChildChanges: Map<correlationKey, Map<childKey, Changes>>`
- `correlationToParentKeys`

### Proposed state model

Switch to scope-based structures:

- `childRegistry: Map<IncludeScopeKey, ChildCollectionEntry>`
- `pendingChildChanges: Map<IncludeScopeKey, Map<childKey, Changes>>`

Remove the need for:

- `correlationToParentKeys`

because a child collection is no longer shared across multiple parents.

### `ChildCollectionEntry`

Add:

- `scopeKey: IncludeScopeKey`

This is needed for nested includes so descendants can build keys relative to the owning entry.

### Output callback

In `setupIncludesOutput`, read tuples as:

- `[childResult, orderByIndex, includeScopeKey]`

and accumulate by:

- `[includeScopeKey, childKey]`

not `[correlationKey, childKey]`.

---

## Parent Attachment Changes

### Top-level flush

In `flushIncludesState`, phase 1 currently computes a correlation value from each parent result and attaches or reuses the child collection for that correlation.

Instead:

- compute `includeScopeKey` from the parent row key and the current parent scope prefix
- create or reuse the child collection entry by `includeScopeKey`
- attach that entry’s collection directly to that parent result

This means there is now exactly one child collection entry per parent row instance at that level.

### Important API shape change inside `flushIncludesState`

I would change the function signature to thread a parent scope prefix:

```ts
function flushIncludesState(
  includesState,
  parentCollection,
  parentId,
  parentChanges,
  parentScopeKeyPrefix?: IncludeScopeKey,
): void
```

Then:

- top-level call uses `undefined`
- recursive nested call uses `entry.scopeKey`

When processing a parent row inside `flushIncludesState`, compute:

- `includeScopeKey = appendIncludeScopeKey(parentScopeKeyPrefix, parentKey)`

This gives:

- top-level parent row -> root scope key
- nested parent row -> ancestor scope + row key

That is the main mechanism that makes nested includes work correctly.

---

## Nested Includes

This is the trickiest part, and the proposal should be explicit here.

### Current nested design

Nested includes use:

- shared buffers
- a routing index that maps nested correlation keys back to parent correlation keys

That design exists because nested child output is currently keyed by correlation value.

### Proposed nested design

Keep the shared-buffer architecture, but convert the routing model to scope keys.

#### Shared nested pipeline output

Nested pipelines should emit the **immediate parent row’s scope key** as their third tuple element.

For example, comments under issues should emit the scope key for the specific issue row instance they belong to, not just `issueId`.

That means the nested buffer becomes:

- `buffer: Map<IncludeScopeKey, Map<childKey, Changes>>`

where the key is the immediate parent row scope key.

#### Routing index

You may still want a routing index to connect descendant scope keys back to the owning `ChildCollectionEntry`, but now the keys are unambiguous.

Update the nested routing structures so they track:

- `nestedScopeKey -> parentEntryScopeKey`
- `parentEntryScopeKey -> Set<nestedScopeKey>`

instead of correlation-based keys.

#### Updating routing index

When processing child changes for a `ChildCollectionEntry`, and a child row with key `childKey` is inserted, compute:

- `nestedScopeKey = appendIncludeScopeKey(entry.scopeKey, childKey)`

Store:

- `nestedRoutingIndex.set(nestedScopeKey, entry.scopeKey)`

This is much more robust than extracting nested correlation values from child results.

#### Draining nested buffers

When a nested buffer entry arrives for `nestedScopeKey`:

1. look up `parentEntryScopeKey`
2. look up the corresponding `ChildCollectionEntry`
3. route the buffered changes into that entry’s per-entry nested includes state
4. store them under `nestedScopeKey`

This preserves the current architecture but removes ambiguity.

---

## What To Remove

If you implement this fully, the following correlation-based runtime structures should likely disappear from includes routing:

- `correlationToParentKeys`
- the assumption that one child `Collection` can be shared by all parents with the same correlation value
- grouping by `__correlationKey` for includes order-by windowing

`correlationKey` should remain only where it is truly needed:

- candidate matching in the join
- maybe internal diagnostics/comments
- nowhere in child collection registry keys

---

## Suggested Step-by-Step Implementation Plan

### Phase 1: Introduce scope key plumbing in compiler

1. Add scope-key helper(s).
2. When building `parentKeyStream`, include `{ includeScopeKey, parentContext }`.
3. After join, tag child rows with `__includeScopeKey`.
4. Change includes result tuples to emit `includeScopeKey`.
5. Change grouped order-by for includes to use `__includeScopeKey`.

At this point the compiler contract is ready, but runtime still expects correlation keys.

### Phase 2: Convert top-level includes runtime to scope-key routing

1. Update `IncludesOutputState` to key registries and pending changes by `IncludeScopeKey`.
2. Update `setupIncludesOutput` to read tuple[2] as `includeScopeKey`.
3. Update `flushIncludesState` to build/attach child entries by `includeScopeKey`.
4. Remove `correlationToParentKeys` from the top-level path.

At this point the duplicate-correlation bug for one-level includes should be fixed.

### Phase 3: Convert nested includes routing

1. Add `scopeKey` to `ChildCollectionEntry`.
2. Thread `parentScopeKeyPrefix` through recursive `flushIncludesState`.
3. Update nested shared buffers to use `IncludeScopeKey`.
4. Update nested routing index to map descendant scope keys to parent entry scope keys.
5. Stop extracting nested routing keys from correlation values.

At this point nested includes should also be parent-instance correct.

---

## Test Plan

Add tests in `packages/db/tests/query/includes.test.ts`.

### New blocking regression test

Two parents share the same correlation key but differ on a parent-referenced filter field.

- parents:
  - `{ id: 1, groupId: 1, createdBy: 'alice' }`
  - `{ id: 2, groupId: 1, createdBy: 'bob' }`
- children:
  - `{ id: 10, groupId: 1, createdBy: 'alice' }`

Query:

- correlate on `groupId`
- filter on `createdBy`

Expect:

- parent 1 gets child 10
- parent 2 gets `[]`

### Reactivity tests

1. Parent field update:
   - change parent 2 `createdBy` from `bob` to `alice`
   - verify child 10 appears only for parent 2 after update if semantics match

2. Child field update:
   - change child 10 `createdBy`
   - verify it moves between parents correctly

3. Duplicate correlation with `ORDER BY + LIMIT`
   - two parents share correlation key
   - ensure each parent gets its own top-k window
   - this validates the `includesGroupKeyFn` fix

### Nested regression test

Use:

- projects include issues
- issues include comments

Construct data where the same issue row can appear in multiple parent-scoped issue collections unless scope routing is correct, and verify nested comment collections stay isolated per parent instance.

---

## Risks / Things To Watch

### Hidden internal fields leaking into results

The compiler currently carries hidden fields like `__correlationKey` on internal row objects. If you add `__includeScopeKey`, make sure it does not become observable in user results when a child query omits `select()`.

This may already be a latent issue for `__correlationKey`, so it is worth checking while you are here.

### Cache assumptions

The query compilation cache should still be fine, because scope keys are runtime row metadata, not structural query IR. But be careful not to bake scope-specific values into cached query objects themselves.

### Collection IDs for child collections

`createChildCollectionEntry` currently uses IDs based on `correlationKey`. That should switch to `scopeKey`, otherwise sibling parents may still collide in child collection IDs.

---

## Implementation Shortcut

If you want the smallest viable implementation that is still correct:

- keep correlation-based join exactly as-is
- introduce `includeScopeKey`
- emit `includeScopeKey` from child pipelines
- key child registries and pending changes by `includeScopeKey`
- group include `ORDER BY` windows by `includeScopeKey`
- thread `scopeKey` through nested routing

That should fix the semantic bug without requiring a larger rewrite of includes compilation.

If you want, I can turn this into a shorter “agent handoff spec” with explicit file-by-file tasks and acceptance criteria.

kevin-dp:
@samwillis i added this to the docs as requested and i also added a unit test that actually tests it.

Issue #1325, opened 2026-03-03
Performance issues when unmounting useLiveQuery due to excessive setTimeout and clearTimeout calls


amiran-gorgazjan:
- [x] I've validated the bug against the latest version of DB packages

**Describe the bug**
Apps with large numbers of live queries, when unmounting them, trigger deferred garbage collection via `setTimeout`. `setTimeout` has a considerable performance impact and looking at the profiler, about half of the entire time spent for a re-render gets spent creating and clearing garbage collection timeouts.

**To Reproduce**

1. Create some view with a large list of data, say tweets paginated, 500 per page
2. Each Tweet component gets the tweet id passed down as a prop and it gets its own data using it's own useLiveQuery instance to optimise for minimal updates when mutating
3. Swap between pages. Each of the 500 tweets will get unmounted and 500 new tweets should be re-mounted.
4. Observe performance via any Profiler
5. The performance is quite bad, depending on the number of tweets making the page almost unusable

**Expected behavior**
I expect the performance to be relatively reasonable.

**Screenshots**
Loads and loads of unmounts:

<img width="1663" height="419" alt="Image" src="https://github.com/user-attachments/assets/6eca6e39-498a-4200-824b-63f2669d59b4" />

And most of that time is spent in setTimeout and clearTimeout calls:

<img width="1032" height="373" alt="Image" src="https://github.com/user-attachments/assets/8af35784-4aa0-4967-a734-d1c9a0817057" />

**Desktop (please complete the following information):**

- OS: OSX
- Browser: Chrome
- Version: 145.0.7632.117 (Official Build) (arm64)

**Additional context**

Can be fixed with a **centralised garbage collection queue**: the function calls could be pushed to an in-memory queue which is triggered for deferred consumption, doing `setTimeout` only once during the unmounting process. The individual `setTimeout` calls will thanks to this happen after the render process.

I tested this in a fork and it gave me 2x performance when using StrictMode where the setTimeouts happen twice per unmount.

kevin-dp:
Hi @amiran-gorgazjan,

Did you test the branch with our WIP for subqueries to see if that solves those performance issues? You can try out subqueries, we currently have 2 open PRs https://github.com/TanStack/db/pull/1298 and https://github.com/TanStack/db/pull/1307

If you need aggregates or `toArray` then you need to use the first PR, if you need to reference parent fields from the subquery then you will need the latter PR. We want to merge those PRs soon and release them but you can already give it a try now to see if that would solve your isses.

[...]

PR #1326, opened 2026-03-03
Use in-memory queue for faster GC registration

amiran-gorgazjan:
# Use in-memory queue for faster GC registration

Improves performance for issue: https://github.com/TanStack/db/issues/1325

## 🎯 Changes

- **Introduced `CleanupQueue`**: Created a singleton in-memory queue (`packages/db/src/collection/cleanup-queue.ts`) to manage garbage collection tasks. It batches task registrations using microtasks and maintains only a single active `setTimeout` for the earliest scheduled task.
- **Updated `CollectionLifecycleManager`**: Replaced direct `setTimeout` and `clearTimeout` calls for GC timers with the new `CleanupQueue.getInstance().schedule()` and `cancel()` methods.
- **Added Tests**: Added comprehensive test coverage for the new queue behavior in `packages/db/tests/cleanup-queue.test.ts`.

**Motivation**: Creating individual `setTimeout` calls for every component unmount or collection inactivity can lead to performance issues when many collections are involved. Using a centralized in-memory queue optimizes scheduling by significantly reducing the number of active timeouts created in the runtime.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/db/src/collection/cleanup-queue.ts
+++ b/packages/db/src/collection/cleanup-queue.ts
@@ -3,6 +3,9 @@ type CleanupTask = {
   callback: () => void
 }
 
+/**
+ * Batches many GC registrations behind a single shared timeout.
+ */
 export class CleanupQueue {
   private static instance: CleanupQueue | null = null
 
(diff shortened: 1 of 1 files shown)


samwillis (approved):
Thanks for dogging into this @amiran-gorgazjan 

I did some quick benchmarking and the speedup is real, and the implementation is great. I've pushed a commit with some jsdoc comments, going to merge it now.

Issue #672, opened 2025-10-13
Add first-class error tracking to base Collection class


KyleAMathews:
## Background

Currently, error state tracking is inconsistent across collection types:

- **query-db-collection**: Tracks `lastError` and `errorCount` in closure variables, exposing them via utility functions (`collection.utils.lastError()`, `collection.utils.isError()`, `collection.utils.errorCount()`)
- **electric-db-collection**: Has no error tracking at all - errors are thrown or logged but not stored
- **Base Collection**: Has `status: 'error'` state but no associated error information

This creates an inconsistent developer experience where error handling differs depending on which collection type you're using.

## Proposal

Make error tracking a first-class concern in the base Collection class, similar to how `status` is handled today.

### API Changes

Add to `CollectionLifecycleManager`:

```typescript
public error: Error | null = null
public errorCount: number = 0

public markError(error?: Error): void {
  if (error) {
    this.error = error
    this.errorCount++
  }
  this.setStatus('error')
}
```

Expose on Collection:

```typescript
// Direct property access (like status)
collection.error // Error | null
collection.errorCount // number

// Events
collection.on('error', (event) => {
  console.log('Error occurred:', event.error)
})
```

### Benefits

1. **Consistency**: All collection types have the same error API
2. **Better DX**: Access `collection.error` directly instead of `collection.utils.lastError()`
3. **Framework integration**: React, Angular, Vue adapters can reactively bind to error state
4. **Debugging**: Errors are preserved and inspectable, not just logged
5. **Retry logic**: `errorCount` enables exponential backoff strategies

### Migration Path

1. Add error tracking to base Collection (non-breaking - new properties)
2. Update `query-db-collection` to use base error tracking instead of closure variables (internal change)
3. Update `electric-db-collection` to pass errors to `markError()` instead of just throwing
4. Deprecate `collection.utils.lastError()` in favor of `collection.error` (breaking in next major)

### Open Questions

- Should `markError()` clear the error on successful recovery, or should we have a separate `clearError()` method?
- Should errors emit events (`collection.on('error', ...)`) in addition to status events?
- What should happen to `errorCount` on recovery - reset to 0 or preserve for analytics?
- Should we track error history (last N errors) or just the most recent?

## Related

- #671 - Fixed error propagation where errors weren't being set on collections at all

---

This came out of the discussion in #671 where we realized error tracking should be baked into the core Collection rather than being add-on utilities.

samwillis:
[...]

PR #1346, opened 2026-03-09
fix(db): reject preload() promise when collection errors so ErrorBoundary is reached

### LINKED ISSUE #1343: `useLiveSuspenseQuery` doesn't work with `ErrorBoundary`

**Describe the bug**
It's not clear what the interaction between `useLiveSuspenseQuery` and `ErrorBoundary` is supposed to look like. Currently, if the fetch function throws, **it isnt caught** by the `ErrorBoundary`

**To Reproduce**

https://stackblitz.com/edit/vitejs-vite-w4tkecqk?file=src%2FApp.tsx

<details>
<summary>Copy pasted snippet of code</summary>

```tsx
import React, { Suspense, useState } from 'react';
import { createCollection, useLiveSuspenseQuery } from '@tanstack/react-db';
import { queryCollectionOptions } from '@tanstack/query-db-collection';
import { QueryClient } from '@tanstack/query-core';
import { ErrorBoundary } from 'react-error-boundary';

interface Todo {
  id: string;
  text: string;
  completed: boolean;
  createdAt: number;
}

const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));

let mockDb: Todo[] = [
  {
    id: '1',
    text: 'Learn TanStack DB',
    completed: false,
    createdAt: Date.now(),
  },
];

const api = {
  getTodos: async (): Promise<Todo[]> => {
    await sleep(1000);
    throw new Error('random error');
    // Uncomment this for normal usage.
    // return [...mockDb];
  },
  addTodo: async (todo: Todo): Promise<Todo> => {
    await sleep(600);
    mockDb.push(todo);
    return todo;
  },
  updateTodo: async (id: string, updates: Partial<Todo>): Promise<Todo> => {
    await sleep(800);
    const index = mockDb.findIndex((t) => t.id === id);
    mockDb[index] = { ...mockDb[index], ...updates };
    return mockDb[index];
  },
};

const queryClient = new QueryClient({
  defaultOptions: {
    queries: {
      retry: false,
      refetchOnWindowFocus: false,
      throwOnError: true,
    },
  },
});

  queryCollectionOptions({
    queryClient: queryClient,
    queryKey: ['todos'],
    queryFn: () => api.getTodos(),
    getKey: (todo: Todo) => todo.id,
      // The transaction contains the modified data
      const { modified } = transaction.mutations[0];
      await api.updateTodo(modified.id, modified);
    },
      // The transaction contains the modified data
      const { modified } = transaction.mutations[0];
      await api.addTodo(modified);
    },
  })
);

function TodoList() {
  const [text, setText] = useState('');

  const { data: todos = [] } = useLiveSuspenseQuery((q) =>
    q
      .orderBy(({ todo }) => todo.createdAt, 'desc')
  );

  const handleSubmit = (e: React.FormEvent) => {
    e.preventDefault();
    if (!text.trim()) return;

    const newTodo: Todo = {
      id: Math.random().toString(36).slice(2, 9),
      text,
      completed: false,
      createdAt: Date.now(),
    };

    // collection.insert updates the UI state instantly
    todoCollection.insert(newTodo);
    setText('');
  };

  const toggleTodo = (todo: Todo) => {
      draft.completed = !draft.completed;
    });
  };

  return (
    <div>
      <form onSubmit={handleSubmit}>
        <input
          value={text}
          onChange={(e) => setText(e.target.value)}
          placeholder="What needs to be done?"
        />
        <button type="submit">Add Todo</button>
      </form>

      <ul>
        {todos.map((todo) => (
          <li
            key={todo.id}
            onClick={() => toggleTodo(todo)}
            style={{
              cursor: 'pointer',
              textDecoration: todo.completed ? 'line-through' : 'none',
            }}
          >
            {todo.text} {todo.completed ? '✅' : '⭕'}
          </li>
        ))}
      </ul>
    </div>
  );
}

  return (
    <div>
      <ErrorBoundary fallback={<div>Something went wrong</div>}>
        <Suspense fallback={<p>Loading...</p>}>
          <TodoList />
        </Suspense>
      </ErrorBoundary>
    </div>
  );
}
```
</details>

<details>
<summary>package.json</summary>

```json
{
  "name": "vite-react-typescript-starter",
  "private": true,
  "version": "0.0.0",
  "type": "module",
  "scripts": {
    "dev": "vite",
    "build": "tsc -b && vite build",
    "lint": "eslint .",
    "preview": "vite preview"
  },
  "dependencies": {
    "@tanstack/query-db-collection": "^1.0.28",
    "@tanstack/react-db": "^0.1.75",
    "@tanstack/react-query": "^5.90.21",
    "react": "^19.2.4",
    "react-dom": "^19.2.4",
    "react-error-boundary": "^6.1.1"
  },
  "devDependencies": {
    "@eslint/js": "^9.39.3",
    "@types/node": "^24.11.0",
    "@types/react": "^19.2.14",
    "@types/react-dom": "^19.2.3",
    "@vitejs/plugin-react": "^5.1.4",
    "eslint": "^9.39.3",
    "eslint-plugin-react-hooks": "^7.0.1",
    "eslint-plugin-react-refresh": "^0.5.2",
    "globals": "^17.4.0",
    "typescript": "~5.9.3",
    "typescript-eslint": "^8.56.1",
    "vite": "^7.3.1"
  }
}
```
</details>

tl;dr
```tsx
<ErrorBoundary fallback={<div>Something went wrong</div>}>
        <Suspense fallback={<p>Loading...</p>}>
          <TodoList />
        </Suspense>
      </ErrorBoundary>

// in TodoList
const { data: todos = [] } = useLiveSuspenseQuery((q) =>
    q
      .orderBy(({ todo }) => todo.createdAt, 'desc')
  );

// in collection
  queryCollectionOptions({
    queryClient: queryClient,
    queryKey: ['todos'],
    queryFn: () => api.getTodos(),

// api.getTodos()
const api = {
  getTodos: async (): Promise<Todo[]> => {
    await sleep(1000);
    throw new Error('random error');
    // Uncomment this for normal usage.
    // return [...mockDb];
  },
```

**Expected behavior**
I expected the error to be caught by the error boundary.

## Workaround
Throwing the collection error if encountered:
```tsx
    const { data:todos } = useLiveSuspenseQuery((q) => q.from({ todo: todoCollection }))

    if (todoCollection.utils.isError) {
        throw todoCollection.utils.lastError
    }
```

sleitor:
## Summary

Fixes #1343

`useLiveSuspenseQuery` was not propagating collection errors to `<ErrorBoundary>` because `preload()` would hang forever when the collection transitioned to the `error` state.

## Root Cause

In `packages/db/src/collection/sync.ts`, `preload()` registers an `onFirstReady` callback that resolves the promise when the collection becomes ready. However, if the collection transitions to `error` while the promise is pending (e.g. because `queryFn` throws), the promise was **never settled** — neither resolved nor rejected. This left any `<Suspense>` boundary in a permanently suspended state and prevented the error from bubbling to an `<ErrorBoundary>`.

## Fix

Two changes:

### 1. `packages/db/src/collection/sync.ts`
Subscribe to `status:change` events inside `preload()`. If the collection enters the `error` state while the promise is pending, reject the promise immediately with a `CollectionIsInErrorStateError`.

### 2. `packages/react-db/src/useLiveSuspenseQuery.ts`
Re-throw the actual error from `collection.utils?.lastError` (available via `query-db-collection`) instead of a generic fallback message, so `<ErrorBoundary>` receives the original error object.

## Testing

All existing tests pass (`@tanstack/db` and `@tanstack/react-db`). The bug is also repro-able with the StackBlitz sandbox in the issue — after this fix the `<ErrorBoundary>` catches the error and renders its fallback.

### DIFF
--- a/.changeset/xtjpqj-ah-fix.md
+++ b/.changeset/xtjpqj-ah-fix.md
@@ -1,0 +1,19 @@
+---
+"@tanstack/db": patch
+"@tanstack/react-db": patch
+---
+
+fix(db): reject preload() promise when collection transitions to error state
+
+Previously, `preload()` only resolved when the collection became ready. If
+the collection transitioned to the `error` state while the promise was pending
+(e.g. because the `queryFn` threw), the promise would hang forever, keeping
+any `<Suspense>` boundary suspended indefinitely and preventing the error from
+reaching an `<ErrorBoundary>`.
+
+Now `preload()` subscribes to `status:change` events and rejects the promise
+when the collection enters the `error` state. `useLiveSuspenseQuery` is also
+updated to re-throw the actual error from `collection.utils?.lastError` instead
+of a generic fallback message, so `<ErrorBoundary>` receives the original error.
+
+Fixes #1343
--- a/.changeset/xtjpqj-ah-fix.md
+++ b/.changeset/xtjpqj-ah-fix.md
@@ -1,6 +1,6 @@
 ---
-"@tanstack/db": patch
-"@tanstack/react-db": patch
+'@tanstack/db': patch
+'@tanstack/react-db': patch
 ---
 
 fix(db): reject preload() promise when collection transitions to error state
--- a/packages/db/src/collection/sync.ts
+++ b/packages/db/src/collection/sync.ts
@@ -271,8 +271,20 @@ export class CollectionSyncManager<
         return
       }
 
+      // Subscribe to status changes to reject the promise when collection errors.
+      // This is necessary because onFirstReady only fires on success, but if the
+      // collection transitions to 'error' while the promise is pending it would
+      // hang forever and keep the Suspense boundary suspended indefinitely.
+      const unsubscribeError = this._events.on(`status:change`, (event) => {
+        if (event.status === `error`) {
+          unsubscribeError()
+          reject(new CollectionIsInErrorStateError())
+        }
+      })
+
       // Register callback BEFORE starting sync to avoid race condition
       this.lifecycle.onFirstReady(() => {
+        unsubscribeError()
         resolve()
       })
 
--- a/packages/react-db/src/useLiveSuspenseQuery.ts
+++ b/packages/react-db/src/useLiveSuspenseQuery.ts
@@ -199,9 +199,7 @@ export function useLiveSuspenseQuery(
     const lastError = (result.collection as any).utils?.lastError
     throw lastError instanceof Error
       ? lastError
-      : new Error(
-          `Collection "${result.collection.id}" failed to load`,
-        )
+      : new Error(`Collection "${result.collection.id}" failed to load`)
   }
 
   if (result.status === `loading` || result.status === `idle`) {
--- a/packages/react-db/src/useLiveSuspenseQuery.ts
+++ b/packages/react-db/src/useLiveSuspenseQuery.ts
@@ -193,9 +193,15 @@ export function useLiveSuspenseQuery(
   // After success, errors surface as stale data (matches TanStack Query behavior)
   if (result.status === `error` && !hasBeenReadyRef.current) {
     promiseRef.current = null
-    // TODO: Once collections hold a reference to their last error object (#671),
-    // we should rethrow that actual error instead of creating a generic message
-    throw new Error(`Collection "${result.collection.id}" failed to load`)
+    // Re-throw the actual error from the collection if available (e.g. from
+    // query-db-collection's utils.lastError), otherwise fall back to a generic
+    // message. Throwing here propagates the error to the nearest ErrorBoundary.
+    const lastError = (result.collection as any).utils?.lastError
+    throw lastError instanceof Error
+      ? lastError
+      : new Error(
+          `Collection "${result.collection.id}" failed to load`,
+        )
   }
 
   if (result.status === `loading` || result.status === `idle`) {


samwillis (changes requested):
Thanks for looking into this @sleitor 


I asked GPT5.2 for a review, its findings are below. I think I agree with it we need to be sure that we surface the original error:

---

I found one substantive issue and one testing gap.

1. `packages/db/src/collection/sync.ts`: `preload()` now loses the original synchronous startup error.

The new `status:change` listener rejects the pending preload promise with `CollectionIsInErrorStateError` as soon as the collection enters `error`. That works for the async transition this PR is targeting, but it also changes behavior when `startSync()` throws synchronously.

Today the sequence is:

- `preload()` installs the `status:change` listener
- `preload()` calls `startSync()`
- `startSync()` catches the original error, sets status to `error`, then rethrows
- setting status to `error` triggers the new listener first
- the promise becomes rejected with `CollectionIsInErrorStateError`
- the original thrown error no longer wins because the promise is already settled

I verified this locally with a small repro: a collection whose `sync.sync()` throws `new Error('boom')` now causes `await collection.preload()` to reject with `CollectionIsInErrorStateError` instead of the original `Error('boom')`.

That also means the React side still won’t always surface the original error object to the nearest `ErrorBoundary`. In `packages/react-db/src/useLiveSuspenseQuery.ts`, the new logic only rethrows the real error when `collection.utils?.lastError` exists. For plain collections, this path still falls back to a generic error message.

Suggestion:
- In `preload()`, make sure synchronous `startSync()` failures reject with the original thrown error.
- A straightforward way to do that is to centralize resolve/reject/unsubscribe in a single settle helper, track whether `preload()` is currently inside the synchronous `startSync()` call, and ignore the `status:change -> error` listener during that window.
- Longer term, if the goal is for `useLiveSuspenseQuery` to always throw the real initial-load error, that error probably needs to live on core collection/lifecycle state rather than only on `query-db-collection` utilities.

2. `packages/react-db/tests/useLiveSuspenseQuery.test.tsx`: the exact regression path still isn’t covered.

I ran the targeted React test suite and it passed, but coverage still showed the changed error branch in `useLiveSuspenseQuery` as uncovered. I don’t see a test that drives a collection into initial `error` while suspended and asserts that the nearest `ErrorBoundary` receives the original error object.

Suggestion:
- Add a core test that `collection.preload()` rejects with the original sync error when `sync.sync()` throws synchronously.
- Add a core test for the async transition this PR is fixing: pending `preload()` rejects once the collection later enters `error`.
- Add a React test that renders `useLiveSuspenseQuery` under `Suspense` + `ErrorBoundary`, triggers an initial load failure, and asserts the boundary receives the original error object rather than a generic fallback.

The fix direction in the PR makes sense, but I think preserving original error identity for synchronous startup failures is important before this merges.

sleitor:
👋 Gentle ping — just checking if this is still on the radar for review. Happy to address any feedback or rebase if needed!

sleitor:
Thanks for the detailed review @samwillis (and GPT-5.2 for the analysis) — this is a solid catch. You're right that the synchronous `startSync()` throw path currently loses the original error identity to the new `status:change` listener, and that `useLiveSuspenseQuery` still falls back to a generic error for plain collections without `utils.lastError`.

I'll rework `preload()` to centralize resolve/reject/unsubscribe in a single settle helper, guard the `status:change -> error` listener while inside the synchronous `startSync()` call so the originally thrown error always wins, and add the three tests you outlined (core sync-throw regression, core async-transition regression, and the React Suspense+ErrorBoundary test). Will push an update shortly.

KyleAMathews:
Thanks @sleitor for the work here. The core preload/error-propagation behavior was independently fixed by https://github.com/TanStack/db/pull/1751 (https://github.com/TanStack/db/commit/c521b5d6503d8fdf03574b9f9791143e59d34204). This branch is now conflicting and its listener changes can regress synchronous error identity, so it should not be merged. Closing as superseded; remaining canonical error-identity work belongs under https://github.com/TanStack/db/issues/672.

sleitor:
thanks for clarification @KyleAMathews. That is makes sence.

PR #1351, opened 2026-03-11
feat(examples): add op-sqlite persistence to React Native offline-transactions demo

kevin-dp:
## Summary

- Adds op-sqlite persistence to the React Native offline-transactions example app, wiring it up with the persistence layer so todos survive app restarts
- Updates the demo's metro config, database setup, server, and UI components to work with local SQLite storage via op-sqlite

## Test plan

- [ ] Run the React Native offline-transactions example on Android/iOS
- [ ] Verify todos persist across app restarts
- [ ] Verify offline mutations are queued and submitted when back online

### DIFF
--- a/examples/react-native/offline-transactions/src/components/TodoList.tsx
+++ b/examples/react-native/offline-transactions/src/components/TodoList.tsx
@@ -34,20 +34,17 @@ export function TodoList({ collection, executor }: TodoListProps) {
     q.from({ todo: collection }).orderBy(({ todo }) => todo.createdAt, `desc`),
   )
 
-  // Monitor network status
+  // Monitor network status for UI display
+  // (The executor's ReactNativeOnlineDetector handles sync retries internally)
   useEffect(() => {
     const unsubscribe = NetInfo.addEventListener((state) => {
       const connected =
         state.isConnected === true && state.isInternetReachable !== false
       setIsOnline(connected)
-
-      if (connected) {
-        executor.notifyOnline()
-      }
     })
 
     return () => unsubscribe()
-  }, [executor])
+  }, [])
 
   // Monitor pending transactions
   useEffect(() => {


samwillis:
Quick review from GPT5.4:

---

## Findings
1. `Medium` - `examples/react-native/offline-transactions/src/components/TodoList.tsx` regresses the initial loading UX. The old version kept a loading state until the query hydrated; this branch now defaults `data` to `[]` and immediately shows the empty state whenever the local SQLite cache starts empty. On a cold start with existing server data, users will briefly see “No todos yet” before the first fetch/refetch completes.

```33:35:examples/react-native/offline-transactions/src/components/TodoList.tsx
const { data: todoList = [] } = useLiveQuery((q) =>
  q.from({ todo: collection }).orderBy(({ todo }) => todo.createdAt, `desc`),
)
```

```177:183:examples/react-native/offline-transactions/src/components/TodoList.tsx
{/* Todo list */}
{todoList.length === 0 ? (
  <View style={styles.emptyContainer}>
    <Text style={styles.emptyText}>No todos yet. Add one above!</Text>
    <Text style={styles.emptySubtext}>
      Todos persist in SQLite and sync to server when online
```

2. `Medium` - `examples/react-native/offline-transactions/src/db/todos.ts` and `examples/react-native/offline-transactions/server/index.ts` still do not make offline inserts idempotent. `syncTodos()` receives an `idempotencyKey` but only logs it; the server accepts client-provided `id` and blindly overwrites any existing row at that id. If an offline insert is retried after a timeout or reconnect race, the second `POST` can silently replace the existing record and reset timestamps instead of being safely deduped.

```66:89:examples/react-native/offline-transactions/src/db/todos.ts
async function syncTodos({
  transaction,
  idempotencyKey,
}: {
  transaction: { mutations: Array<PendingMutation> }
  idempotencyKey: string
}) {
  const mutations = transaction.mutations

  console.log(
    `[Sync] Processing ${mutations.length} mutations`,
    idempotencyKey,
  )

  for (const mutation of mutations) {
    try {
      switch (mutation.type) {
        case `insert`: {
          const todoData = mutation.modified as Todo
          await todoApi.create({
            id: todoData.id,
            text: todoData.text,
            completed: todoData.completed,
```

```65:85:examples/react-native/offline-transactions/server/index.ts
// POST create todo
app.post('/api/todos', async (req, res) => {
  console.log('POST /api/todos', req.body)
  await delay(200)

  const { id, text, completed } = req.body
  if (!text || text.trim() === '') {
    return res.status(400).json({ error: 'Todo text is required' })
  }

  const now = new Date().toISOString()
  const todo: Todo = {
    id: id || generateId(),
    text,
    completed: completed ?? false,
    createdAt: now,
    updatedAt: now,
  }
  todosStore.set(todo.id, todo)
```

3. `Low` - `examples/react-native/offline-transactions/server/index.ts` persists runtime state to `server/todos.json` inside the example source tree, but there is no ignore rule for that generated file. Running the demo will dirty the repo and future runs will depend on leftover local state unless the file is manually cleaned up.

```1:10:examples/react-native/offline-transactions/server/index.ts
import { readFileSync, writeFileSync } from 'node:fs'
import { dirname, join } from 'node:path'
import { fileURLToPath } from 'node:url'
import cors from 'cors'
import express from 'express'

const app = express()
const PORT = 3001
const DATA_FILE = join(dirname(fileURLToPath(import.meta.url)), 'todos.json')
```

No additional React Native example findings stood out beyond those on this pass.

kevin-dp:
@samwillis Thanks for the review!

**Finding 1 (loading UX)** — Good catch, fixed in 8ed3fd37. Restored `isLoading` from `useLiveQuery` and gated the empty state so we show a loading indicator while hydrating instead of flashing "No todos yet."

**Finding 2 (idempotent inserts)** — I think this is fine as-is for a demo. The server's `todosStore.set(todo.id, todo)` is effectively idempotent since re-inserting the same ID just overwrites with the same data. The only side effect is timestamp reset, which is benign here. Adding idempotency-key tracking to an in-memory Map demo server would add complexity that distracts from the example's purpose (showing the offline-transactions API and the SQLite persistence API surface). The `idempotencyKey` is provided by the framework for consumers to use in their production backends.

**Finding 3 (todos.json)** — Good catch, fixed in 970bc7dc. Added a `.gitignore` for `server/todos.json`.

PR #1353, opened 2026-03-11
Make indexing opt-in

kevin-dp:
## Summary

Makes indexing explicit and opt-in rather than automatic, allowing users to choose the right index type for their use case and enabling better tree-shaking.

Based on #950 with the `ReadOptimizedIndex`/`WriteOptimizedIndex` renaming reverted — keeps the original `BasicIndex`/`BTreeIndex` names.

### Key changes

- `autoIndex` now defaults to `off` instead of `eager`
- To use `createIndex()` or `autoIndex: 'eager'`, you must set `defaultIndexType` on the collection
- Dev mode suggestions warn when indexes would help

### Bundle size impact

- No indexing: ~30% smaller bundle
- BasicIndex only: ~5 KB (~1.3 KB gzipped)
- BTreeIndex: ~33 KB (~7.8 KB gzipped)

## Test plan
- [x] Build passes
- [x] All 2033 tests pass
- [x] Tests updated to set `defaultIndexType` where needed


KyleAMathews:
## PR Review Summary

Overall this is a well-designed simplification. Replacing `IndexResolver` (sync + async union) with `IndexConstructor` eliminates the 3-state lifecycle and ~250 lines of `LazyIndexWrapper`/`IndexProxy`. The tree-shaking story via `@tanstack/db/indexing` is clean. A few issues worth addressing before merge:

---

### Critical

**`autoIndex: 'eager'` without `defaultIndexType` silently does nothing in production**
`packages/db/src/indexes/auto-index.ts:18-27` — When a user explicitly requests `autoIndex: 'eager'` but forgets `defaultIndexType`, the `console.warn` is gated behind `isDevModeEnabled()` which returns `false` in production. The function silently returns `false` and indexing never happens — no error, no warning.

This is a deliberately-requested feature that silently degrades. Consider throwing `CollectionConfigurationError` at collection construction time (`packages/db/src/collection/index.ts:~322`) when this invalid configuration is detected:

```ts
if (config.autoIndex === 'eager' && !config.defaultIndexType) {
  throw new CollectionConfigurationError(
    `autoIndex: 'eager' requires defaultIndexType to be set. ` +
    `Import an index type and set it:\n` +
    `  import { BasicIndex } from '@tanstack/db/indexing'\n` +
    `  createCollection({ defaultIndexType: BasicIndex, autoIndex: 'eager', ... })`
  )
}
```

---

### Important

1. **`emitIndexSuggestion` doesn't protect against `onSuggestion` callback errors** — `packages/db/src/indexes/index-registry.ts:76-77`. User-provided callback is called without try-catch. A buggy callback could crash query execution with a confusing stack trace.

2. **`BasicIndex.remove()` swallows errors, leaving index inconsistent** — `packages/db/src/indexes/basic-index.ts:113-123`. When `evaluateIndexExpression` throws during removal, the catch block returns without cleaning up `indexedKeys`, leaving phantom keys in the index. Should at minimum call `this.indexedKeys.delete(key)` in the catch block.

3. **`createIndex` throws generic `Error` instead of `CollectionConfigurationError`** — `packages/db/src/collection/indexes.ts:65-71`. The project has a well-established error hierarchy; this should use `CollectionConfigurationError` to match the pattern and let users catch it with `instanceof`.

---

### Suggestions

- **Type-level enforcement of `autoIndex` + `defaultIndexType` coupling** — Currently `{ autoIndex: 'eager', defaultIndexType: undefined }` compiles without error. A discriminated union config type could catch this at compile time:
  ```ts
  | { autoIndex?: 'off'; defaultIndexType?: IndexConstructor<TKey> }
  | { autoIndex: 'eager'; defaultIndexType: IndexConstructor<TKey> }
  ```

- **`BasicIndex` error re-throws lose original stack trace** — `basic-index.ts:82-86, 162-165`. Using `{ cause: error }` would preserve the original stack for debugging.

- **`setSizeCallback` is now set unconditionally** in `order-by.ts:282-288`. Previously only set when an index existed. In the no-index path, `loadNextItems` calls `requestSnapshot` with a fixed limit, but there's no deduplication guard (the `pendingOrderedLoadPromise` only covers the indexed path). Could cause repeated identical snapshot requests.

---

### Test Coverage Gaps

| Priority | Gap |
|----------|-----|
| 1 | `createIndex()` error when no `defaultIndexType` and no explicit `indexType` — not tested |
| 2 | `autoIndex: 'eager'` without `defaultIndexType` behavior — not tested |
| 3 | Dev mode suggestion system (`index-registry.ts`, 170 lines) — zero test coverage |
| 4 | `BasicIndex` (503 lines, recommended default) — no dedicated tests |

---

### Strengths

- Excellent architectural simplification — single `indexes` map replaces the dual `lazyIndexes`/`resolvedIndexes` maps
- Clean entry point separation for tree-shaking (~30% smaller without indexing)
- Actionable error messages in `createIndex` that tell users exactly what to import
- Well-structured test updates with `describe.each` for both `autoIndex` modes

kevin-dp:
## Addressed review feedback

### Critical
- **`autoIndex: 'eager'` without `defaultIndexType` now throws at construction time** (`collection/index.ts`) — instead of silently degrading in production, this now throws `CollectionConfigurationError` with an actionable message. The dead warning code in `auto-index.ts` was removed since it's no longer reachable.

### Important
1. **`emitIndexSuggestion` callback errors** (`index-registry.ts`) — wrapped `onSuggestion` callback in try-catch so a buggy callback can't crash query execution.
2. **`BasicIndex.remove()` phantom key cleanup** (`basic-index.ts`) — when `evaluateIndexExpression` throws during removal, the catch block now calls `this.indexedKeys.delete(key)` and `this.updateTimestamp()` to keep the index consistent.
3. **`createIndex` error type** (`collection/indexes.ts`) — changed from generic `Error` to `CollectionConfigurationError` to match the project's error hierarchy.

### Suggestions
- **`{ cause: error }` on re-throws** (`basic-index.ts`) — added to both `add()` and `build()` error paths to preserve original stack traces.

### Test coverage
- Added test: `autoIndex: 'eager'` without `defaultIndexType` throws `CollectionConfigurationError`
- Added test: `createIndex()` without `indexType` or `defaultIndexType` throws `CollectionConfigurationError`
- Updated existing tests in `query-while-syncing.test.ts` and `collection-change-events.test.ts` to provide `defaultIndexType` where needed

### Not addressed (needs discussion)
- **Type-level discriminated union for `autoIndex` + `defaultIndexType`** — non-trivial config type change, runtime validation covers this now
- **`setSizeCallback` deduplication in no-index path** — needs deeper investigation
- **Test coverage for dev mode suggestion system and BasicIndex** — larger effort, tracked separately

PR #1368, opened 2026-03-13
fix(db-ivm): hash Temporal objects by value instead of identity

### LINKED ISSUE #1367: Temporal dates break live query updates when using joins
## Bug

Live queries with joins don't update when a Temporal date field changes on a row. Simple queries (no joins) work fine.

## Cause

The hashing function in `db-ivm` uses `Object.keys()` to structurally hash objects. Temporal dates have no enumerable own properties, so `Object.keys()` returns `[]` and every Temporal date hashes to the same value regardless of the actual date.

When a Temporal date changes on a row, the join index sees `hash(oldRow) === hash(newRow)`, treats them as identical, and the -1 and +1 multiplicities cancel out. The update gets silently swallowed.

## Reproduction

Works: `q.from({ todo: todoCollection })`

Broken: `q.from({ todo: todoCollection }).innerJoin({ list: listCollection }, ({ todo, list }) => eq(todo.list_id, list.id)).select(({ todo, list }) => ({ todo, list }))`

Updating a Temporal date field (e.g. `dueDate`) on a todo will not propagate through the live query when a join is involved.

## Proposed fix

In `hashObject()` in `packages/db-ivm/src/hashing/hash.ts`, detect Temporal objects via `Symbol.toStringTag` and hash them by their `toString()` representation instead of by enumerable keys.

goatrenterguy:
## Summary

Fixes #1367

Temporal objects (`PlainDate`, `ZonedDateTime`, `Instant`, etc.) have no enumerable own properties, so `Object.keys()` returns `[]`. This caused the IVM `hash()` function to produce identical hashes for all Temporal instances, which meant the join `Index` treated old and new rows as equal and silently swallowed updates when only a Temporal field changed.

This adds a Temporal-specific branch in `hashObject()` that detects Temporal objects via `Symbol.toStringTag` (the same detection approach already used in `packages/db/src/utils.ts`) and hashes them by their type tag and `toString()` representation.

### Changes

- `packages/db-ivm/src/hashing/hash.ts` - Added `isTemporal()` detection and `hashTemporal()` hashing before the plain object fallback
- `packages/db-ivm/tests/utils.test.ts` - Unit tests for Temporal hash correctness (same value, different value, cross-type distinction)
- `packages/db/tests/query/join.test.ts` - Regression test verifying Temporal field updates propagate through joined live queries

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a critical issue where live join queries would silently ignore updates to Temporal date/time fields. Previously, when only Temporal fields like PlainDate or ZonedDateTime changed, live queries wouldn't reflect those updates. Now these changes are properly detected, ensuring live queries accurately reflect all data changes in real-time.

### DIFF
--- a/.changeset/fix-temporal-hash.md
+++ b/.changeset/fix-temporal-hash.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/db-ivm': patch
+---
+
+Fix Temporal objects (PlainDate, ZonedDateTime, etc.) producing identical hashes in the IVM hash function. Temporal objects have no enumerable own properties, so Object.keys() returns [] and all instances were hashed identically. This caused join live queries to silently swallow updates when only a Temporal field changed. Temporal objects are now hashed by their type tag and string representation.
--- a/packages/db-ivm/src/hashing/hash.ts
+++ b/packages/db-ivm/src/hashing/hash.ts
@@ -18,6 +18,7 @@ const ARRAY_MARKER = randomHash()
 const MAP_MARKER = randomHash()
 const SET_MARKER = randomHash()
 const UINT8ARRAY_MARKER = randomHash()
+const TEMPORAL_MARKER = randomHash()
 
 // Maximum byte length for Uint8Arrays to hash by content instead of reference
 // Arrays smaller than this will be hashed by content, allowing proper equality comparisons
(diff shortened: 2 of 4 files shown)


KyleAMathews:
Heh we both wrote PRs 😆 I just pulled yours into mine at https://github.com/TanStack/db/pull/1370 so I'll close this one.

PR #1370, opened 2026-03-13
fix(db, db-ivm): support Temporal objects in join hashing and normalization

### LINKED ISSUE #1367: Temporal dates break live query updates when using joins

KyleAMathews:
Temporal objects (e.g. `Temporal.PlainDate`) break live query updates when joins are involved. Simple queries without joins work fine.

Incorporates #1368 by @goatrenterguy (Ben Guericke) which provides the core hash fix and tests.

## Root Cause

The structural hash function in `db-ivm` uses `Object.keys()` to hash objects. Temporal objects store data in non-enumerable internal slots, so `Object.keys()` returns `[]` — every Temporal value hashes identically. When a Temporal field changes on a row in a join index, `hash(oldRow) === hash(newRow)`, the `-1` and `+1` multiplicities cancel out, and the update is silently swallowed.

## Approach

Three related gaps in Temporal support along the join/query pipeline:

**1. Hashing (`db-ivm`)** — The primary bug (from #1368). Detect Temporal objects via `Symbol.toStringTag` and hash by `TEMPORAL_MARKER + typeTag + toString()` instead of falling through to `hashPlainObject`. Follows the same pattern used for `Date` (marker + value representation).

**2. Value normalization (`db`)** — `normalizeValue()` converts special types to primitives for `Map` key equality in join key matching. Added Temporal → string conversion (`__temporal__${tag}__${toString}`), following the existing `Date → getTime()` and `Uint8Array → __u8__` patterns.

**3. Comparator (`db`)** — `ascComparator()` had no Temporal handling, so `ORDER BY` on Temporal fields would sort by arbitrary object reference IDs. Added string-based comparison via `toString()`, which produces lexicographically sortable ISO 8601 strings for date/time types.

Also made `isTemporal()` defensive (null guard on the exported API) and converted the `temporalTypes` list from `Array` to `Set` for consistency with the `db-ivm` copy.

## Key Invariants

- Different Temporal values must produce different hashes
- Same Temporal value (different instances) must produce the same hash
- Temporal join keys must match across both sides of a join via `Map` lookup
- Temporal sort order must be chronological, not reference-based

## Non-goals

- Deduplicating the `temporalTypes`/`isTemporal` definitions across `db` and `db-ivm` packages (would require cross-package dependency changes)
- Handling `Temporal.Duration` sort ordering (durations aren't lexicographically sortable via `toString()` — a separate concern)

## Verification

```bash
pnpm run build
pnpm --filter @tanstack/db-ivm test -- --run   # 322 tests pass
pnpm --filter @tanstack/db test -- --run        # 2082 tests pass
```

## Files Changed

| File | Change |
|------|--------|
| `packages/db-ivm/src/hashing/hash.ts` | Add `TEMPORAL_MARKER`, `temporalTypes`, `isTemporal()`, `hashTemporal()` |
| `packages/db-ivm/tests/utils.test.ts` | Unit tests for Temporal hash correctness (from #1368) |
| `packages/db/tests/query/join.test.ts` | Regression test for Temporal join updates (from #1368) |
| `packages/db/src/utils/comparison.ts` | Add Temporal branch in `normalizeValue()` and `ascComparator()` |
| `packages/db/src/utils.ts` | Add null guard to `isTemporal()`, convert `temporalTypes` to Set |
| `.changeset/fix-temporal-join-hashing.md` | Patch changeset for both packages |

---

Fixes #1367
Incorporates #1368

## Summary by CodeRabbit

* **Bug Fixes**
  * Temporal date/time objects (PlainDate, PlainDateTime, PlainTime, Instant) now hash, normalize, compare, and sort correctly in joins so live query updates to Temporal fields propagate reliably.
* **Tests**
  * Added regression tests covering Temporal hashing, comparison, join behavior, and live query update propagation.
* **Chores**
  * Added a changeset documenting the Temporal-related fixes.

### DIFF
--- a/packages/db-ivm/package.json
+++ b/packages/db-ivm/package.json
@@ -55,6 +55,7 @@
   },
   "devDependencies": {
     "@types/debug": "^4.1.12",
-    "@vitest/coverage-istanbul": "^3.2.4"
+    "@vitest/coverage-istanbul": "^3.2.4",
+    "temporal-polyfill": "^0.3.0"
   }
 }
--- a/packages/db-ivm/tests/utils.test.ts
+++ b/packages/db-ivm/tests/utils.test.ts
@@ -1,18 +1,7 @@
 import { describe, expect, it } from 'vitest'
+import { Temporal } from 'temporal-polyfill'
 import { DefaultMap } from '../src/utils.js'
 import { hash } from '../src/hashing/index.js'
-
-// Minimal mock that mimics Temporal objects: Symbol.toStringTag + toString()
-// without requiring the temporal-polyfill dependency.
-function createTemporalLike(
-  tag: string,
-  value: string,
-): { toString: () => string; [Symbol.toStringTag]: string } {
-  return Object.create(null, {
-    [Symbol.toStringTag]: { value: tag },
-    toString: { value: () => value },
-  })
-}
 
 describe(`DefaultMap`, () => {
   it(`should return default value for missing keys`, () => {
@@ -183,9 +172,9 @@ describe(`hash`, () => {
     })
 
     it(`should hash Temporal objects by value`, () => {
-      const date1 = createTemporalLike(`Temporal.PlainDate`, `2024-01-15`)
-      const date2 = createTemporalLike(`Temporal.PlainDate`, `2024-01-15`)
-      const date3 = createTemporalLike(`Temporal.PlainDate`, `2024-06-15`)
+      const date1 = Temporal.PlainDate.from(`2024-01-15`)
+      const date2 = Temporal.PlainDate.from(`2024-01-15`)
+      const date3 = Temporal.PlainDate.from(`2024-06-15`)
 
       const hash1 = hash(date1)
       const hash2 = hash(date2)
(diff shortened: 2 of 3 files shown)


samwillis (changes requested):
Looks like we don't have the types for Temporal in the typescript setup. We should add them rather than cast as `any`.

samwillis (approved):
:shipit:

Issue #350, opened 2025-07-31
Expose QueryFunctionContext properties (signal, meta) in query-db-collection


KyleAMathews:
### Background

TanStack Query provides a `QueryFunctionContext` to query functions with useful properties like `signal` for cancellation and `meta` for passing additional context. Currently, query-db-collection passes this context to the queryFn but doesn't fully utilize or expose these capabilities.

### Current State

```typescript
// TanStack Query provides:
interface QueryFunctionContext {
  queryKey: QueryKey
  signal?: AbortSignal  // For cancellation
  meta?: Record<string, unknown>  // Metadata
  pageParam?: unknown  // For infinite queries
}

// Current usage in query-db-collection:
queryFn: async (context) => {
  // context is passed but signal and meta aren't commonly used
  return api.fetchTodos()
}
```

### Problems

1. **No request cancellation**: Can't cancel in-flight requests when collections are cleaned up
2. **No metadata support**: Can't pass additional context through the query system
3. **Missed optimization**: Unnecessary network requests continue even after cleanup

#### 1. Document and encourage signal usage

```typescript
const todoCollection = createCollection(
  queryCollectionOptions({
    queryKey: ['todos'],
    queryFn: async (context) => {
      // Use signal for cancellable requests
      const response = await fetch('/api/todos', {
        signal: context.signal
      })
      
      if (\!response.ok) throw new Error('Failed to fetch')
      return response.json()
    },
    getKey: (item) => item.id,
    queryClient,
  })
)
```

#### 2. Add meta support to QueryCollectionConfig

```typescript
export interface QueryCollectionConfig<TItem, TError, TQueryKey> {
  // ... existing options ...
  
  /**
   * Metadata to pass to the query
   * Available in queryFn via context.meta
   */
  meta?: Record<string, unknown>
}
```

#### Example 1: Cancellable Fetch with AbortSignal

```typescript
const searchCollection = createCollection(
  queryCollectionOptions({
    queryKey: ['search', searchTerm],
    queryFn: async (context) => {
      try {
        const response = await fetch(`/api/search?q=${searchTerm}`, {
          signal: context.signal
        })
        return response.json()
      } catch (error) {
        if (error.name === 'AbortError') {
          console.log('Search cancelled')
          return [] // Return empty results for cancelled requests
        }
        throw error
      }
    },
    getKey: (item) => item.id,
    queryClient,
  })
)
```

#### Example 2: Using Meta for Error Context

```typescript
const userCollection = createCollection(
[...]

PR #1375, opened 2026-03-14
feat: Add meta in queries

Jastrzebowski:
### Summary
* Add a chainable `.meta()` method to pass dynamic runtime context (tenant IDs, scopes, feature flags) to query functions
* Ensure query-level metadata is properly serialised in cache keys to prevent data isolation issues in multi-tenant and authorisation scenarios
* Include comprehensive documentation and tests covering meta propagation, merging semantics, and real-world use cases

Discussion and issues related with passing meta in queries: 
* https://github.com/TanStack/db/discussions/1129
* https://github.com/TanStack/db/issues/350

### Problem
Applications need to pass runtime context to their query functions — such as tenant IDs for multi-tenancy, authorisation scopes, or feature flags — without embedding these values in query predicates. Previously, there was no way to attach this metadata to queries, forcing developers to either hardcode context or use workarounds. 

Additionally, when different runtime contexts share the same predicates, caching and deduplication logic must account for meta differences, otherwise, queries from different tenants or users with identical predicates would incorrectly share cache entries, causing data isolation bugs.

### Approach
The `.meta()` method accepts arbitrary context data and integrates seamlessly with the query builder, following the same chainable pattern as existing methods like `.limit()`. Query-level metadata is merged with collection-level metadata (with query-level taking precedence). 

Meta is included in the cache key serialisation, ensuring that queries with identical predicates but different runtime contexts receive separate cache entries. The `isPredicateSubset()` deduplication logic also checks meta equality, preventing data corruption in multi-tenant scenarios where two requests with the same predicates but different tenant scopes would otherwise reuse cached data from the wrong tenant.

The implementation includes tests that validate meta preservation throughout the query lifecycle, correct merging semantics, proper cache key differentiation, and deduplication behaviour. 

Documentation covers usage in live queries and extends the existing `meta` documentation for collections.

## ✅ Checklist

## 🚀 Release Impact


Jastrzebowski:
Hey TanStack team, my first contribution here. I didn't include the generated docs as they generate a lot of changes. What's the general rule: include generated docs in the PR or create a follow-up PR for docs only?

kevin-dp:
> Hey TanStack team, my first contribution here. I didn't include the generated docs as they generate a lot of changes. What's the general rule: include generated docs in the PR or create a follow-up PR for docs only?

Hi @Jastrzebowski, changes to docs should be made in a separate PR because the docs are built and updated by CI when merged into main. So we typically have an implementation PR and a separate docs PR and we only merge the docs PR when the corresponding implementation is released.

Jastrzebowski:
Thank you, @kevin-dp. I understand your concerns, and I agree it's important we don't introduce single-collection features to the core library. Do you see a cleaner way to extend `queryCollectionOptions` to pass parameters at query time?

PR #1379, opened 2026-03-17
feat: persisted query metadata for warm-start consistency

kevin-dp:
## Summary

Implements Phase 1 of persisted query metadata to fix a warm-start timing race where rows hydrated from SQLite persistence are incorrectly deleted by disjoint queries.

**The bug:** When two disjoint queries share a persisted collection (e.g., "history" and "live" message queries), the first query to return can delete rows that belong to the other query's domain. The live query correctly returns `[]` (no live messages), but the query result handler interprets this as "every unclaimed row should be deleted," wiping out the history rows.

**The fix (Phase 1):** Per-row reference counting with persistence support.

### Changes

- **`@tanstack/db`** — Added `MetadataStorage` interface and optional `metadataStorage` on sync params, so any sync layer can persist metadata
- **`query-db-collection`** — `addRow` now tracks new claims and increments a ref count; `removeRow` only decrements if the query actually owned the row (returns `false` if query never loaded it); `cleanupQueryInternal` uses ref counts consistently; ref counts are loaded from and persisted to `metadataStorage`
- **`db-sqlite-persisted-collection-core`** — Added `ref_count` column on the row data table; `loadSubset` returns `refCount`; added `updateRefCounts`, `loadMetadata`, `storeMetadata`, `deleteMetadata` methods; added `collection_metadata` table; `createWrappedSyncConfig` injects `metadataStorage` into sync params

### Phases 2 & 3 (future)

- **Phase 2:** Persist full `queryToRows`/`rowToQueries` ownership maps so the first query result can properly diff against the previous session's state
- **Phase 3:** Persist per-query GC settings with absolute expiration timestamps for placeholder query lifecycle management

See `rfcs/001-persisted-query-metadata.md` (not included in this PR) for the full design.

## Test plan

- [x] New test: "should retain pre-hydrated rows when a disjoint query correctly returns empty results" — reproduces the bug and verifies the fix
- [x] All 86 query-db-collection tests pass
- [x] All 39 persisted-collection-core tests pass
- [x] No regressions in db package tests

### DIFF
--- a/packages/db-sqlite-persisted-collection-core/src/persisted.ts
+++ b/packages/db-sqlite-persisted-collection-core/src/persisted.ts
@@ -223,10 +223,7 @@ export interface PersistenceAdapter<
     key: string,
     value: unknown,
   ) => Promise<void>
-  deleteMetadata?: (
-    collectionId: string,
-    key: string,
-  ) => Promise<void>
+  deleteMetadata?: (collectionId: string, key: string) => Promise<void>
 }
 
 export interface SQLiteDriver {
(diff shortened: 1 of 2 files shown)


kevin-dp:
Closing in favor of https://github.com/TanStack/db/pull/1380

PR #1382, opened 2026-03-17
feat: add stringAgg aggregate function

KyleAMathews:
## Summary

Adds a `stringAgg` aggregate function that concatenates string values within groups, with configurable separators and ordering. Available across the full stack: query builder, compiler, and IVM engine.

## Approach

**Two-tier IVM architecture:**

- **Stateful incremental path** (used by the query compiler): Maintains a sorted array of entries per group with O(log n) binary search for insertion/removal. Fast-path string slicing for head/tail changes avoids full rebuilds — only middle-position mutations trigger a rebuild from the ordered entries.
- **Stateless fallback path** (no `rowKeyExtractor`): Simple sort-and-join on each update. Used when row identity isn't available.

**Query builder overloads** disambiguate the flexible API surface:
```ts
stringAgg(value)                        // default order, no separator
stringAgg(value, orderBy)               // ordered, no separator  
stringAgg(value, separator)             // separator, default order
stringAgg(value, separator, orderBy)    // both
```

The compiler distinguishes separator (literal string) from orderBy (column reference) by expression type, and always provides a `rowKeyExtractor` to enable the incremental path.

## Key invariants

- `entriesByKey` and `orderedEntries` must stay synchronized — `removeStringAggEntry` now throws on desynchronization rather than silently returning
- Fast-path text splicing is only used for first/last position changes; middle mutations set `textDirty` for a full rebuild
- Null/undefined values are excluded from concatenation (consistent with SQL `string_agg` semantics)

## Non-goals

- Custom comparators for ordering (uses built-in comparison with Date normalization)
- Streaming/lazy concatenation for very large groups

## Verification

```bash
pnpm vitest run packages/db-ivm/tests/operators/groupBy.test.ts packages/db/tests/query/builder/functions.test.ts packages/db/tests/query/group-by.test.ts packages/db/tests/query/group-by.test-d.ts
```

All 140 tests pass, including new coverage for:
- Incremental inserts/removes/updates with ordering
- Group deletion + re-creation (cleanup path)
- Fallback path (no `rowKeyExtractor`)
- Values containing the separator string
- Builder overload disambiguation
- Type-level tests for return types

## Files changed

| File | Change |
|------|--------|
| `packages/db-ivm/src/operators/groupBy.ts` | Core `stringAgg` implementation with binary search, incremental text maintenance, and fallback path |
| `packages/db-ivm/src/operators/reduce.ts` | Pass group key to reduction function |
[...]

PR #1384, opened 2026-03-18
feat: support concat(toArray(...)) scalar includes

KyleAMathews:
## Summary

- add scalar includes materialization so `toArray(subquery.select(({ c }) => c.text))` produces scalar arrays
- support `concat(toArray(...))` by lowering it through includes materialization instead of introducing a new aggregate
- make scalar child `select()` type-check without user-facing casts while preserving existing object-select inference

## Design

- extend includes materialization from `collection` to `collection | array | concat`
- wrap scalar child selects in an internal `__includes_scalar__` field so the existing includes pipeline can keep using object-shaped child rows
- materialize inline array/concat values in the live query output layer and re-emit parent rows when child chunks change
- add builder/type support for scalar `select()` results without breaking spread/object projections

## Benchmarks

Actual live-query benchmark for:
- `content: concat(toArray(q.from({ c: chunkCollection }).where(...).orderBy(...).select(({ c }) => c.text)))`

```text
1 message x 2000 chunks
concatToArray  preload 17.52ms  append 0.94ms  mid update 0.95ms
stringAgg      preload 13.21ms  append 1.29ms  mid update 1.53ms

100 messages x 100 chunks
concatToArray  preload 70.55ms  append 1.14ms  mid update 1.06ms
stringAgg      preload 65.46ms  append 0.88ms  mid update 1.00ms
```

Update performance is competitive, and the single-message incremental case is slightly better on the concat/toArray path. Preload is still a bit slower than the purpose-built aggregate.

## Verification

```bash
pnpm exec tsc -p packages/db/tsconfig.json --noEmit --pretty false
pnpm vitest --run packages/db/tests/query/builder/functions.test.ts packages/db/tests/query/includes.test.ts packages/db/tests/query/group-by.test.ts
```

## Notes

- This replaces the design from #1382 rather than extending `groupBy` with a new aggregate.
- The separate `.select(...).fn.select(...)` idea is still blocked because `fn.select` runs before includes materialize.

### DIFF
--- a/packages/db/src/query/builder/functions.ts
+++ b/packages/db/src/query/builder/functions.ts
@@ -2,7 +2,12 @@ import { Aggregate, Func } from '../ir'
 import { toExpression } from './ref-proxy.js'
 import type { BasicExpression } from '../ir'
 import type { RefProxy } from './ref-proxy.js'
-import type { Context, GetRawResult, RefLeaf } from './types.js'
+import type {
+  Context,
+  GetRawResult,
+  RefLeaf,
+  StringifiableScalar,
+} from './types.js'
 import type { QueryBuilder } from './index.js'
 
 type StringRef =
@@ -38,8 +43,20 @@ type ComparisonOperandPrimitive<T extends string | number | boolean> =
   | undefined
   | null
 
-// Helper type for any expression-like value
-type ExpressionLike = BasicExpression | RefProxy<any> | RefLeaf<any> | any
+// Helper type for values that can be lowered to expressions.
+type ExpressionLike =
+  | Aggregate
+  | BasicExpression
+  | RefProxy<any>
+  | RefLeaf<any>
+  | string
+  | number
+  | boolean
+  | bigint
+  | Date
+  | null
+  | undefined
+  | Array<unknown>
 
 // Helper type to extract the underlying type from various expression types
 type ExtractType<T> =
@@ -277,10 +294,12 @@ export function length<T extends ExpressionLike>(
   return new Func(`length`, [toExpression(arg)]) as NumericFunctionReturnType<T>
 }
 
-export function concat<T>(arg: ToArrayWrapper<T>): ConcatToArrayWrapper<T>
+export function concat<T extends StringifiableScalar>(
+  arg: ToArrayWrapper<T>,
+): ConcatToArrayWrapper<T>
 export function concat(...args: Array<ExpressionLike>): BasicExpression<string>
 export function concat(
-  ...args: Array<ExpressionLike>
+  ...args: Array<ExpressionLike | ToArrayWrapper<any>>
 ): BasicExpression<string> | ConcatToArrayWrapper<any> {
   const toArrayArg = args.find(
     (arg): arg is ToArrayWrapper<any> => arg instanceof ToArrayWrapper,
@@ -417,13 +436,15 @@ export const operators = [
 
 export type OperatorName = (typeof operators)[number]
 
-export class ToArrayWrapper<_T = any> {
+export class ToArrayWrapper<_T = unknown> {
   declare readonly _type: `toArray`
+  declare readonly _result: _T
   constructor(public readonly query: QueryBuilder<any>) {}
 }
 
-export class ConcatToArrayWrapper<_T = any> {
+export class ConcatToArrayWrapper<_T = unknown> {
   declare readonly _type: `concatToArray`
+  declare readonly _result: _T
   constructor(public readonly query: QueryBuilder<any>) {}
 }
 
--- a/packages/db/src/query/builder/types.ts
+++ b/packages/db/src/query/builder/types.ts
@@ -8,7 +8,7 @@ import type {
   PropRef,
   Value,
 } from '../ir.js'
-import type { QueryBuilder } from './index.js'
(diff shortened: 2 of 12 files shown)


packages/db/src/query/builder/functions.ts
```diff
@@ -277,9 +277,24 @@ export function length<T extends ExpressionLike>(
   return new Func(`length`, [toExpression(arg)]) as NumericFunctionReturnType<T>
 }
 
+export function concat<T>(arg: ToArrayWrapper<T>): ConcatToArrayWrapper<T>
+export function concat(...args: Array<ExpressionLike>): BasicExpression<string>
 export function concat(
   ...args: Array<ExpressionLike>
-): BasicExpression<string> {
+): BasicExpression<string> | ConcatToArrayWrapper<any> {
```
samwillis, line 303: Note to self / check: the `any` here made me think we were loosing type information, but I think the function overload above for `ConcatToArrayWrapper<T>` means we don't?

samwillis:
I pulled this branch locally and worked through the typing/runtime boundary issues in detail.

I’ve now pushed an update that keeps the intended runtime work here, but tightens the public surface so we don’t claim support beyond what the compiler/live-query path actually handles.

## What changed

### 1. Kept the scalar includes runtime path
The runtime/materialization work in this PR remains the core of the implementation:

- scalar child selects in `toArray(...)` still work
- `concat(toArray(...))` still lowers through includes materialization
- the `INCLUDES_SCALAR_FIELD` wrapping and inline materialization path are preserved
- runtime tests for scalar array materialization and concat re-emits are in place

### 2. Tightened the root scalar boundary
The broad scalar `select()` typing is no longer allowed to leak through root consumers.

Concretely:
- top-level scalar `select(({ users }) => users.country)` is no longer advertised as supported in docs/examples
- `createLiveQueryCollection(...)` and `queryOnce(...)` now reject scalar-result builders at the type level
- JS users also get a clear runtime error at that same boundary instead of silent mismatch

That means scalar `select()` is still usable where this PR actually needs it, but not exposed as a generally supported top-level query result shape.

### 3. Removed the `select()` catch-all `any` escape hatch
The public catch-all `select()` overload that returned `any` has been removed, so ambiguous callbacks no longer silently widen the builder surface.

### 4. Tightened `concat(toArray(...))`
`concat(toArray(...))` is now restricted at the type level to stringifiable scalar child results.

So invalid cases like:
- `concat(toArray(q.from(...)))`
- `concat(toArray(q.from(...).select(() => ({ ... }))))`

now fail in the type system instead of only later at runtime.

## What we folded in from #1276
I also pulled in the useful part of [#1276](https://github.com/TanStack/db/pull/1276), but in a narrower form.

What we now support:
- returning a top-level alias ref directly, e.g. `select(({ users }) => users)`, behaves like the spread form and projects the full row

What we intentionally do not do:
- we do **not** rewrite arbitrary direct refs
- scalar leaf refs like `select(({ c }) => c.text)` remain true scalar selections

That narrowing was important because this PR depends on scalar child selects remaining scalar for the `toArray(...)` / `concat(toArray(...))` path.

So effectively:
- alias/object ref direct returns are treated as spread-equivalent
- scalar leaf refs are left alone

## Coverage added
I added:
- positive type tests for scalar child `toArray(...)`
- positive type tests for scalar `concat(toArray(...))`
- a composability type test showing a scalar-selecting builder can be assigned to a variable and then used in `toArray(...)`
- negative type tests for invalid `concat(toArray(...))`
- negative type tests for root scalar queries
- runtime coverage for the root scalar rejection boundary
- runtime and type coverage for the narrowed direct-alias-return behavior

## Validation
Locally I ran:

- `pnpm exec tsc -p packages/db/tsconfig.json --noEmit --pretty false`
- targeted query/runtime suites for includes, select-spread, and builder function coverage

At this point I think the branch is in a much better place:
- the intended feature is still here
- the unsupported surface area is no longer being widened
- and the safe portion of #1276 is folded in without conflicting with the scalar-select design

packages/db/src/query/builder/functions.ts
```diff
@@ -38,8 +43,20 @@ type ComparisonOperandPrimitive<T extends string | number | boolean> =
...
+  | RefProxy<any>
+  | RefLeaf<any>
+  | string
+  | number
+  | boolean
+  | bigint
+  | Date
+  | null
+  | undefined
+  | Array<unknown>
```
kevin-dp: Would `Array<ExpressionLike>` be more precise?

packages/db/src/query/builder/functions.ts
kevin-dp: I don't think we do, we have 2 concrete signatures above (overloads) and this one is just the implementation of that overloaded function

packages/db/src/query/builder/types.ts
```diff
@@ -192,6 +206,69 @@ type SelectShape = { [key: string]: SelectValue | SelectShape }
...
+        ? false
+        : true
+      : false
+export type NonScalarSelectObject<T> = T extends SelectObject
+  ? IsScalarSelectLike<T> extends true
+    ? never
+    : T
+  : never
+
+export type ResultTypeFromSelectValue<TSelectValue> =
```
kevin-dp, line 227: Gotta love this 😆

Issue #1394, opened 2026-03-20
Feature: effect cursor / startAfter for durable replay


KyleAMathews:
## Use Case

We're building a tool where entities wake, run effects, and go idle. On re-wake, the runtime replays effects from persisted manifest entries. The problem: replayed effects fire `onEnter` for all existing rows that match the query, including rows that were already processed on a previous wake.

### The scenario

1. Entity wakes, creates an effect watching a child entity's `runs` collection
2. Effect's `onEnter` fires when a child's run completes → sends a "child-done" message
3. Entity goes idle
4. Entity re-wakes (new message arrives)
5. Runtime replays the effect from the manifest
6. Effect fires `onEnter` again for the already-completed run → duplicate "child-done" message

### What we need

A way to tell an effect: "only fire callbacks for mutations that happen **after** this point." Something like:

```ts
createEffect({
  query: (q) => q.from({ runs: childDb.collections.runs }),
  startAfter: cursor, // <-- "skip everything before this point"
  onEnter: (event) => {
    // Only fires for new completions, not ones from previous wakes
  },
})
```

### Current workarounds

- `skipInitial: true` — too aggressive. Skips ALL initial data, including events that arrived while the entity was idle (which are exactly the ones we need to process).
- Making every effect handler idempotent (check state before acting) — works but pushes complexity to every effect author. We'd prefer the runtime handles this.

## The Challenge

TanStack DB doesn't have a global offset/sequence number on collection mutations. Effects work on collection state (rows matching a query), not on an event log. So "start after offset X" doesn't have a natural mapping.

### Possible approaches

1. **Row-level versioning**: Each row gets a monotonic version number on mutation. Effects with `startAfter` skip rows whose version is ≤ the cursor. New mutations (inserts, updates, deletes) after the cursor fire normally.

2. **Subscription-level cursor**: The effect subscription could be configured to treat the current snapshot as "already seen" and only fire callbacks for future changes. This is similar to `skipInitial` but more precise — it skips the specific rows that exist at creation time, not all initial load data.

3. **Collection-level epoch**: Collections track a mutation counter. `startAfter: epoch` means "fire onEnter/onExit only for mutations with epoch > startAfter." The runtime would store the epoch when the entity goes idle and pass it back when replaying effects.

4. **External filtering in onEnter**: The caller filters in the callback (our current workaround). Not ideal but works.

### What we're doing with this

[...]

PR #1405, opened 2026-03-25
feat: cursor-aware effect replay with startAfter gating

KyleAMathews:
## Summary

Add `startAfter` cursor option to `createEffect` and `useLiveQueryEffect`, enabling effects to skip callbacks during historical replay while still hydrating query state. Supports both single-source effects (scalar cursor) and join queries (`Record<string, CollectionCursor>` for per-source gating).

## Root Cause

When replaying a stream of sync changes (e.g., after reconnection or page reload), `createEffect` fires callbacks for every historical change. There was no way to say "I've already processed everything up to cursor X — only fire for new changes." This forced consumers to either accept duplicate side-effects during replay or build their own deduplication layer.

## Approach

Thread a `CollectionCursor` (`string | number`) through three layers:

**1. Sync layer** (`sync.ts`, `state.ts`): `sync.write()` accepts an optional `cursor`. The collection state manager propagates it through `PendingSyncedTransaction` → `eventCursors` → emitted `ChangeMessage` events.

**2. Effect pipeline** (`effect.ts`): New `startAfter` config option. When set, the `EffectPipelineRunner` gates callbacks per-alias: changes at or before `startAfter[alias]` still hydrate the D2 query graph (so query state is correct), but `flushPendingChanges` discards events until a live source contributes changes past its cursor boundary. For single-source effects, a scalar `startAfter` value is automatically normalized to the single alias.

**3. DeltaEvent enrichment** (`effect.ts`): Each emitted `DeltaEvent` carries:
- `cursor`: batch-level high-water cursor (max across all source cursors)
- `cursors`: per-source cursor map (for gated effects)
- `triggeringSource`: which source alias drove this batch (for gated effects)

## Key Invariants

- Pre-cursor changes **must** hydrate query state — otherwise `previousValue` on the first live `update` event would be wrong
- Each per-alias gate opens exactly once and never re-closes
- Cursor-less changes after replay open the gate (sync provider moved past cursor-based protocol)
- Uncursored changes adjacent to a gate-opening cursor are treated as live, not replay
- `compareCollectionCursors` type mismatches route through `onSourceError` for graceful disposal
- Join tap callbacks during graph runs (lazy source loads) inherit the root batch's live status and don't affect cursor gates
- A scalar `startAfter` with multi-source effects throws (use Record form for joins)

## Non-goals

- Per-row cursor on DeltaEvents (batch-level is intentional — the batch is the checkpoint unit)
- Parameterizing `CollectionCursor` by primitive type (runtime check is sufficient for now)

## Trade-offs

The cursor is **batch-level**, using the highest cursor seen per-source in the batch. This means the cleanest semantics come when replay writes are cursor-ordered and each sync commit maps to one cursor boundary. This was chosen over per-row cursors because the batch is the natural checkpoint granularity — saving a cursor after processing a batch is the common consumer pattern.

Per-alias gating adds complexity to the effect pipeline, but the alternative (requiring consumers to manually filter events by source) would push that complexity to every call site and couldn't protect `previousValue` correctness.

## Verification

```bash
pnpm vitest run packages/db/tests/effect.test.ts --pool-options.threads.maxThreads=2
pnpm vitest run packages/db/tests/collection.test.ts --pool-options.threads.maxThreads=2
pnpm exec tsc -p packages/db/tsconfig.json --noEmit
pnpm exec tsc -p packages/react-db/tsconfig.json --noEmit
```

## Files changed

| File | Change |
|------|--------|
| `packages/db/src/types.ts` | Add `CollectionCursor` type, `cursor` field on `ChangeMessage` |
| `packages/db/src/collection/state.ts` | `PendingCursorWrite` type, cursor propagation through `eventCursors` → emitted change events |
| `packages/db/src/collection/sync.ts` | Record cursor from sync writes into `rowCursorWrites` |
| `packages/db/src/query/effect.ts` | Per-alias `startAfter` gating, `DeltaEvent` with `cursor`/`cursors`/`triggeringSource`, join-tap-aware graph run isolation, `compareCollectionCursors` |
| `packages/db/src/query/live/utils.ts` | Propagate cursor/metadata through `splitUpdates` via spread |
| `packages/react-db/src/useLiveQueryEffect.ts` | Pass `startAfter` through to `createEffect` |
| `packages/electric-db-collection/src/electric.ts` | Pass `stream.lastOffset` as cursor in sync writes |
| `packages/db/tests/effect.test.ts` | 11 new tests: cursor gating, per-source joins, triggeringSource, type mismatches, batch straddling, monotonicity |
| `packages/db/tests/collection.test.ts` | 1 new test: sync cursor subscription propagation |

### DIFF
--- a/.changeset/cursor-aware-effect-replay.md
+++ b/.changeset/cursor-aware-effect-replay.md
@@ -3,4 +3,4 @@
 '@tanstack/react-db': patch
 ---
 
-Add cursor-aware effect replay via `startAfter` option on `createEffect` and `useLiveQueryEffect`. Sync writes can now carry an opaque sortable cursor that propagates through `ChangeMessage` and `DeltaEvent`, enabling effects to suppress callbacks during historical replay while still hydrating internal query state.
+Add cursor-aware effect replay via `startAfter` option on `createEffect` and `useLiveQueryEffect`. Sync writes can now carry an opaque sortable cursor that propagates through `ChangeMessage` and `DeltaEvent`, enabling effects to suppress callbacks during historical replay while still hydrating internal query state. `startAfter` accepts either a scalar cursor (single-source) or a `Record<string, CollectionCursor>` for per-source gating in join queries. `DeltaEvent` now includes `triggeringSource` and `cursors` fields for gated effects.
--- a/packages/db/src/query/effect.ts
+++ b/packages/db/src/query/effect.ts
@@ -51,6 +51,10 @@ export type DeltaEvent<
       value: TRow
       /** Batch-level high-water cursor — the maximum cursor observed across all source changes in this batch. Use for checkpointing via startAfter. */
       cursor?: CollectionCursor
+      /** Per-source cursor map. Present when cursor gating (startAfter) is active. */
+      cursors?: Record<string, CollectionCursor>
+      /** Source alias whose changes triggered this batch. Present when cursor gating is active. */
+      triggeringSource?: string
       metadata?: Record<string, unknown>
     }
   | {
@@ -60,6 +64,10 @@ export type DeltaEvent<
       value: TRow
       /** Batch-level high-water cursor — the maximum cursor observed across all source changes in this batch. Use for checkpointing via startAfter. */
       cursor?: CollectionCursor
+      /** Per-source cursor map. Present when cursor gating (startAfter) is active. */
+      cursors?: Record<string, CollectionCursor>
+      /** Source alias whose changes triggered this batch. Present when cursor gating is active. */
+      triggeringSource?: string
       metadata?: Record<string, unknown>
     }
   | {
@@ -71,6 +79,10 @@ export type DeltaEvent<
       previousValue: TRow
       /** Batch-level high-water cursor — the maximum cursor observed across all source changes in this batch. Use for checkpointing via startAfter. */
       cursor?: CollectionCursor
+      /** Per-source cursor map. Present when cursor gating (startAfter) is active. */
+      cursors?: Record<string, CollectionCursor>
+      /** Source alias whose changes triggered this batch. Present when cursor gating is active. */
+      triggeringSource?: string
       metadata?: Record<string, unknown>
     }
 
@@ -145,9 +157,14 @@ export interface EffectConfig<
    * Historical changes at or before the cursor still update internal query state.
    *
    * Requires the sync source to provide a monotonic cursor on every
-   * `sync.write()` call. Only supported for single-source (non-join) effects.
+   * `sync.write()` call.
+   *
+   * - **Scalar value**: applies to the single source alias. Throws for
+   *   multi-source (join) effects.
+   * - **Record**: maps source aliases to their respective cursors, enabling
+   *   independent per-source gating for join queries.
    */
-  startAfter?: CollectionCursor
+  startAfter?: CollectionCursor | Record<string, CollectionCursor>
 }
 
 /** Handle returned by createEffect */
@@ -324,7 +341,7 @@ interface EffectPipelineRunnerConfig<
 > {
   query: EffectQueryInput<any>
   skipInitial: boolean
-  startAfter?: CollectionCursor
+  startAfter?: CollectionCursor | Record<string, CollectionCursor>
   onBatchProcessed: (events: Array<DeltaEvent<TRow, TKey>>) => void
   /** Called when a source collection enters error or cleaned-up state */
   onSourceError: (error: Error) => void
@@ -379,9 +396,13 @@ class EffectPipelineRunner<TRow extends object, TKey extends string | number> {
   // skipInitial state
   private readonly skipInitial: boolean
   private initialLoadComplete = false
-  private readonly startAfter: CollectionCursor | undefined
-  private cursorGateOpen: boolean
-  private pendingBatchCursor: CollectionCursor | undefined
+
+  // Per-alias cursor gating
+  private readonly startAfterByAlias: Map<string, CollectionCursor> = new Map()
+  private readonly cursorGateOpenByAlias: Map<string, boolean> = new Map()
+  private readonly pendingBatchCursorByAlias: Map<string, CollectionCursor> =
+    new Map()
+  private readonly liveAliasesInBatch: Set<string> = new Set()
 
   // Scheduler integration
   private subscribedToAllCollections = false
@@ -401,8 +422,6 @@ class EffectPipelineRunner<TRow extends object, TKey extends string | number> {
 
   constructor(config: EffectPipelineRunnerConfig<TRow, TKey>) {
     this.skipInitial = config.skipInitial
-    this.startAfter = config.startAfter
-    this.cursorGateOpen = config.startAfter === undefined
     this.onBatchProcessed = config.onBatchProcessed
     this.onSourceError = config.onSourceError
 
@@ -413,16 +432,6 @@ class EffectPipelineRunner<TRow extends object, TKey extends string | number> {
     this.collections = extractCollectionsFromQuery(this.query)
     const aliasesById = extractCollectionAliases(this.query)
 
-    if (
-      config.startAfter !== undefined &&
-      Object.keys(this.collections).length > 1
-    ) {
-      throw new Error(
-        `startAfter is only supported for single-source effects. ` +
-          `This effect queries ${Object.keys(this.collections).length} collections.`,
-      )
-    }
-
     // Build alias → collection map
     this.collectionByAlias = {}
     for (const [collectionId, aliases] of aliasesById.entries()) {
@@ -430,6 +439,29 @@ class EffectPipelineRunner<TRow extends object, TKey extends string | number> {
       if (!collection) continue
       for (const alias of aliases) {
         this.collectionByAlias[alias] = collection
+      }
+    }
+
+    // Normalize startAfter into per-alias map
+    if (config.startAfter !== undefined) {
+      if (typeof config.startAfter === `object`) {
+        // Record<string, CollectionCursor> — per-alias cursors
+        for (const [alias, cursor] of Object.entries(config.startAfter)) {
+          this.startAfterByAlias.set(alias, cursor)
+          this.cursorGateOpenByAlias.set(alias, false)
+        }
+      } else {
+        // Scalar — single-source shorthand
+        const aliases = Object.keys(this.collectionByAlias)
+        if (aliases.length !== 1) {
+          throw new Error(
+            `A scalar startAfter value is only supported for single-source effects. ` +
+              `Use a Record<string, CollectionCursor> to map cursors to source aliases. ` +
+              `This effect queries ${aliases.length} collections.`,
+          )
+        }
+        this.startAfterByAlias.set(aliases[0]!, config.startAfter)
+        this.cursorGateOpenByAlias.set(aliases[0]!, false)
       }
     }
 
@@ -650,10 +682,10 @@ class EffectPipelineRunner<TRow extends object, TKey extends string | number> {
         if (orderByInfo) {
           this.trackSentValues(alias, changes, orderByInfo.comparator)
           const split = [...splitUpdates(changes)]
-          this.recordPendingBatchCursor(split)
+          this.recordPendingBatchCursor(alias, split)
           this.sendChangesToD2(alias, split)
         } else {
-          this.recordPendingBatchCursor(changes)
+          this.recordPendingBatchCursor(alias, changes)
           this.sendChangesToD2(alias, changes)
         }
       }
@@ -681,12 +713,28 @@ class EffectPipelineRunner<TRow extends object, TKey extends string | number> {
       return
     }
 
-    if (!this.cursorGateOpen) {
+    // During a graph run, changes arrive from join tap operators (lazy source
+    // loads via requestSnapshot). These inherit the root batch's live status
+    // and must not affect cursor gates or live-source tracking.
+    if (this.isGraphRunning) {
+      this.sendChangesToD2(alias, changes)
+      return
+    }
+
+    // Check per-alias cursor gate
+    if (
+      this.startAfterByAlias.has(alias) &&
+      this.cursorGateOpenByAlias.get(alias) === false
+    ) {
       this.handleSourceChangesBeforeCursorGate(alias, changes)
       return
     }
 
-    this.recordPendingBatchCursor(changes)
+    // Gate is open or alias has no gate — mark as live
+    if (this.startAfterByAlias.size > 0) {
+      this.liveAliasesInBatch.add(alias)
+    }
+    this.recordPendingBatchCursor(alias, changes)
     this.sendChangesToD2(alias, changes)
     this.scheduleGraphRun(alias)
   }
@@ -699,9 +747,10 @@ class EffectPipelineRunner<TRow extends object, TKey extends string | number> {
     alias: string,
     changes: Array<ChangeMessage<any, string | number>>,
   ): void {
+    const startAfter = this.startAfterByAlias.get(alias)!
     let firstLiveIndex: number
     try {
-      firstLiveIndex = findFirstChangeAfterCursor(changes, this.startAfter!)
+      firstLiveIndex = findFirstChangeAfterCursor(changes, startAfter)
     } catch (error) {
       this.onSourceError(
         error instanceof Error ? error : new Error(String(error)),
@@ -724,15 +773,17 @@ class EffectPipelineRunner<TRow extends object, TKey extends string | number> {
       this.runGraph()
     }
 
-    this.cursorGateOpen = true
+    this.cursorGateOpenByAlias.set(alias, true)
+    this.liveAliasesInBatch.add(alias)
 
     const liveChanges = changes.slice(firstLiveIndex)
-    this.recordPendingBatchCursor(liveChanges)
+    this.recordPendingBatchCursor(alias, liveChanges)
     this.sendChangesToD2(alias, liveChanges)
     this.scheduleGraphRun(alias)
   }
 
   private recordPendingBatchCursor(
+    alias: string,
     changes: Array<ChangeMessage<any, string | number>>,
   ): void {
     for (const change of changes) {
@@ -741,18 +792,19 @@ class EffectPipelineRunner<TRow extends object, TKey extends string | number> {
         continue
       }
       try {
-        if (this.pendingBatchCursor === undefined) {
-          this.pendingBatchCursor = cursor
+        const current = this.pendingBatchCursorByAlias.get(alias)
+        if (current === undefined) {
+          this.pendingBatchCursorByAlias.set(alias, cursor)
         } else {
-          const cmp = compareCollectionCursors(cursor, this.pendingBatchCursor)
+          const cmp = compareCollectionCursors(cursor, current)
           if (cmp < 0) {
             throw new Error(
               `Cursors within a sync batch must be monotonically ordered. ` +
-                `Saw ${String(cursor)} after ${String(this.pendingBatchCursor)}.`,
+                `Saw ${String(cursor)} after ${String(current)}.`,
             )
           }
           if (cmp > 0) {
-            this.pendingBatchCursor = cursor
+            this.pendingBatchCursorByAlias.set(alias, cursor)
           }
         }
       } catch (error) {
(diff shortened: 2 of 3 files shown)


samwillis:
I had both Opus and GPT5.2 review this, with the context of our discussion, and the comment on #1394.

The key thing here is that we are trying to bolt on log replay *triggers* on top of the query engine, and there are multiple layers where event collapsing can happen. This api is giving the impression that you will get a true replay of all events, but depending on how the source collections emits them (how it handles up-to-date/markReady), and what other collection are joined they may be collapsed. 

`createEffect` is operating on **visible state deltas** not source events. If it's running on a live stream it will trigger on each change to the visible state, but needs coordination from the source collection to ensure in a replay mode that it will see each change as a transaction that runs `runGraph` and emits the changes. There is a second comment after the review that looks at this.

I do wander if we need a separate `createTrigger` that doesn't take a query, only a collection and operate on that.

---

## Deep Review of `startAfter` Cursor Gating (PR #1405)

I've traced the full path from sync writes through the collection state manager, subscription system, D2 graph engine, and effect pipeline. A second independent review confirmed the same core findings and surfaced additional detail on the Electric adapter's control-point batching and the subscription layer's event rewriting. Here are the combined findings, ordered from most critical to least.

---

### 1. CRITICAL: Buffer Drain Path Bypasses Cursor Gating Entirely

During `start()`, incoming source changes are buffered in `pendingBuffers` until all subscriptions are established. When the buffer drains (lines 672-692), it calls `recordPendingBatchCursor` + `sendChangesToD2` directly — it does **not** call `handleSourceChanges` and therefore completely bypasses `handleSourceChangesBeforeCursorGate`.

```672:692:packages/db/src/query/effect.ts
    for (const [alias] of pendingBuffers) {
      const buffer = pendingBuffers.get(alias)!
      pendingBuffers.delete(alias)

      const orderByInfo = this.getOrderByInfoForAlias(alias)

      // Drain all buffered batches. Since we deleted the alias from
      // pendingBuffers above, any new changes arriving during drain go
      // through handleSourceChanges directly (not back into this buffer).
      for (const changes of buffer) {
        if (orderByInfo) {
          this.trackSentValues(alias, changes, orderByInfo.comparator)
          const split = [...splitUpdates(changes)]
          this.recordPendingBatchCursor(alias, split)
          this.sendChangesToD2(alias, split)
        } else {
          this.recordPendingBatchCursor(alias, changes)
          this.sendChangesToD2(alias, changes)
        }
      }
    }
```

Then `flushPendingChanges` discards the whole batch because no alias was marked live:

```947:955:packages/db/src/query/effect.ts
    if (
      (this.skipInitial && !this.initialLoadComplete) ||
      (hasCursorGating && this.liveAliasesInBatch.size === 0)
    ) {
      this.pendingChanges = new Map()
      this.pendingBatchCursorByAlias.clear()
      this.liveAliasesInBatch.clear()
      return
    }
```

**Concrete failure scenario:** a sync provider that writes everything synchronously inside the `sync` function:

```typescript
  begin()
  write({ type: 'insert', value: A, cursor: 1 }) // replay
  write({ type: 'insert', value: B, cursor: 3 }) // live (past startAfter=2)
  commit()
  markReady()
}
```

The `commit()` triggers `commitPendingTransactions` which emits through the subscription callback. But the callback is in "buffering" mode during `start()`, so both changes end up in the buffer. The buffer drain sends them directly to D2. The subsequent `runGraph()` produces DeltaEvents. `flushPendingChanges` checks `liveAliasesInBatch.size === 0` (since the drain path never populates it) and **discards all events** — including the live data at cursor=3 that should have fired.

The cursor gate for this alias also never opens (`cursorGateOpenByAlias` stays `false`), so if no more data arrives, the effect is permanently stuck.

All existing tests use `mockSyncCollectionOptionsNoInitialState` which avoids this path, so the bug is untested. A test covering "buffered startup batch straddles `startAfter` before all aliases are subscribed" should be added.

**Fix**: The buffer drain loop should route through `handleSourceChanges` (or at least through the cursor gate logic) instead of calling `sendChangesToD2` directly.

---

### 2. CRITICAL: This Does Not, and Cannot With the Current Architecture, Guarantee True Replayability of Distinct Source Events

The branch is consistent with the PR writeup, but the semantics are **"replay-safe checkpointing of net query deltas"**, not **"replay every source event after cursor"**. There are **three** independent collapsing layers that each reduce the number of observable events:

**Layer 1 — `CollectionStateManager`**: `commitPendingTransactions` computes a **visible-state diff**, not a per-write event log. Multiple `write()` calls to the same key in one `begin()`/`commit()` collapse into one emitted `ChangeMessage`. The `rowCursorWrites` map uses `Map.set(key, ...)`, keeping only the **last** cursor written for a given key:

```1179:1287:packages/db/src/collection/state.ts
      for (const key of changedKeys) {
        const cursor = eventCursors.get(key)
        const previousVisibleValue = currentVisibleState.get(key)
        const newVisibleValue = this.get(key)
        // ... one net event per key based on diff
      }
```

**Layer 2 — `CollectionSubscription`**: `filterAndFlipChanges` rewrites unseen `update` into `insert` and drops unseen `delete`. The query engine does not receive the raw sync event stream:

```638:650:packages/db/src/collection/subscription.ts
// (from filterAndFlipChanges)
if (!keyInSentKeys) {
  if (change.type === `update`) {
    newChange = { ...change, type: `insert`, previousValue: undefined }
  } else if (change.type === `delete`) {
    if (!skipDeleteFilter) {
      continue  // dropped entirely
    }
  }
  this.sentKeys.add(change.key)
}
```

**Layer 3 — Effect pipeline**: `accumulateEffectChanges` accumulates all D2 output multiplicities until graph quiescence, then `classifyDelta` produces **one** `enter`/`exit`/`update` per key:

```1258:1284:packages/db/src/query/effect.ts
function accumulateEffectChanges<T>(
  acc: Map<unknown, EffectChanges<T>>,
  [[key, tupleData], multiplicity]: ...
): Map<unknown, EffectChanges<T>> {
  // ...
  if (multiplicity < 0) {
    changes.deletes += Math.abs(multiplicity)
    changes.deleteValue ??= value
  } else if (multiplicity > 0) {
    changes.inserts += multiplicity
    changes.insertValue = value
  }
  // ...
}
```

This means **joins, `orderBy`/`topK`, subquery/includes materialisation, and same-key multiple writes in one sync transaction can all collapse multiple source mutations into one effect event.** If the requirement from issue #1394 is "only process mutations after cursor, preserving each distinct source mutation", this branch does not provide that contract.

This is the fundamental tension Sam identified in the issue: **TanStack DB is a state-delta system, not an event log.** The `startAfter` feature works correctly for state deltas (the visible query result after a given cursor is correct), but cannot guarantee that every source-level event produces a corresponding DeltaEvent. This matters most when the sync layer models **append-only events** as mutable rows.

---

### 3. SIGNIFICANT: `isRedundantSync` Can Swallow Cursored Events

In `commitPendingTransactions` (state.ts), there's a redundancy check:

```1207:1232:packages/db/src/collection/state.ts
        const completedOp = completedOptimisticOps.get(key)
        let isRedundantSync = false

        if (completedOp) {
          if (
            completedOp.type === `delete` &&
            previousVisibleValue !== undefined &&
            newVisibleValue === undefined &&
            deepEquals(completedOp.value, previousVisibleValue)
          ) {
            isRedundantSync = true
          } else if (
            newVisibleValue !== undefined &&
            deepEquals(completedOp.value, newVisibleValue)
          ) {
            isRedundantSync = true
          }
        }
        // ...
        if (isRedundantSync && !shouldEmitVirtualUpdate) {
          continue  // <-- event suppressed, cursor lost
        }
```

When a completed optimistic operation matches the sync confirmation, the event is suppressed. But `eventCursors` has the cursor for that key. Since the event is skipped, the cursor never reaches the subscription, and the effect's cursor gate never sees it.

Scenario: user makes an optimistic insert at key=1, sync confirms with cursor=5, values match → event suppressed → effect never sees cursor=5 → cursor gate never opens if `startAfter=4`.

This interaction between optimistic operations and cursor gating is probably rare in the target use case (server-side durable execution), but it's a landmine.

---

### 4. SIGNIFICANT: Electric Adapter Control-Point Batching Shapes What the Effect Sees

The Electric adapter's `subscribe` handler uses `up-to-date` and `subset-end` control messages as commit boundaries. The first progressive `up-to-date` triggers an atomic truncate + replay of all buffered initial messages. This means the number of source messages that get collapsed into one collection transaction is controlled by Electric's control-point timing, not by the individual `write()` calls.

This is a critical layer that sits between the raw stream and the collection, and it directly affects how many distinct source events are visible to the effect. If Electric batches 50 row changes into one `begin()`/`commit()` cycle triggered by a single `up-to-date`, all 50 writes collapse per the state-diff logic above. The cursor on each would be `stream.lastOffset` (the same batch-level offset for all 50), so `startAfter` granularity is limited to Electric batch boundaries.

```1459:1467:packages/electric-db-collection/src/electric.ts
        write({
          type: isDuplicateInsert ? `update` : operation,
          value: changeMessage.value,
          cursor: stream.lastOffset,
          metadata: {
            ...changeMessage.headers,
          },
        })
```

This is consistent with the PR's "batch is the checkpoint unit" design, but it constrains replay precision to the sync provider's commit boundaries.

---

### 5. DESIGN: Transaction-Scoped Batching Can Leak Replay Into Output

When `handleSourceChangesBeforeCursorGate` splits a batch at the cursor boundary, it calls `scheduleGraphRun` twice — once for replay data, once for live data:

```767:783:packages/db/src/query/effect.ts
    if (firstLiveIndex > 0) {
      const replayChanges = changes.slice(0, firstLiveIndex)
      this.sendChangesToD2(alias, replayChanges)
      this.scheduleGraphRun(alias)
    } else if (!getActiveTransaction() && this.graph?.pendingWork()) {
      // Best-effort boundary preservation outside transaction-scoped flushes.
      this.runGraph()
    }

    this.cursorGateOpenByAlias.set(alias, true)
    this.liveAliasesInBatch.add(alias)

    const liveChanges = changes.slice(firstLiveIndex)
    this.recordPendingBatchCursor(alias, liveChanges)
    this.sendChangesToD2(alias, liveChanges)
    this.scheduleGraphRun(alias)
```

Without an active transaction, `scheduleGraphRun` runs the graph immediately, so the replay graph run completes (and output is discarded) before the live changes are sent. This is correct.

But **if there is an active TanStack DB transaction**, both `scheduleGraphRun` calls are deferred to the same scheduler job. When the scheduler finally runs, **both replay and live data are queued in D2's inputs** and processed in **one** `graph.run()`. The D2 graph merges all queued multisets at the join operator level (JoinOperator drains its entire input queue per `run()`). The accumulated output includes contributions from both replay and live data. Since `liveAliasesInBatch` has the alias, `flushPendingChanges` emits **everything**, including output caused by replay changes.

The code acknowledges this with the comment "Best-effort boundary preservation outside transaction-scoped flushes." In practice, sync callbacks don't run inside TanStack DB transactions, so the immediate (correct) path is taken. But this is fragile and undocumented.

---

### 6. DESIGN: Uncursored Batches Are Treated as Fully Live Once Gating Is Active

`findFirstChangeAfterCursor()` returns `0` when a batch contains no cursors at all:

```1352:1358:packages/db/src/query/effect.ts
  if (!anyCursor) {
    return 0
  }

  return -1
```

This opens the gate and emits the entire batch. It matches the test "should eventually open cursor gate when cursor-less changes follow cursored replay", but it also means **`startAfter` is only safe if the adapter puts a monotonic cursor on every relevant `sync.write()` call**. Any missed cursor in an adapter or edge path becomes a duplicate-side-effect hole instead of a hard failure. This isn't validated at the `startAfter` contract level — there's no warning or error when the sync provider stops sending cursors partway through replay.

---

### 7. DESIGN: `triggeringSource` / `cursors` Are Not Reliable Causal Metadata

Two reasons:

**First**, `triggeringSource` is just the first alias inserted into `liveAliasesInBatch`:

```983:985:packages/db/src/query/effect.ts
      if (this.liveAliasesInBatch.size > 0) {
        triggeringSource = this.liveAliasesInBatch.values().next().value
      }
```

For join queries where multiple sources contribute live changes in the same batch, only the first-inserted alias is reported.

**Second**, in-graph join tap / lazy-load changes skip both `recordPendingBatchCursor()` and `liveAliasesInBatch` (because they hit the `isGraphRunning` guard at line 719). So result changes can be influenced by secondary-source loads without that source being represented in `cursors` or `triggeringSource` at all:

```718:721:packages/db/src/query/effect.ts
    if (this.isGraphRunning) {
      this.sendChangesToD2(alias, changes)
      return
    }
```

Checkpoint routing logic should not assume these fields are authoritative for multi-source batches.

---

### 8. DESIGN: "Exit Without Enter" Is Possible After Replay

If a replayed change causes a row to enter the query result, that `enter` event is discarded (cursor gate closed). If a subsequent live change causes the same row to exit the query result, the `exit` event is emitted. The consumer sees an `exit` for a row they never saw `enter` for.

This is semantically correct for `startAfter` (the consumer already processed the enter before checkpointing), but it's a sharp edge. The `previousValue` on an `update` event could reference state that the consumer never observed through this effect instance. This should be documented.

---

### 9. CODE: No Cross-Batch Monotonicity Enforcement

`recordPendingBatchCursor` enforces monotonicity **within** a single batch of changes. But across batches, cursors can go backwards without detection — `pendingBatchCursorByAlias` is cleared in `flushPendingChanges` between batches, so the comparison against `current` restarts from scratch each time.

This may be acceptable (the user's sync provider is responsible for ordering), but it contradicts the "monotonically ordered" invariant implied by the within-batch enforcement.

---

### 10. CONTEXT: `markReady()` Does Not Create Row Deltas

`markReady()` itself does **not** mutate the transaction payload seen by the query engine. It only transitions the lifecycle status and emits an empty notification:

```134:157:packages/db/src/collection/lifecycle.ts
  public markReady(): void {
    // ...
    if (this.status === `loading`) {
      this.setStatus(`ready`, true)
      // ...
      if (this.changes.changeSubscriptions.size > 0) {
        this.changes.emitEmptyReadyEvent()
      }
    }
  }
```

However, what **does** shape the transaction boundary is the sync adapter's control-point handling. In the Electric adapter, `subset-end` and `up-to-date` are the commit triggers, and the first progressive `up-to-date` does an atomic truncate + replay of all buffered initial messages. So control messages absolutely can change how many source messages get collapsed into one collection transaction before the effect layer sees them. This is the mechanism by which `markReady` timing **indirectly** affects cursor gating — not by creating events, but by determining when batches are committed and therefore how much collapsing occurs.

---

### 11. Alignment with Sam's `$version` Proposal

Sam's comment on the issue proposes a `$version` virtual prop approach instead. The key differences:

| Aspect | `startAfter` (this PR) | `$version` (Sam's proposal) |
|--------|----------------------|---------------------------|
| Where gating happens | Effect pipeline (before output) | User code (in handler or WHERE) |
| Scope | Per-source alias | Per-row |
| Works with event collapsing | Partially (state-delta semantics) | Yes (user controls filtering) |
| Works with mutable rows | Fragile (intermediate states lost) | Explicit (user decides) |
| Complexity | Internal to effect engine | External, user-controlled |

Sam's proposal explicitly acknowledges: "If user data collapses actions into mutable state, intermediate actions are not reconstructable from `$version` alone." The `startAfter` PR doesn't surface this limitation as clearly.

The `$version` approach is more honest about what TanStack DB can and cannot guarantee. The `startAfter` approach provides better DX for the common case (append-only events with monotonic cursors) but creates an illusion of exact replay guarantees that the pipeline can't always deliver.

---

### 12. Summary of Recommended Actions

**Must fix:**
1. Buffer drain path must respect cursor gating (Issue #1)

**Should fix:**
2. Document that `startAfter` provides state-delta replay, not event-log replay — the semantics are "hydrate historical state, then emit future net query-result transitions"
3. Document the "exit without enter" behaviour
4. Make `triggeringSource` an array, or document that it's the first contributing alias and that lazy-loaded join sources are invisible to it
5. Consider whether uncursored batches opening the gate should emit a warning, since it silently breaks the monotonic-cursor contract

**Should consider:**
6. The `isRedundantSync` interaction with cursor gating (Issue #3)
7. Whether `startAfter` is the right abstraction vs. Sam's `$version` proposal — particularly given the three independent collapsing layers that make exact replay impossible for mutable rows
8. Adding cross-batch monotonicity enforcement or documenting it as a caller responsibility

**Tests to add:**
9. Buffered startup batch straddles `startAfter` before all aliases are subscribed — this is the gap most likely to hide a real bug

samwillis:
After the review above, I asked Opus:
> I think it's also important to consider if there should be a collection level config to tell it to "commit" after every change in a "replay mode" and how this could make things work?

---

## What a per-change commit mode would look like

The idea: add a config flag on the collection (or on the sync config) that makes every `write()` auto-commit, so the sync adapter doesn't need to batch. Something like:

```typescript
createCollection({
  sync: {
    replayCommitMode: 'per-change', // new option
      // adapter writes normally — collection auto-wraps each write
      // in its own begin/commit cycle
    }
  }
})
```

Let me trace what this would fix and what it wouldn't.

---

## What it fixes

**Layer 1 collapsing (same-key state diffing) is eliminated.** Today, `commitPendingTransactions` applies all operations in the transaction to `syncedData`, then diffs previous vs. final visible state per key. If key=1 gets insert(cursor=1) then update(cursor=2) in one transaction, the collection emits one event with the final value.

With per-change commits, each write gets its own `commitPendingTransactions` call. Insert(key=1, cursor=1) produces one insert event. Update(key=1, cursor=2) produces one update event. Both carry their own cursor. The effect sees both.

**Cursor precision improves.** Each event carries its own cursor, so the consumer can checkpoint at per-event granularity. No more straddling batches — each change is either before or after the `startAfter` boundary, never split across a boundary within a single commit.

**The buffer drain bug (Issue #1) becomes less severe.** Since each change is its own commit, even if changes arrive during the buffer phase, each produces a single event with a clear cursor position. The gate logic still needs fixing, but the "straddling" scenario becomes impossible.

---

## What it does NOT fix

Even with per-change commits, two other collapsing layers remain:

**Layer 2 — `CollectionSubscription.filterAndFlipChanges` still rewrites events.** If the subscription hasn't seen key=1 yet (not in `sentKeys`), an `update` to key=1 gets rewritten to `insert`. A `delete` for an unseen key gets dropped entirely. This happens regardless of commit granularity. So the effect's D2 input still doesn't match the raw sync stream 1:1.

**Layer 3 — The D2 graph + effect accumulation.** Even with per-change commits, the `transactionScopedScheduler` can coalesce multiple synchronous commits into one graph run. If the sync adapter fires:

```typescript
begin(); write({ key: 1, cursor: 1 }); commit();
begin(); write({ key: 1, cursor: 2 }); commit();
```

Both `commit()` calls trigger subscription callbacks synchronously. Each calls `handleSourceChanges` → `scheduleGraphRun`. Without an active TanStack DB transaction, each `scheduleGraphRun` runs the graph immediately and separately — so this case actually works and produces two separate DeltaEvent batches.

But if something wraps these in a parent context (or if the scheduler defers for any reason), the two graph runs merge and `accumulateEffectChanges` produces one net delta for key=1.

More importantly, **queries with JOINs, WHERE, or aggregation still transform the relationship between source events and output events.** A source insert might not produce any output (WHERE filters it). A source update to the join key might cause both an exit and an enter for different output keys. Per-change commits at the collection level don't affect this — it's a property of the query pipeline.

---

## The performance problem

Per-change commits during initial sync would be devastating. Consider an Electric shape with 10,000 rows in the initial snapshot. Today, this is one `begin()` → 10,000 `write()` calls → one `commit()` → one visible-state diff → one subscription callback. With per-change commits, it would be 10,000 individual `commitPendingTransactions` calls, each doing a state diff + event emission + subscription callback. Plus 10,000 D2 graph runs.

You could mitigate this by only enabling per-change mode after the initial sync, or only for changes past the `startAfter` boundary. But this adds significant complexity:

```typescript
// Hypothetical: only per-change commit for post-cursor changes
sync: {
  replayCommitMode: {
    mode: 'per-change',
    afterCursor: savedCursor, // only applies past this point
  }
}
```

This would require the collection to know about cursor semantics, which currently lives only in the effect layer. It would leak effect-level concerns into the collection.

---

## The atomicity problem

Some sync providers intentionally batch changes that should be atomic. Electric's `up-to-date` control message means "the preceding changes form a consistent snapshot." Breaking this into per-change commits means consumers see intermediate states that were never independently consistent on the server. For state-delta effects (the current `createEffect` semantics), this is usually fine. But it breaks the implicit contract that each event represents a consistent state transition.

---

## What would actually make things work end-to-end

Given the three collapsing layers, no single config at the collection level fully solves the problem. Here's what a complete solution would require:

**For same-key collapsing (Layer 1):** Per-change commits, or modifying `commitPendingTransactions` to emit events per operation instead of per state diff.

**For subscription rewriting (Layer 2):** The effect would need to subscribe with `includeInitialState: false` and `skipFiltering: true` (via `markAllStateAsSeen`), which it already does for `mockSyncCollectionOptionsNoInitialState` in tests. But for collections with existing state, this means the effect would miss the initial snapshot — which is actually fine for `startAfter` since the whole point is "I already have state up to this cursor."

**For D2 accumulation (Layer 3):** The effect would need to run the graph and flush after every single source change, not accumulate across a scheduler run. This means `scheduleGraphRun` must run immediately (no transaction batching), and `runGraph` must be called per change, not per scheduler cycle.

Combining all three, you'd need something like:

```typescript
createCollection({
  sync: {
    commitMode: 'per-change', // Layer 1: no same-key collapsing
    sync: ...
  }
})

createEffect({
  query: (q) => q.from({ events: collection }),
  startAfter: savedCursor,
  // implicitly: subscribe with skipFiltering, flush per change
  onEnter: (event) => { ... }
})
```

This is achievable but represents a fundamentally different operating mode for the effect pipeline — closer to an event-stream processor than a reactive query result observer.

---

## The deeper question: is this the right layer?

I think the tension here is that `startAfter` is trying to add event-log semantics to a system designed around state deltas. There are two honest paths forward:

**Path A — Accept state-delta semantics.** `startAfter` means "hydrate historical state, then emit future net query-result transitions." Document that intermediate events may be collapsed. This is what the current PR does, and it's correct for its stated semantics. Sam's `$version` proposal fits here — it gives the user a tool to implement their own filtering without the system pretending it can replay individual events.

**Path B — Add a true event-replay mode.** Per-change commits + skip subscription filtering + per-change graph flushes. This is a real operating mode, but it should probably be a separate API (maybe `createEventProcessor` or a config like `replayMode: 'event-log'`) rather than bolted onto the existing `createEffect`. It would only work for simple queries (no joins, no aggregation), because the D2 pipeline inherently transforms the event stream.

A per-change commit config at the collection level is a necessary building block for Path B, but it's not sufficient on its own. And it comes with real costs (performance, atomicity) that need to be weighed against the alternative of just using Sam's `$version` approach where the consumer handles idempotency explicitly.

My recommendation: if the team wants to pursue Path B, design it as a distinct mode with clear constraints (single-source, no aggregation, per-change semantics) rather than trying to make the general-purpose `startAfter` on `createEffect` guarantee something the architecture can't deliver in the general case.

PR #1431, opened 2026-03-30
feat: $pendingOperation virtual prop

marbemac:
## 🎯 Changes

Adds a new `$pendingOperation` virtual property to collection rows (`'insert' | 'update' | 'delete' | null`). This tells you what type of optimistic mutation is pending for each row, which is useful for building draft/review UIs where you want to show git-style change indicators, amongst other things.

Note, large-ish diff but most of it is new tests 😅.

The key feature: items deleted in a pending transaction can now stay visible in query results. By default nothing changes — deleted items still vanish from queries. But if your query references `$pendingOperation` in a `.where()` clause, the implicit filter is disabled and you can see pending-delete items inline:

```ts
// Default — deleted items hidden (same as before)
q.from({ task: tasks }).where(({ task }) => eq(task.projectId, projectId))

// Opt in — deleted items visible with $pendingOperation: 'delete'
q.from({ task: tasks }).where(({ task }) =>
  and(
    eq(task.projectId, projectId),
    or(isNull(task.$pendingOperation), not(isNull(task.$pendingOperation))),
  ),
)
```

Works with live queries, `createEffect`, joins/subqueries, GROUP BY, ordered/paginated queries, and selective where clauses like `not(isNull($pendingOperation))` for "show only pending changes" views.

This was discussed with @samwillis beforehand — the approach is to keep deletes as deletes at the collection layer and convert them to updates at the subscription layer when opted in.

I also went ahead and built and linked into our actual app to test out the functionality. It's damn neat! For example, can show indicators inline for as yet uncommitted transaction. Imagine CRM app, and app with data grids w rows that can be updated/removed/etc, or really any productivity app that wants to allow humans and agents to safely make changes, and represent those changes inline in the application without having to change the data layer at all (note `R` and `A` badges on the files on the left, and the red bordered "install" that has been deleted but not yet "committed").

Screenshot 2026-03-29 at 7 45 41 PM

### Notable behavioral change

`isRowSynced()` now also checks `pendingOptimisticUpserts` and `pendingOptimisticDeletes`, so `$synced` is `false` during the completed-but-awaiting-sync window. Previously it could briefly show `$synced: true` while a pending operation was still awaiting sync confirmation.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Added a `$pendingOperation` virtual property to every row, exposing pending optimistic operations (`'insert' | 'update' | 'delete' | null`).
  * When queries filter by `$pendingOperation`, deleted-but-pending rows remain visible in live queries, ordered/paginated results, joins/subqueries, and `GROUP BY` output.
  * Subscriptions and effects now automatically include pending deletes when `$pendingOperation` is referenced.
* **Tests**
  * Expanded coverage for `$pendingOperation` across transactions, rollbacks/commits, event behavior, live queries, and `createEffect`.

### DIFF
--- a/.changeset/pending-operation-virtual-prop.md
+++ b/.changeset/pending-operation-virtual-prop.md
@@ -1,5 +1,5 @@
 ---
-"@tanstack/db": minor
+'@tanstack/db': minor
 ---
 
 Add `$pendingOperation` virtual property to track optimistic mutation type
--- a/.changeset/pending-operation-virtual-prop.md
+++ b/.changeset/pending-operation-virtual-prop.md
@@ -1,0 +1,9 @@
+---
+"@tanstack/db": minor
+---
+
+Add `$pendingOperation` virtual property to track optimistic mutation type
+
+- New virtual property `$pendingOperation` on every collection row: `'insert' | 'update' | 'delete' | null`
+- Items deleted in pending transactions can stay visible in query results when `$pendingOperation` is referenced in a `.where()` clause
+- Works with live queries, `createEffect`, joins/subqueries, GROUP BY, and ordered/paginated queries
--- a/packages/db-sqlite-persistence-core/tests/persisted.test.ts
+++ b/packages/db-sqlite-persistence-core/tests/persisted.test.ts
@@ -247,6 +247,7 @@ const stripVirtualProps = <T extends Record<string, any> | undefined>(
     $origin: _origin,
     $key: _key,
     $collectionId: _collectionId,
+    $pendingOperation: _pendingOperation,
     ...rest
   } = value as Record<string, unknown>
   return rest as T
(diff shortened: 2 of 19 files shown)


marbemac:
One thing I considered but left out because wasn't sure if ya'll would be ok w a new API method. A `.includePendingDeletes()` method on the query builder. Since the tautology pattern is a little awkward for the "include everything including deleted items" use case, instead of:

```ts
.where(({ task }) =>
  and(
    eq(task.projectId, projectId),
    or(isNull(task.$pendingOperation), not(isNull(task.$pendingOperation))),
  ),
)
```

User could do:

```ts
.where(({ task }) => eq(task.projectId, projectId))
.includePendingDeletes()
```

OR alternative, just a simple exported util:

```ts
.where(({ task }) => includeDeletedItems(eq(task.projectId, projectId)))
```

If that's something you guys would consider, I can try and add it - don't think it'd be much code.

PR #1445, opened 2026-04-02
feat(db): allow externally-provided previousValue in sync update chan…

SamJB123:
## 🎯 Changes

This change allows consumers the option of providing direct previousValue inputs, improving integration with other sync engines such as Yjs v14, without affecting standard behavior.

When collection values are live-reading proxy objects (e.g. Yjs Y.Type proxies via getAttr()), deepEquals(previousVisibleValue, newVisibleValue) always returns true — both sides read through the already-mutated proxy and see identical current values. First updates on proxy-backed rows are silently dropped.

ChangeMessage already has an optional previousValue field, and sync sources like Yjs deltas already know the pre-mutation state (SetAttrOp.prevValue). This change lets commitPendingTransactions use a sync-provided previousValue for the comparison instead of the captured currentVisibleState, which is unreadable when the value is a live proxy.

When previousValue is not provided, behavior is unchanged.

## ✅ Checklist

- [✅] I have tested this code locally with `pnpm test`.
- [✅] I have tested this code in runtime implementations using yjs v14 successfully.

## 🚀 Release Impact

- [✅] This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).


KyleAMathews:
Closing as superseded by merged PR #1835, which fixes immutable previous-value publication with broader batch, overlay, replacement, and rollback coverage while preserving the original contributor credit.

PR #1448, opened 2026-04-03
fix: deduplicate and filter null join keys in lazy join subset queries

KyleAMathews:
## Summary

Fix bloated `ANY()` SQL params in lazy join subset queries. When multiple rows reference the same foreign key or have null foreign keys, the join key array now gets deduplicated and null-filtered before being sent to Electric.

## Root Cause

In `joins.ts`, the lazy join path collects foreign keys from all rows via `.map(([[joinKey]]) => joinKey)` and passes them directly to `inArray()`. With real data where multiple rows share a foreign key (e.g., 5 tasks in the same project) or have nullable foreign keys, the resulting `ANY()` param contains repeated IDs and NULLs:

```sql
WHERE "id" = ANY('{NULL,NULL,"proj-1","proj-1","proj-1","proj-2","proj-2"}')
```

## Approach

Three-line fix at the collection point (`joins.ts:305-317`):
1. `.filter((key) => key != null)` — remove null/undefined keys (loose equality catches both)
2. `new Set(...)` — deduplicate remaining keys
3. Early return when no valid keys remain — skip the snapshot request entirely

## Key Invariants

- Join semantics are unchanged: left joins still produce `undefined` for unmatched rows
- The early return on empty keys prevents pointless `inArray(ref, [])` calls to the sync layer
- `Set` equality works correctly for the primitive types used as join keys (strings, numbers)

## Non-goals

- The pre-existing silent deoptimization fallback (lines 325-328) where a failed index lookup loads the entire collection was not addressed — that's a separate concern

## Verification

```bash
pnpm -C packages/db exec vitest run -t "lazy join key deduplication"  # 2 new tests
pnpm -C packages/db exec vitest run                                   # full suite: 2256 passed
pnpm -C packages/db exec tsc --noEmit                                 # clean
```

## Files Changed

| File | Change |
|------|--------|
| `packages/db/src/query/compiler/joins.ts` | Dedup + null-filter join keys, early return on empty |
| `packages/db/tests/query/live-query-collection.test.ts` | 2 tests: mixed keys (dedup+nulls) and all-null keys (early return) |
| `.changeset/fix-lazy-join-key-dedup.md` | Changeset for patch release |

---

### DIFF
--- a/packages/db/tests/query/live-query-collection.test.ts
+++ b/packages/db/tests/query/live-query-collection.test.ts
@@ -2779,21 +2779,9 @@ describe(`createLiveQueryCollection`, () => {
       await liveQuery.preload()
       await flushPromises()
 
-      // No loadSubset call should contain an inArray expression
+      // No loadSubset call should have been made for the lazy join
       // since all keys were null and filtered out
-      const findInArrayExpr = (expr: LoadSubsetOptions[`where`]): boolean => {
-        if (!(expr instanceof Func)) return false
-        if (expr.name === `in`) return true
-        if (expr.name === `and` || expr.name === `or`) {
-          return expr.args.some((arg) => findInArrayExpr(arg))
-        }
-        return false
-      }
-
-      const hasInArrayCall = capturedOptions.some((opt) =>
-        findInArrayExpr(opt.where),
-      )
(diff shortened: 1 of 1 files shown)


samwillis (approved):
:shipit:

PR #1453, opened 2026-04-04
fix(db-sqlite-persistence-core): add schema-aware overloads to persistedCollectionOptions

### LINKED ISSUE #1452: persistedCollectionOptions: schema type parameter not inferred, result incompatible with createCollection

**Describe the bug**

When passing a `schema` to `persistedCollectionOptions`, TypeScript fails to infer the `TSchema` type parameter. This causes two problems:

1. **Schema type not inferred**: `options.schema` resolves to `undefined` instead of the actual schema type (e.g. a Zod schema). `TSchema` defaults to `never`.
2. **Incompatible with `createCollection`**: The result cannot be passed to `createCollection` because:
   - `createCollection`'s schema overloads require `{ schema: T }` (required)
   - `createCollection`'s no-schema overloads require `{ schema?: never }`
   - The result has `schema?: TSchema | undefined` (optional), which matches neither

**Root cause**

`persistedCollectionOptions` lacks schema-aware overloads. Compare with `localOnlyCollectionOptions` (which works correctly) — it has separate overloads for the schema and no-schema cases with `& { schema: T }` / `& { schema?: never }` on both input and output types.

`persistedCollectionOptions` only has two overloads (sync-present vs sync-absent), neither of which distinguishes the schema case.

**To Reproduce**

```ts
import { z } from 'zod'
import { createCollection } from '@tanstack/db'
import { persistedCollectionOptions } from '@tanstack/db-sqlite-persistence-core'

const todoSchema = z.object({
  id: z.string(),
  title: z.string(),
})

const adapter = {
  loadSubset: () => Promise.resolve([]),
  applyCommittedTx: () => Promise.resolve(),
  ensureIndex: () => Promise.resolve(),
}

// TSchema defaults to `never` — schema type is lost
const options = persistedCollectionOptions({
  id: 'test',
  schema: todoSchema,
  schemaVersion: 1,
  getKey: (item) => item.id,
  persistence: { adapter },
})

// options.schema is `undefined` instead of `typeof todoSchema`

// This errors: "No overload matches this call"
const collection = createCollection(options)
```

**Expected behavior**

`persistedCollectionOptions` should infer the schema type and produce a result compatible with `createCollection`, matching the behavior of `localOnlyCollectionOptions`.

**Additional context**

The fix requires adding schema-aware overloads to `persistedCollectionOptions` (for both sync-present and sync-absent modes), following the same pattern used by `localOnlyCollectionOptions` in `packages/db/src/local-only.ts`.

merodiro:
Closes #1452

## Summary

- persistedCollectionOptions was missing schema-aware overloads, so TSchema defaulted to never when a schema was passed, making the result incompatible with createCollection
- Added four schema/no-schema overloads matching the pattern used by localOnlyCollectionOptions and localStorageCollectionOptions
- Added type tests covering schema inference and createCollection compatibility for both sync-present and sync-absent modes

## Note on error messages

Adding schema-aware overloads changes the order TypeScript evaluates overloads when a call is invalid. Previously, calling persistedCollectionOptions without persistence in sync-present mode would report the error on the function call itself. Now TypeScript reports it on the sync property inside the object literal, since the last overload it tries is the sync-absent one (which Omits sync). The type safety is unchanged - the call still fails to compile without persistence - only the error location and message differ. This is the same behavior seen in createCollection, which has 8 overloads.

## Test plan

- [x] npx tsc --noEmit passes with zero errors
- [x] npx vitest run --typecheck passes (8 type tests in persisted.test-d.ts)
- [x] Existing tests unchanged and passing
- [x] Verified no breakage in other packages using persistedCollectionOptions with 2 or 4 explicit type args

### DIFF
--- a/.changeset/calm-toes-approve.md
+++ b/.changeset/calm-toes-approve.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/db-sqlite-persistence-core': patch
+---
+
+Add schema-aware overloads to `persistedCollectionOptions` so schema-based calls infer the correct types and remain compatible with `createCollection`.
--- a/packages/db-sqlite-persistence-core/src/persisted.ts
+++ b/packages/db-sqlite-persistence-core/src/persisted.ts
@@ -14,6 +14,7 @@ import type {
   CollectionConfig,
   CollectionIndexMetadata,
   DeleteMutationFnParams,
+  InferSchemaOutput,
   InsertMutationFnParams,
   LoadSubsetOptions,
   PendingMutation,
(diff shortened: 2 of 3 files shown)


samwillis (approved):
Perfect, thanks @merodiro 🚀

KyleAMathews:
Closing as superseded by merged PR #1866, which fixes the same #1452 schema-inference and createCollection composition bug for both sync-wrapped and local-only persisted collections and adds focused type-oracle coverage.

PR #1455, opened 2026-04-04
fix: duplicate alias in sibling includes silently breaks nested children

### LINKED ISSUE #1454: Duplicate alias in sibling includes silently breaks nested children

**Describe the bug**
When two sibling includes in `.select()` use the same alias letter, nested child collections silently produce empty results. No error or warning is produced — data is silently lost.

For example, if an `issues` include uses `{ i: issues }` and a sibling `tags` include uses `{ i: tags }`, any nested grandchild inside `issues` (e.g., `comments`) returns empty. Changing the tags alias to `{ t: tags }` fixes it.

Reproducible with [this test](https://github.com/ibirrer/db/blob/fix/includes-duplicate-alias/packages/db/tests/query/includes.test.ts#L4088-L4137) in PR #1455.

**Steps to Reproduce**

```typescript
    id: p.id,
    issues: q
      .from({ i: issues })          // alias "i"
      .where(({ i }) => eq(i.projectId, p.id))
      .select(({ i }) => ({
        id: i.id,
        comments: q                  // nested grandchild
          .from({ c: comments })
          .where(({ c }) => eq(c.issueId, i.id))
          .select(({ c }) => ({ id: c.id, body: c.body })),
      })),
    tags: q
      .from({ i: tags })            // same alias "i" — breaks comments above
      .where(({ i }) => eq(i.projectId, p.id))
      .select(({ i }) => ({ id: i.id, label: i.label })),
  })),
)

await collection.preload()

const alpha = collection.get(1)
const issue10 = alpha.issues.get(10)
issue10.comments // empty — should contain comments for issue 10
```

**Observed Behavior**

- `tags` is populated correctly
- `issues` contains rows, but they come from the wrong collection (tags data routed to the issues pipeline)
- Nested `comments` inside issues is always empty
- No error or warning is thrown

**Expected Behavior**

Aliases in sibling includes should be independent scopes. Using the same alias letter in different sibling subqueries should just work — each gets its own data pipeline.

ibirrer:
## Summary

When two sibling includes in `.select()` use the same alias (e.g., `{ i: issues }` and `{ i: tags }`), nested child collections silently produce empty results. The root cause is that all includes aliases are flattened into a single namespace — sharing one D2 graph input and one subscription, so the second sibling's collection data overwrites the first.

## Fix

Give each includes subquery its own independent D2 graph input:

- `extractCollectionAliases` no longer traverses into `IncludesSubquery` nodes, keeping `collectionByAlias` scoped to the top-level query
- `compileQuery` accepts a `createInput` factory; when processing includes, each child gets fresh inputs for its source aliases
- `compileBasePipeline` merges the new inputs into `inputsCache` and `compiledAliasToCollectionId` under unique keys (`__inc_N_alias`), so each gets its own subscription feeding the correct collection

## Test

Regression test added to `includes.test.ts`: sibling includes with duplicate alias `"i"` — nested comments are populated. Fails without fix, passes with fix.

Fixes #1454

### DIFF
--- a/packages/db/src/query/compiler/index.ts
+++ b/packages/db/src/query/compiler/index.ts
@@ -141,6 +141,9 @@ export function compileQuery(
   // For includes: parent key stream to inner-join with this query's FROM
   parentKeyStream?: KeyedStream,
   childCorrelationField?: PropRef,
+  // Factory to create a fresh D2 input for an includes subquery alias.
+  // Each sibling gets its own input to avoid alias collisions.
+  createInput?: (alias: string, collectionId: string) => KeyedStream,
 ): CompilationResult {
   // Check if the original raw query has already been compiled
   const cachedResult = cache.get(rawQuery)
@@ -391,10 +394,16 @@ export function compileQuery(
             }
           : subquery.query
 
+      // Give each includes child its own D2 inputs so that sibling
+      // subqueries using the same alias letter get independent streams.
+      const childInputs = createInput
+        ? createInputsForSources(childQuery, allInputs, createInput)
+        : allInputs
+
       // Recursively compile child query WITH the parent key stream
       const childResult = compileQuery(
         childQuery,
-        allInputs,
+        childInputs,
         collections,
         subscriptions,
         callbacks,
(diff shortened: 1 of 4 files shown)


KyleAMathews:
Thanks for isolating the duplicate sibling-alias failure. https://github.com/TanStack/db/pull/1740 (https://github.com/TanStack/db/commit/5d9335d0d42c1cc1ec2b92be8ce40ae8abe42827) replaced alias-keyed include materialization with opaque source identities; the current alpha-renaming oracle and nested Collection-valued reproduction pass. Closing this conflicting draft as superseded. The original fixture remains valuable regression coverage.

PR #1465, opened 2026-04-08
fix(db): clean up optimistic state when server returns a different key

### LINKED ISSUE #1442: Optimistic insert not removed when server returns a different key
### Description

When an `onInsert` handler syncs server data back to the collection (via `writeInsert`, `refetch`, or any sync path), the optimistic insert under the original client key is never removed if the server returns a different key. This causes:

1. **Duplication**: Both the client-key item (`$synced: false`, `$origin: local`) and the server-key item (`$synced: true`, `$origin: remote`) appear in the collection. The client-key item persists forever.
2. **Stale `$synced` (same-key variant)**: When client and server use the same key, `$synced` stays `false` permanently and the optimistic data shadows the server data.

Both violate the 0.6 contract: _"when your mutation handler promise resolves, the optimistic state is removed."_

### Reproduction

**Different key — via writeInsert:**

```ts
  queryCollectionOptions({
    id: 'assignments',
    getKey: (item) => item.id,
    queryKey: ['assignments'],
    queryFn: async () => api.getAssignments(),
    queryClient,

      const newItems = transaction.mutations.map((m) => m.modified)
      // Client uses id: -Date.now(), server returns sequential DB id
      const serverItems = await api.createAssignments(newItems)
      collection.utils.writeInsert(serverItems)
      return { refetch: false }
    },
  }),
)

collection.insert({ id: -Date.now(), resource_id: 42, name: 'New' })
// Result: two items visible — one under client key, one under server key
```

**Different key — via refetch:**

```ts
  const newItems = transaction.mutations.map((m) => m.modified)
  const serverItems = await api.createAssignments(newItems)
  // No writeInsert — just refetch. Server returns the item under a different key.
  await collection.utils.refetch()
},
```

Same result: the optimistic item under the client key persists alongside the refetched server item.

**Same key — via writeInsert:**

```ts
  const newItems = transaction.mutations.map((m) => m.modified)
  // Server returns same key but adds server-computed fields
  const serverItems = await api.createAssignments(newItems)
  collection.utils.writeInsert(serverItems)
  return { refetch: false }
},
```

Result: single item, but `$synced` stays `false` and the optimistic data (without server-computed fields) shadows the server data from `writeInsert`.

### Failing Tests

Three reproduction tests in `packages/query-db-collection/tests/query.test.ts`:

- `should not duplicate items when writeInsert uses a different key than the optimistic insert`
- `should mark item as synced when writeInsert uses the same key as the optimistic insert`
- `should not duplicate items when refetch returns a different key than the optimistic insert`

All fail on current `main` with the `@tanstack/db` dist rebuilt.

### Root Cause

Introduced in `9952921e` ("Virtual props implementation #1213"). The `pendingOptimisticDirectUpserts` set was added to keep optimistic state visible between transaction completion and sync confirmation (for correct `$synced` tracking).

The sequence:

1. `collection.insert()` → creates direct transaction, mutation key = `clientKey`
2. `onInsert` runs (tx state = `persisting`)
3. Sync data is committed (via `writeInsert` or `refetch`) — `commitPendingTransactions` clears `pendingOptimisticDirectUpserts` for the **sync key** (which may differ from `clientKey`)
4. Handler returns → `commit()` sets state to `completed` → `touchCollection()` → `recomputeOptimisticState(false)`
5. **In `recomputeOptimisticState`**: the completed-transaction loop **unconditionally re-adds** `clientKey` to `pendingOptimisticDirectUpserts` (line 503)
6. `isPersisted.resolve()` → microtask: `scheduleTransactionCleanup` removes the transaction from the map
7. Any subsequent `recomputeOptimisticState` (from observer refetch, etc.) finds `clientKey` in `pendingOptimisticDirectUpserts` but no transaction to process → key persists forever, resurrected into `optimisticUpserts` via the seeding step (line 555)

### Proposed Fix: Track which direct transactions had sync writes

Add a `Set<string>` (`directTransactionsWithSyncWrites`) to `CollectionStateManager`. When `commitPendingTransactions` processes an `immediate` sync transaction (which only comes from `writeInsert`/`writeUpdate`/`writeDelete` called inside a handler), record the ID of any persisting direct transaction. Then in `recomputeOptimisticState`, skip re-adding to `pendingOptimisticDirectUpserts` for transactions in this set.

This only changes behavior when a sync write was committed during the handler — the non-`writeInsert` flows are unaffected.

**Trade-off**: This approach uses an indirect signal — it infers "writeInsert was called" by observing that an `immediate` sync transaction was processed while a direct transaction was persisting. If a future change introduces another source of `immediate` sync transactions, it could incorrectly suppress the pending optimistic state. It also adds a `Set` that must stay in sync across three locations (`commitPendingTransactions`, `recomputeOptimisticState`, `scheduleTransactionCleanup`).

A potentially cleaner alternative would be a first-class flag on the transaction itself (e.g., `transaction.hasSyncWrites = true`) set explicitly by `writeInsert`/`writeUpdate`/`writeDelete`. That's a direct signal rather than an inferred one, but requires touching `transactions.ts` as well.

| Flow                             | Result    |
| -------------------------------- | --------- |
| writeInsert + different key      | **fixed** |
| writeInsert + same key           | **fixed** |
| refetch + different key          | **fixed** |
| No writeInsert + default refetch | **safe**  |
| No writeInsert + refetch: false  | **safe**  |

### Environment

- `@tanstack/db@0.6.1`
- `@tanstack/query-db-collection@1.0.32`
- `@tanstack/react-db@0.1.79`

goatrenterguy:
## Summary

Fixes #1442. When an `onInsert`/`onUpdate`/`onDelete` handler syncs server data back to the collection (via `writeInsert`, `writeUpdate`, `writeUpsert`, `writeDelete`, or `refetch`), the optimistic state under the original client key is now correctly removed if the server returns a different key.

## Root Cause

Introduced in `9952921e` (Virtual props implementation #1213). The `pendingOptimisticDirectUpserts` set was added to keep optimistic state visible between transaction completion and sync confirmation for correct `$synced` tracking. However, in `recomputeOptimisticState()`, completed direct transactions **unconditionally re-add** their mutation keys to this set on every call — even after `commitPendingTransactions` already cleared the key. The set is additive and never fully cleared, so entries persist forever when the server confirms under a different key.

## Approach

Two coordinated mechanisms in `packages/db/src/collection/state.ts`:

**Mechanism A — Prevent re-adding keys for transactions with sync writes:**
- `directTransactionsWithSyncWrites: Set<string>` tracks transaction IDs whose handlers committed immediate sync writes (via `writeInsert`/`writeUpdate`/`writeDelete`)
- `processedCompletedTransactions: Set<string>` ensures keys are added to `pendingOptimisticDirectUpserts` at most once per transaction (add-once guard)
- Together, these prevent `recomputeOptimisticState` from re-adding keys after sync already confirmed the data

**Mechanism B — Orphan cleanup for refetch-with-different-key:**
- After `commitPendingTransactions` processes sync, keys in `pendingOptimisticDirectUpserts` that belong to completed direct transactions but aren't in `syncedData` are removed
- `scheduleTransactionCleanup` processes queued syncs **before** deleting the transaction, so the orphan cleanup can find the completed transaction to identify its keys

The `else` branch in the completed-transaction loop is split into three cases (direct+first-time, non-direct, direct+already-processed) to prevent concurrent direct transactions from clobbering each other's entries.

| Flow | Result |
| --- | --- |
| writeInsert + different key | **fixed** — Mechanism A prevents re-adding |
| writeInsert + same key | **fixed** — add-once guard prevents re-adding after line 990 clears it |
| writeUpdate in onUpdate handler | **fixed** — same mechanism as writeInsert |
| writeBatch with different keys | **fixed** — same mechanism |
| refetch + different key | **fixed** — Mechanism B removes orphaned key |
| No writeInsert + default refetch | **safe** — unchanged behavior |
| No writeInsert + refetch: false | **safe** — key survives until external sync |

## Test plan

Five new tests in `packages/query-db-collection/tests/query.test.ts`:

- [x] `should not duplicate items when writeInsert uses a different key than the optimistic insert`
- [x] `should mark item as synced when writeInsert uses the same key as the optimistic insert`
- [x] `should not duplicate items when refetch returns a different key than the optimistic insert`
- [x] `should clean up optimistic state when writeUpdate is called in onUpdate handler`
- [x] `should not duplicate items when writeBatch uses different keys than the optimistic inserts`

Full test suites pass:
- `@tanstack/db`: 94 files, 2275 passed, 0 failed
- `@tanstack/query-db-collection`: 2 files, 200 passed, 0 failed

### DIFF
--- a/.changeset/fix-optimistic-insert-cleanup.md
+++ b/.changeset/fix-optimistic-insert-cleanup.md
@@ -1,0 +1,7 @@
+---
+"@tanstack/db": patch
+---
+
+fix: clean up optimistic state when server returns a different key than the optimistic insert
+
+When an `onInsert`/`onUpdate`/`onDelete` handler syncs server data back to the collection (via `writeInsert`, `writeUpdate`, `writeUpsert`, `writeDelete`, or `refetch`), the optimistic state under the original client key is now correctly removed if the server returns a different key. Previously, the client-key item persisted forever alongside the server-key item, causing duplication and stale `$synced: false` state.
--- a/.changeset/fix-optimistic-insert-cleanup.md
+++ b/.changeset/fix-optimistic-insert-cleanup.md
@@ -1,5 +1,5 @@
 ---
-"@tanstack/db": patch
+'@tanstack/db': patch
 ---
 
 fix: clean up optimistic state when server returns a different key than the optimistic insert
--- a/packages/db/src/collection/state.ts
+++ b/packages/db/src/collection/state.ts
@@ -492,7 +492,10 @@ export class CollectionStateManager<
           isDirectTransaction &&
           !this.processedCompletedTransactions.has(transaction.id) &&
           !this.directTransactionsWithSyncWrites.has(transaction.id)
-        if (isDirectTransaction && !this.processedCompletedTransactions.has(transaction.id)) {
+        if (
+          isDirectTransaction &&
+          !this.processedCompletedTransactions.has(transaction.id)
+        ) {
           this.processedCompletedTransactions.add(transaction.id)
         }
 
--- a/packages/db/src/collection/state.ts
+++ b/packages/db/src/collection/state.ts
@@ -83,6 +83,8 @@ export class CollectionStateManager<
   public pendingOptimisticDeletes = new Set<TKey>()
   public pendingOptimisticDirectUpserts = new Set<TKey>()
   public pendingOptimisticDirectDeletes = new Set<TKey>()
+  public directTransactionsWithSyncWrites = new Set<string>()
+  public processedCompletedTransactions = new Set<string>()
 
   /**
    * Tracks the origin of confirmed changes for each row.
@@ -483,6 +485,17 @@ export class CollectionStateManager<
       const isDirectTransaction =
         transaction.metadata[DIRECT_TRANSACTION_METADATA_KEY] === true
       if (transaction.state === `completed`) {
+        // Add-once guard: only process direct transactions for pendingOptimisticDirect*
+        // on their first completion. This prevents re-adding keys that were already
+        // cleaned up by commitPendingTransactions (e.g., after writeInsert with same key).
+        const isFirstProcessing =
+          isDirectTransaction &&
+          !this.processedCompletedTransactions.has(transaction.id) &&
+          !this.directTransactionsWithSyncWrites.has(transaction.id)
+        if (isDirectTransaction && !this.processedCompletedTransactions.has(transaction.id)) {
+          this.processedCompletedTransactions.add(transaction.id)
+        }
+
         for (const mutation of transaction.mutations) {
           if (!this.isThisCollection(mutation.collection)) {
             continue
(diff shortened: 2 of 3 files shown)


KyleAMathews:
Closing as superseded by #1547, which fixed the concrete stale optimistic-row bug reported in #1442 and shipped in @tanstack/db 0.6.7. The linked issue is already closed with confirmation from the original reproducer.\n\nThe broader problem of correlating a temporary client key with a different server-assigned key cannot be inferred safely by collection core from the absence or arrival of rows. That identity mapping must come explicitly from the mutation/backend or adapter contract; otherwise delayed sync, filtering, deletion, and an unrelated insert are indistinguishable. The additional transaction-tracking heuristics in this candidate therefore should not be retained as a generic reconciliation mechanism.\n\nThanks @goatrenterguy for the detailed diagnosis, tests, and exploration—the concrete failure helped drive the smaller fix in #1547.

Issue #1458, opened 2026-04-07
sqlite-compiler: bare "id" in trigger WHEN clauses breaks DELETE for on-demand PowerSync collections


simonboisset:
- [x] I've validated the bug against the latest version of DB packages (`@tanstack/powersync-db-collection@0.1.40`, `@tanstack/db@0.6.2`)
 
---
 
## Describe the bug
 
In the `@tanstack/powersync-db-collection` package, the `compileSQLite` function in `sqlite-compiler.ts` generates incorrect SQL when compiling the `id` column reference inside trigger `WHEN` clauses for on-demand sync collections.
 
When `powerSyncCollectionOptions` is used with `syncMode: 'on-demand'`, the package creates TEMP AFTER triggers on the internal `ps_data__*` tables to track diffs. These triggers use `compileSQLite` with `{ jsonColumn: 'OLD.data' }` or `{ jsonColumn: 'NEW.data' }` to build WHEN clauses.
 
For non-`id` columns, the output is correct: `json_extract(OLD.data, '$.column_name')`. But for the `id` column, the code skips `json_extract` (since `id` lives on the row itself, not inside the JSON `data` column) and falls through to `quoteIdentifier(columnName)`, producing a bare `"id"` — without the required `OLD.` or `NEW.` prefix.
 
SQLite cannot resolve a bare column reference inside a trigger body. It needs `OLD."id"` or `NEW."id"`. As a result, the TEMP diff-tracking triggers silently fail, and **DELETE operations are never applied to the PowerSync CRUD queue** for any on-demand collection whose stream WHERE clause references `id`.
 
**The bug is completely silent — no error is thrown, no console warning.** The row disappears optimistically from TanStack DB's in-memory state, then reappears after the next PowerSync sync cycle because the delete was never persisted to SQLite.
 
---
 
## Root cause
 
In [`sqlite-compiler.ts`, lines 97–101](https://github.com/TanStack/db/blob/main/packages/powersync-db-collection/src/sqlite-compiler.ts#L97-L101):
 
```typescript
const columnName = exp.path[0]!
if (compileOptions?.jsonColumn && columnName !== `id`) {
  return `json_extract(${compileOptions.jsonColumn}, '$.${columnName}')`
}
return quoteIdentifier(columnName)
```
 
[...]

PR #1470, opened 2026-04-13
fix(powersync-db-collection): "id" in trigger WHEN clauses breaks DELETE for on-demand PowerSync collections

Chriztiaan:
## 🎯 Changes

Addresses https://github.com/TanStack/db/issues/1458. 

When `compileSQLite` builds WHEN clauses for diff-tracking triggers on on-demand sync collections, it correctly emits `json_extract(OLD.data, '$.column_name')` for regular columns but falls through to a bare "id" for the id column - missing the required `OLD.`/`NEW.` prefix. SQLite can't resolve an unqualified column reference inside a trigger body, so the diff-tracking triggers silently fail and DELETE operations never reach the PowerSync CRUD queue.

Because no error is thrown, the failure is invisible. Deleted rows disappear from TanStack DB's in-memory state as expected, but since the delete isn't persisted to SQLite, the rows reappear after the next sync cycle. This affects any on-demand collection whose stream WHERE clause references id.

As mentioned in the original issue, the minimal fix is to extract the OLD/NEW prefix from jsonColumn when falling through for the id column. I have also added a few tests to verify that the data correctly enters the upload queue.

The issue touches on making the prefixes configurable, but I don't think that is needed at the moment.

## ✅ Checklist

## 🚀 Release Impact


Chriztiaan:
Hey @samwillis , could I possibly get eyes on this?
We have already tested and reviewed it internally :)

samwillis (approved):

PR #1471, opened 2026-04-13
fix: lazy load includes child collections in on-demand sync mode

### LINKED ISSUE #1441: Query Driven sync doesn't pass filters when being used in includes

**Describe the bug**
When using the nested UI queries as announced [here](https://tanstack.com/blog/tanstack-db-0.6-app-ready-with-persistence-and-includes#includes-project-your-data-into-the-same-shape-as-your-ui), the where clause is not passed to the `queryFn` via `loadSubsetOptions`. This means all the data needs to be fetched thus eliminating the benefits of partial data fetching.

**To Reproduce**
```typescript
const root = createCollection(
  queryCollectionOptions({
    queryKey: ['root'],
    syncMode: 'on-demand',
    queryFn: async (context) => {
      return getRoots();
    },
    queryClient,
    getKey: (item) => item.id,
  })
);

const items = createCollection(
  queryCollectionOptions({
    queryKey: ['root-items'],
    syncMode: 'on-demand',
    queryFn: async (context) => {
      const { filters } = parseLoadSubsetOptions(
        context.meta?.loadSubsetOptions
      );
      const rootId = filters[0].value; // <- This is never defined
      const items = await getRootItems(filters[0].value)
    },
    queryClient,
    getKey: (item) => item.id,
  })
);


const collection = await queryOnce((q) => {
  const c = q
    .from({ r: root })
    .select(({ r }) => ({
      id: collection.id,
      name: collection.title,
      children: toArray(
        q
          .from({ items: items })
          .where(({ items }) => eq(items.rootId, r.id))
          .select(({ items }) => ({
            id: items.id,
            type: items.type,
            name: items.title,
          }))
      ),
    }));
  return c;
});
```

**Expected behavior**
`rootId` should be defined in the `queryFn`

- OS: [e.g. iOS]: macOS Tahoe 26.2
- Browser [e.g. chrome, safari]: chrome
- Version [e.g. 22]: 146.0.7680.167

**Additional context**

kevin-dp:
## Summary

- Includes child collections now use lazy loading (matching the existing join optimization), so only rows matching parent correlation keys are loaded on-demand via `requestSnapshot({ where: inArray(field, keys) })`
- Previously, includes child collections loaded ALL data upfront because the correlation WHERE clause was consumed by the inner join and no filter was passed to the sync layer's `loadSubset`/`queryFn`
- Adds 6 new tests verifying the lazy loading behavior for includes

Fixes #1441

## Test plan

- [x] New tests in `includes-lazy-loading.test.ts` verify:
  - Correlation filter (`inArray`) is passed to child collection's `loadSubset`
  - Query results remain correct with lazy-loaded includes
  - Child source is marked lazy (no eager full load without filter)
  - Reactive loading when new parent rows arrive
  - Works with both `toArray` and `Collection` materialization
- [x] All 112 existing includes tests pass
- [x] Full test suite passes (2071 tests)
- [x] Build passes

### DIFF
--- a/.changeset/includes-child-where-clauses.md
+++ b/.changeset/includes-child-where-clauses.md
@@ -1,0 +1,7 @@
+---
+'@tanstack/db': patch
+---
+
+fix: pass child where clauses to loadSubset in includes
+
+Pure-child WHERE clauses on includes subqueries (e.g., `.where(({ item }) => eq(item.status, 'active'))`) are now passed through to the child collection's `loadSubset`/`queryFn`, enabling server-side filtering. Previously only the correlation filter reached the sync layer; additional child filters were applied client-side only.
--- a/.changeset/includes-lazy-loading.md
+++ b/.changeset/includes-lazy-loading.md
@@ -1,0 +1,7 @@
+---
+'@tanstack/db': patch
+---
+
+fix: lazy load includes child collections in on-demand sync mode
+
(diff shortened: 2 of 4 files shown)


samwillis (approved):
LGTM

PR #1472, opened 2026-04-13
fix: pass child where clauses to loadSubset in includes

kevin-dp:
## Summary

- Merges child `sourceWhereClauses` into the parent's map during includes compilation, so pure-child WHERE clauses (e.g., `eq(item.status, 'active')`) are picked up by the child collection's `CollectionSubscriber` and AND-combined with the `inArray` correlation filter in `loadSubset`
- Previously, only `aliasToCollectionId` and `aliasRemapping` were merged from the child compilation result — `sourceWhereClauses` was silently dropped
- Adds 4 new tests verifying child WHERE clauses reach `loadSubset`

Builds on #1471

## Test plan

- [x] New tests verify:
  - Single pure-child WHERE clause is included in `loadSubset` alongside correlation filter
  - Multiple pure-child WHERE clauses are all included
  - Query results are correct with child WHERE filtering
  - `and()` syntax combining correlation + child filter works
- [x] All 10 includes lazy loading tests pass
- [x] Full test suite passes (2075 tests)
- [x] Build and typecheck pass

### DIFF
--- a/.changeset/includes-child-where-clauses.md
+++ b/.changeset/includes-child-where-clauses.md
@@ -1,0 +1,7 @@
+---
+'@tanstack/db': patch
+---
+
+fix: pass child where clauses to loadSubset in includes
+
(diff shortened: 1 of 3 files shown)


samwillis (approved):
Lgtm

PR #1474, opened 2026-04-14
docs(react-db): add CRITICAL common mistakes to skill

balegas:
## Summary
- Add two CRITICAL common mistakes to the `react-db` skill that AI agents hit frequently
- **Using `===` instead of `eq()` in `.where()`** — throws `InvalidWhereExpressionError` at runtime
- **Assigning `useLiveQuery()` directly instead of destructuring `{ data }`** — causes "map is not a function"

Both are documented in `db-core/live-queries` but agents typically only read the `react-db` skill. Adding the warnings where they'll be seen.

## Context
These are the two most common runtime errors we see when AI agents build TanStack DB apps. The `live-queries` sub-skill documents them, but agents read `react-db/SKILL.md` as their primary reference and miss the sub-skill cross-references.

## Test plan
- [ ] Verify skill renders correctly
- [ ] Run agent session building a CRUD app — confirm correct `eq()` usage and `{ data }` destructuring

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/react-db/skills/react-db/SKILL.md
+++ b/packages/react-db/skills/react-db/SKILL.md
@@ -307,8 +307,11 @@ See meta-framework/SKILL.md for full preloading patterns.
 Wrong — throws `InvalidWhereExpressionError` at runtime:
 
 ```tsx
-const { data } = useLiveQuery((q) =>
-  q.from({ todo: todoCollection }).where(({ todo }) => todo.completed === true),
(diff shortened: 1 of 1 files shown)


samwillis (approved):

KyleAMathews (approved):
Nice! Could you add a changeset too so it can be released?

PR #1493, opened 2026-04-24
fix(electric-db-collection): preserve persisted rows on progressive resume

### LINKED ISSUE #1478: Persisted collection with progressive sync mode wiped on reconnect

## Describe the bug

When using `electricCollectionOptions` with `syncMode: "progressive"` and `persistedCollectionOptions` (browser SQLite persistence), data hydrated from persistence gets wiped when the Electric stream reconnects after going offline. The collection drops to 0 items and never recovers.

Persistence and hydration work correctly on their own. The data loads from SQLite after a refresh while offline. The wipe happens the moment the stream comes back online.

## To Reproduce

1. Create a persisted Electric collection with `syncMode: "progressive"`.
2. Load the page. Data syncs and is persisted to SQLite.
3. Go offline.
4. Refresh the page. Data hydrates from SQLite. All rows are visible.
5. Come back online.
6. The collection is wiped to 0 items.

## Expected behavior

The collection should keep the hydrated rows and resume syncing changes from the stored offset. The stream does resume correctly (the offset and handle are sent in the reconnection request), but the in-memory state is wiped before the stream has a chance to deliver any data.

## Desktop

- OS: macOS
- Browser: Chromium (Electron)

## Additional context

```
@tanstack/db: 0.6.4
@tanstack/electric-db-collection: 0.3.2
@tanstack/react-db: 0.1.82
@tanstack/browser-db-sqlite-persistence: 0.1.8
@electric-sql/client: 1.5.15
```

pkudinov:
## Summary

Fixes TanStack/db#1478.

Progressive Electric collections wrapped with SQLite persistence can hydrate rows from the persisted store, then wipe them when the resumed Electric stream sends an `up-to-date` control message without replaying snapshot rows.

This changes progressive startup so a valid persisted Electric resume point is treated as already past the initial atomic-swap phase. That preserves hydrated persisted rows while still allowing normal progressive initial sync to use the existing buffer/truncate/apply behavior.

## Root cause

Progressive mode uses the first `up-to-date` message to finish initial sync by truncating the collection and replaying buffered stream rows. With persisted resume metadata, the stream starts from a saved `handle`/`offset`, so it may legitimately send `up-to-date` with no old rows. The collection had already hydrated those rows from persistence, but the initial atomic swap ran anyway and truncated them.

## Changes

- Initialize progressive `hasReceivedUpToDate` when using a compatible persisted Electric resume point.
- Add a regression test that hydrates persisted rows, resumes with saved Electric metadata, receives `up-to-date` without snapshot rows, and verifies rows remain in memory and persistence.
- Add a patch changeset for `@tanstack/electric-db-collection`.

## Validation

```bash
pnpm --filter @tanstack/electric-db-collection test -- --coverage.enabled false
pnpm exec eslint packages/electric-db-collection/src/electric.ts packages/electric-db-collection/tests/electric.test.ts --max-warnings=0
```

I also verified this patch in a real Next.js app using browser OPFS SQLite persistence: after resetting local SQLite, loading a persisted progressive Electric collection, and refreshing/warm-loading twice, the collection stayed populated instead of dropping to zero rows.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed progressive resume behavior to prevent saved collection data from being truncated, ensuring persisted rows remain intact when resuming from previously stored Electric shape metadata.

* **Tests**
  * Added an integration test covering persisted progressive resume, verifying hydrated persisted rows stay visible through the initial progressive phase and update correctly after resumed change and up-to-date control messages.

* **Chores**
  * Published a patch release entry for the collection package to deliver the fix.


KyleAMathews (approved):
thanks!

PR #1496, opened 2026-04-25
fix(db): includes orderBy ignored after optimistic update on child collection

### LINKED ISSUE #1444: TanStack DB "includes" ignores orderBy after optimistic update
I have a simple todo list with statuses and tasks, both sorted by position.

Example query:

 ```
 const { data: todoData } = useLiveQuery((q) => {
    return q
      .from({
        status: statusCollection,
      })
      .orderBy(({ status }) => status.position, "asc")
      .select(({ status }) => ({
        id: status.id,
        name: status.name,
        position: status.position,
        tasks: q
          .from({ task: taskCollection })
          .where(({ task }) =>
            eq(task.status_id, status.id),
          )
          .orderBy(({ task }) => task.position, "asc")
          .select(({ task }) => ({
            id: task.id,
            name: task.name,
            position: task.position,
          })),
      }));
  });
```

After an optimistic update on the child collection (tasks), the orderBy is ignored. The list is only sorted correctly after the server confirmation (second render).
If I perform an optimistic update on the parent collection (status), everything works as expected - sorting is applied immediately.

Before the action (task position change):

```
[{id: 2, name: 'Test', position: 0}
{id: 1, name: 'Hello', position: 1} ...]
```

After optimistic update:

```
[{id: 2, name: 'Test', position: 1}
{id: 1, name: 'Hello', position: 0} ...]
```

After server confirmation:

```
[{id: 1, name: 'Hello', position: 0}
{id: 2, name: 'Test', position: 1} ...]
```

Tested on:
```
    "@tanstack/electric-db-collection": "^0.2.43",
    "@tanstack/react-db": "^0.1.79",
```

comet08:
Closes #1444

## 🎯 Changes

When a child collection item is updated, the D2 graph emits a `(-1, +1)` pair. If the delete (`-1`) message arrives first, an accumulator entry is created with the **old** `orderByIndex`. The subsequent insert (`+1`) message updated `value` correctly but never updated `orderByIndex` — so the child collection's comparator kept using the stale positional index.

This caused incorrect sort order immediately after optimistic updates on child collections, with ordering only correcting itself after server confirmation.

The parent pipeline (`accumulateChanges`) already handled this correctly. This fix applies the same `orderByIndex` update logic to the child pipeline accumulator and the nested pipeline accumulator for consistency.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved `orderBy` sorting for included child collections when child rows are updated, preventing ordering from being lost or overwritten during aggregation.

* **Tests**
  * Added coverage to verify included child ordering remains correct after both optimistic client-side changes and server-backed sync confirmation.


KyleAMathews:
The reported optimistic child-reorder bug no longer reproduces on current main. Its exact regression is green and retained in packages/db/tests/query/includes-oracle.property.test.ts, so this fix PR is no longer needed. Closing as superseded by the current implementation.

PR #1497, opened 2026-04-26
fix: invalidate stale predicate-scoped cache entries on manual-sync writes

marbemac:
Manual-sync writes in on-demand Query Collections now preserve predicate-, order-, and pagination-scoped Query cache results. Active queries revalidate from their source, while inactive or disabled entries are removed so later subscriptions fetch fresh data.

## Root cause

Each on-demand Query cache entry belongs to one `queryFn` result and may encode a distinct `where`, `orderBy`, `limit`, or `offset`. Manual writes replaced every entry under the collection's base key with the full normalized collection snapshot. That changed the shape and membership of scoped cache entries.

An inactive poisoned entry could later be treated as a valid cache hit. Its rows would be reconciled through the wrong query ownership scope, which surfaced in production as “not found” results for rows that still existed in Postgres. Active limited or ordered queries had the same flaw: patching them from the normalized snapshot could widen the cache or fail to replenish a deleted row.

## Approach

- Keep eager behavior unchanged: eager collections patch their full-result cache in place.
- Never write the full normalized snapshot into on-demand cache entries.
- Refetch active, fetchable observers directly. This also revalidates queries with `staleTime: 'static'`.
- Remove inactive and disabled entries in one cache operation so their next subscription runs `queryFn`.
- Ignore success notifications that still carry the exact pre-write `dataUpdateCount`, then accept the next authoritative result.
- Clear the notification marker on observer cleanup and Query reset or recreation paths.
- Wait for an optimistic transaction's deferred-refresh barrier before refetching, and revalidate after either barrier outcome.

## Key invariants

- An on-demand Query cache entry only contains data shaped by its own `queryFn`.
- A direct write remains visible in the synced store immediately.
- Fetch and invalidation notifications cannot replace that write with unchanged cached data.
- A later authoritative result always reconciles the collection, including after Query state resets.

## Non-goals and trade-offs

This does not change eager cache updates or the direct-write API. Active on-demand queries now incur a network refetch after a manual write; that cost is required because a normalized collection snapshot cannot reconstruct ordered, limited, paginated, or predicate-scoped results.

## Verification

```bash
pnpm exec vitest run --config packages/query-db-collection/vitest.config.ts packages/query-db-collection/tests/query.test.ts --maxWorkers=2
pnpm exec tsc --noEmit -p packages/query-db-collection/tsconfig.json
pnpm exec eslint packages/query-db-collection/src/query.ts packages/query-db-collection/tests/query.test.ts
pnpm test:docs
```

The package suite passes all 146 tests. New regressions cover inactive cache hits, overlapping scopes, active limited and static queries, disabled entries, in-flight initial loads, Query resets, and rejected deferred-refresh barriers.

## Files changed

- `packages/query-db-collection/src/query.ts`: revalidate or remove on-demand cache entries and guard stale notifications.
- `packages/query-db-collection/tests/query.test.ts`: add red/green coverage for scoped-cache and lifecycle cases.
- `docs/collections/query-collection.md`: document eager and on-demand direct-write behavior.
- `docs/reference/query-db-collection/interfaces/QueryCollectionUtils.md`: update generated API reference text.
- `.changeset/fix-cache-poisoning-on-manual-writes.md`: add the patch release note.

## Release impact

- [x] This change affects published code and includes a changeset.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed manual writes in on-demand collections so they no longer overwrite filtered, sorted, or paginated cache results with a full snapshot.
  - Active queries now refresh with current data, while inactive or disabled entries are removed.
  - Prevented stale or incorrect results after direct inserts, updates, and deletes.

- **Documentation**
  - Clarified cache behavior for direct writes across eager and on-demand collections.
  - Added guidance for scoped queries, persistence handlers, and deferred refreshes.


andyfelder16:
any news on this? we hit the same bug :( — for now we are avoiding the `write*` utils entirely on on-demand collections and reconciling with `refetch()` instead, so it would be great to see this land

KyleAMathews:
@marbemac @andyfelder16 could you test this against my latest pushes? I merged in origin/main & did a few fixes

KyleAMathews:
Superseded by #1826, which rebases the work onto current main and includes the additional cache-ownership and lifecycle fixes found through oracle testing and review.

PR #1504, opened 2026-04-30
feat: add materialize() helper for includes subqueries

### LINKED ISSUE #1481: feat: `toArray()` equivalent for `findOne` subqueries
When using includes (nested queries) we can use `toArray` to materialize the results of the nested query into an array instead of a collection. However, when the nested query uses `findOne` to return a single element it is annoying to have to deal with a singleton array. We want a `toArray` equivalent for this use case that simply returns that one element.

I'm questioning if "toArray" is a good name because it means we now have to introduce another function for subqueries that return a single element. But conceptually both are doing the same: materializing the underlying collection. So perhaps `materialize(subquery)` would be  a better name and it will return an array if the underlying query returns multiple results and a single element if it's a singular query.

kevin-dp:
Closes #1481.

## Summary

- Adds `materialize()` — a single helper for materializing includes subqueries onto parent rows. It resolves to `Array<T>` when the wrapped subquery returns multiple rows, and to `T | undefined` when the subquery is `findOne()`. The intent is to spare callers from unwrapping a singleton array whenever they know the child query yields at most one row.
- `toArray()` is unchanged — it still always produces `Array<T>`. `materialize()` is purely additive.
- Implementation: extends the existing `IncludesMaterialization` enum with `'singleton'`, threads it through the IR / compiler / runtime, and updates the `materializeIncludedValue` helper to return the first child row (or `undefined`) for the singleton case. The reactive re-emit machinery routes through that helper, so insert/update/delete/no-match transitions all flow naturally.
- Type inference: `materialize(q.from(...).findOne())` infers `T | undefined`, `materialize(q.from(...))` infers `Array<T>`. Detected at the type level via a phantom `IsSingle` flag on `MaterializeWrapper` that's set from `TContext extends SingleResult`.
- `materialize()` is also rejected inside expressions (`coalesce()`, `eq()`, …) with the same kind of error message that `toArray()` produces.

## Example

```ts
// Multi-row → Array<Issue>
select(({ p }) => ({
  ...p,
  issues: materialize(
    q.from({ i: issues }).where(({ i }) => eq(i.projectId, p.id))
  ),
}))

// Singleton → Project | undefined
select(({ i }) => ({
  ...i,
  project: materialize(
    q.from({ p: projects }).where(({ p }) => eq(p.id, i.projectId)).findOne()
  ),
}))
```

A changeset will follow in a separate commit.

## Test plan

- [x] New runtime tests in `tests/query/includes.test.ts` cover singleton attach, missing-match → `undefined`, insert / update / delete reactivity, two parents sharing a correlation key, multi-row fallback, scalar singleton, and nested `materialize` (array of singletons).
- [x] New type tests in `tests/query/includes.test-d.ts` cover array, singleton (with select / without select / scalar), and a nested array-of-singleton case.
- [x] Full `packages/db` test suite passes (1502 tests, 0 type errors).
- [x] Package builds cleanly.

## Summary by CodeRabbit

* **New Features**
  * Introduced `materialize()` helper function for wrapping subqueries in SELECT statements, providing clean snapshot materialization.
  * Automatically returns `T | undefined` for single-row queries and `Array<T>` for multi-row queries.
  * Supports nested materialization for complex query structures.

* **Tests**
  * Added comprehensive test coverage for materialize functionality.

### DIFF
--- a/packages/db/src/query/builder/functions.ts
+++ b/packages/db/src/query/builder/functions.ts
@@ -2,6 +2,7 @@ import { Aggregate, Func } from '../ir'
 import { toExpression } from './ref-proxy.js'
 import type { BasicExpression } from '../ir'
 import type { RefProxy } from './ref-proxy.js'
+import type { SingleResult } from '../../types.js'
 import type {
   Context,
   GetRawResult,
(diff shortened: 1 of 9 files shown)


andrewcoelho:
This is much needed as doing `toArray` and then data.[0] everywhere really clutters everything

samwillis (approved):

Jaime02:
The tanstack db docs are currently missing the new `materialize` function, right? 😅

PR #1508, opened 2026-05-02
Delete CONTRIBUTING.md reference from README

Jaime02:
The file does not exist 😅


KyleAMathews:
Closing as obsolete. The premise was that CONTRIBUTING.md did not exist, but current main now contains CONTRIBUTING.md and the README link resolves to it, so removing the link would now be incorrect.

PR #1512, opened 2026-05-04
fix(db): preserve union types in select result extraction

### LINKED ISSUE #1511: types: `.select()` collapses discriminated union types on object fields

**Describe the bug**

`.select()` collapses discriminated union types on object fields into a single type with only the intersection of keys. Fields that are unique to individual union variants become inaccessible (`never`) in the result type, while the union is correctly preserved when no `.select()` is used.

**To Reproduce**

```ts
import { createCollection, createLiveQueryCollection } from '@tanstack/db'

type Document =
  | { type: 'pdf'; url: string; pages: number }
  | { type: 'image'; url: string; width: number; height: number }
  | { type: 'legacy'; path: string }

type Item = { id: number; name: string; document: Document }

const items = createCollection<Item, number>({ id: 'items', getKey: (i) => i.id })

// ✅ Without .select() — union is preserved
const col1 = createLiveQueryCollection((q) => q.from({ i: items }))
col1.toArray[0]!.document // type: Document (correct union)

// ❌ With .select() — union is collapsed
const col2 = createLiveQueryCollection((q) =>
  q.from({ i: items }).select(({ i }) => ({
    id: i.id,
    document: i.document,
  })),
)
col2.toArray[0]!.document
// Actual type: { type: "pdf" | "image" | "legacy" }
// Variant-specific keys (url, pages, width, height, path) are lost
```

**Expected behavior**

`result.document` should have type `Document` (the original discriminated union) regardless of whether `.select()` is used. The union should be preserved so that narrowing (e.g., `if (doc.type === 'pdf')`) works correctly downstream.

nathan-muir:
Fixes #1511

## 🎯 Changes

When a collection field is a discriminated union of object types, selecting that field via `.select()` collapses the union into a single type with only the intersection of keys.

```ts
type Document =
  | { type: 'pdf'; url: string; pages: number }
  | { type: 'image'; url: string; width: number; height: number }
  | { type: 'legacy'; path: string }

type Item = { id: number; document: Document }

// Before: result.document is { type: string } — union collapsed, variant keys lost
// After:  result.document is Document — full discriminated union preserved
q.from({ i: items }).select(({ i }) => ({ id: i.id, document: i.document }))
```

The root cause is in `ResultTypeFromSelect`: when a `Ref<T>` holds a union type, `ExtractRef` recurses into the mapped keys via `keyof`/`Prettify`, which collapse the union to only common keys.

The fix detects when a `Ref`'s branded type (`RefLeaf<T>`) is a union and extracts it directly, bypassing `ExtractRef`. This preserves `Ref<T>`'s mapped type structure so that common keys of the union remain accessible in `.where()` and `.orderBy()` callbacks (e.g., `eq(i.document.type, 'pdf')`).

New helpers:
- `IsUnion<T>` — detects union types
- `ExtractRefBrand<T>` — extracts the raw type from a `RefLeaf`'s brand

Tests added:
- Select preserves union types on object fields
- Where clause works on common keys of union fields
- Select preserves union when collection type is a union

## ✅ Checklist

## 🚀 Release Impact

-  ~This change is docs/CI/dev-only (no release).~

### DIFF
--- a/.changeset/fix-ref-union-collapse.md
+++ b/.changeset/fix-ref-union-collapse.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/db': patch
+---
+
+fix: preserve discriminated union types when selecting object fields via `.select()`
--- a/packages/db/src/query/builder/types.ts
+++ b/packages/db/src/query/builder/types.ts
@@ -322,9 +322,16 @@ export type ResultTypeFromSelect<TSelectObject> =
                 : // includes subquery (bare QueryBuilder) — produces a child Collection
                   TSelectObject[K] extends QueryBuilder<infer TChildContext>
                   ? Collection<GetResult<TChildContext>>
-                  : // Ref (full object ref or spread with RefBrand) - recursively process properties
+                  : // Ref (full object ref or spread with RefBrand)
                     TSelectObject[K] extends Ref<infer _T>
-                    ? ExtractRef<TSelectObject[K]>
+                    ? // When the branded type is a union, extract it directly to
+                      // preserve the discriminated union. ExtractRef would collapse
+                      // it via keyof/Prettify which only yield common keys.
+                      true extends IsUnion<ExtractRefBrand<TSelectObject[K]>>
+                      ? IsNullableRef<TSelectObject[K]> extends true
+                        ? ExtractRefBrand<TSelectObject[K]> | undefined
+                        : ExtractRefBrand<TSelectObject[K]>
+                      : ExtractRef<TSelectObject[K]>
                     : // RefLeaf (simple property ref like user.name)
                       TSelectObject[K] extends RefLeaf<infer T>
                       ? IsNullableRef<TSelectObject[K]> extends true
@@ -371,6 +378,9 @@ export type SelectResult<TSelect> =
   IsPlainObject<TSelect> extends true
     ? ResultTypeFromSelect<TSelect>
     : ResultTypeFromSelectValue<TSelect>
+
+// Extract the raw type stored in a RefLeaf's brand
+type ExtractRefBrand<T> = T extends RefLeaf<infer U> ? U : never
 
 // Extract Ref or subobject with a spread or a Ref
 type ExtractRef<T> = Prettify<ResultTypeFromSelect<WithoutRefBrand<T>>>
@@ -1067,6 +1077,14 @@ type IsPlainObject<T> = T extends unknown
   : false
 
 type IsAny<T> = 0 extends 1 & T ? true : false
+
+// Detects whether T is a union type (e.g., A | B | C returns true, A returns false)
+type IsUnion<T, U = T> = T extends unknown
+  ? [U] extends [T]
+    ? false
+    : true
+  : false
+
 
 /**
  * JsBuiltIns - List of JavaScript built-ins
--- a/packages/db/tests/query/select.test-d.ts
+++ b/packages/db/tests/query/select.test-d.ts
@@ -1,6 +1,6 @@
 import { describe, expectTypeOf, test } from 'vitest'
 import { createCollection } from '../../src/collection/index.js'
-import { createLiveQueryCollection } from '../../src/query/index.js'
+import { createLiveQueryCollection, eq } from '../../src/query/index.js'
 import { mockSyncCollectionOptions } from '../utils.js'
 import { upper } from '../../src/query/builder/functions.js'
 import type { OutputWithVirtual } from '../utils.js'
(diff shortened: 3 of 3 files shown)


kevin-dp (changes requested):
Thanks for digging into this — the diagnosis is correct and the helpers (`IsUnion`, `ExtractRefBrand`) are useful. However, after reproducing locally I think the fix as currently written only addresses the **shallow** case. It doesn't preserve the discriminated union when the union sits inside a nested object that itself is selected.

## Failing repro on top of this PR

Adding the following test to `packages/db/tests/query/select.test-d.ts` (alongside the two tests this PR already adds) fails *even with this PR applied*:

```ts
test(`select preserves union when nested under another field`, () => {
  type Payload =
    | { kind: 'text'; body: string }
    | { kind: 'binary'; bytes: number; mime: string }

  type Envelope = { id: number; payload: { inner: Payload } }

  const envelopes = createCollection(
    mockSyncCollectionOptions<Envelope>({
      id: `nested-union-envelopes`,
      getKey: (e) => e.id,
      initialData: [],
    }),
  )

  // Selecting a nested object whose field is a discriminated union
  // must preserve the union (not collapse to the intersection of keys).
  const col = createLiveQueryCollection((q) =>
    q.from({ e: envelopes }).select(({ e }) => ({
      id: e.id,
      payload: e.payload,
    })),
  )
  const r = col.toArray[0]!
  expectTypeOf(r.payload).toEqualTypeOf<{ inner: Payload }>()
  expectTypeOf(r.payload.inner).toEqualTypeOf<Payload>()
})
```

Result on this PR's branch (running `pnpm vitest run tests/query/select.test-d.ts --typecheck` from `packages/db/`):

```
× select types > select preserves union when nested under another field
  → Type '{ inner: Payload; }' does not satisfy the constraint
    '{ inner: { kind: "binary" | "text"; body: never; bytes: never; mime: never }; ... }'
  → Type 'Payload' does not satisfy the constraint
    '{ kind: "binary" | "text"; body: never; bytes: never; mime: never }'
```

So `r.payload.inner` still resolves to the keys-intersected, `never`-filled object — the exact symptom #1511 is about, just one level deeper.

## Why the current patch isn't sufficient

The fix lives in `ResultTypeFromSelect` and short-circuits to `ExtractRefBrand<TSelectObject[K]>` **only when the top-level branded type of the selected field is itself a union**:

```ts
TSelectObject[K] extends Ref<infer _T>
  ? true extends IsUnion<ExtractRefBrand<TSelectObject[K]>>
    ? ...ExtractRefBrand<TSelectObject[K]>...
    : ExtractRef<TSelectObject[K]>
```

In the shallow case (`document: ItemDocument`), `ExtractRefBrand` returns `ItemDocument` directly, `IsUnion` is `true`, and we short-circuit. ✅

In the nested case (`payload: { inner: Payload }`), `ExtractRefBrand` returns `{ inner: Payload }` — **not** a union at the top level — so `IsUnion` is `false` and we fall through to the normal `ExtractRef` path. That path goes:

1. `ExtractRef<Ref<{ inner: Payload }>>` → `Prettify<ResultTypeFromSelect<WithoutRefBrand<Ref<{ inner: Payload }>>>>`
2. Inside the resulting object the `inner` field type is `Ref<Payload>` (a `Ref` over a union)
3. `Ref<Payload>` expands its mapped type `{ [K in keyof Payload]: ... }` — but `keyof (A | B)` yields only common keys, so variant-specific keys collapse to `never`
4. The `IsUnion` gate isn't re-checked recursively at this nested level, so the short-circuit doesn't fire again

So the root structural cause (`Ref<T>`'s mapped type collapsing unions via `keyof`) is unfixed; the patch only catches the case where the union happens to be exactly at the selected field's top level.

## What a complete fix probably needs

In rough order of how invasively they touch the codebase:

1. **Recurse the `IsUnion`/`ExtractRefBrand` gate through nested objects.** Whenever we descend through an object-shaped Ref, re-check each nested field for the same condition. This stays close to the spirit of the current PR but extends it to arbitrary depth.

2. **Or** distribute `Ref<T>` itself over `T` (`T extends unknown ? RefSingle<T, Nullable> : never`) so `Ref<A | B>` becomes `Ref<A> | Ref<B>` at construction time, and then make `ResultTypeFromSelect` / `ExtractRef` / `WithoutRefBrand` distribute consistently. This is structurally cleaner but has a non-trivial blast radius — several downstream branches in `ResultTypeFromSelect` pattern-match on whole-union shapes like `RefLeaf<T> | undefined` (the encoding TS uses for schema-optional fields). Distributing breaks those matches and regresses optional-field handling unless optionality is re-encoded as a marker brand instead of a union.

PR #1514, opened 2026-05-04
feat(electric-db-collection): support nested property references in SQL queries

miguelrk:
## 🎯 Changes

Extends `@tanstack/electric-db-collection` SQL compilation so multi-segment IR (intermediate representation) refs (nested property paths) work in `WHERE` and `ORDER BY` instead of throwing. Paths compile to PostgreSQL JSON / jsonb operators (`->` for intermediate keys, `->>` for the final key so the compared value is text).

**Scope / limitations:** nested paths assume JSON / jsonb extraction from a single root column; if the physical column isn’t JSON / jsonb, SQL may fail at runtime.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/.changeset/electric-sql-nested-json-refs.md
+++ b/.changeset/electric-sql-nested-json-refs.md
@@ -1,0 +1,11 @@
+---
+'@tanstack/electric-db-collection': patch
+---
+
+Support nested property refs in WHERE/ORDER BY by compiling to PostgreSQL JSON/`jsonb` operators (`->` / `->>`).
+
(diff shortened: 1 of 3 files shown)


KyleAMathews:
Closing because this capability cannot be implemented solely in the TanStack Electric collection compiler. The generated nested-path filter uses PostgreSQL JSON operators (`->` / `->>`), but Electric's Shape server validates and evaluates WHERE expressions with its own expression engine, which does not support those JSON extraction operators. We are not planning the required Electric server work.\n\nThe current candidate also cannot safely land independently: final `->>` extraction coerces JSON numbers to text, changing numeric filtering and ordering semantics. Thanks @miguelrk for exploring the compiler side and documenting the intended behavior.

Issue #126, opened 2025-05-30
Add helpers for fine grained reactivity in React from useLiveQuery


thruflo:
Eg:

- custom <For /> component
- or useLazyLiveQuery
- or in the sub component a custom hook that you use to trigger re-rendering when the sub-item changes

As per Legend State, etc.

KyleAMathews:
This was specifically for React right @thruflo ? https://legendapp.com/open-source/state/v2/react/fine-grained-reactivity/

thruflo:
Yes.

Issue #50, opened 2025-05-06
Optional lazy results for the useLiveQuery hook


samwillis:
Like the non-lazy version, but only triggers a re-render when the length of the results change, or a key moves/changes.
You then pass the result item through to a sub component and `useLazyItem()` which triggers just that component to re-render when the item itself changes.

It's a bit like memoizing the sub components, but without even the outer component rendering the list at all for small changes.

PR #1515, opened 2026-05-05
[DB] feat: add keyed collection subscriptions

lalitkapoor:
## 🎯 Changes

Adds `collection.subscribeKeyChanges(key, callback)` as a fine-grained subscription primitive for future changes to a single collection key. This lets callers pair `collection.get(key)` with a keyed subscription without creating a live query for row-level rendering.

The implementation reuses `CollectionSubscription` lifecycle behavior, registers keyed subscribers before sync starts so synchronous sync writes are not missed, and emits only changes matching the subscribed key.

## ✅ Checklist

- [ ] I have tested this code locally with `pnpm test`.

Tested with:

- `pnpm --filter @tanstack/db exec vitest --run tests/collection-subscribe-changes.test.ts tests/collection.test-d.ts --typecheck.enabled --typecheck.ignoreSourceErrors`

## 🚀 Release Impact

### DIFF
--- a/.changeset/chatty-files-hear.md
+++ b/.changeset/chatty-files-hear.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+Add `subscribeKeyChanges` for subscribing to future changes for a single collection key.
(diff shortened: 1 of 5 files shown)


KyleAMathews:
Closing this stale and conflicting draft implementation. The fine-grained and keyed subscription feature requests remain open in #50 and #126. A future API should start from the current shared observer and collection lifecycle contracts rather than rebasing this pre-observer branch.

PR #1516, opened 2026-05-05
fix(react-db): refresh live query snapshot after subscribe

lalitkapoor:
## 🎯 Changes

Fixes a `useLiveQuery` subscription timing bug where the client could fail to see rows that were already in the collection.

The failure mode was:

1. `useLiveQuery` read a render-time snapshot while the collection was still empty.
2. Rows were written into the collection before React attached the `useSyncExternalStore` subscription. In our app, this is the shape of local/persisted rows hydrating while upstream sync is still pending.
3. Because the collection status was still `loading`, the hook skipped its post-subscribe refresh.
4. React kept rendering the stale empty snapshot, even though the collection had rows available. The client did not see those rows until a later collection change or ready transition happened.

The fix is to force one snapshot refresh immediately after subscribing regardless of collection status. That makes rows written during the render-to-subscribe gap visible while `isLoading` can still remain true for upstream sync.

Adds a regression test with a real collection that writes a row through sync callbacks during that gap and verifies `useLiveQuery` returns the row while still loading.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/.changeset/bright-beds-reflect.md
+++ b/.changeset/bright-beds-reflect.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/react-db': patch
+---
+
+Refresh live query snapshots immediately after subscribing, even when the collection is still loading.
--- a/packages/react-db/src/useLiveQuery.ts
+++ b/packages/react-db/src/useLiveQuery.ts
@@ -439,11 +439,13 @@ export function useLiveQuery(
         versionRef.current += 1
         onStoreChange()
       })
-      // Collection may be ready and will not receive initial `subscribeChanges()`
-      if (collectionRef.current.status === `ready`) {
-        versionRef.current += 1
-        onStoreChange()
-      }
+      // The collection may have changed between render-time getSnapshot() and
+      // this subscription being attached. This is common when an on-demand query
+      // hydrates local rows immediately but keeps the collection loading while
+      // remote sync finishes. Force one post-subscribe snapshot refresh
+      // regardless of status so React sees those rows.
+      versionRef.current += 1
+      onStoreChange()
       return () => {
         subscription.unsubscribe()
       }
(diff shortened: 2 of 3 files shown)


kevin-dp:
Hi @lalitkapoor,

Thank you for opening this PR.
I like how small the change is, but i do believe it has a number of implications we need to carefully consider (numbers 1 and 2 below). This probably requires some more discussion with @samwillis and @KyleAMathews 

Here are the concerns:

**1. Extra render on every subscribe (minor performance).**
Previously, mounting a hook against a `loading` collection skipped the refresh. Now every mount/dep-change triggers an extra `onStoreChange` → `getSnapshot` → render. Bounded (one per subscribe), almost certainly fine, but worth being aware of in components that mount many `useLiveQuery` hooks.

**2. Unconditional bump even when nothing changed.**
If no rows landed during the render→subscribe gap, the bump still forces a re-render with an identical data set (new `snapshotRef` object, same content). React's reconciliation will short-circuit downstream, but consumers using `result` as a dependency in `useEffect` / `useMemo` could see a referentially-different object even when nothing changed. Probably acceptable, but a behavior change worth calling out in the changeset.

**3. Test fragility — write-during-render.**
The test performs `syncBegin/syncWrite/syncCommit` *inside the render callback* gated by a `didWriteAfterSnapshot` flag. This works because `renderHook` invokes the function during render (before effects/subscribe), but it's an unusual pattern that depends on React Testing Library's internals. Two suggestions:

- Consider doing the write inside a `useLayoutEffect` of a child or via a more explicit "after render, before commit" hook — or at minimum, comment that the placement is load-bearing.
- The flag approach risks confusing future maintainers; a comment like "write must happen during render — moving this to an effect breaks the repro" would help.

**4. Test coverage gap.**
The test verifies the row appears, but doesn't verify the *absence* of regressions for the existing path:
- A collection that's already `ready` at subscribe time — does it still get exactly one refresh (not two)?
- A subsequent `syncWrite` after the initial refresh — does normal change propagation still work?

Both are likely fine given how minimal the change is, but a one-line follow-up assertion would lock this in.

KyleAMathews:
Closing as superseded by the shared live-query observer work in #1642 and current React external-store coverage. The historical render-to-subscribe concern was valid, but the current snapshot reread closes that gap without the unconditional callback introduced here. Please open a fresh current-main reproduction if a gap remains.

PR #1517, opened 2026-05-05
fix(db): a key in BasicIndex/BTreeIndex lives in at most one bucket

kevin-dp:
## Summary

Both `BasicIndex` and `BTreeIndex` could leave a key in two buckets at once whenever `update`/`remove` was called with an `oldItem` that evaluated to a value other than the one the index actually had recorded for that key — `remove` would silently no-op on the bucket map while the subsequent `add` placed the key in a new bucket. Calling `add` twice for the same key (e.g. when an upstream "update" change message arrives without a `previousValue`, so the collection treats it as an insert) had the same effect.

This is the underlying cause of the user-reported "stale auto-index after optimistic→synced update on the indexed column" bug, where a kanban card snaps back to its old column because the auto-index still has the row in the old stage's bucket.

The fix: each index now tracks every key's currently-indexed value internally and uses that on `remove`/`update` instead of trusting the caller-supplied `oldItem`. `add` also drops the key from its current bucket first if it's already indexed.

The first commit adds a regression test that demonstrates the invariant violation (and fails on `main`); the second commit lands the fix.

## Test plan

- [x] New `tests/index-key-bucket-tracking.test.ts` fails on `main` and passes after the fix
- [x] All existing `@tanstack/db` index, auto-index, subscribe-changes tests still pass
- [x] Full `@tanstack/db` + `@tanstack/offline-transactions` + `@tanstack/electric-db-collection` suites still pass locally

### DIFF
--- a/.changeset/fix-index-stale-bucket.md
+++ b/.changeset/fix-index-stale-bucket.md
@@ -1,0 +1,9 @@
+---
+'@tanstack/db': patch
+---
+
+fix(db): prevent indexes from leaving a key in two buckets at once
+
+`BasicIndex` and `BTreeIndex` now track each key's currently-indexed value internally and use that on `remove`/`update`, instead of trusting the caller-supplied `oldItem`. If `oldItem` evaluated to a value that didn't match the bucket the key actually lived in, the previous implementation silently no-oped on the bucket map while the subsequent `add` placed the key in a new bucket, leaving the key in two buckets simultaneously. Calling `add` for a key that's already indexed also now drops the previous bucket first, so a missed `remove` can't double-bucket the key.
+
+This was the underlying cause of stale auto-index buckets after an optimistic→synced update on the indexed column (e.g. a kanban card snapping back to its old column).
--- a/packages/db/src/indexes/basic-index.ts
+++ b/packages/db/src/indexes/basic-index.ts
@@ -54,6 +54,11 @@ export class BasicIndex<
   private sortedValues: Array<any> = []
   // Set of all indexed PKs
   private indexedKeys = new Set<TKey>()
+  // Reverse map: key -> currently indexed value. Lets `remove`/`update`
+  // find the bucket a key actually lives in instead of trusting the
+  // caller-supplied old item, which can be stale across optimistic→synced
+  // lifecycle races.
+  private keyToValue = new Map<TKey, any>()
   // Comparator function
   private compareFn: (a: any, b: any) => number = defaultComparator
 
@@ -88,6 +93,13 @@ export class BasicIndex<
 
     const normalizedValue = normalizeValue(indexedValue)
 
+    // If the key is already indexed, drop it from its current bucket first so
+    // an `add` after a stale or missed `remove` can't leave the key in two
+    // buckets at once.
+    if (this.indexedKeys.has(key)) {
+      this.removeFromBucket(key, this.keyToValue.get(key))
+    }
+
     if (this.valueMap.has(normalizedValue)) {
       // Value already exists, just add the key to the set
       this.valueMap.get(normalizedValue)!.add(key)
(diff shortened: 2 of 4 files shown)


KyleAMathews:
I reviewed this alongside #1691 and the stacked follow-up in https://github.com/K-Mistele/db/pull/1. This PR has the right core invariant, but it can be made both faster and more thoroughly tested once that stack lands.

## Suggested implementation after #1691

Treat the reverse `keyToValue` map as the authoritative old indexed value:

1. Evaluate and normalize `newItem` before mutating any state.
2. Read the current normalized value from `keyToValue`; do not trust or evaluate caller-supplied `oldItem` for a tracked key.
3. Return early only when the recorded value matches the new value and the key is present in the expected bucket/key set.
4. For a real move, remove the key directly from its recorded bucket, add it to the new bucket, update `keyToValue`, and update the timestamp once.
5. Make repeated `add(key, item)` remove the key from its recorded bucket first.
6. Make `remove(key, item)` use the recorded bucket when available.
7. Maintain `keyToValue` in `build()` and `clear()`.

This fits #1691's direct bucket helpers and avoids the old public `remove()` + `add()` update path. It should also remove one expression evaluation from every tracked update: only `newItem` needs evaluation. The evaluator cache and primitive normalization fast path in the stacked follow-up reduce the remaining work.

Evaluate `newItem` before removing the old bucket so an evaluation failure can leave the existing index entry intact. That is safer than partially removing the key and then throwing, but it changes the old fallback semantics and should be explicit in the tests.

The cost remains one extra map entry per indexed key, per index. That memory trade-off buys an authoritative one-key/one-bucket invariant and avoids scanning buckets.

## Property-test expansion

The stacked follow-up adds model-based property tests for both `BasicIndex` and `BTreeIndex`. They currently generate valid add/update/delete sequences and compare, after every action:

- `keyCount` and the indexed key set
- exact value buckets and ordered entries
- equality and range queries
- rebuilt index state

For this PR, extend the action generator with:

- updates whose supplied `oldItem` is stale
- repeated `add()` calls for an existing key
- removals whose supplied item is stale
- expression-evaluation failures

After every action, also assert directly that each modeled key occurs in exactly one bucket and that its bucket agrees with the reference model. Keep the optimistic-to-synced collection regression in this PR as the end-to-end proof.

## Interaction with the stacked follow-up

The follow-up currently has tests that pin the legacy evaluator-failure fallback and a narrow `BasicIndex` repair for bucket membership disagreeing with `indexedKeys`. Those tests will need revision once tracked removals no longer evaluate stale or throwing `oldItem` values. The extra no-op membership checks can remain as cheap defense-in-depth.

## Landing

My plan is to pick this up after #1691 and https://github.com/K-Mistele/db/pull/1 land, then adapt this branch with normal follow-up commits. That keeps the performance work and the one-key/one-bucket correctness change independently reviewable while letting this PR use the new property-test and update infrastructure.

MILLERMARRU (commented):
I read through basic-index.ts on main and the bug is a straightforward trust issue: `remove()` re-derives the bucket to clean up by evaluating the `item` argument the caller hands it, rather than tracking what value was actually indexed for that key. Since collections call `remove`/`update` with whatever `oldItem` a transaction snapshot happens to hold, an optimistic-to-synced race can pass a stale `oldItem` and the index silently no-ops on the wrong bucket while a subsequent `add` inserts the key under a new value, leaving it duplicated across two buckets. The `keyToValue` reverse map sidesteps the whole class of bug by making the index the source of truth for its own state instead of re-deriving it from caller input every time. The test `update moves the key out of its actual bucket, regardless of the oldItem passed` is a precise repro of exactly this failure mode, and the kanban-card scenario in the changeset (optimistic drag-and-drop snapping back) is a believable real-world trigger.

PR #1519, opened 2026-05-07
fix(db): support Temporal values in query comparison operators

obeattie:
## 🎯 Changes

Fixes a `TypeError` when query filters use comparison operators (`gt`/`gte`/`lt`/`lte`) on Temporal values. The evaluators applied JS's native `>`/`<` directly, which throws on Temporal types because their `valueOf()` is [designed to throw](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/ZonedDateTime/valueOf) — a guard against silent miscomparison.

Adds a `compareValues(a, b)` helper that dispatches to the Temporal types' static `.compare()` when both operands share a type. Mixed Temporal types and types with no defined ordering (`PlainMonthDay`) throw a descriptive `TypeError` rather than fall back to string-lex comparison. Non-Temporal values continue to use native operators.

`eq` is unchanged — it still goes through `normalizeValue`, so `ZonedDateTime` equality treats the zone as part of identity while ordering compares by instant only, mirroring `.equals()` vs `.compare()` in the spec.

Tests cover all eight Temporal types, the same-instant/different-zone asymmetry for `ZonedDateTime`, and `Duration`'s equivalent-forms case (`PT60M` vs `PT1H`: equal under `.compare()` but not `.equals()`).

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Added semantic, type-aware ordering for Temporal values in comparisons and sorting.
* **Bug Fixes**
  * Fixed `gt`, `gte`, `lt`, `lte`, and `orderBy` to use Temporal-native ordering semantics for consistent filtering and sorting (including `Duration` and `ZonedDateTime` equivalence by value/instant).
  * Ordering now throws a `TypeError` for unorderable Temporal types (e.g., `PlainMonthDay`) and for mixed Temporal types.
* **Tests**
  * Expanded Temporal comparison and edge-case coverage, including stable behavior for `NaN`.

### DIFF
--- a/.changeset/temporal-comparators.md
+++ b/.changeset/temporal-comparators.md
@@ -2,4 +2,11 @@
 '@tanstack/db': patch
 ---
 
-Support Temporal values in `gt`/`gte`/`lt`/`lte` query operators, using `Temporal.*.compare()` for correct semantic ordering. `orderBy` now uses the same logic, fixing inconsistencies with `Duration` and `ZonedDateTime` values.
+Support Temporal values in the `gt`/`gte`/`lt`/`lte` query operators. Comparisons now dispatch to the Temporal types' static `compare()` instead of the native relational operators, which throw on Temporal objects (`valueOf()` is designed to throw). `orderBy` uses the same logic, so filtering and ordering now agree — previously `orderBy` compared Temporal values lexicographically by their `toString()`, which mis-ordered equivalent `Duration` forms (`PT60M` vs `PT1H`) and same-instant `ZonedDateTime` values in different zones.
+
+Note two intentional behavior changes for `orderBy` over Temporal columns:
+
+- Ordering `Temporal.PlainMonthDay` values now throws a `TypeError`, since the type has no defined ordering (previously they were silently ordered by string).
+- Ordering mixed Temporal types (e.g. `PlainDate` vs `PlainDateTime`) now throws a `TypeError` instead of comparing their string forms.
+
+Equality (`eq`) is unchanged: `ZonedDateTime` equality still treats the zone as part of identity, and equivalent `Duration` forms remain unequal, mirroring Temporal's `.equals()` vs `.compare()` semantics.
--- a/.changeset/temporal-comparators.md
+++ b/.changeset/temporal-comparators.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/db': patch
+---
+
+Support Temporal values in `gt`/`gte`/`lt`/`lte` query operators, using `Temporal.*.compare()` for correct semantic ordering. `orderBy` now uses the same logic, fixing inconsistencies with `Duration` and `ZonedDateTime` values.
--- a/packages/db/src/utils/comparison.ts
+++ b/packages/db/src/utils/comparison.ts
@@ -253,9 +253,9 @@ const temporalCompareByTag = new Map<
  *
  * Callers must ensure both arguments are Temporal objects; mixed types throw.
  */
-export function compareTemporalValues(a: any, b: any): number {
-  const aTag = a[Symbol.toStringTag] as string
-  const bTag = b[Symbol.toStringTag] as string
+export function compareTemporalValues(a: unknown, b: unknown): number {
+  const aTag = (a as Record<symbol, unknown>)[Symbol.toStringTag] as string
+  const bTag = (b as Record<symbol, unknown>)[Symbol.toStringTag] as string
   if (aTag !== bTag) {
     throw new TypeError(
       `Cannot order Temporal values of different types: ${aTag} vs ${bTag}`,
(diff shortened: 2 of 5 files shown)


packages/db/src/utils/comparison.ts
```diff
@@ -212,6 +212,41 @@ export function denormalizeUndefined(value: any): any {
...
+        `Cannot order Temporal values of different types: ${aTag} vs ${bTag}`,
+      )
+    }
+    const compare = (
+      a.constructor as { compare?: (x: unknown, y: unknown) => number }
+    ).compare
+    if (typeof compare !== `function`) {
+      throw new TypeError(`${aTag} has no defined ordering`)
+    }
+    return compare(a, b)
```
kevin-dp: Let's extract this into a `compareTemporalValues` function in order to keep this `compareValues` function clean.

kevin-dp (changes requested):
Thank you for your PR. I had a look through it and have some concerns.

## Main concern — inconsistency with `orderBy`

The codebase now has two different orderings for Temporal values:

- Filter operators (`gt`/`lt`/…) → new `compareValues` → Temporal `.compare()` (by instant/value).
- `ORDER BY` → `ascComparator` in `comparison.ts:59-65` → lexicographic `a.toString()` comparison.

These diverge in observable ways:

| Case | `WHERE` (new) | `ORDER BY` (existing) |
|---|---|---|
| `Duration` `PT1H` vs `PT60M` | equal | `"PT1H" < "PT60M"` (wrong) |
| `ZonedDateTime` same instant, diff zone | equal | ordered by zone-name string |
| `PlainMonthDay` | throws | lex-compares fine |
| mixed Temporal types | throws | lex-compares fine |

A user could write `.where(d => gt(d.dur, oneHour))` and `.orderBy(d => d.dur)` over the same column and get inconsistent semantics — and a `PlainMonthDay` `orderBy` "works" while a `PlainMonthDay` filter throws. Reusing `compareValues` inside `ascComparator` would unify them (and incidentally fix the lexicographic Duration bug).

## Other concerns

1. `gte`/`lte` behavior change for incomparable primitives (`comparison.ts`, the `a < b ? -1 : a > b ? 1 : 0` fallback). When neither `<` nor `>` holds but values aren't truly equal, this returns `0`, so `gte`/`lte` now return `true` where the old native `a >= b` returned `false`. The concrete case is `NaN`: previously `gte(NaN, NaN) === false`, now `true` (since `isUnknown` filters only null/undefined, not NaN). `gt`/`lt` are unaffected. Not covered by the tests — worth a guard or at least a test pinning the behavior.

2. `a.constructor.compare` dispatch works in practice (tests pass on `temporal-polyfill`, and native Temporal works too), but it's mildly fragile across realms / if `constructor` is shadowed. Dispatching off the already-computed `Symbol.toStringTag` would be more robust and consistent with how `isTemporal` and `areValuesEqual` already brand-check.

3. Temporal-vs-non-Temporal operand (e.g. `gt(plainDate, "2024-01-15")`) falls into the native-operator branch and throws an opaque `valueOf` `TypeError`. Same as before this PR, so no regression — but if you're already producing descriptive errors here, it'd be a natural one to catch too.

4. This PR is missing a changeset

obeattie:
@kevin-dp Thanks! I've made those changes so would appreciate another look.

.changeset/temporal-comparators.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tanstack/db': patch
+---
+
+Support Temporal values in `gt`/`gte`/`lt`/`lte` query operators, using `Temporal.*.compare()` for correct semantic ordering. `orderBy` now uses the same logic, fixing inconsistencies with `Duration` and `ZonedDateTime` values.
```
kevin-dp: Can we extend the changeset a bit to acknowledge the behavioral change. Something like this could work:

```
---
'@tanstack/db': patch
---

Support Temporal values in the `gt`/`gte`/`lt`/`lte` query operators. Comparisons now dispatch to the Temporal types' static `compare()` instead of the native relational operators, which throw on Temporal objects (`valueOf()` is designed to throw). `orderBy` uses the same logic, so filtering and ordering now agree — previously `orderBy` compared Temporal values lexicographically by their `toString()`, which mis-ordered equivalent `Duration` forms (`PT60M` vs `PT1H`) and same-instant `ZonedDateTime` values in different zones.

Note two intentional behavior changes for `orderBy` over Temporal columns:

- Ordering `Temporal.PlainMonthDay` values now throws a `TypeError`, since the type has no defined ordering (previously they were silently ordered by string).
- Ordering mixed Temporal types (e.g. `PlainDate` vs `PlainDateTime`) now throws a `TypeError` instead of comparing their string forms.

Equality (`eq`) is unchanged: `ZonedDateTime` equality still treats the zone as part of identity, and equivalent `Duration` forms remain unequal, mirroring Temporal's `.equals()` vs `.compare()` semantics.

```

kevin-dp (approved):
Hi @obeattie,

Thanks for addressing my review!
The PR is good to go now. Could you just extend the changeset according to my comment please.

.changeset/temporal-comparators.md
obeattie: Sure, done ✅

kevin-dp (approved):

PR #1520, opened 2026-05-07
fix(db): avoid passing computed orderBy aliases to loadSubset

andrewcoelho:
## 🎯 Changes

TLDR: Fixes incorrect `loadSubset` optimization hints when a subquery orders by a computed selected field e.g. `sortKey: coalesce(...)`.

I ran into this when trying to use an `orderBy` field that is a projected/computed alias. That was causing subset loading to use an invalid ordering hint, so the top-N window is built from the wrong subset and I observed query results that looked like limit was being applied before orderBy.

This change makes ref resolution stop at computed select expressions (only pass-through `ref` projections are followed), and only forwards `orderBy` hints when each `orderBy` ref resolves to a real source field.

- Updated `followRef` in `packages/db/src/query/ir.ts` to return `undefined` for computed selected fields.
- Applied the same `followRef` fix in `packages/db/src/query/compiler/index.ts` to keep compiler behavior consistent.
- Tightened `computeSubscriptionOrderByHints` in `packages/db/src/query/live/utils.ts` to validate each `orderBy` ref via `followRef` before passing `orderBy`/`limit` hints to `loadSubset`.
- Added a regression test in `packages/db/tests/query/load-subset-subquery.test.ts` to ensure computed subquery `orderBy` aliases do not forward `orderBy`/`limit` hints.

ETA sample query that shows the issue:

```ts
const pagedQuery = useLiveQuery(
  (q) => {
    // 1) Aggregate child records per parent (derived subquery)
    const latestChildByParent = q
      .from({ child: childCollection })
      .groupBy(({ child }) => child.parentId)
      .select(({ child }) => ({
        parentId: child.parentId,
        latestAt: max(child.createdAt),
      }))

    // 2) Join aggregate back to parent and compute a projected sort alias
    const parentsWithComputedSort = q
      .from({ parent: parentCollection })
      .join(
        { latest: latestChildByParent },
        ({ parent, latest }) => eq(parent.id, latest.parentId),
      )
      .select(({ parent, latest }) => ({
        id: parent.id,
        // computed/projection field (not a raw source column)
        computedSortKey: coalesce(latest.latestAt, parent.createdAt),
      }))

    // 3) Filter + order by computed alias + paginate
    return q
      .from({ row: parentsWithComputedSort })
      .where(({ row }) => gt(row.computedSortKey, anchorDate))
      .orderBy(({ row }) => row.computedSortKey, 'asc')
      .select(({ row }) => ({
        id: row.id,
        computedSortKey: row.computedSortKey,
      }))
      .limit(3)
  },
)

// Expected (ascending by computedSortKey):
// A
// B
// C

// Actual:
// A
// D
// E

// If I increase pageSize to include all rows:
// A
// B
// C
// D
// E
// ...
```

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/.changeset/fix-load-subset-order-by-hints.md
+++ b/.changeset/fix-load-subset-order-by-hints.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/db': patch
+---
+
+Fix `loadSubset` orderBy hints for subqueries with computed selected fields. We now stop ref-following on non-ref select expressions and only pass `orderBy`/`limit` hints when each `orderBy` ref resolves to a real source field.
--- a/packages/db/src/query/compiler/index.ts
+++ b/packages/db/src/query/compiler/index.ts
@@ -1107,15 +1107,17 @@ export function followRef(
     // is it part of the select clause?
     if (query.select) {
       const selectedField = query.select[field]
-      if (selectedField && selectedField.type === `ref`) {
-        return followRef(query, selectedField, collection)
+      if (selectedField) {
+        // Only pass-through refs can be followed to a raw collection field.
+        // Computed select expressions do not map to a source column.
+        if (selectedField.type === `ref`) {
+          return followRef(query, selectedField, collection)
+        }
+        return
       }
     }
 
-    // Either this field is not part of the select clause
-    // and thus it must be part of the collection itself
-    // or it is part of the select but is not a reference
-    // so we can stop here and don't have to follow it
+    // Field is not projected by select, so it belongs to the collection itself.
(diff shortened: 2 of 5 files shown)


KyleAMathews:
Closing as superseded per RFC #1657 Phase 0 cleanup. The computed orderBy alias bug no longer reproduces on current main: the PR regression case passes without the production patch, and the alias correctly withholds orderBy/limit pushdown. The regression test remains useful and can be preserved separately as a test-only addition. Thank you for investigating and contributing the fix.

PR #1522, opened 2026-05-11
test(transactions): demonstrate ambient tx loss across await in mutate()

marius-se:
## 🎯 Changes

Adds a test that demonstrates a mismatch between `Transaction#mutate`'s documented contract and its implementation for async callbacks.

### The contract

The docstring on `Transaction#mutate` (`packages/db/src/transactions.ts:248-251`) states:

> If the callback returns a Promise, the transaction context will remain active until the promise settles, allowing optimistic writes after `await` boundaries.

### What actually happens

`mutate()` invokes the callback synchronously and pops the transaction off the ambient stack in a `finally` block (`packages/db/src/transactions.ts:287-298`):

```ts
mutate(callback: () => void): Transaction<T> {
  // ...
  registerTransaction(this)
  try { callback() }                  // not awaited
  finally { unregisterTransaction(this) }
  // ...
}
```

When the callback is `async`, `callback()` returns its Promise at the first `await`, the `finally` fires immediately, and `unregisterTransaction(this)` removes the transaction from `transactionStack` before any post-`await` code runs. Collection operations after the await see no ambient transaction and either throw `MissingInsertHandlerError` (no `onInsert`) or create a brand-new direct transaction via the collection's handler.

Additional signals that the async path is unsupported today:

- The callback parameter is typed `() => void`, not `() => void | Promise<void>`.
- No existing test in `packages/db/tests/` exercises an async `mutate()` callback. The async tests in `transactions.test.ts` cover `mutationFn` and `commit()`, not the `mutate()` callback.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/db/tests/transactions.test.ts
+++ b/packages/db/tests/transactions.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it } from 'vitest'
+import { describe, expect, it, vi } from 'vitest'
 import { createTransaction } from '../src/transactions'
 import { createCollection } from '../src/collection/index.js'
 import {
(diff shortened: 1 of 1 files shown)


KyleAMathews:
updated docs in https://github.com/TanStack/db/pull/1538

PR #1529, opened 2026-05-17
feat(db): add caseWhen query operator

samwillis:
## Summary
- Adds `caseWhen(...)` as a unified conditional query helper inspired by SQL `CASE WHEN`, with scalar expression support in `select`, `where`, `orderBy`, `groupBy`, `having`, and equality join operands.
- Adds projection-valued `caseWhen` support for `select()` branch values, including nested projection objects, ref spreads, `toArray(...)`, and Collection includes.
- Implements guarded include routing so includes inside inactive `caseWhen` branches are not materialized for that parent row.
- Documents `caseWhen` in the live query guide and adds a changeset for the new operator.

## Why This Is Useful
`caseWhen` lets queries express conditional logic without dropping down to functional callbacks, so the query compiler can keep optimizing and incrementally maintaining the result.

A common scalar use case is categorizing rows while preserving literal result types:

```ts
const userCategories = createLiveQueryCollection((q) =>
  q
    .select(({ user }) => ({
      id: user.id,
      category: caseWhen(
        gt(user.age, 65),
        'senior',
        gt(user.age, 18),
        'adult',
        'minor',
      ),
    }))
)
```

Because scalar `caseWhen` is a normal query expression, it can also drive filtering, grouping, sorting, and join operands:

```ts
const activeFirst = createLiveQueryCollection((q) =>
  q
    .where(({ user }) => caseWhen(eq(user.deleted, true), false, true))
    .orderBy(({ user }) => caseWhen(eq(user.active, true), 0, 1))
    .select(({ user }) => ({
      id: user.id,
      name: user.name,
    }))
)
```

Projection-valued branches make optional result shapes easier to model. For example, a joined row can project an optional nested object only when the joined source exists:

```ts
const usersWithOptionalPost = createLiveQueryCollection((q) =>
  q
    .leftJoin({ post: postsCollection }, ({ user, post }) =>
      eq(user.id, post.userId),
    )
    .select(({ user, post }) => ({
      id: user.id,
      post: caseWhen(post, {
        title: post.title,
      }),
    }))
)
```

It also enables guarded includes, where child collections are materialized only for rows whose branch is active:

```ts
const adultProfiles = createLiveQueryCollection((q) =>
  q
    .select(({ user }) => ({
      id: user.id,
      adultProfile: caseWhen(gt(user.age, 18), {
        ...user,
        posts: q
          .from({ post: postsCollection })
          .where(({ post }) => eq(post.userId, user.id))
          .select(({ post }) => ({
            title: post.title,
          })),
      }),
    }))
)
```

## Design
`caseWhen` is one public API with two internal paths selected by branch value shape, not call location:

- When every branch/default value is expression-like, the builder lowers to a scalar `Func('caseWhen', ...)`. The evaluator walks condition/value pairs left to right, returns the first matching branch, evaluates only the active value branch, and returns `null` when there is no default.
- When any branch/default value is select-only, the builder returns a `CaseWhenWrapper` that is lowered by `select()` into `ConditionalSelect` IR. The select compiler evaluates branch conditions lazily and returns `undefined` when there is no matching projection branch and no default.
- Includes under `ConditionalSelect` branches carry guard metadata through include extraction/routing. Parent keys are emitted only when the guard chain matches, which keeps inactive branch includes from loading or attaching.

Full design notes: https://gist.github.com/samwillis/60d8d0d937cf23267d0f8383bea3af92

## Test plan
- `pnpm exec vitest --run tests/query/case-when.test.ts tests/query/case-when.test-d.ts` from `packages/db`
- `pnpm test` from `packages/db`
- `pnpm test` from the repo root was also run; `packages/db` passed, but the workspace run failed in unrelated `@tanstack/expo-db-sqlite-persistence` type-checking for untracked `tests/helpers/expo-runtime-node-preload.cjs`.

## Summary by CodeRabbit

* **New Features**
  * Added a conditional expression operator for queries that evaluates multiple conditions and returns matching scalar or projection results; supports an optional default.
  * Conditional projections can include guarded includes that are materialized only when their branch conditions pass, and child include values can be written into nested result paths.
* **Documentation**
  * Added guide and reference for using conditional expressions.
* **Tests**
  * New test suites covering conditional expressions across queries, joins, grouping, and includes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/db/pull/1529?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


samwillis:
Closing in favor of a replacement PR from the upstream TanStack/db branch so stacked PRs can use upstream branches directly.

PR #1531, opened 2026-05-17
feat(db): support multi-source from queries

samwillis:
## Stack

This PR is stacked on #1529 (`caseWhen`). Please review/merge #1529 first.

Because #1529 is a cross-repository PR from `samwillis:case-when`, GitHub does not allow this upstream PR to use that fork branch as its base. Until #1529 merges, this PR's diff may include the `caseWhen` commits as ancestry. The multi-source-specific commits start after #1529's head (`a6d86d04`).

Implementation plan: https://gist.github.com/samwillis/df7e2376c6c2bfd007e53dec797fca12

## Summary

This PR adds multi-source `from` support to the DB query engine. It lets a query union multiple independent sources into one live query result without requiring an explicit join between those sources.

## What Changed

- Adds a `UnionFrom` IR node for `from({ a, b, ... })` with multiple sources.
- Updates the query builder so `from` accepts multiple collection or subquery sources, while joins still accept one source.
- Compiles each source independently, namespaces rows by alias, prefixes row keys by source alias to avoid collisions, and concatenates the source streams.
- Preserves no-select result typing as an exclusive union, so only the active branch is required on each row.
- Supports subquery branches, including pre-filtered and joined subqueries.
- Keeps global `where` semantics after the union, while allowing users to pre-filter a branch by making it a subquery.
- Updates optimizer/live-query traversal so all union branches are discovered for subscriptions, aliases, docs, and collection extraction.
- Handles joins after multi-source `from`, including branch-dependent join keys via `coalesce`.
- Teaches join lazy loading to expand safe union-backed subquery projections into concrete per-branch lazy loads, while deopting computed projections and limited/offset branches.
- Documents multi-source `from` and ordering behavior in the live queries guide.

## Examples

Basic multi-source query:

```ts
const feed = createLiveQueryCollection((q) =>
  q
    .from({
      message: messages,
      toolCall: toolCalls,
    })
    .orderBy(({ message, toolCall }) =>
      coalesce(message.timestamp, toolCall.timestamp),
    ),
)
```

Without `select`, the result is an exclusive union:

```ts
type FeedRow =
  | { message: Message; toolCall?: undefined }
  | { message?: undefined; toolCall: ToolCall }
```

Pre-filter one branch with a subquery:

```ts
const feed = createLiveQueryCollection((q) => {
  const visibleMessages = q
    .from({ message: messages })
    .where(({ message }) => eq(message.kind, `visible`))

  return q.from({
    message: visibleMessages,
    toolCall: toolCalls,
  })
})
```

Join after a multi-source `from` with branch-dependent keys:

```ts
const feedWithUsers = createLiveQueryCollection((q) =>
  q
    .from({
      message: messages,
      toolCall: toolCalls,
    })
    .join({ user: users }, ({ message, toolCall, user }) =>
      eq(coalesce(message.userId, toolCall.userId), user.id),
    )
    .select(({ message, toolCall, user }) => ({
      userName: user.name,
      timestamp: coalesce(message.timestamp, toolCall.timestamp),
    })),
)
```

Guard branch-specific projections and includes with `caseWhen` from #1529:

```ts
const feed = createLiveQueryCollection((q) =>
  q
    .from({ message: messages, toolCall: toolCalls })
    .select(({ message, toolCall }) => ({
      message: caseWhen(message.id, {
        id: message.id,
        text: message.text,
        chunks: toArray(
          q
            .from({ chunk: chunks })
            .where(({ chunk }) => eq(chunk.messageId, message.id))
            .select(({ chunk }) => chunk.text),
        ),
      }),
      toolCall: caseWhen(toolCall.id, {
        id: toolCall.id,
        name: toolCall.name,
      }),
      timestamp: coalesce(message.timestamp, toolCall.timestamp),
    })),
)
```

## Test Plan

- `pnpm vitest run packages/db/tests/query/multi-source-from.test.ts`
- Focused type test file passes with `pnpm vitest --run --typecheck tests/query/multi-source-from.test-d.ts --coverage.enabled=false` from `packages/db`; Vitest still reports unrelated existing `includes.test.ts` type errors in the wider typecheck pass.

## Summary by CodeRabbit

* **New Features**
  * Added caseWhen for conditional expressions and conditional select projections.
  * Added multi-source from() to build union-style queries with namespaced rows and per-branch optional fields.
  * Includes can be conditionally materialized and written to nested result paths; lazy-loading targets handle multi-branch cases.

* **Documentation**
  * Updated guides for caseWhen, multi-source from(), multi-source ordering, and include behavior.

* **Tests**
  * New/expanded test suites covering caseWhen, multi-source from, grouping, ordering, includes, and live/reactivity.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/db/pull/1531?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


samwillis:
Update: pushed another fix for the nested multi-source include streaming path.

What changed:
- Propagates nested include materialization through `QueryRef`/multi-source sources so `concat(toArray(...))` values no longer stay as `null` in child include rows.
- Preserves correct child row keys for inline materialization updates (`text:<id>` rather than `undefined`).
- Uses `ChangeMessage.key` when feeding live-query/effect D2 inputs, avoiding key recomputation from WeakMap-backed include rows.
- Emits cloned next values for inline include materialization updates so downstream live queries do not suppress changes due to shared object identity.
- Carries outer `caseWhen` guards when remapping propagated source includes, and assigns unique internal routing keys to avoid collisions.
- Skips propagated source-include remapping for `fnSelect`, where there is no stable selected output path to materialize into.

Regression coverage was added to `packages/db/tests/query/multi-source-from.test.ts` for the app-shaped `runs -> items -> multi-source runItemsSource -> textContent: concat(toArray(textDeltas))` query. It now asserts direct `run.items` changes have stable keys and that a live query over `run.items` receives incremental streaming updates while inactive `caseWhen` branches remain untouched.

Verified locally:
- `pnpm vitest run packages/db/tests/query/multi-source-from.test.ts`
- `git diff --check`

samwillis:
Closing in favor of a replacement PR from the upstream TanStack/db branch so the stack can use upstream branches directly.

PR #1536, opened 2026-05-18
feat(db): add caseWhen query operator

samwillis:
## Summary
- Adds `caseWhen(...)` as a unified conditional query helper inspired by SQL `CASE WHEN`, with scalar expression support in `select`, `where`, `orderBy`, `groupBy`, `having`, and equality join operands.
- Adds projection-valued `caseWhen` support for `select()` branch values, including nested projection objects, ref spreads, `toArray(...)`, and Collection includes.
- Implements guarded include routing so includes inside inactive `caseWhen` branches are not materialized for that parent row.
- Documents `caseWhen` in the live query guide and adds a changeset for the new operator.

## Why This Is Useful
`caseWhen` lets queries express conditional logic without dropping down to functional callbacks, so the query compiler can keep optimizing and incrementally maintaining the result.

A common scalar use case is categorizing rows while preserving literal result types:

```ts
const userCategories = createLiveQueryCollection((q) =>
  q
    .select(({ user }) => ({
      id: user.id,
      category: caseWhen(
        gt(user.age, 65),
        'senior',
        gt(user.age, 18),
        'adult',
        'minor',
      ),
    }))
)
```

Because scalar `caseWhen` is a normal query expression, it can also drive filtering, grouping, sorting, and join operands:

```ts
const activeFirst = createLiveQueryCollection((q) =>
  q
    .where(({ user }) => caseWhen(eq(user.deleted, true), false, true))
    .orderBy(({ user }) => caseWhen(eq(user.active, true), 0, 1))
    .select(({ user }) => ({
      id: user.id,
      name: user.name,
    }))
)
```

Projection-valued branches make optional result shapes easier to model. For example, a joined row can project an optional nested object only when the joined source exists:

```ts
const usersWithOptionalPost = createLiveQueryCollection((q) =>
  q
    .leftJoin({ post: postsCollection }, ({ user, post }) =>
      eq(user.id, post.userId),
    )
    .select(({ user, post }) => ({
      id: user.id,
      post: caseWhen(post, {
        title: post.title,
      }),
    }))
)
```

It also enables guarded includes, where child collections are materialized only for rows whose branch is active:

```ts
const adultProfiles = createLiveQueryCollection((q) =>
  q
    .select(({ user }) => ({
      id: user.id,
      adultProfile: caseWhen(gt(user.age, 18), {
        ...user,
        posts: q
          .from({ post: postsCollection })
          .where(({ post }) => eq(post.userId, user.id))
          .select(({ post }) => ({
            title: post.title,
          })),
      }),
    }))
)
```

## Design
`caseWhen` is one public API with two internal paths selected by branch value shape, not call location:

- When every branch/default value is expression-like, the builder lowers to a scalar `Func('caseWhen', ...)`. The evaluator walks condition/value pairs left to right, returns the first matching branch, evaluates only the active value branch, and returns `null` when there is no default.
- When any branch/default value is select-only, the builder returns a `CaseWhenWrapper` that is lowered by `select()` into `ConditionalSelect` IR. The select compiler evaluates branch conditions lazily and returns `undefined` when there is no matching projection branch and no default.
- Includes under `ConditionalSelect` branches carry guard metadata through include extraction/routing. Parent keys are emitted only when the guard chain matches, which keeps inactive branch includes from loading or attaching.

Full design notes: https://gist.github.com/samwillis/60d8d0d937cf23267d0f8383bea3af92

## Test plan
- `pnpm exec vitest --run tests/query/case-when.test.ts tests/query/case-when.test-d.ts` from `packages/db`
- `pnpm test` from `packages/db`
- `pnpm test` from the repo root was also run; `packages/db` passed, but the workspace run failed in unrelated `@tanstack/expo-db-sqlite-persistence` type-checking for untracked `tests/helpers/expo-runtime-node-preload.cjs`.

## Summary by CodeRabbit

* **New Features**
  * Added caseWhen() conditional expressions for scalar and object projections; usable in selects, filters, grouping, joins, and nested selects.
  * Conditional selects now compile and evaluate with ordered branch/default semantics.

* **Includes / Materialization**
  * Guarded includes: nested includes are materialized only when their branch guards match; supports nested result paths for placing included results.

* **Aggregation / Grouping**
  * caseWhen() works inside GROUP BY and aggregate-bearing projections.

* **Tests**
  * Added comprehensive type and runtime tests covering branching, nested includes, grouping, and joins.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/db/pull/1536?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/db/src/query/builder/functions.ts
```diff
@@ -351,6 +373,583 @@ export function coalesce<T extends [ExpressionLike, ...Array<ExpressionLike>]>(
...
+  value6: V6,
+  condition7: C7,
+  value7: V7,
+  condition8: C8,
+  value8: V8,
+  condition9: C9,
+  value9: V9,
+  condition10: C10,
+  value10: V10,
+): CaseWhenResult<[V1, V2, V3, V4, V5, V6, V7, V8, V9, V10], false>
```
kevin-dp: Are we exaggerating the overload by going up to 10 values?

packages/db/src/query/builder/types.ts
```diff
@@ -372,6 +390,10 @@ export type SelectResult<TSelect> =
     ? ResultTypeFromSelect<TSelect>
     : ResultTypeFromSelectValue<TSelect>
 
+type ResultTypeFromCaseWhen<T> = T extends unknown
```
kevin-dp, line 395: I find it weird that this type is branching based on whether `T` is `unknown`. Can you explain why this is the case? I find it especially weird that when it is unknown we return a type that wraps it and otherwise we return `never`. I would have expected the opposite (type as `never` when `T` is unknown).

packages/db/src/query/compiler/group-by.ts
```diff
@@ -683,16 +709,186 @@ function extractAndReplaceAggregates(
...
+function isConditionalSelect(value: unknown): value is ConditionalSelect {
+  return (
+    value instanceof ConditionalSelect ||
+    (value != null &&
+      typeof value === `object` &&
+      (value as { type?: string }).type === `conditionalSelect`)
+  )
+}
+
+function isCaseWhenConditionTrue(value: any): boolean {
```
kevin-dp: This function is duplicated from `packages/db/src/query/compiler/evaluators.ts`. Let's not duplicate it.

packages/db/src/query/compiler/index.ts
```diff
@@ -1211,4 +1392,36 @@ function getNestedValue(obj: any, path: Array<string>): any {
   return value
 }
 
+function isCaseWhenConditionTrue(value: any): boolean {
```
kevin-dp: Third time this function is duplicated!

packages/db/src/query/compiler/select.ts
```diff
@@ -139,6 +149,105 @@ export function processSelect(
...
+      if (isCaseWhenConditionTrue(branch.condition(row))) {
+        return branch.value(row)
+      }
+    }
+
+    return defaultValue ? defaultValue(row) : undefined
+  }
+}
+
+function isCaseWhenConditionTrue(value: any): boolean {
```
kevin-dp: And a fourth time...

kevin-dp (changes requested):
I left some comments.
There's quite some duplication. I found 4 copies of the same function spread across the codebase. There are probably other functions that are duplicated too that i didn't spot. Please have an agent go through it to find all duplicated functions and extract them into a single shared function.

packages/db/src/query/builder/functions.ts
samwillis: Dropped to 5 overloads. If we have users needing more, we can add them then. This overload pattern is much easer than a recursive one.

packages/db/src/query/builder/types.ts
samwillis: From GPT 5.5:

I tried removing the `T extends unknown ? ... : never` and replacing it with:

```ts
type ResultTypeFromCaseWhen<T> = ResultTypeFromSelectValue<T>
```

That broke the type test for variadic projection branches. Without the conditional, TypeScript treats the union of branch projection objects as one merged object-ish input to `ResultTypeFromSelectValue`, which collapses the union shape incorrectly.

So I restored it and added the comment at `packages/db/src/query/builder/types.ts` explaining that this is intentionally a distributive conditional type. `T extends unknown` is not checking whether `T` is literally `unknown`; it’s the standard TypeScript pattern to distribute the inner type transform across each member of a union.

In other words:

```ts
A | B -> ResultTypeFromSelectValue<A> | ResultTypeFromSelectValue<B>
```

instead of:

```ts
ResultTypeFromSelectValue<A | B>
```

packages/db/src/query/builder/functions.ts
```diff
@@ -351,6 +373,249 @@ export function coalesce<T extends [ExpressionLike, ...Array<ExpressionLike>]>(
...
+  value3: V3,
+  condition4: C4,
+  value4: V4,
+  condition5: C5,
+  value5: V5,
+  condition6: ExpressionLike,
+  value6: CaseWhenValue,
+  ...rest: Array<CaseWhenValue>
+): any
+export function caseWhen(...args: Array<CaseWhenValue>): any {
```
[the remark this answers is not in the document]
samwillis: Leaving this unchanged intentionally. We reduced the precise overload surface to 1-5 branches in response to review feedback, then added this broad fallback only so uncommon 6+ branch calls continue to type-check. Preserving alternating condition/value validation and precise result inference for arbitrary-length calls requires the recursive variadic tuple helper we deliberately removed because it made the public type surface harder to read than the explicit overloads. Runtime validation still rejects invalid condition slots, and common calls retain precise inference through the fixed overloads.

packages/db/src/query/builder/functions.ts
```diff
@@ -450,8 +716,64 @@ export class ConcatToArrayWrapper<_T = unknown> {
...
+  }
+  if (value instanceof Date || Array.isArray(value)) return true
+  if (typeof value === `object` && `type` in value) {
+    return (
+      value.type === `agg` ||
+      value.type === `func` ||
+      value.type === `ref` ||
+      value.type === `val`
+    )
+  }
```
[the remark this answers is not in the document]
samwillis: I tightened `isExpressionValue()` in `packages/db/src/query/builder/functions.ts` so it no longer treats any object with `type: 'val' | 'func' | 'agg' | 'ref'` as an IR expression. It now checks the expected IR shape, for example `func`/`agg` need `name` and `args`, `ref` needs `path`, and `val` needs a `value` property.

I also tightened the shared `isExpressionLike()` in `packages/db/src/query/ir.ts`, because the same loose structural check affected select compilation.

Added regression coverage in `packages/db/tests/query/case-when.test.ts` for projection objects like:

```ts
{ type: `val`, label: user.name }
```

Committed and pushed as `a746f73e fix(db): tighten caseWhen expression detection`.

kevin-dp (approved):
Thanks for addressing the comments. The type overload with the var args is great and avoids the need for too many overloads 💯

PR #1547, opened 2026-05-22
fix(db): drop stale direct optimistic rows after txid sync

KyleAMathews:
## Summary

Fix stale optimistic rows persisting when sync confirms a different server-generated key (e.g., Postgres `GENERATED ALWAYS AS IDENTITY`). After a direct `collection.insert()` with a temporary client ID, the server-assigned row would appear alongside the stale optimistic row instead of replacing it.

## Root Cause

`CollectionStateManager` maintained separate `pendingOptimisticDirectUpserts` / `pendingOptimisticDirectDeletes` sets that exempted direct transactions from stale-row cleanup. When sync confirmed a row under a different key than the client's temp key, the stale optimistic row could never be cleaned up — no sync transaction would ever reference the old client key, but the direct-transaction exemption kept it alive indefinitely.

## Approach

Remove the direct-transaction exemption entirely. All completed optimistic mutations now go through the same stale-row check: kept only if `pendingSyncKeys.has(key)`. This is safe because direct transactions stay in `persisting` state (keeping the optimistic row visible via the active-transaction loop) until the sync backend delivers the matching txid. By the time the transaction completes, the sync data is already available.

Also removed the now-dead `DIRECT_TRANSACTION_METADATA_KEY` constant and its usage in `mutations.ts`.

### Key Invariants

- A direct transaction's `mutationFn` (e.g., `awaitTxId()`) does not resolve until sync delivers the matching txid
- While a transaction is in `persisting` state, its optimistic rows are kept alive by the active-transaction loop, not the pending-optimistic sets
- Once `completed`, the pending-optimistic stale check is the sole gatekeeper

### Non-goals

- Adding observability/logging to `state.ts` (systemic gap, separate effort)
- Covering optimistic delete with server-generated keys (deletes always target known keys)

## Verification

```bash
pnpm vitest run --pool-options.threads.maxThreads=2 packages/db/tests packages/electric-db-collection/tests
```

All 2364 tests pass. 2 pre-existing failures in `collection-subscribe-changes.test.ts` (also fail on main).

## Files Changed

- **`packages/db/src/collection/state.ts`** — Remove `pendingOptimisticDirect*` sets, simplify stale-row check to `pendingSyncKeys.has(key)` only
- **`packages/db/src/collection/mutations.ts`** — Remove `DIRECT_TRANSACTION_METADATA_KEY` import and metadata tagging
- **`packages/db/src/collection/transaction-metadata.ts`** — Deleted (orphaned)
- **`packages/db/tests/collection.test.ts`** — Two new tests: optimistic insert replaced by server key, live query reflects the key swap
- **`packages/electric-db-collection/tests/electric.test.ts`** — One new test: txid sync confirms different server-generated key

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed stale optimistic rows persisting when sync confirms a different server-generated key during insert operations.

* **Tests**
  * Added comprehensive tests for optimistic insert sync behavior, verifying proper handling when server-generated keys replace client-provided keys.

* **Chores**
  * Updated CI to use Node.js 22.13.

* **Documentation**
  * Added Git and PR hygiene guidelines.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/db/pull/1547?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/fix-stale-optimistic-server-key.md
+++ b/.changeset/fix-stale-optimistic-server-key.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/db': patch
+---
+
+Fix stale optimistic rows persisting when sync confirms a different server-generated key. Previously, direct transactions (from `collection.insert()` etc.) had their optimistic rows exempted from stale-row cleanup, which prevented temp-key rows from being removed when the server returned a different primary key.
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
-          node-version: '20'
+          node-version: '22.13'
           cache: 'pnpm'
 
       - name: Install dependencies
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -14,6 +14,7 @@ This guide provides principles and patterns for AI agents contributing to the Ta
 8. [Function Design](#function-design)
 9. [Modern JavaScript Patterns](#modern-javascript-patterns)
 10. [Edge Cases and Corner Cases](#edge-cases-and-corner-cases)
+11. [Git and PR Hygiene](#git-and-pr-hygiene)
 
 ## Type Safety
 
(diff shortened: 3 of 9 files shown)


samwillis (approved):
LGTM, and makes perfect sense!

PR #1550, opened 2026-05-25
fix(db): require all or() branches to be index-optimizable

### LINKED ISSUE #1502: optimizeOrExpression silently drops items when one OR branch references an unindexed field
Below is an AI generated summary of a bug I came across today:

## Summary

`optimizeOrExpression` in `packages/db/src/utils/index-optimization.ts` is unsound. When a `where` clause contains an `or(...)` whose branches reference a mix of indexed and unindexed fields, the optimizer returns `canOptimize: true` with a result set that is the **union of only the indexed branches** — silently dropping items that would match the unindexed branch.

The result is that `useLiveQuery` / `currentStateAsChanges` returns an incomplete subset of the matching rows, with no warning or error. Callers don't fall back to a full scan because `canOptimize` is `true`.

## Where the bug is

`packages/db/src/utils/index-optimization.ts` (verified on `main` and on the published 0.6.5 tarball — identical code):

```ts
function optimizeOrExpression(expression, collection) {
  if (expression.type !== 'func' || expression.args.length < 2) {
    return { canOptimize: false, matchingKeys: new Set() }
  }

  const results = []

  // Try to optimize each part, keep the optimizable ones
  for (const arg of expression.args) {
    const result = optimizeQueryRecursive(arg, collection)
    if (result.canOptimize) {
      results.push(result)
    }
    // ⚠️ non-optimizable args are dropped on the floor
  }

  if (results.length > 0) {
    const allMatchingSets = results.map((r) => r.matchingKeys)
    const unionedKeys = unionSets(allMatchingSets)
    return { canOptimize: true, matchingKeys: unionedKeys }
    // ⚠️ canOptimize: true even though some branches were skipped
  }

  return { canOptimize: false, matchingKeys: new Set() }
}
```

For OR semantics this is unsound. If any branch can't be optimized, the optimizer can't soundly compute the answer from the indexes alone — the missing branch could match items not in any indexed branch's result set. The function should return `canOptimize: false` and let the consumer fall back to a full scan.

The corresponding `optimizeAndExpression` is fine because for AND, dropping a non-optimizable arg only widens the result set (the consumer is expected to apply the full predicate to that superset). For OR it narrows it.

## Repro

Conceptual scenario that triggers it in practice:

1. A collection has items where some match `field_A = X`, others match `field_B = Y`, with no overlap.
2. Some prior query has caused `field_A` to be auto-indexed (`autoIndex: 'eager'` is the default).
3. `field_B` has not been queried in isolation, so no index exists on it.
4. Run a live query with `where: or(eq(field_A, X), eq(field_B, Y))`.

Expected: rows matching either branch.
Actual: only rows matching `field_A = X`. Rows matching `field_B = Y` (and not `field_A = X`) are dropped.

If you reorder the queries so a query touches `field_B` first (causing auto-indexing on `field_B`), the bug disappears, because now both branches are optimizable.

## Suggested fix

```ts
function optimizeOrExpression(expression, collection) {
  if (expression.type !== 'func' || expression.args.length < 2) {
    return { canOptimize: false, matchingKeys: new Set() }
  }

  const results = []
  for (const arg of expression.args) {
    const result = optimizeQueryRecursive(arg, collection)
    if (!result.canOptimize) {
      // OR with any non-optimizable branch cannot be soundly index-optimized
      return { canOptimize: false, matchingKeys: new Set() }
    }
    results.push(result)
  }

  const unionedKeys = unionSets(results.map((r) => r.matchingKeys))
  return { canOptimize: true, matchingKeys: unionedKeys }
}
```

v-anton:
## 🎯 Changes

Fixes #1502.

`optimizeOrExpression` returned `canOptimize: true` when only *some* `or()` branches had matching indexes, producing a partial union that silently dropped rows from unindexed branches.

The fix requires all branches to be optimizable (`results.length === expression.args.length`) before returning the index result. If any branch can't be optimized, fall back to full scan.

## ✅ Checklist

## 🚀 Release Impact

- [x] This change affects published code, and I have generated a changeset.

## Summary by CodeRabbit

* **Bug Fixes**
  * OR-style queries now only use index results when every branch is index-optimizable, preventing partial index use that could produce incomplete results.

* **Tests**
  * Added tests to validate live-query behavior for OR queries when only some branches are indexed, ensuring correct and consistent results.

* **Chores**
  * Release metadata updated to mark a patch release.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/db/pull/1550?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/fix-or-partial-union.md
+++ b/.changeset/fix-or-partial-union.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/db": patch
+---
+
+fix(db): require all or() branches to be index-optimizable before using index result
--- a/packages/db/src/utils/index-optimization.ts
+++ b/packages/db/src/utils/index-optimization.ts
@@ -469,22 +469,17 @@ function optimizeOrExpression<T extends object, TKey extends string | number>(
 
   const results: Array<OptimizationResult<TKey>> = []
 
-  // Try to optimize each part, keep the optimizable ones
   for (const arg of expression.args) {
     const result = optimizeQueryRecursive(arg, collection)
-    if (result.canOptimize) {
-      results.push(result)
+    if (!result.canOptimize) {
+      return { canOptimize: false, matchingKeys: new Set() }
     }
+    results.push(result)
   }
 
-  if (results.length === expression.args.length) {
-    // Use unionSets utility for OR logic
-    const allMatchingSets = results.map((r) => r.matchingKeys)
-    const unionedKeys = unionSets(allMatchingSets)
-    return { canOptimize: true, matchingKeys: unionedKeys }
-  }
-
-  return { canOptimize: false, matchingKeys: new Set() }
+  const allMatchingSets = results.map((r) => r.matchingKeys)
+  const unionedKeys = unionSets(allMatchingSets)
+  return { canOptimize: true, matchingKeys: unionedKeys }
 }
 
 /**
--- a/packages/db/src/utils/index-optimization.ts
+++ b/packages/db/src/utils/index-optimization.ts
@@ -498,8 +498,8 @@ function canOptimizeOrExpression<
     return false
   }
 
-  // If any argument can be optimized, we can gain some speedup
-  return expression.args.some((arg) => canOptimizeExpression(arg, collection))
+  // All branches must be optimizable — partial OR optimization is unsound
+  return expression.args.every((arg) => canOptimizeExpression(arg, collection))
 }
 
 /**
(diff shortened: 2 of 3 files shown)


v-anton:
For context: #1562 stacks on top of this PR and extends partial optimization to AND via a residual predicate, fixing an analogous soundness bug without forfeiting the index speedup. Open as draft pending direction here.

KyleAMathews:
Superseded by merged #1582. Main contains the same all-disjunct OR guard plus the broader exactness/re-filtering contract, and the focused mixed/all/none strategy checks pass on current main.

PR #1557, opened 2026-05-25
fix(db): fall back to crypto.getRandomValues when randomUUID is unavailable

### LINKED ISSUE #1541: @tanstack/db assumes crypto.randomUUID exists in non-secure browser contexts
### Describe the bug

In browser contexts served over non-localhost HTTP, such as a dev server accessed from another machine via a LAN IP address, `crypto.randomUUID` can be unavailable because it is a secure-context-only API. `crypto.getRandomValues` remains available in insecure contexts.

`@tanstack/db` calls `crypto.randomUUID()` directly in collection, transaction, and mutation paths. That causes collection writes to throw before application code can handle it.

### Version

- `@tanstack/db`: `0.6.5`
- Browser: Chromium/Chrome behavior consistent with the Web Crypto secure-context restriction

### Reproduction

Serve a browser app over non-localhost HTTP, for example:

```text
http://<LAN-IP>:<port>
```

In DevTools on that page:

```js
window.isSecureContext // false
typeof crypto.randomUUID // "undefined"
typeof crypto.getRandomValues // "function"
```

Then run code equivalent to:

```js
import { createCollection, localOnlyCollectionOptions } from '@tanstack/db'

  localOnlyCollectionOptions({
    id: 'items',
    getKey: (item) => item.id,
  }),
)

collection.insert({ id: '1' })
```

### Expected behavior

The collection write succeeds, or `@tanstack/db` falls back to UUID generation via `crypto.getRandomValues()` when `crypto.randomUUID()` is not present.

### Actual behavior

The write throws:

```text
TypeError: crypto.randomUUID is not a function
```

### Affected code paths

Current source has direct `crypto.randomUUID()` calls in several places, including:

- `packages/db/src/collection/mutations.ts` for mutation IDs on insert/update/delete
- `packages/db/src/transactions.ts` for transaction IDs
- `packages/db/src/local-only.ts` for local-only collection IDs
- `packages/db/src/collection/index.ts` when a collection ID is omitted

### Suggested fix

Centralize UUID generation and use native `crypto.randomUUID()` when present. Otherwise, generate an RFC 4122 version 4 UUID using `crypto.getRandomValues()` when available.

If neither API is available, throw an explicit error explaining that UUID generation requires Web Crypto / a secure-enough browser environment.

Relevant docs:

- `crypto.randomUUID()` is secure-context-only: https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID
- `crypto.getRandomValues()` is available in insecure contexts: https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues

spokodev:
## Problem

`crypto.randomUUID()` is restricted to secure contexts per the Web Crypto spec, so it is unavailable on pages served over plain HTTP from a non-localhost host (for example a dev server reached via a LAN IP). `@tanstack/db` calls it directly from collections, transactions, mutations, local-only, and local-storage, so the first write throws

```text
TypeError: crypto.randomUUID is not a function
```

before user code can handle it. `crypto.getRandomValues()` remains available in non-secure contexts, so a v4 UUID is still reachable; we just have to assemble it ourselves when the higher-level API is missing.

## Repro

```text
http://<LAN-IP>:<port>
```

```js
window.isSecureContext // false
typeof crypto.randomUUID // "undefined"
typeof crypto.getRandomValues // "function"
```

```ts
import { createCollection, localOnlyCollectionOptions } from '@tanstack/db'

  localOnlyCollectionOptions({
    id: 'items',
    getKey: (item) => item.id,
  }),
)

collection.insert({ id: '1' }) // throws TypeError: crypto.randomUUID is not a function
```

## Fix

Add a `safeRandomUUID()` helper in `packages/db/src/utils.ts`:

1. Use `crypto.randomUUID()` when available (the common case).
2. Fall back to RFC 4122 v4 generation via `crypto.getRandomValues()` with the version/variant bits set per spec §4.4.
3. Throw an explicit `Error` if neither Web Crypto API is available.

Replace the seven direct call sites with the helper:

- `packages/db/src/collection/index.ts` (default collection id)
- `packages/db/src/collection/mutations.ts` (insert / update / delete mutation ids; three sites)
- `packages/db/src/local-only.ts` (local-only collection id)
- `packages/db/src/local-storage.ts` (version-tracking uuid)
- `packages/db/src/transactions.ts` (transaction id)

The issue body listed four files; while replacing those I also found three additional `crypto.randomUUID()` call sites in `mutations.ts` (the per-operation `mutationId`) and one in `local-storage.ts` that have the same failure mode, and rolled them into the helper as well so the fallback is consistent across `@tanstack/db`.

## Tests

`packages/db/tests/safe-random-uuid.test.ts` covers both axes:

- **Delegate path**: when `crypto.randomUUID` exists, the helper returns its value unchanged and the stub is called exactly once.
- **Fallback path**: when only `crypto.getRandomValues` is present, the result matches a strict RFC 4122 v4 regex (`/^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/`).
- **Fallback uniqueness**: two successive fallback calls with deterministic but distinct byte input produce different UUIDs, ensuring the helper does not collapse to a constant when `randomUUID` is missing.
- **No crypto at all**: throws an explicit error mentioning the Web Crypto API.
- **Partial crypto** (`{}` object, no `randomUUID`, no `getRandomValues`): same explicit error, so we never silently fall through to native crypto.

All five new tests pass with the patched helper (`pnpm --filter @tanstack/db test`). The pre-existing 163 test failures on `main` (`localStorage.getItem is not a function` in `union-all.test.ts` and related jsdom interactions) are not introduced by this change; the new test count of `2213 passed` matches `2208 baseline + 5 new`.

Lint clean (`pnpm --filter @tanstack/db lint`, no new warnings). Build clean (`pnpm --filter @tanstack/db build`).

Closes #1541

## Note on disclosure

Bug discovery and reproduction were independent of this contributor (issue author `@kj55-dev` filed a clean writeup with a suggested fix). I implemented the helper, located the additional call sites in `mutations.ts` and `local-storage.ts`, wrote the tests, and reviewed every change. AI assistance was used while drafting parts of the test cases and this PR body, in line with the AGENTS.md guidance in the repo.

### DIFF
--- a/.changeset/safe-random-uuid-non-secure-context.md
+++ b/.changeset/safe-random-uuid-non-secure-context.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/db': patch
+---
+
+Fix `@tanstack/db` throwing `TypeError: crypto.randomUUID is not a function` in non-secure browser contexts. `crypto.randomUUID()` is restricted to secure contexts, so pages served over plain HTTP from a non-localhost host (such as a dev server reached via a LAN IP) could not insert into a collection, run a mutation, or open a transaction. UUID generation now centralises in `safeRandomUUID()` which prefers `crypto.randomUUID()` when available and falls back to RFC 4122 v4 via `crypto.getRandomValues()`.
--- a/packages/db/src/collection/index.ts
+++ b/packages/db/src/collection/index.ts
@@ -3,6 +3,7 @@ import {
   CollectionRequiresConfigError,
   CollectionRequiresSyncConfigError,
 } from '../errors'
+import { safeRandomUUID } from '../utils'
 import { currentStateAsChanges } from './change-events'
 
 import { CollectionStateManager } from './state'
@@ -329,7 +330,7 @@ export class CollectionImpl<
     if (config.id) {
       this.id = config.id
     } else {
-      this.id = crypto.randomUUID()
+      this.id = safeRandomUUID()
     }
 
     // Set default values for optional config properties
--- a/packages/db/src/collection/mutations.ts
+++ b/packages/db/src/collection/mutations.ts
@@ -17,6 +17,7 @@ import {
   UndefinedKeyError,
   UpdateKeyNotFoundError,
 } from '../errors'
+import { safeRandomUUID } from '../utils'
 import { DIRECT_TRANSACTION_METADATA_KEY } from './transaction-metadata.js'
 import type { Collection, CollectionImpl } from './index.js'
 import type { StandardSchemaV1 } from '@standard-schema/spec'
@@ -193,7 +194,7 @@ export class CollectionMutationsManager<
       const globalKey = this.generateGlobalKey(key, item)
 
       const mutation: PendingMutation<TOutput, `insert`> = {
-        mutationId: crypto.randomUUID(),
+        mutationId: safeRandomUUID(),
         original: {},
         modified: validatedData,
         // Pick the values from validatedData based on what's passed in - this is for cases
@@ -366,7 +367,7 @@ export class CollectionMutationsManager<
         const globalKey = this.generateGlobalKey(modifiedItemId, modifiedItem)
 
         return {
-          mutationId: crypto.randomUUID(),
+          mutationId: safeRandomUUID(),
           original: originalItem,
           modified: modifiedItem,
           // Pick the values from modifiedItem based on what's passed in - this is for cases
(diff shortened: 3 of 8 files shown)


kevin-dp:
Thanks so much for putting this together! 🙏 Closing in favor of #1593, which takes the same approach (helper renamed to `safeRandomUUID` to match your naming) and additionally rolls out the fix across the other packages that call `crypto.randomUUID()` directly (`db-sqlite-persistence-core`, `browser-db-sqlite-persistence`, `electron-db-sqlite-persistence`, and `offline-transactions`), plus adds unit tests for the fallback and missing-`crypto` paths. Really appreciate the contribution and the prompt to get this fixed!

PR #1561, opened 2026-05-27
test(db): reproduce stale buffered sync when optimistic snapback races a persisting transaction

kevin-dp:
## Summary

Adds a failing reproduction test for a stale-data bug originally reported as "auto-index returns the row in the wrong bucket after an optimistic→synced update", and confirms that the same symptom still reproduces against the fix in #1517.

The repro narrows the bug down to a different mechanism than the index-bucket lifecycle issue #1517 addressed: it lives in `commitPendingTransactions`'s gating on `hasPersistingTransaction`, not in `BasicIndex`/`BTreeIndex` or in `isRedundantSync`.

CI is intentionally red: the failing test describes the *correct* behaviour, and stays failing until the bug is fixed.

## What the test reproduces

Models the user's "fast successive drags" kanban scenario:

1. User drags card P from stage A → stage B → optimistic transaction `T1` starts persisting.
2. User immediately drags card Q from stage A → stage B → optimistic transaction `T2` starts persisting.
3. `T1`'s `mutationFn` resolves (server-confirmed). In the user's app, `awaitTxId` races out via the snapshot path before the shape stream has delivered P's row change — so when the optimistic delta lifts, no buffered sync exists for P, and `recomputeOptimisticState` drops P's optimistic upsert as "stale". Visible state for P snaps from B back to A.
4. The shape stream now delivers P's row change. But `T2` is still persisting, so `commitPendingTransactions` is gated on `hasPersistingTransaction` and the sync message sits buffered in `pendingSyncedTransactions` indefinitely.
5. While `T2` keeps persisting, anyone reading the live query sees P stuck in stage A — even though the server has confirmed it as stage B, and the sync message for that change is sitting in `pendingSyncedTransactions`.

In the user's app, with continuous drag traffic the "another transaction is persisting" condition is essentially always true, so the buffered sync never flushes → "stays until reload".

The test also includes a baseline case (single drag, no concurrent traffic) that passes — demonstrating that the bug only surfaces when a concurrent persisting transaction is blocking sync application for unrelated keys.

## Relationship to #1517

The user originally suspected `BasicIndex` was leaving keys in stale buckets; #1517 addressed that by tracking key→bucket and dropping prior buckets on `add`. They retested with #1517 applied and the snapback still reproduced. I ran this repro on both `main` and `kevin/stale-index-bucket-repro` (the #1517 branch) — same failure mode either way. So this is a separate bug; the user's `awaitMatch` workaround sidesteps it by preventing the premature optimistic lift in step 3.

This PR does not propose a fix. The probable direction is for `commitPendingTransactions` to be able to apply syncs for keys not touched by any currently-persisting transaction, instead of all-or-nothing gating on `hasPersistingTransaction`.

## Test plan

- [x] Run `tests/auto-index-snapback.test.ts` locally — baseline passes, bug-repro fails as expected.
- [x] Confirm same failure mode on the `kevin/stale-index-bucket-repro` (#1517) branch.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test suite for database concurrent transaction behavior with optimistic updates, including baseline scenarios and edge case validation.
  * Identified and documented a known issue with specific concurrent transaction sequences that requires resolution.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/db/pull/1561?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/db/tests/auto-index-snapback.test.ts
+++ b/packages/db/tests/auto-index-snapback.test.ts
@@ -1,0 +1,179 @@
+import { describe, expect, it } from 'vitest'
+import { createCollection } from '../src/collection/index.js'
+import { createTransaction } from '../src/transactions.js'
+import { BasicIndex } from '../src/indexes/basic-index.js'
+import { createLiveQueryCollection } from '../src/query/live-query-collection.js'
+import { inArray } from '../src/query/builder/functions.js'
+import { mockSyncCollectionOptions, stripVirtualProps } from './utils.js'
+
+type Row = { id: string; stage: string }
+
+const flushTasks = () => new Promise((r) => setTimeout(r, 0))
+
+describe(`auto-index snapback after optimistic + sync, with concurrent in-flight transactions`, () => {
+  it(`single drag, no concurrent traffic — works correctly (baseline)`, async () => {
+    const options = mockSyncCollectionOptions<Row>({
+      id: `baseline`,
+      getKey: (item) => item.id,
+      initialData: [
+        { id: `P`, stage: `A` },
+        { id: `Q`, stage: `A` },
+      ],
+      autoIndex: `eager`,
+      defaultIndexType: BasicIndex,
+    })
+    const collection = createCollection(options)
+    await collection.stateWhenReady()
+
+    const liveA = createLiveQueryCollection({
+      query: (q: any) =>
+        q
+          .from({ p: collection })
+          .where(({ p }: any) => inArray(p.stage, [`A`])),
+      startSync: true,
+    })
+    await liveA.stateWhenReady()
+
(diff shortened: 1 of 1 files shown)


kevin-dp:
Closing this issue because the behaviour is expected as we don't provide transactional guarantees across shapes. The problem here comes from the fact that each column has its own shape. And moving an item between different columns is a transactions that span 2 shapes. This means that the tx ID will be observed on both shape streams and when the first one notices it it marks that tx ID as seen but actually we have only seen the changes of that transaction to one of the shapes and not yet for the second shape. So this can cause these kind of snapbacks as optimistic state is being dropped etc.

PR #1566, opened 2026-06-02
fix(db): canonicalize inArray value order in normalizeExpressionPaths

v-anton:
## Problem

`inArray` is set membership, but the predicate handed to a collection's `loadSubset` was not canonical in value order. `normalizeExpressionPaths` — the pass that canonicalizes predicates for subscriptions — only normalized ref *paths*, not `inArray` value order. And the join lazy-load predicate (`inArray(joinKey, [...])`, built in `joins.ts`) was passed to `requestSnapshot` **without** going through `normalizeExpressionPaths` at all.

So the same value set arriving in a different order — a join re-emitting its lazy-load keys across pipeline batches, or two subscriptions for the same data — serialized to a **distinct** predicate. Adapters that dedupe by serialized predicate / queryKey (e.g. `@tanstack/query-db-collection`) then treat it as a new subset and **refetch identical data** (and miss CDN cache cells keyed on the request URL).

## Fix

- `normalizeExpressionPaths` now also canonicalizes set-membership value order — it sorts `in` value arrays. `in` is unordered, and `isPredicateSubset` / index lookups already treat it order-independently, so sorting is safe.
- The join lazy-load predicate is routed through `normalizeExpressionPaths` in `joins.ts` (using the `target.alias` already in scope), so its keys are canonicalized like every other predicate — fixing the inconsistency where it bypassed normalization entirely.

## Tests

- Added `tests/query/canonicalize-in-values.test.ts`: an on-demand collection queried with an out-of-order `inArray` literal asserts `loadSubset` receives the values sorted. Verified it **fails** without the change (`loadSubset` gets `[3, 1, 2]`) and **passes** with it.
- Full `@tanstack/db` suite green (2372 tests).

Changeset included (`patch`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Canonicalized `in`-filter value ordering so identical value sets produce the same query/cache keys, including for join lazy-load queries—eliminates duplicate queries and improves cache consistency.

* **Tests**
  * Added tests verifying `in` filter values are consistently sorted in query requests.

* **Documentation**
  * Clarified that `in` predicates are canonicalized by sorting their value lists.

### DIFF
--- a/.changeset/fix-canonicalize-in-values.md
+++ b/.changeset/fix-canonicalize-in-values.md
@@ -1,0 +1,13 @@
+---
+'@tanstack/db': patch
+---
+
+fix(db): canonicalize `inArray` value order in `normalizeExpressionPaths`
+
+`inArray` is set membership, but `normalizeExpressionPaths` only canonicalized
+ref paths, not value order — so the same value set in a different order produced
+a distinct serialized predicate (and `loadSubset` queryKey / cache key for
+adapters that key by it) and refetched identical data. `normalizeExpressionPaths`
+now also sorts `in` value arrays, and the join lazy-load predicate (previously
+built and combined without normalization) is routed through it so its keys are
+canonicalized too.
--- a/packages/db/src/query/compiler/expressions.ts
+++ b/packages/db/src/query/compiler/expressions.ts
@@ -1,18 +1,21 @@
 import { Func, PropRef, Value } from '../ir.js'
+import { defaultComparator } from '../../utils/comparison.js'
 import type { BasicExpression, OrderBy } from '../ir.js'
 
 /**
- * Normalizes a WHERE clause expression by removing table aliases from property references.
+ * Normalizes a WHERE clause expression into a canonical form.
  *
- * This function recursively traverses an expression tree and creates new BasicExpression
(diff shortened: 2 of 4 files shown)


v-anton:
Thanks @coderabbitai — the suggested diff references `canonicalizeInValues`, which was from an earlier revision; this PR moved the canonicalization into `normalizeExpressionPaths` instead. That means `requestLimitedSnapshot` is already covered: it forwards `this.options.whereExpression`, which is canonicalized at subscription creation (the `in`-value sort now happens *inside* `normalizeExpressionPaths`), so no per-call step is needed on the ordered/limited path. Added a regression test (`sorts inArray values for ordered/limited queries too`) that exercises the `orderBy`+`limit` path and asserts `loadSubset` receives sorted values.

KyleAMathews:
Thanks for identifying the demand-identity problem. This was superseded by https://github.com/TanStack/db/pull/1768 (https://github.com/TanStack/db/commit/d8defd2a8eb96162cbd4e24970d519eac217bb95), which canonicalizes `IN` values as an equality-semantic set for demand identity while preserving raw predicate order. The old lazy-join site has also been replaced. Closing this conflicting draft without merge.

PR #1568, opened 2026-06-04
fix(query-db-collection): forward gcTime from queryCollectionOptions to the underlying query

### LINKED ISSUE #1546: query-collection gcTime not set to underlaying query

**Describe the bug**

Hi, I am not sure this is an issue or intended behavior. Setting `gcTime`on the `queryCollectionOptions` does not override the default  `gcTime` on the tanstack/query.

This is my collection definition:
```ts
  queryCollectionOptions({
    queryKey: ["test"],
    staleTime: Infinity,
    gcTime: Infinity,
   queryClient,
```

Checking tanstack query devtools the query is still using the `gcTime`from the `queryClient`.

I do get this warning when the query is cleaned up:
```
[cleanupQueryIfIdle] Invariant violation: refcount=1 but no listeners. Cleaning up to prevent leak. {hashedQueryKey: '["test"]'}
```

spokodev:
The `queryCollectionOptions` config exposes a `gcTime` knob in its type, but the value never reaches the underlying TanStack Query observer. As @fehmer noted on #1546, the devtools panel shows the query falling back to the `queryClient` default instead of the override.

At `packages/query-db-collection/src/query.ts:1167`, the `observerOptions` block conditionally spreads `enabled`, `refetchInterval`, `retry`, `retryDelay`, and `staleTime` from the config, but `gcTime` is missing from the destructure on line 580 and from the conditional spread. Adding it preserves the documented behavior.

```ts
// src/query.ts
   const {
     ...
     staleTime,
+    gcTime,
     persistedGcTime,
     ...
   } = config

   // ...

   const observerOptions: QueryObserverOptions<...> = {
     ...
     ...(staleTime !== undefined && { staleTime }),
+    ...(gcTime !== undefined && { gcTime }),
   }
```

The public `QueryCollectionConfig` type also gets a `gcTime` field next to `staleTime`, mirroring the same `QueryObserverOptions[gcTime]` reference.

Tests cover the forwarding path (override reaches the observer), the default fallback (omitted gcTime → queryClient default preserved), and `gcTime: Infinity` (the exact value @fehmer reported using). The forwarding and Infinity tests fail on master and pass with the fix; the fallback test passes on both because the existing behavior is preserved.

Closes #1546

## Release Notes

* **New Features**
  * The `gcTime` configuration option is now properly forwarded to TanStack Query, enabling control over cache garbage collection timing per query collection.

* **Tests**
  * Added test coverage for `gcTime` forwarding behavior, including explicit values and fallback to default settings.

### DIFF
--- a/.changeset/query-db-collection-gctime-forward.md
+++ b/.changeset/query-db-collection-gctime-forward.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/query-db-collection': patch
+---
+
+Forward `gcTime` from `queryCollectionOptions` to the underlying TanStack Query observer. The `gcTime` option was previously documented in the config shape but silently dropped before reaching the observer, leaving consumers stuck on the `queryClient` default. Closes #1546.
--- a/packages/query-db-collection/src/query.ts
+++ b/packages/query-db-collection/src/query.ts
@@ -119,6 +119,13 @@ export interface QueryCollectionConfig<
     TQueryData,
     TQueryKey
   >[`staleTime`]
+  gcTime?: QueryObserverOptions<
+    TQueryData,
+    TError,
+    Array<T>,
+    TQueryData,
+    TQueryKey
+  >[`gcTime`]
   persistedGcTime?: number
 
   /**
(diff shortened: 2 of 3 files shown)


kevin-dp:
Thank you for fixing this issue, PR is great so i'll merge away!

fehmer:
Thanks @spokodev

PR #1575, opened 2026-06-08
fix(electric): bound refresh wait for on-demand subsets

KyleAMathews:
## Summary

On-demand Electric live queries no longer wait indefinitely for `forceDisconnectAndRefresh()` before requesting subset snapshots. This prevents React Native/Expo-style native fetch implementations that don't promptly abort long polls from keeping live queries in `loading` until the Electric live poll naturally times out.

## Root Cause

For on-demand sync, `@tanstack/electric-db-collection` asks Electric for a subset snapshot when a live query needs data. If the underlying `ShapeStream` is already up-to-date, the collection first calls:

```ts
await stream.forceDisconnectAndRefresh()
```

That is intended to abort an in-flight live long-poll and force a fresh stream round-trip before `requestSnapshot()`. In browsers this normally returns quickly because `fetch` aborts the long-poll promptly.

Some native fetch implementations, especially React Native/Expo, may not reject/abort long-poll requests promptly. In that case, `forceDisconnectAndRefresh()` can remain pending until the long-poll times out. Because the live query tracks the subset load promise, the query can keep reporting `loading` even after rows are present locally.

## Approach

Bound the refresh wait with a small timeout:

```ts
await Promise.race([
  stream.forceDisconnectAndRefresh(),
  new Promise<void>((resolve) => {
    timeoutId = setTimeout(
      resolve,
      FORCE_DISCONNECT_AND_REFRESH_TIMEOUT_MS,
    )
  }),
])
```

If the refresh finishes quickly, behavior is unchanged. If it does not finish within `250ms`, the collection proceeds to `requestSnapshot()` instead of waiting for the long-poll to finish naturally.

## Key Invariants

- `requestSnapshot()` still runs after the bounded refresh attempt.
- Refresh errors still go through the existing `handleSnapshotError()` path.
- The timeout is only a fallback for the pre-snapshot refresh; it does not change the snapshot request itself.
- Browser behavior should remain effectively unchanged because aborting the live poll is expected to complete before the timeout.

## Non-goals

- This does not change Electric's long-poll timeout behavior.
- This does not attempt to detect React Native/Expo specifically.
- This does not replace or wrap the user's `fetchClient`.
- This does not alter eager or progressive initial sync behavior; the affected path is on-demand subset loading after the stream is already up-to-date.

## Trade-offs

A fully awaited refresh is slightly stricter, but it can turn native fetch abort quirks into user-visible multi-second or multi-minute loading states. A bounded wait preserves the intended fast path while making the on-demand path resilient when abort is slow or ignored.

The timeout is intentionally short (`250ms`) because the refresh is only a pre-snapshot optimization. If it cannot complete quickly, making progress with `requestSnapshot()` is better than holding the live query in loading until the live poll naturally returns.

## Verification

```bash
pnpm --filter @tanstack/electric-db-collection test
```

Both passed.

Note: in a fresh worktree, running the filtered test before building produced type-resolution errors for e2e suite imports of `@tanstack/electric-db-collection`. Building the package first fixed that, and the package tests passed.

## Files changed

- `packages/electric-db-collection/src/electric.ts`
  - Adds `FORCE_DISCONNECT_AND_REFRESH_TIMEOUT_MS`.
  - Wraps `stream.forceDisconnectAndRefresh()` in a bounded `Promise.race()` before on-demand subset snapshot requests.
- `.changeset/tame-rn-live-poll-refresh.md`
  - Adds a patch changeset for `@tanstack/electric-db-collection`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved live query loading when stream refreshes take too long or fail to stop promptly.
  - On-demand data requests now proceed after a brief refresh timeout instead of waiting for long-poll operations to finish.
  - Prevented delayed refresh results from interfering with subsequent snapshot requests.
- **Tests**
  - Added coverage for delayed, rejected, and promptly completed stream refreshes, including timer cleanup.


KyleAMathews:
### Code review

Found 1 issue:

1. The new 250ms refresh-timeout path has no deterministic regression coverage. Existing tests cover refresh fulfillment and rejection, but not a refresh that remains pending past the bound, fallthrough to `requestSnapshot()`, late settlement, or timer cleanup. AGENTS.md explicitly requires regression tests for bug fixes and async race conditions.

https://github.com/TanStack/db/blob/56f2bf3cdb988e6a261783e929f7698ef89f4787/packages/electric-db-collection/src/electric.ts#L488-L507

kevin-dp (approved):

PR #1581, opened 2026-06-10
test: failing tests for AND/OR queries mixing indexed and non-indexed conditions

kevin-dp:
## Summary

Adds seven unit tests to `collection-indexes.test.ts` that assert the expected results of `currentStateAsChanges` for `where` clauses that the index optimizer currently gets wrong. All tests are written purely in terms of expected behaviour.

**Mixing indexed and non-indexed conditions** (e.g. `length(name) > 6`):
1. **OR** — the result must be the union of rows matching either condition. Currently rows matched only by the non-indexed condition are missing.
2. **AND** — every condition must be enforced. Currently rows failing the non-indexed condition are included.
3. **Compound range + condition on another (indexed) field** — all conditions must be enforced. Currently only the range is applied.

**Range boundary handling:**
4. **Strictest lower bound** — `gte(age, 25) AND gt(age, 25)` must reduce to `age > 25` regardless of argument order. Currently the inclusive bound wins.
5. **Strictest upper bound** — `lte(age, 30) AND lt(age, 30)` must reduce to `age < 30`. Currently the inclusive bound wins.
6. **Date bounds at the same value** — distinct `Date` instances representing the same time must be treated as equal when picking the strictest bound. (Also exercises one-sided compound ranges, which currently return an empty result.)
7. **Strict comparisons on date fields** — `gt(createdAt, date)` must exclude the boundary row. Currently the BTree index includes it because it compares the normalized indexed value against the raw `Date`.

⚠️ **These tests fail on `main` by design** — they pin down the expected behaviour. The follow-up PR #1582 provides the fixes that make them pass.

## Summary by CodeRabbit

* **Tests**
  * Added new coverage to ensure strict boundary behavior for date range queries (values equal to the boundary are excluded).
  * Expanded assertions for complex query optimization, validating correct `and`/`or` behavior when mixing indexed and computed predicates, including “strictest bound wins” logic.
  * Added edge-case and parity tests for full-scan consistency, including handling of `undefined` values and `NaN` semantics in equality and `IN` queries.

### DIFF
--- a/packages/db/tests/collection-indexes.test.ts
+++ b/packages/db/tests/collection-indexes.test.ts
@@ -625,6 +625,19 @@ describe(`Collection Indexes`, () => {
       })
     })
 
+    it(`should exclude the boundary value from greater than queries on dates`, () => {
+      // gt must be strict for date fields: Bob was created exactly on
+      // 2023-01-02, so only rows created strictly later may be returned.
+      collection.createIndex((row) => row.createdAt)
+
+      const result = collection.currentStateAsChanges({
+        where: gt(new PropRef([`createdAt`]), new Date(`2023-01-02`)),
+      })!
+
+      const names = result.map((r) => r.value.name).sort()
+      expect(names).toEqual([`Charlie`, `Diana`, `Eve`])
+    })
+
     it(`should perform greater than or equal queries`, () => {
       withIndexTracking(collection, (tracker) => {
         const result = collection.currentStateAsChanges({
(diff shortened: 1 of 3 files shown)


kevin-dp:
Superseded by #1582.

These failing tests were split out purely to show the red → green progression. They're now included directly in #1582 (passing) alongside the fixes that make them pass, plus the PostgreSQL `NaN`-semantics work that superseded #1617. #1582 is retargeted to `main` and is self-contained, so this PR is no longer needed.

Closing in favor of #1582.

PR #1582, opened 2026-06-10
fix: enforce all where conditions when index optimization is partial

kevin-dp:
## Summary

Follow-up to #1581 — fixes the index optimizer so that the seven failing tests added there pass, and adopts **PostgreSQL float semantics for `NaN`** so the `where` evaluator, ordering, and indexes all agree on `NaN`/invalid Dates. All fixes are correctness fixes for `where` clauses served from indexes.

> **Supersedes #1581.** That PR added the failing tests demonstrating the bugs; those tests are included here (now passing) alongside the fixes, so this PR is self-contained against `main`. #1581 has been closed in favor of this one.

> **Supersedes #1617.** That proposal PR is folded in here (see the NaN-semantics section below) and has been closed, so the index-optimization correctness work is built on the final, coherent contract rather than the interim JS-semantics handling.

### Commit 1 — enforce all conditions when index optimization is partial

`currentStateAsChanges` treats the optimizer's `matchingKeys` as the exact result set, but the optimizer silently dropped conditions it couldn't serve from an index:

- **OR** unioned only the optimizable disjuncts → rows matched only by a non-indexed disjunct were **missing**.
- **AND** intersected only the optimizable conjuncts → non-indexed conjuncts were never enforced, returning **non-matching rows**.
- **Compound ranges** (`age > 5 AND age < 10`) early-returned and **discarded all sibling conjuncts**, even indexed ones.

This also affected the initial state of filtered subscriptions / live queries (`requestSnapshot` sends snapshots without re-filtering), while subsequent change events were evaluated correctly — producing inconsistent, timing-dependent state.

Fix: `OptimizationResult` gains an `isExact` flag. **OR** now requires every disjunct to be optimizable (missing rows can't be recovered by post-filtering), else falls back to a full scan. **AND** keeps partial index optimization but marks the result inexact when any conjunct was skipped; the compound-range path reports which conjuncts it covered so the rest are processed. `currentStateAsChanges` re-checks candidate rows against the full expression only when the result is inexact — fully-indexed queries take the same fast path as before.

### Commit 2 — range boundary fixes (from CodeRabbit review + what it surfaced)

- **Strictest bound at equal values** ([CodeRabbit finding](https://github.com/TanStack/db/pull/1582#pullrequestreview-4466646635)): `gte(x, 5) AND gt(x, 5)` kept the inclusive bound when the non-strict condition appeared first. Bounds are now merged strictness-aware, comparing values with `makeComparator` (the same comparison semantics the indexes use) so dates and locale strings behave correctly — `===` would silently fail for equal-valued `Date` instances.
- **One-sided compound ranges returned empty results**: the optimizer always passed `{from, to, ...}` to `rangeQuery`, but `rangeQuery` distinguishes an absent bound from an explicit `undefined` (which becomes the undefined-sentinel and produces an empty interval). Bounds are now only passed when they exist.
- **BTree exclusive bounds were broken for normalized values**: `rangeQuery`'s exclusive-lower-bound check compared the normalized indexed value (dates stored as timestamps) against the raw query `Date` — never equal, so plain `gt(createdAt, date)` included the boundary row. It now compares against the normalized key.

### PostgreSQL float semantics for `NaN` (supersedes #1617)

`NaN === NaN` is `false` in JavaScript, so a stored `NaN` had no consistent order — it couldn't be sorted or indexed deterministically, and `eq(x, NaN)` matched nothing while joins/`groupBy`/`distinct` (which key on a hash index) already treated `NaN = NaN`. Earlier commits on this branch worked around that defensively by giving `NaN`/invalid Dates a stable sort position *for sorting only* while the evaluator still rejected them and the optimizer re-filtered them out — correct, but a JS/SQL hybrid that left the filter-vs-join inconsistency in place.

This adopts one coherent, SQL-faithful model end-to-end. Following PostgreSQL — which overrides IEEE-754 precisely so floats can be sorted and indexed — `NaN` (and invalid `Date` values, whose timestamp is `NaN`) is now **equal to itself** and **greater than every other non-null value**:

- `ascComparator` orders `NaN`/invalid Dates as the greatest non-null value (was: alongside nulls).
- The `where` evaluator's `eq`/`gt`/`gte`/`lt`/`lte`/`in` implement the same.
- Because the index ordering and the evaluator now **agree**, the optimizer treats `NaN`/invalid Dates as **exact** (no re-filter), and invalid Dates are no longer treated as range-divergent. This resolves the reviewer's [invalid-Date `eq`/`IN` finding](https://github.com/TanStack/db/pull/1582#issuecomment-4800755889) — index-served and full-scan queries now return identical results.
- `null`/`undefined` are unchanged: still three-valued logic (a comparison with `null` yields `UNKNOWN`), still ordered by `NULLS FIRST`/`NULLS LAST`.

End-to-end consistency (WHERE, ordering, joins, `groupBy`, `distinct`) now all agree on `NaN`. This is a user-facing semantics change, documented under "Comparison semantics" in the live-queries guide, with a dedicated changeset.

> **Out of scope (tracked separately):** two pre-existing index-candidate-set issues surfaced in the same review thread — range predicates on a field with a heterogeneous value domain (e.g. arrays mixed with scalars), and content equality for `Uint8Array`s larger than 128 bytes — are not addressed here and will be handled in follow-up PRs.

### Testing

- The 7 expected-behaviour tests from #1581 pass.
- New `tests/nan-semantics.test.ts` asserts indexed == full-scan for `eq`/`gt`/`lt`/`in`/`orderBy` on `NaN` and invalid Dates; `tests/comparison.test.ts` and the evaluator suite cover the PG ordering.
- Full `@tanstack/db` suite: 102 files, 2422 tests passing (5 skipped), no type errors.
- Changesets included (patch — we don't cut minors before 1.0).

## Summary by CodeRabbit

* **New Features**
  * Query comparisons now follow PostgreSQL-like behavior for `NaN`, including matching, sorting, and range checks.
  * Comparison behavior for `null` and missing values is now documented more clearly.

* **Bug Fixes**
  * Fixed index-backed queries so mixed `AND`/`OR` conditions return the same results as full scans.
  * Improved range filtering for boundary cases, including strict bounds, `undefined`, invalid dates, and custom sort orderings.
  * Ensured results stay consistent when using indexed queries, compound ranges, and `IN` checks.


KyleAMathews:
### Code review

Found 3 issues:

1. Compound range optimization loses explicit `undefined` bounds while marking the result exact. `from`/`to` use `undefined` both as “no bound selected” and as a valid bound value, then `rangeOptions` only includes a bound when `from !== undefined` / `to !== undefined`. That conflicts with `rangeQuery`'s explicit distinction between an absent bound and an explicitly provided `undefined` bound. I verified this with a red test: `and(gt(value, undefined), lt(value, 10))` over an indexed field returned both the row with `value: 5` and the `value: undefined` boundary row, even though the optimizer treated the result as exact and skipped re-filtering. Please track `hasFromBound` / `hasToBound` separately from the bound values and use those booleans for comparison initialization and `rangeOptions` construction.

https://github.com/TanStack/db/blob/559a10df0894fc39f3776229d4723e4edcb61e71/packages/db/src/utils/index-optimization.ts#L288-L329

2. `BTreeIndex.rangeQuery` can apply the exclusive lower-bound exclusion even when no lower bound was provided. When `from` is absent, `fromKey` is set to `minKey()`, but the callback still drops rows equal to `fromKey` if `fromInclusive: false` is present. I verified this with a red test: `rangeQuery({ to: 10, fromInclusive: false })` on keys `1, 5, 10` returned only `5, 10`, dropping the minimum key despite there being no lower bound. The exclusion should be guarded by `hasFrom`, e.g. `if (hasFrom && !fromInclusive && ...)`.

https://github.com/TanStack/db/blob/559a10df0894fc39f3776229d4723e4edcb61e71/packages/db/src/indexes/btree-index.ts#L231-L255

3. `canOptimizeExpression` can still over-report optimizability for indexes that do not support the requested operation. The optimizer checks `index.supports(indexOperation)` for simple comparisons, and `in` falls back only when `in` or `eq` are supported, but the `canOptimize*` helpers only check that an index exists. I verified this with a red test using an `eq`-only index: `optimizeExpressionWithIndexes(or(gt(value, 1), lt(value, 10)), collection).canOptimize` was `false`, while `canOptimizeExpression(...)` returned `true`. Please align the helper predicates with the optimizer's operation-support checks.

https://github.com/TanStack/db/blob/559a10df0894fc39f3776229d4723e4edcb61e71/packages/db/src/utils/index-optimization.ts#L426-L449

https://github.com/TanStack/db/blob/559a10df0894fc39f3776229d4723e4edcb61e71/packages/db/src/utils/index-optimization.ts#L628-L647

kevin-dp:
@KyleAMathews Thanks for the review — all three findings are accurate. Here's how I've addressed each:

**1. Compound range loses explicit `undefined` bounds while marking the result exact** — Fixed. You're right that conflating `undefined`-as-"no bound" with `undefined`-as-a-bound-value is the core problem, so I now track `hasFromBound`/`hasToBound` separately from the values and use those for both the comparison initialization and `rangeOptions` construction.

There's also a subtlety worth calling out: a comparison against `null`/`undefined` is never true under the filter semantics, but in the index those values sort as the smallest key, so a range query fundamentally can't represent such a bound — even with correct bound tracking, `gt(field, undefined)` would still return index-ordered rows. So in addition to the bound tracking, the compound-range result is now marked `isExact: false` when any bound value is `null`/`undefined`, which forces the caller to re-filter against the full expression (matching a full scan). I also had to propagate that inexactness through `optimizeAndExpression`, which was previously ignoring the compound range's own exactness. Covered by a red→green test asserting `and(gt(score, undefined), lt(score, 90))` returns the same as a full scan (empty).

**2. `BTreeIndex.rangeQuery` applies the exclusive lower-bound exclusion even when no lower bound was provided** — Fixed exactly as suggested: the exclusion is now guarded by `hasFrom` (`if (hasFrom && !fromInclusive && ...)`). Good catch — this was a regression I introduced earlier in this PR when I changed the exclusion to compare against the normalized `fromKey` (which defaults to `minKey()` when no lower bound is given). Added a direct `rangeQuery({ to: 10, fromInclusive: false })` unit test as a regression guard.

**3. `canOptimizeExpression` can over-report optimizability for unsupported operations** — Acknowledged and accurate, but I'm intentionally leaving this one out of scope for this PR. It's pre-existing (the `canOptimize*` helpers have never checked operation support, independent of this PR's changes), and `canOptimizeExpression` has no internal callers — the actual query path goes through `optimizeExpressionWithIndexes`, which does check `index.supports(...)`. So there's no behavioral impact today; the only consequence of the over-report would be an attempted optimization that correctly falls back. Happy to tighten the helpers to mirror the optimizer's support checks in a follow-up.

Each of 1 and 2 has a separate reproduction commit (red) and fix commit so the failing tests are visible in history.

KyleAMathews:
### Code review follow-up

Found 1 additional issue:

1. `isExact` is still too optimistic for nullish indexed values. The full predicate evaluator rejects comparisons involving `undefined` (and `IN` with an `undefined` field value), but BTree indexes retain and return those keys. I verified this with red snapshot-vs-full-scan tests on the current head: `eq(score, undefined)` returned the `undefined` row from the indexed snapshot while full scan returned `[]`; `inArray(score, [undefined, 1])` returned `undefined` plus `1` while full scan returned only `1`; and `lt(score, 10)` returned the `undefined` row in addition to numeric matches. Since these paths all currently return `isExact: true`, `currentStateAsChanges` skips the full predicate re-check and leaks the false positives into snapshots. Please make nullish equality/IN results inexact, and conservatively mark range results inexact (or otherwise prove/exclude nullish candidates) so indexed snapshots stay equivalent to full predicate evaluation.

https://github.com/TanStack/db/blob/cd1ad9af93555caeb34c6abc58abf004ddb3fc2c/packages/db/src/utils/index-optimization.ts#L417-L435

https://github.com/TanStack/db/blob/cd1ad9af93555caeb34c6abc58abf004ddb3fc2c/packages/db/src/utils/index-optimization.ts#L621-L637

kevin-dp:
Thanks — agreed, this is a real leak and the same root cause as the earlier nullish-bound issue, just in the paths my previous fix didn't cover. A comparison against `null`/`undefined` evaluates to UNKNOWN in the predicate evaluator's three-valued logic (so the row is filtered out), but a BTree index stores and returns nullish-keyed rows (they sort as the smallest key) — so the index result is a superset and those paths must not claim `isExact: true`.

I reproduced all three reported cases (red snapshot-vs-expected tests in #1581), then fixed them here. While reproducing I also found a fourth instance of the same bug that wasn't reported: a compound **upper-bound-only** range (e.g. `lte(score, 90) AND lte(score, 95)`) leaks the nullish row too, because the open lower bound includes the bottom-sorted nullish keys. I covered that as well.

`isExact` is now computed as:

- **`eq`**: inexact when the query value is nullish.
- **`gt` / `gte`**: inexact when the bound is nullish; a non-null lower bound stays exact, since it sorts above the nullish keys and excludes them.
- **`lt` / `lte`**: conservatively inexact — the open lower bound always includes the bottom-sorted nullish rows.
- **`IN`**: inexact when any listed value is nullish.
- **compound range**: exact only when a non-null lower bound is present (`hasFromBound && !hasNullBound`); a to-only range is inexact.

In all inexact cases the index still narrows the candidate set, and `currentStateAsChanges` re-checks each candidate against the full expression, so indexed snapshots stay equivalent to a full predicate scan. Verified: the four reproduction tests are red in #1581 CI and green here, and the full `@tanstack/db` suite passes.

KyleAMathews:
Two more issues that reproduce with failing tests.

1. Locale string range optimization can violate the optimizer’s superset invariant. With `stringSort: 'locale', locale: 'en'`, a row `{ name: 'ö' }` satisfies the JS predicate `name > 'z' && name > 'ä'`, but the optimized index candidate set omits it. Since the row is omitted from `matchingKeys`, post-filtering cannot recover it. This is blocking correctness-wise.

One nuance: disabling only compound range collapsing may not be sufficient if simple locale-backed string range predicates still use BTree range lookup while the WHERE evaluator uses JS relational operators. A single `gt(name, 'z')` can have the same semantic mismatch. The robust fix is to either make range predicate evaluation use the same comparator as the index, or avoid locale-backed string range optimization for JS relational predicates.

Test

```ts
const compareOptions = {
  stringSort: 'locale',
  locale: 'en',
  direction: 'asc',
  nulls: 'first',
}

index.add('row-umlaut-o', { name: 'ö' })

const expression = and(
  gt(new PropRef(['name']), 'z'),
  gt(new PropRef(['name']), 'ä'),
)

expect('ö' > 'z' && 'ö' > 'ä').toBe(true)

const result = optimizeExpressionWithIndexes(expression, collection)

expect(result.canOptimize).toBe(true)
expect(result.matchingKeys.has('row-umlaut-o')).toBe(true)
```

2. `NaN` exactness also reproduces. For both `eq(score, NaN)` and `inArray(score, [NaN])`, the optimizer currently returns `isExact: true`, while evaluator equality semantics reject `NaN === NaN`. That can leak indexed rows because full-expression filtering is skipped. Marking normalized `NaN` values as inexact, or unifying equality semantics, fixes the reproduced case.

```ts
index.add('row-nan', { score: NaN })

const expression = eq(new PropRef(['score']), NaN)

expect(NaN === NaN).toBe(false)

const result = optimizeExpressionWithIndexes(expression, collection)

expect(result.canOptimize).toBe(true)
expect(result.isExact).toBe(false)
```

kevin-dp:
Thanks — both reproduce and are fixed. Reproductions are in #1581 (red), fixes here (green).

**2. NaN exactness** — straightforward, same superset class as the earlier nullish fix. `NaN` is never equal to itself in the evaluator, but the index still returns NaN-keyed rows (Map SameValueZero equality), so `eq`/`IN` against `NaN` now report `isExact: false` and the caller re-filters. Extended the existing exact-value check to treat `NaN` like a nullish value (also covers a NaN range bound in the compound path).

**1. Locale string range** — this one is a subset violation, so it needs different handling from the inexact/re-filter approach: re-filtering can't recover rows the index never returned. As you note, the mismatch is that a BTree string index orders with `localeCompare` (default `stringSort: 'locale'`) while the WHERE evaluator uses JS relational operators (code-point order), and there's no index range that's a superset of the JS-relational result under locale ordering.

I went with your second option — **avoid locale-backed string range optimization** — rather than changing the evaluator to use the index comparator. The reason: making range evaluation locale-aware would change query *semantics* for every string range predicate (indexed or not), which feels like a deliberate product decision rather than something to fold into this correctness fix. So `gt`/`gte`/`lt`/`lte` on a string value under locale collation now fall back to a full scan (`canOptimize: false`); `eq`/`IN` use exact equality and are unaffected, and lexical collation is untouched (its order matches the evaluator). This covers both the compound-collapsing path and simple single-predicate string ranges, per your nuance.

Tradeoff worth flagging: with the default locale collation, string range predicates no longer use the index. If locale-aware range *semantics* are actually desired, switching the evaluator to the index comparator (your first option) would let these stay indexed — happy to do that as a follow-up if that's the preferred direction.

My reproduction tests assert full-scan equivalence via `currentStateAsChanges` rather than the internal `canOptimize`/`matchingKeys` flags (since under this fix `canOptimize` is `false` for the locale case), but they encode the same requirement your snapshot-vs-full-scan tests do. Verified: the three new tests are red in #1581 and green here, and the full `@tanstack/db` suite passes.

KyleAMathews:
Thanks for the updates here — I red-tested the remaining indexing-correctness claims from the external review against `4b97808`, and I was able to reproduce them.

Confirmed issues:

1. **Range safety still only considers the query bound, not the indexed value domain.**
   A stored array value can satisfy the JS relational predicate while being outside the index range candidate set:

   ```ts
   type Row = { id: string; value: number | number[] }
   // stored row: { id: 'array', value: [0] }
   // query: lt(value, 10)
   // JS: [0] < 10 === true
   ```

   Red test result: `currentStateAsChanges({ where: lt(new PropRef(['value']), 10) })` returned `[]`, but should include `['array']`.

   So the range optimization needs to account for the indexed value domain as well as the bound value. A conservative fix is to disable range optimization when any indexed value has ordering semantics that diverge from the WHERE evaluator; a more precise fix would track/prove a range-compatible homogeneous domain for the index.

2. **Invalid Date equality/IN is still treated as exact.**
   Invalid Dates normalize to `NaN`, and Map lookup can find the stored invalid-date row, but the evaluator does not consider invalid Dates equal.

   Red tests:

   ```ts
   eq(ref.createdAt, new Date('invalid'))
   inArray(ref.createdAt, [new Date('invalid'), validDate])
   ```

   Results at `4b97808`:
   - `eq(invalidDate, invalidDate)` returned the invalid-date row, but should return `[]`.
   - `inArray([invalidDate, validDate])` returned `['invalid', 'valid']`, but should return only `['valid']`.

   The minimal fix is to make `isExactComparisonValue()` treat invalid Dates like `NaN`:

   ```ts
   if (value instanceof Date && Number.isNaN(value.getTime())) return false
   ```

3. **Related pre-existing issue: large Uint8Array equality with indexes.**
   I also confirmed the larger-binary mismatch. The evaluator compares `Uint8Array`s by content, but indexed equality uses normalized Map keys; arrays over 128 bytes are currently left as object keys, so two distinct 129-byte arrays with identical content miss through the index.

   Red test result: querying `eq(bytes, new Uint8Array(stored129ByteArray))` returned `[]`, but the full evaluator semantics say it should match the stored row.

   This one may be reasonable to track separately if you want to keep this PR scoped, but it is the same candidate-set invariant violation:

   ```txt
   full-scan matches ⊆ index candidates
   ```

I also locally checked that the above tests go green with the minimal invalid-date fix, a conservative stored-domain range fallback, and content-based normalization for large Uint8Arrays.

kevin-dp:
@KyleAMathews I had a look at your latest comment about the three issues:

**2. Invalid Date `eq`/`IN` — fixed ✅ (though via the opposite resolution to the minimal patch).**

Rather than making `isExactComparisonValue()` *reject* invalid Dates (the JS-semantics fix), we folded in the PostgreSQL float semantics from #1617: `NaN` and invalid Dates are now **equal to themselves and ordered as the greatest non-null value**, end-to-end across the evaluator, comparator, and index optimizer. The maintainers wanted one coherent SQL-faithful model rather than the JS/SQL hybrid — and it also removes the pre-existing filter-vs-join inconsistency (joins/`groupBy`/`distinct` already treated `NaN = NaN`).

So the candidate-set invariant you were testing now holds — index-served and full-scan results are identical — but the *expected values flip* from your red tests:

- `eq(createdAt, new Date('invalid'))` → `['invalid']` (you'd asserted `[]`)
- `inArray(createdAt, [new Date('invalid'), valid])` → `['invalid', 'valid']` (you'd asserted `['valid']`)

A new `nan-semantics.test.ts` asserts `indexed == full-scan` for both `NaN` and invalid Dates across `eq`/`gt`/`lt`/`in`/`orderBy`. #1617 is closed as superseded by this PR.

**1 & 3 — confirmed, but out of scope for this PR.**

I reproduced both, and they're real, but they're narrower corner cases that I don't think should ride along with this change:

- **Range safety vs. the indexed value domain** (your `lt(value, 10)` with a stored `[0]`) really hinges on a semantics question we haven't settled: what *should* a relational query over a field with a heterogeneous value domain (arrays mixed with scalars) even mean? The index-level "make it match the JS evaluator" fix is one answer, but I'd rather we agree on the intended contract first than encode it implicitly here.
- **Large `Uint8Array` equality** (`>128`-byte arrays falling back to identity keys) is a pre-existing limitation, unrelated to the optimizer changes, and similarly deserves its own discussion — content-normalize/hash large arrays vs. documenting the threshold as a known limitation is a real tradeoff, not an obvious fix.

Both are orthogonal to the `NaN`/optimizer work here, so I'm leaving them out to keep this PR focused, and we'll pick them up separately once we've discussed the semantics we want.

Net: **2 is resolved here; 1 and 3 are confirmed but deferred** pending a semantics discussion, so this PR stays scoped to the optimizer correctness + `NaN` semantics.

KyleAMathews (approved):
:shipit:

PR #1585, opened 2026-06-11
fix(powersync): Don't flush records when diff trigger isn't setup yet when using `on-demand` mode.

Chriztiaan:
## 🎯 Changes

Users reported an error that occurs during startup when using `on-demand`: `An error has been detected in the sync handler Error: no such table: __XXXX_tracking_6bed0e00`. During initialisation we were flushing records before the diff trigger was setup. 

This change simply skips flushing if the diff trigger hasn't been setup yet.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced on-demand sync log spam by avoiding premature diff flushes before tracking is initialized.
  * Prevented incorrect flush/disposal behavior during unload/disposal, avoiding attempts to flush tracking after it’s no longer valid.
  * Improved first-time subset loading to skip flush/dispose when sync tracking isn’t configured yet.


Chriztiaan:
@samwillis  - could you review this one please?

tannerlinsley:
Thanks for this. I pulled it into a stack with #1688 and ran the PowerSync package tests. The initial guard is pointed in the right direction, but `disposeTracking` stays truthy after the tracking table has been disposed, so unload/cleanup can still try to flush the deleted table and log `no such table` under PowerSync 2.

Can you clear the tracking state as part of disposal and add a regression test that fails on the old code, including the unload/cleanup path? Once that and CI are green, this should be good to merge.

Chriztiaan:
> Thanks for this. I pulled it into a stack with #1688 and ran the PowerSync package tests. The initial guard is pointed in the right direction, but `disposeTracking` stays truthy after the tracking table has been disposed, so unload/cleanup can still try to flush the deleted table and log `no such table` under PowerSync 2.
> 
> Can you clear the tracking state as part of disposal and add a regression test that fails on the old code, including the unload/cleanup path? Once that and CI are green, this should be good to merge.

Done, thanks for the feedback!

Issue #1178, opened 2026-01-23
react-db's useLiveQuery seems to leave an extra subscriber in StrictMode


tec27:
- [x] I've validated the bug against the latest version of DB packages

**Describe the bug**
In a pretty simple setup with a conditional query, I'm seeing 2 subscribers get created and only 1 get cleaned up when the component unmounts, which leaves the collection syncing permanently.

**To Reproduce**

1) Set up a collection:

```tsx
export const buildLogsCollection = createCollection<BuildLog>(
  electricCollectionOptions({
    id: "build-logs",
    shapeOptions: {
      url: `${BASE_URL}/shapes/build-logs`,
    },
    syncMode: "on-demand",
    getKey: (item) => item.id,
    gcTime: 1,
  }),
);

// Debug: track collection lifecycle
buildLogsCollection.on("status:change", (event) => {
  console.log("[build-logs] status:", event.status, "prev:", event.previousStatus);
});
buildLogsCollection.on("subscribers:change", (event) => {
  console.log(
    "[build-logs] subscribers:",
    event.subscriberCount,
    "prev:",
    event.previousSubscriberCount,
  );
});
```

2) Create a hook to use the collection data:

```tsx
export const useBuildLogs = (ownerId?: string) => {
  const { data } = useLiveQuery(
    (query) => {
      if (!ownerId) return undefined;

      return query
        .from({ log: buildLogsCollection })
        .where(({ log }) => eq(log.ownerId, ownerId))
        .orderBy(({ log }) => log.createdAt);
    },
    [ownerId],
  );
  return data;
};
```

3) Use the hook in a component:

```tsx
function App() {
  return <StrictMode><AppContent /></StrictMode>
}

function AppContent() {
  const [showLogs, setShowLogs] = useState(false);
  return <div>
    <button onClick={() => setShowLogs(v => !v)}>Toggle Logs</button>
    {showLogs && <BuildLogs />}
  </div>
}

function BuildLogs() {
    const logs = useBuildLogs("foo");
    return <div>{JSON.stringify(logs)}</div>
}
```

4) Run the app in development mode (so StrictMode is used)

**Expected behavior**
After toggling logs on and then back off, the `build-logs` URL should stop being retrieved once `gcTime` expires, and should log as having 0 subscribers.

**Desktop (please complete the following information):**

- Browser: Chrome
- Version: 144

**Additional context**
With StrictMode off I see the correct behavior (1 subscriber added, 1 subscriber removed)

[...]

Issue #690, opened 2025-10-17
useLiveQuery (react) triggers extra mount re-render when collection is already ready


john-rtr:
- [x] I've validated the bug against the latest version of DB packages

### Describe the bug

`useLiveQuery` (from `@tanstack/react-db`) forces a React re-render on mount even when the local collection is already `ready` and no data has changed. Additionally, the returned `data` reference changes, which prevents relying on referential stability. At minimum, when nothing changes, the `data` object reference should remain stable.

### To Reproduce

Reproduction is the following `TestRerender` component (local-only collection + live query with `startSync: true`):

```tsx
import { useEffect, useRef } from 'react'
import z from 'zod'
import {
  createCollection,
  createLiveQueryCollection,
  localOnlyCollectionOptions,
  useLiveQuery
} from '@tanstack/react-db'

const collection = createCollection(
  localOnlyCollectionOptions({
    id: 'local',
    getKey: (item: { id: string }) => item.id,
    schema: z.object({ id: z.string(), value: z.string() }),
    initialData: [
      { id: 'a', value: 'x' },
      { id: 'a', value: 'y' }
    ]
  })
)

const liveQuery = createLiveQueryCollection({
  startSync: true,
  query: (q) => q.from({ item: collection })
})

console.log('LiveQuery status before render:', liveQuery.status)

export function TestRerender() {
  const { data } = useLiveQuery(liveQuery)

  const renderCountRef = useRef(0)
  renderCountRef.current += 1

  useEffect(() => {
    console.log('renderCount', renderCountRef.current)
    prevDataRef.current = data
  })

  return null
}
```

### Expected behavior

- If the collection is already `ready` and no data has changed, mounting `useLiveQuery` should not trigger an additional render beyond the initial React mount.
- At minimum, the `data` object reference should remain stable if no data has changed.

### Actual behavior

- An extra render occurs on mount and the `data` reference changes even though no data has changed.

### Environment

- OS: macOS 15 (darwin 24.0.0)
- Framework: React 19.0.0
- @tanstack/db: 0.4.11
- @tanstack/react-db: 0.1.31
- Tooling: Vite 6.2.1, TypeScript 5.8.2

john-rtr:
Not sure, but this might be related [this line](https://github.com/TanStack/db/blob/3f4e154e7063f751da7d9a1f0945bd2c5258c239/packages/react-db/src/useLiveQuery.ts#L427C7-L431C8) that bumps version if collection is ready, even though it was already ready

[...]

Issue #493, opened 2025-09-03
Add Playwright tests for react-db useLiveQuery


samwillis:
React in the browser does concurrent rendering, this caused a bug in `useLiveQuery` where it would not show the initial state despite our tests checking this exact scenario. In order to truly validate the hook we need to test in a browser environment.

Relate issues #483 #437 that were fixed in #485 but without a test that reproduced the bug.

PR #1588, opened 2026-06-12
fix(react-db): defer eager onStoreChange to a microtask in useLiveQuery

### LINKED ISSUE #1587: react-db useLiveQuery can notify React before mount when collection is already ready

**Describe the bug**

`@tanstack/react-db`'s `useLiveQuery` can synchronously notify React's `useSyncExternalStore` subscriber before the component has committed when the collection is already `ready` during the initial mount.

In a React 19 + Vite app, this shows up as a development console error during a cold, CPU-throttled Lighthouse run:

> Can't perform a React state update on a component that hasn't mounted yet. This indicates that you have a side-effect in your render function that asynchronously tries to update the component. Move this work to useEffect instead.

The mechanism appears to be:

1. `useLiveQuery` calls `collection.startSyncImmediate()` during render when it receives a collection directly:
   - `packages/react-db/src/useLiveQuery.ts`, currently around lines 368-369.
2. In the `useSyncExternalStore` subscribe function, `useLiveQuery` eagerly calls `onStoreChange()` synchronously if `collectionRef.current.status === "ready"`:
   - `packages/react-db/src/useLiveQuery.ts`, currently around lines 431-445.
3. If `startSyncImmediate()` can resolve from already-warm data and flip the collection to `ready` before React commits, that eager `onStoreChange()` can land in the render-to-commit window.

This seems related to, but not the same as:

- #690 (`useLiveQuery` triggers extra mount re-render when collection is already ready)
- #1178 (`useLiveQuery` leaves an extra subscriber in StrictMode; comments discuss render-time sync)
- #493 (browser Playwright coverage for `useLiveQuery` concurrent rendering behavior)

**To Reproduce**

I have not reduced this to a standalone public reproduction yet, but the scenario is:

1. Use React 19 with `@tanstack/react-db@0.1.86` / `@tanstack/db@0.6.8`.
2. Pre-warm the data backing a set of live query collections so their sync path can complete immediately.
3. Cold-load a component tree that mounts one or more `useLiveQuery(collection)` calls.
4. Run with a slow/cold browser load or Lighthouse CPU throttling so there is a larger render-to-commit window.
5. Observe the React pre-mount state update warning in the browser console.

In the app where this was observed, the warning was captured by Lighthouse's `errors-in-console` audit during a cold desktop run. It does not reproduce consistently on warm interactive loads, which is why I suspect the race depends on the collection becoming `ready` before commit plus the eager synchronous subscribe notification.

**Expected behavior**

Mounting `useLiveQuery` against an already-ready collection should not synchronously trigger a React store update before the component has committed.

The initial ready state should be reflected through the initial snapshot, or any eager ready notification should be deferred until after the commit/subscription phase.

**Screenshots**

Not applicable.

- OS: macOS
- Browser: Chrome via Lighthouse 12 desktop run
- Package versions:
  - `@tanstack/react-db`: `0.1.86`
  - `@tanstack/db`: `0.6.8`
  - `react`: `19.2.7`

Not tested.

**Additional context**

One possible direction would be to avoid calling `onStoreChange()` synchronously from the initial subscribe path when the collection is already `ready` (for example by deferring it with `queueMicrotask`, or by moving the sync/ready notification into a post-commit subscription phase). I am not sure which option best preserves the intended behavior from #485/#690, but the synchronous notification appears to be what React is objecting to.

tsushanth:
## Summary

Closes #1587.

`useLiveQuery`'s `subscribeRef` calls `onStoreChange()` synchronously inside the `useSyncExternalStore` subscribe function when the underlying collection is already `ready`:

```ts
// Collection may be ready and will not receive initial `subscribeChanges()`
if (collectionRef.current.status === \`ready\`) {
  versionRef.current += 1
  onStoreChange()
}
```

That synchronous notification lands during the render-to-commit window when subscribe runs under StrictMode double-render or cold/throttled loads, which React surfaces as:

> Can't perform a React state update on a component that hasn't mounted yet. This indicates that you have a side-effect in your render function that asynchronously tries to update the component. Move this work to useEffect instead.

The reporter observed this consistently in a Lighthouse-cold-desktop run with React 19 + `@tanstack/react-db@0.1.86` + `@tanstack/db@0.6.8`.

## Fix

Defer the eager notification to a microtask so it lands after the current commit:

```diff
-      if (collectionRef.current.status === \`ready\`) {
-        versionRef.current += 1
-        onStoreChange()
-      }
+      if (collectionRef.current.status === \`ready\`) {
+        queueMicrotask(() => {
+          if (unsubscribed) return
+          versionRef.current += 1
+          onStoreChange()
+        })
+      }
```

While doing so, also guard the late-notify paths against an in-flight `subscribeChanges` callback firing after React unsubscribes. Track a local `unsubscribed` flag set by the cleanup, and drop both the eager microtask and any in-flight subscription event after teardown, so React never sees a state update post-unsubscribe:

```diff
+      let unsubscribed = false
+
       const subscription = collectionRef.current.subscribeChanges(() => {
+        if (unsubscribed) return
         versionRef.current += 1
         onStoreChange()
       })
       ...
       return () => {
+        unsubscribed = true
         subscription.unsubscribe()
       }
```

## API impact

None. The contract of `useLiveQuery` is preserved — an already-ready collection still notifies React once after mount, just asynchronously instead of mid-commit. Callers observe identical state transitions; the only difference is the timing of the *first* notification when the collection is already warm at hook-init.

## Test plan

- [x] `pnpm test` in `packages/react-db` — **94/94 pass, no type errors** (existing suite unchanged)
- [x] Coverage on `useLiveQuery.ts` is 96.47% lines / 95.65% branches after the change; the new microtask branch is hit by the existing tests that exercise the ready-collection path
- [ ] The original repro is a Lighthouse-cold-desktop run with React 19 + StrictMode double-render; existing tests don't reproduce that specific race directly (it's a render-to-commit window timing condition). The microtask deferral is correct by React's `useSyncExternalStore` contract regardless of repro

## Notes for review

- I considered using `Promise.resolve().then(...)` vs `queueMicrotask(...)`. Both land in the same microtask queue, but `queueMicrotask` is the more semantically explicit "do this after the current sync work" primitive. Either works.
- I deliberately did NOT switch to `useEffect`-based subscription, which the React warning message suggests, because the existing architecture commits to `useSyncExternalStore`. The microtask deferral is the minimal, contract-faithful fix.
- Related-but-distinct issues mentioned in #1587: #690 (extra mount re-render), #1178 (StrictMode extra subscriber), #493 (Playwright concurrent rendering coverage). This PR doesn't address those; it only fixes the pre-mount state-update warning case.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened subscription safety in live queries with guard checks to prevent notifications after unsubscription.
  * Modified state update timing for ready collection status, deferring updates through asynchronous task queuing.
  * Enhanced consistency in collection state notifications during component lifecycle transitions.


tsushanth:
Rebased onto current main.

tsushanth:
Rebased onto current main.

PR #1593, opened 2026-06-17
fix: safe randomUUID helper for non-secure browser contexts

### LINKED ISSUE #1541: @tanstack/db assumes crypto.randomUUID exists in non-secure browser contexts

kevin-dp:
Fixes #1541.

In non-secure browser contexts (e.g. a Vite dev server accessed via a LAN IP over plain HTTP), `crypto.randomUUID` is `undefined`, causing TanStack DB to throw when it tries to mint IDs. `crypto.getRandomValues` *is* still available in those contexts, so we can build a valid RFC 4122 v4 UUID from it as a fallback.

## Changes

- Add `randomUUID()` helper in `@tanstack/db` (`src/utils/uuid.ts`) that prefers `crypto.randomUUID()` and falls back to `crypto.getRandomValues` when unavailable. Throws if neither is available.
- Export it from the `@tanstack/db` public entry, and re-export from `@tanstack/db-sqlite-persistence-core` so the browser/electron adapters (which don't directly depend on `@tanstack/db`) can use it too.
- Replace every `crypto.randomUUID()` call site across the affected packages.

## Affected packages (one commit each)

- `@tanstack/db` — collection, mutations, local-only, local-storage, transactions
- `@tanstack/db-sqlite-persistence-core` — persisted.ts
- `@tanstack/browser-db-sqlite-persistence` — browser-coordinator.ts
- `@tanstack/electron-db-sqlite-persistence` — electron-coordinator.ts
- `@tanstack/offline-transactions` — OfflineExecutor, BroadcastChannelLeader, OfflineTransaction

## Tests

Added `packages/db/tests/uuid.test.ts` covering:
- delegation to `crypto.randomUUID` when available
- fallback to `crypto.getRandomValues` (verifies RFC 4122 version/variant bits)
- error when neither API is available

All affected package test suites pass locally.

A changeset (`patch` for each affected package) is included.

## Summary by CodeRabbit

* **New Features**
  * Introduced an enhanced UUID generation mechanism with automatic fallback support, deployed across database, transaction, offline, and persistence modules to ensure consistent identifier generation in non-secure contexts.

* **Tests**
  * Added comprehensive test coverage validating UUID generation behavior, fallback scenarios, uniqueness, and format compliance across multiple runtime conditions.

### DIFF
--- a/.changeset/safe-random-uuid.md
+++ b/.changeset/safe-random-uuid.md
@@ -1,9 +1,9 @@
 ---
-"@tanstack/db": patch
-"@tanstack/browser-db-sqlite-persistence": patch
-"@tanstack/offline-transactions": patch
-"@tanstack/db-sqlite-persistence-core": patch
-"@tanstack/electron-db-sqlite-persistence": patch
+'@tanstack/db': patch
+'@tanstack/browser-db-sqlite-persistence': patch
+'@tanstack/offline-transactions': patch
+'@tanstack/db-sqlite-persistence-core': patch
+'@tanstack/electron-db-sqlite-persistence': patch
 ---
 
 Use a safe `randomUUID` helper that falls back to `crypto.getRandomValues` when `crypto.randomUUID` is unavailable (non-secure browser contexts such as dev servers reached via a LAN IP over HTTP). Fixes #1541.
--- a/.changeset/safe-random-uuid.md
+++ b/.changeset/safe-random-uuid.md
@@ -1,0 +1,9 @@
+---
+"@tanstack/db": patch
+"@tanstack/browser-db-sqlite-persistence": patch
+"@tanstack/offline-transactions": patch
+"@tanstack/db-sqlite-persistence-core": patch
+"@tanstack/electron-db-sqlite-persistence": patch
+---
+
+Use a safe `randomUUID` helper that falls back to `crypto.getRandomValues` when `crypto.randomUUID` is unavailable (non-secure browser contexts such as dev servers reached via a LAN IP over HTTP). Fixes #1541.
--- a/packages/browser-db-sqlite-persistence/src/browser-coordinator.ts
+++ b/packages/browser-db-sqlite-persistence/src/browser-coordinator.ts
@@ -1,4 +1,4 @@
-import { randomUUID } from '@tanstack/db-sqlite-persistence-core'
+import { safeRandomUUID } from '@tanstack/db-sqlite-persistence-core'
 import type {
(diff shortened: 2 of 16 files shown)


KyleAMathews (approved):
:shipit:

PR #1597, opened 2026-06-18
fix(db): preserve discriminated union types through .select() (#1511)

### LINKED ISSUE #1511: types: `.select()` collapses discriminated union types on object fields

kevin-dp:
Fixes #1511.

## Problem

`.select()` collapses a discriminated-union field (e.g. `{ type: 'pdf'; url: string } | { type: 'image'; width: number }`) to the intersection of common keys. Two failing-on-main type tests in `packages/db/tests/query/select.test-d.ts` demonstrate both the shallow case and the nested case (union inside `payload: { inner: Payload }`).

## Root cause

Two compounding issues in `packages/db/src/query/builder/types.ts`:

1. `Ref<T, Nullable>` is a non-distributive mapped type. When `T` is a discriminated union, `[K in keyof T]` only iterates the common keys — the union is lost before per-key processing ever runs.
2. `ExtractRef<T>` projects the branded shape back through `ResultTypeFromSelect`, which drags the `VirtualPropsRef` intersection (`$synced`, `$origin`, `$key`, `$collectionId`) into the user-facing result whenever a Ref to a sub-object is selected. This is a pre-existing leak that the nested-union test makes visible.

## Fix

- Make `Ref<T, Nullable>` distribute over `T` via a `T extends unknown ? RefBranch<T, Nullable> : never` wrapper. `Ref<A | B>` becomes `Ref<A> | Ref<B>`, so `keyof` never sees the union.
- Reshape `ExtractRef<T>` to distinguish a *true* `Ref` (brand carrier whose own keys are exactly the brand symbols + virtual props + `keyof U`) from a *spread-produced inline object* (which inherits the brand symbol from the spread but also carries extra keys). True Refs return the underlying user type `U` directly; spreads fall through to the original `Prettify<ResultTypeFromSelect<WithoutRefBrand<T>>>` projection (preserving virtual props as the spread tests require).
- Add `DeepNullable<U>` so the `Nullable=true` flag (left/right/full joins) keeps propagating `| undefined` into every leaf, matching what the prior `ResultTypeFromSelect` recursion produced for nullable joins. This preserves all existing join-test behavior.

No changes outside `packages/db/src/query/builder/types.ts`. ~36 LOC.

## Why not approach (a) (PR #1512)

PR #1512 adds an `IsUnion`-gated shortcut in the middle of the `ResultTypeFromSelect` cascade. It makes the shallow test pass under vitest's typecheck runner, but `tsc --noEmit` still flags both new tests as failing because virtual props leak into the projected union, and the nested case still collapses. This PR fixes both cases under `tsc`.

## Verification

```
$ pnpm tsc --noEmit -p packages/db/tsconfig.json
# 0 errors in src/ or tests/ (one pre-existing unrelated `fractional-indexing`
# module-resolution error in the linked db-ivm package remains, identical on main)

$ pnpm vitest run --typecheck     # packages/db
Tests       867 passed (867)
Type Errors no errors
```

Both new union tests pass under both `tsc` and vitest. No regressions in the existing select / select-spread / join type-tests.

## IDE display

Hovering a discriminated-union Ref field now shows `Ref<A> | Ref<B> | Ref<C>` instead of the previous collapsed form. This is an improvement (it exposes the union the user wrote) but is a visible change for anyone who was reading the collapsed display.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `.select()` type inference so discriminated-union selections no longer collapse into the intersection of shared keys.
  * Improved preservation of discriminated unions in both top-level and nested selections.
  * Enhanced nullable discriminated-union handling to propagate nullability correctly through projections.
* **Tests**
  * Added/expanded type-inference checks to verify discriminated-union behavior, nested unions, and object spread projection outcomes.
* **Documentation**
  * Updated the release note entry describing the patch.


KyleAMathews:
Thanks for the fix here — the distributive `RefBranch` change looks like it addresses the direct/nested union cases from #1511. I found one blocking regression in the new `IsTrueRef` fast path around spread projections.

`IsTrueRef<T>` currently checks only that `T` has no keys outside the brand keys, nullable brand, virtual row props, and `keyof U`:

```ts
type IsTrueRef<T> =
  T extends RefLeaf<infer U>
    ? [
        Exclude<
          keyof T,
          | typeof RefBrand
          | typeof NullableBrand
          | keyof VirtualRowProps
          | keyof U
        >,
      ] extends [never]
      ? true
      : false
    : false
```

That proves “no extra keys”, but it does not prove that `T` is exactly the canonical `Ref<U>` shape. A spread-derived object can still have only keys from `keyof U` while having changed or removed fields. In those cases this branch returns `U` from `ExtractRef` and discards the actual projection.

I reproduced two failing type cases against this PR.

### 1. Same-key compatible/narrower override

```ts
type User = {
  id: number
  code: string | number
  slug: string
  nickname?: string
}

const col = createLiveQueryCollection((q) =>
  q.from({ u: users }).select(({ u }) => ({
    narrowed: { ...u, code: u.slug },
  })),
)

const result = col.toArray[0]!
expectTypeOf(result.narrowed.code).toEqualTypeOf<string>()
```

`u.slug` projects to `string`, so `result.narrowed.code` should be `string`. On this PR, the type assertion fails because the projection is classified as the original ref shape rather than recursively projected.

### 2. Omitted optional property is reintroduced

```ts
type User = {
  id: number
  code: string | number
  slug: string
  nickname?: string
}

const col = createLiveQueryCollection((q) =>
  q.from({ u: users }).select(({ u }) => {
    const { nickname, ...withoutNickname } = u
    return { trimmed: withoutNickname }
  }),
)

const result = col.toArray[0]!
type HasNickname = `nickname` extends keyof typeof result.trimmed ? true : false
expectTypeOf<HasNickname>().toEqualTypeOf<false>()
```

Since `nickname` was destructured out, `result.trimmed` should not have a `nickname` key. On this PR the assertion fails: `nickname` is still present because `withoutNickname` is treated as an untouched `Ref<User>`.

The existing spread regression test for `{ ...u.profile }` covers an unchanged spread, but it does not cover modified same-key fields or omitted optional fields.

I also verified locally that tightening `IsTrueRef` to a bidirectional/full-shape equivalence check against the canonical `Ref<U, IsNullableRef<T>>` shape makes these new regression tests pass. The exact implementation may be worth tuning for readability/compiler performance, but the important part is that key-set inclusion is not sufficient here — this needs to prove the projected member ref types and key presence match the canonical ref shape before `ExtractRef` can safely return `U`.

Could you add regression tests for the two cases above and update `IsTrueRef` to use a full-shape/equivalence check rather than the current subset/no-extra-keys check?

samwillis (approved):

kevin-dp:
@KyleAMathews I've pushed a fix that replaces the subset/no-extra-keys check in `IsTrueRef` with a strict structural-equivalence check against the canonical `Ref<U, IsNullableRef<T>>` shape:

```ts
type IsTrueRef<T> =
  T extends RefLeaf<infer U>
    ? RefShapeMatches<T, Ref<U, IsNullableRef<T>>> extends true
      ? true
      : false
    : false

// Strict structural equivalence (Equals idiom)
type RefShapeMatches<A, B> =
  (<G>() => G extends A ? 1 : 2) extends <G>() => G extends B ? 1 : 2
    ? true
    : false
```

A couple of notes on the choice of check:
- I went with strict equality rather than bidirectional assignability because assignability is blind to optional-key *presence* — your 2nd case (`const { nickname, ...withoutNickname } = u`) stays mutually assignable to the canonical shape, so a mutual-assignability check still reintroduces `nickname`. The `Equals` form is sensitive to key presence and handles it correctly.
- Genuine refs (direct refs, discriminated-union members) are exactly the canonical `Ref` shape, so they still take the fast path and unions are preserved. Anything spread-derived now falls through to `ResultTypeFromSelect`, which reconstructs the right type.

I also added the two regression tests you provided to `select.test-d.ts`. I verified they fail on the previous commit and pass after the fix, and that the full `db` package plus the dependent adapter packages (`query-db-collection`, `vue-db`, `angular-db`, `electric-db-collection`) all typecheck with 0 errors. Compile cost is negligible (~0.08% more instantiations, no `TS2589`).

PR #1600, opened 2026-06-18
fix: reconcile duplicate live query child inserts

### LINKED ISSUE #1599: TypeError: Cannot read properties of undefined (reading 'Symbol(liveQueryInternal)') when calling .update() on a queryCollection
- [ X ] I've validated the bug against the latest version of DB packages

**Describe the bug**
Calling .update() on a collection created with createCollection(queryCollectionOptions(...)) throws a TypeError at sync.ts:156 during the sync reconciliation step. The error occurs because the sync code unconditionally accesses utils[LIVE_QUERY_INTERNAL] on the collection config, but LIVE_QUERY_INTERNAL is only present on collections created with liveQueryCollectionOptions — not queryCollectionOptions.

Despite the error, the mutation does reach the server — changes persist on reload — so this appears to be a client-side reconciliation issue only.

**To Reproduce**
const surplusProductCollection = createCollection(queryCollectionOptions({ myConfig }));

const {data: modelDetails} = useLiveQuery(q => {
        if (!selectedModelId) {
            return undefined;
        }
        return q
        .from({model: surplusModelCollection})
        .where(({model}) => eq(model.id, selectedModelId))
        .join({product: surplusProductCollection}, ({product, model}) => eq(
            model.id, product.model_id
        ))
})
const product = modelDetails?.product;
// Calling .update() triggers the error
surplusProductCollection.update(product.id, (draft) => {
  draft.category_id = Number(value);
});

**Expected behavior**
.update() should work on queryCollection without throwing. The LIVE_QUERY_INTERNAL symbol access should be guarded — if the collection was not created with liveQueryCollectionOptions, the sync reconciliation path should treat hasCustomGetKey, hasJoins, and hasDistinct as false (or skip the check entirely).

**Error**
Uncaught TypeError: Cannot read properties of undefined (reading 'Symbol(liveQueryInternal)')
    at Object.write (sync.ts:156:36)
    at flushIncludesState (collection-config-builder.ts:1675:33)
    at Object.flushPendingChanges (collection-config-builder.ts:816:7)
    at CollectionConfigBuilder.maybeRunGraph (collection-config-builder.ts:381:21)
    at CollectionConfigBuilder.executeGraphRun (collection-config-builder.ts:578:10)
    at run (collection-config-builder.ts:506:23)
    at Scheduler.schedule (scheduler.ts:79:7)
    at CollectionConfigBuilder.scheduleGraphRun (collection-config-builder.ts:502:32)
    at CollectionSubscriber.sendChangesToPipeline (collection-subscriber.ts:176:34)
    at sendChanges (collection-subscriber.ts:189:12)
    at CollectionSubscription.callbackWithSentKeysTracking (subscription.ts:126:7)
    at CollectionSubscription.emitEvents (subscription.ts:331:12)
    at CollectionChangesManager.emitEvents (changes.ts:113:20)
    at CollectionStateManager.recomputeOptimisticState (state.ts:695:20)
    at CollectionMutationsManager.update (mutations.ts:455:11)
    at CollectionImpl.update (index.ts:764:28)
    at Object.onUpdate (WithDetailedModelCard.tsx:346:58)
    at handleSave (main.tsx:208:19)

- Device: PC
- OS: Windows
- Browser Chrome
- Package:     @tanstack/react-db
- Version 0.1.86
**Additional context**
Relevant source code

const utils = this.config.utils as Partial<LiveQueryCollectionUtils>
const internal = utils[LIVE_QUERY_INTERNAL]   // undefined for queryCollection
throw new DuplicateKeySyncError(key, this.id, {
  hasCustomGetKey: internal?.hasCustomGetKey ?? false,  // TypeError here
  hasJoins: internal?.hasJoins ?? false,
  hasDistinct: internal?.hasDistinct ?? false,
})

KyleAMathews:
Fix live-query includes reconciliation so `.update()` flows that re-emit an existing joined child row update the internal child collection instead of attempting a duplicate insert.

User-visible impact: updating a `queryCollectionOptions` source used in a live query with joined/included child rows no longer crashes with either the secondary `LIVE_QUERY_INTERNAL` `TypeError` or the underlying duplicate child insert error.

## Root Cause

The user action is `.update()`, but that update triggers live-query graph reconciliation. During reconciliation, `flushIncludesState` writes changes into internal child collections such as:

```text
__child-collection:detailed-surplus-agreement-line-items-product-30587
```

From that internal collection's perspective, a derived/joined row can be emitted as an `insert` even though the user operation was an update. When the child collection already contained that key, sync correctly rejected the duplicate insert:

```text
Cannot insert document with key "302935" from sync because it already exists in the collection "__child-collection:..."
```

There was also a secondary issue in the duplicate-key error path: generic/internal collection configs do not necessarily have live-query internals in `config.utils`, but `sync.ts` assumed they did while constructing `DuplicateKeySyncError` context.

## Approach

The fix has two parts:

1. In `flushIncludesState`, compute whether the internal child collection already has the child key before writing an insert-only child change. If the child already exists, write an internal `update` instead of an `insert`.

   ```ts
   const key = entry.syncMethods.collection.getKeyFromItem(change.value)
   const childAlreadyExists = entry.syncMethods.collection.has(key)

   if (change.inserts > 0 && change.deletes === 0) {
     entry.syncMethods.write({
       value: change.value,
       type: childAlreadyExists ? `update` : `insert`,
     })
   }
   ```

2. In the duplicate-key sync error path, guard optional `config.utils` before reading `LIVE_QUERY_INTERNAL`, preserving enhanced context when present and defaulting the flags when absent.

## Key Invariants

- Existing internal child rows must not receive duplicate sync inserts during live-query includes reconciliation.
- New internal child rows should still be inserted normally.
- Existing live-query-specific duplicate-key diagnostics should be preserved when `LIVE_QUERY_INTERNAL` is present.
- Collection configs without `utils` must not throw a secondary `TypeError` while constructing duplicate-key errors.

## Non-goals

- This does not change public `.update()` semantics.
- This does not change query collection refetch behavior.
- This does not redesign includes materialization or child collection identity.

## Trade-offs

The targeted fix keeps reconciliation local to the internal child collection write decision. A broader alternative would be changing upstream D2/includes change classification, but the child collection is the place with direct knowledge of whether a given child key is already materialized.

## Verification

```bash
cd packages/db
pnpm vitest run tests/collection.test.ts -t "config has no utils" --coverage.enabled=false
pnpm vitest run tests/query/includes.test.ts --coverage.enabled=false --pool-options.threads.maxThreads=2
```

```bash
cd packages/query-db-collection
pnpm vitest run tests/query.test.ts -t "duplicate insert into includes child" --coverage.enabled=false
```

CI is green for PR #1600, including:

- Test
- Run E2E Tests
- Build Example Site
- Preview

## Files changed

- `packages/db/src/collection/sync.ts`
  - Guard access to optional `config.utils` before reading `LIVE_QUERY_INTERNAL`.
  - Convert insert-only internal child changes to updates when the child key already exists.
- `packages/db/tests/collection.test.ts`
  - Add regression coverage for duplicate-key sync errors when collection config has no `utils`.
- `packages/db/tests/query/includes.test.ts`
  - Add coverage for joined includes child reconciliation.
- `packages/query-db-collection/tests/query.test.ts`
  - Add coverage for query collection update/refetch feeding a live-query includes child collection.
- `.changeset/fix-sync-duplicate-key-utils.md`
  - Add patch changeset for `@tanstack/db`.

---

Fixes #1599

## Bug Fixes
* Improved duplicate-key sync failure handling in live queries, preventing unexpected TypeErrors and ensuring the correct duplicate-key error is raised.
* Refined live query includes propagation for joined child collections to avoid duplicate child-row insertions during update flows.
* Improved child write behavior to correctly choose between inserts and updates based on whether the child already exists.

## Tests
* Added regression tests covering duplicate-key sync behavior and joined include update scenarios.
* Extended existing test assertions and imports to validate the new error and propagation behavior.

### DIFF
--- a/.changeset/fix-sync-duplicate-key-utils.md
+++ b/.changeset/fix-sync-duplicate-key-utils.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/db': patch
+---
+
+Fix duplicate-key sync reconciliation for collection configs without live query internals so it reports the intended duplicate key error instead of throwing a TypeError.
--- a/.changeset/fix-sync-duplicate-key-utils.md
+++ b/.changeset/fix-sync-duplicate-key-utils.md
@@ -2,4 +2,4 @@
 '@tanstack/db': patch
 ---
 
-Fix duplicate-key sync reconciliation for collection configs without live query internals so it reports the intended duplicate key error instead of throwing a TypeError.
+Fix live query includes reconciliation so updates that re-emit existing child rows update internal child collections instead of attempting duplicate inserts, and ensure duplicate-key sync errors handle collection configs without live query internals.
--- a/packages/db/src/collection/sync.ts
+++ b/packages/db/src/collection/sync.ts
@@ -151,9 +151,10 @@ export class CollectionSyncManager<
                   // throwing a duplicate-key error during reconciliation.
                   messageType = `update`
                 } else {
-                  const utils = this.config
-                    .utils as Partial<LiveQueryCollectionUtils>
-                  const internal = utils[LIVE_QUERY_INTERNAL]
+                  const utils = this.config.utils as
+                    | Partial<LiveQueryCollectionUtils>
+                    | undefined
+                  const internal = utils?.[LIVE_QUERY_INTERNAL]
                   throw new DuplicateKeySyncError(key, this.id, {
                     hasCustomGetKey: internal?.hasCustomGetKey ?? false,
                     hasJoins: internal?.hasJoins ?? false,
--- a/packages/db/src/query/live/collection-config-builder.ts
+++ b/packages/db/src/query/live/collection-config-builder.ts
@@ -1671,14 +1671,19 @@ function flushIncludesState(
             if (entry.orderByIndices && change.orderByIndex !== undefined) {
               entry.orderByIndices.set(change.value, change.orderByIndex)
             }
+            const key = entry.syncMethods.collection.getKeyFromItem(
+              change.value,
+            )
+            const childAlreadyExists = entry.syncMethods.collection.has(key)
+
             if (change.inserts > 0 && change.deletes === 0) {
-              entry.syncMethods.write({ value: change.value, type: `insert` })
+              entry.syncMethods.write({
+                value: change.value,
+                type: childAlreadyExists ? `update` : `insert`,
+              })
             } else if (
               change.inserts > change.deletes ||
-              (change.inserts === change.deletes &&
-                entry.syncMethods.collection.has(
-                  entry.syncMethods.collection.getKeyFromItem(change.value),
-                ))
+              (change.inserts === change.deletes && childAlreadyExists)
             ) {
               entry.syncMethods.write({ value: change.value, type: `update` })
             } else if (change.deletes > 0) {
--- a/packages/db/tests/collection.test.ts
+++ b/packages/db/tests/collection.test.ts
@@ -51,7 +51,7 @@ describe(`Collection`, () => {
   it(`throws DuplicateKeySyncError instead of TypeError when config has no utils`, async () => {
     let begin!: () => void
     let write!: Parameters<
-      SyncConfig<{ id: number; text: string }>[`sync`]
+      SyncConfig<{ id: number; text: string }, number>[`sync`]
     >[0][`write`]
 
     const collection = createCollection<{ id: number; text: string }, number>({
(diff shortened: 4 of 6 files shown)


mwalkersigma:
Installed : https://pkg.pr.new/@tanstack/react-db@c183272


This seems to done the trick i no longer see the error in the console and updates are working again thanks for the quick response!!

kevin-dp (approved):
Good fix!

PR #1603, opened 2026-06-19
feat(offline-transactions): confirm writes off the serial drain path (confirmWrite hook)

### LINKED ISSUE #1602: offline-transactions: confirm writes off the serial drain (hold optimistic state through the post-commit sync window)
## Summary

`@tanstack/offline-transactions` drains the outbox **serially** — one `mutationFn` at a time — which is what preserves create-then-update / FK ordering. But there is no built-in way to keep a row's optimistic state painted through the **gap between "the server committed the write" and "the sync stream echoed it back."** Both available workarounds are bad, and this hurts real apps using an async confirmation stream (e.g. ElectricSQL's `awaitTxId`).

## The gap

The executor drops a transaction's optimistic overlay the instant its `mutationFn` resolves (`resolveTransaction` → `cleanupRestorationTransaction`). If the synced data hasn't caught up yet, the row **flickers** (disappears, then reappears when sync delivers it).

To avoid the flicker today, you must `await` the confirmation **inside** the `mutationFn`:

```ts
mutationFn: async ({ transaction }) => {
  const { txid } = await postToServer(transaction)
  await collection.utils.awaitTxId(txid) // runs ON the serial path
}
```

But the drain is serial, so this `await` blocks the **next** write. With an Electric shape stream whose `awaitTxId` budget is ~10s, drain throughput collapses to ~1 write / 10s, and each settled write tends to spawn a duplicate "already-applied" resend. So you're forced to choose between **slow drain** (await inline) and **UI flicker** (don't await). Neither is acceptable.

Notably, the library *already* has the right primitive — `TransactionExecutor.restoreOptimisticState` creates a standalone `createTransaction({ autoCommit: false })`, applies the mutations, registers it on each collection's `_state`, and `cleanupRestorationTransaction` tears it down — but it's private and only wired for *rehydrate-on-load*, not for the post-commit confirm window.

## Proposed API

An opt-in `OfflineConfig.confirmWrite` hook that runs **after** the write commits and its outbox entry is removed, but **off** the serial drain path:

```ts
startOfflineExecutor({
  // ...
  confirmWrite: async ({ mutations, result }) => {
    // result = whatever your mutationFn returned (e.g. a server txid)
    await Promise.all(
      collectionsOf(mutations).map((c) => c.utils.awaitTxId(result.txid)),
    )
  },
})
```

While the returned promise is pending, the library keeps the committed mutations' optimistic overlay painted (reusing the existing hold primitive) and releases it when the hook settles. The serial chain still serializes the **POSTs** (ordering preserved); only the confirmation moves off it.

Key semantics:

- **Never rolls back** — the write is already durably committed, so a rejected/timed-out hook just releases the overlay early (possible brief flicker), never data loss. Timeout / verify-by-state logic lives inside the hook.
- **Never throws into the drain** — a throw can't make the executor retry an already-committed write.
- **No flicker** — the hold is registered synchronously, before the original overlay is dropped.
- **Bounded** — a `maxConfirmationHolds` cap avoids O(n²) optimistic-recompute churn on a large, fast drain.
- **Fully opt-in** — with no `confirmWrite`, behavior is unchanged.

## Context

We hit exactly this in production (offline-first app on Electric). We currently work around it with an external module that reaches into `collection._state` and re-implements the library's private hold/teardown — fragile, and it duplicates internals. A first-class hook would let that module collapse to the small callback above.

I have a PR ready that implements this (new `confirmWrite`/`maxConfirmationHolds`/`getActiveConfirmationHoldCount`, the create/release primitive factored into `OptimisticHold.ts` and shared with `restoreOptimisticState`, tests, changeset). Happy to bikeshed the hook name (`confirmWrite` / `awaitSync` / `confirmTransaction`) and the default cap. Will link it here.

TomasGonzalez:
Closes #1602

## 🎯 Changes

Adds an opt-in **`OfflineConfig.confirmWrite`** hook to `@tanstack/offline-transactions` that keeps a just-committed write's optimistic state painted across the **post-commit confirmation window** — running **off** the serial drain path.

### Motivation

The executor drains the outbox **serially** (one `mutationFn` at a time), which is what preserves create-then-update / FK ordering. But there's no built-in way to keep a row painted between *"the server committed the write"* and *"the sync stream echoed it back"*. The only option today is to `await` that confirmation **inside** the `mutationFn` — e.g.:

```ts
mutationFn: async ({ transaction }) => {
  const { txid } = await postToServer(transaction)
  await collection.utils.awaitTxId(txid) // ⛔ runs ON the serial path
}
```

Because the drain is serial, that `await` blocks the **next** write. With an ElectricSQL shape stream whose `awaitTxId` budget is ~10s, throughput collapses to ~1 write / 10s and each settled write tends to spawn a duplicate "already-applied" resend. The alternative — dropping the `await` — makes rows **flicker** (disappear then reappear) in the commit→sync gap, because the executor drops the transaction's optimistic overlay the instant `mutationFn` resolves.

### What this adds

`confirmWrite` runs **after** the write commits and its outbox entry is removed, but **off** the serial path — it never blocks the next `mutationFn`:

```ts
startOfflineExecutor({
  // ...
  confirmWrite: async ({ mutations, result }) => {
    // result is whatever your mutationFn returned (e.g. a server txid)
    await Promise.all(
      collectionsOf(mutations).map((c) => c.utils.awaitTxId(result.txid)),
    )
  },
})
```

While the returned promise is pending, the library keeps the committed mutations' optimistic overlay painted via an internal **hold transaction** — the *same* primitive `restoreOptimisticState` already uses to re-show pending writes after a reload — then releases it when the hook settles. The serial chain still serializes the **POSTs** (ordering preserved); only the confirmation moved off it.

Design points:

- **Never rolls back.** The write is already durably committed, so a rejected/timed-out `confirmWrite` just releases the overlay early (a possible brief flicker), never data loss. Any timeout / verify-by-state logic lives inside the hook.
- **Never throws into the drain.** Hold creation and the hook are fully guarded; a throw can't make the executor retry an already-committed write.
- **No flicker.** The hold is registered *synchronously*, before `resolveTransaction` drops the original overlay, so the overlay is owned continuously.
- **Bounded.** `maxConfirmationHolds` (default `1000`) caps concurrent holds to avoid O(n²) optimistic-recompute churn on a large, fast drain; beyond the cap the hold is skipped (degrades to today's behavior). `getActiveConfirmationHoldCount()` exposes the live count.
- **Released on teardown.** Holds are dropped on `clear()` and `dispose()`.

Internals: the hold create/release pattern is factored into a new `executor/OptimisticHold.ts` and reused by `restoreOptimisticState`, so there's a single primitive. The hook is fully **opt-in** — with no `confirmWrite` configured, behavior is byte-for-byte unchanged.

### Drive-by fix

`runMutationFn` previously `await`ed the `mutationFn` and **discarded** its return value, so `waitForTransactionCompletion` always resolved `undefined`. The result is now threaded through to the completion promise and to `confirmWrite` (so the hook can read e.g. a server txid). One existing e2e assertion was updated to reflect this.

### Context

This generalizes a pattern we shipped app-side at Autarc (an external `offlineConfirmationHolds.ts` that reaches into `collection._state` and re-implements the library's private hold/teardown). With this hook, that ~360-line module collapses to the small `confirmWrite` callback above. Happy to bikeshed the name (`confirmWrite` / `awaitSync` / `confirmTransaction`) and the default cap.

## ✅ Checklist

- [x] I have tested this code locally with `pnpm test`. (`@tanstack/offline-transactions`: 71 passing + 6 new in `tests/confirm-write.test.ts`, 1 pre-existing skip; full suite green. `eslint src` and `prettier --check` clean. `tsc` introduces zero new errors vs `main`.)

## 🚀 Release Impact

- [x] This change affects published code, and I have generated a changeset (`.changeset/confirm-write-hook.md`, `minor`).

## Summary by CodeRabbit

* **New Features**
  * Added optional `confirmWrite` post-commit hook to control when optimistic overlays are released.
  * Introduced `maxConfirmationHolds` to cap concurrent confirmation overlays.
  * Added `getActiveConfirmationHoldCount()` for visibility into active confirmation holds.
  * Confirmation handling was decoupled from the serial outbox drain to improve write throughput; optimistic overlays persist through the confirmation window.

* **Tests**
  * Added/expanded Vitest coverage for `confirmWrite`, hold limits, clear/dispose behavior, and non-blocking confirmations.

* **Documentation**
  * Updated README and skill docs to describe the new post-commit confirmation flow and configuration options.


KyleAMathews:
Thanks for the detailed implementation and evidence. We are closing this PR for the pre-1.0 cleanup because the off-drain confirmation design still needs a supported core optimistic-hold ownership contract. The current approach adds confirmWrite, hold lifecycle/cap/introspection semantics, and cross-package private-state coupling before those guarantees are settled. Issue #1602 and the offline RFC remain the durable place to choose that contract; if the RFC selects it, we can reuse the tests and prior art here in a fresh implementation.

Issue #1571, opened 2026-06-04
Include value updates break with solidjs


Tarmean:
- [x] I've validated the bug against the latest version of DB packages (@tanstack/solid-db@0.2.1 with @tanstack/db@0.6.7)

**Describe the bug**

In solid-db's useLiveQuery, toArray includes are not reactive. For query includes, if the initial include is empty it starts as a non-observable null value. In this case the child query is also not reactive.

**To Reproduce**

- Create two empty collections
- Create a query which has an include between them
- Create a solidjs subscription on the query
- Fill the empty collections (not in a single solidjs batch so solid sees an intermediate value)

For a query with an toArray include the initial collection does not have to be empty, the include is never reactive. 

**Expected behavior**
That solidjs updates when the include changes

**Additional context**

flushIncludesState clones the rows for the events.
   
But solidjs reads  currentCollection.values() on any update, and these rows are updated in place. Since the pointers are stable, reconcile prunes them and nothing downstream sees the update.
This is relevant when the update is a value (e.g. toArray), or if the initial include collection was empty. In that case the include field starts as null and solidjs doesn't see the change to a non-null reference.

https://github.com/TanStack/db/blob/be656be5c5fa65cc1c2bf76bc0b60be3d9d2a577/packages/solid-db/src/useLiveQuery.ts#L349

If an include is initially empty, the proxy is not attached to the parent. It remains null. This is why the initial child field can be null, breaking reactivity for collection includes as well.

https://github.com/TanStack/db/blob/be656be5c5fa65cc1c2bf76bc0b60be3d9d2a577/packages/db/src/query/live/collection-config-builder.ts#L1851-L1867

KyleAMathews:
Closing as completed by https://github.com/TanStack/db/pull/1785 (https://github.com/TanStack/db/commit/a378bd3a65f6b9ed0c9a85f793b7dc2e2a59a313). Solid now republishes rendered `toArray` child-value changes and reacts when an initially empty Collection-valued include receives its first row. Both cases have permanent regression coverage.

Issue #1524, opened 2026-05-12
solid-db rendering issues when not using key named `id`


fehmer:
- [x] I've validated the bug against the latest version of DB packages

**Describe the bug**
Using a queryCollection with items not having an `id` property causing problems on rendering.

**To Reproduce**

Full demo available at https://github.com/fehmer/tanstack-demo/blob/main/src/CollectionIssue.tsx

1. create a queryCollection with items of type `{_id:string, name:string}`
2. create a live query based on this collection, use `orderBy` to order by name
3. display the content of the collection using  `<For each={query()}>`
4. rename the last item with a name to be ordered on the top
5. notice all items are shown with name Alvin

Collection definition:
```tsx
const collection = createCollection(
  queryCollectionOptions({
    queryKey: ["collection"],
    queryClient,
    getKey: (it) => it._id,
    queryFn: async () => {
      return [
        { name: "Bob", _id: "bob1" },
        { name: "Kevin", _id: "kevin1" },
        { name: "Stuart", _id: "stuart1" },
      ] as Item[];
    },
  }),
);
```

Live query:
```tsx
function useItemsLiveQuery() {
  return useLiveQuery((q) => {
    return q.from({ item: collection }).orderBy(({ item }) => item.name, "asc");
  });
}
```

Render:
```tsx
export const CollectionIssue: Component = () => {
  const items = useItemsLiveQuery();
  const idItems = useIdItemsLiveQuery();
  return (
    <>
      <h2>Collection with _id</h2>
      <ol>
        <For each={items()}>{(item) => <li>{item.name}</li>}</For>
      </ol>

      <button
        onClick={() => {
          void updateName({ _id: "stuart1", name: "Alvin" });
        }}
      >
        rename stuart
      </button>
    </>
  );
};



const updateName = createOptimisticAction<Item>({
  onMutate: ({ _id, name }) => {
    collection.update(_id, (tag) => {
      tag.name = name;
    });
  },
  mutationFn: async ({ _id, name }) => {
    collection.utils.writeUpdate({ _id, name });
  },
});

```

It does work fine in case the `Item` has a property called `id` without underscore. 

***Screenshots***
before rename

<img width="258" height="300" alt="Image" src="https://github.com/user-attachments/assets/1a5b2adf-7d60-4041-9832-4349b41a4139" />

after rename

[...]

PR #1604, opened 2026-06-21
fix(solid-db): clone rows in syncDataFromCollection so reconcile sees in-place mutations

tsushanth:
Fixes part of #1571.

## What

`useLiveQuery`'s rendered `data` array missed updates whenever the underlying collection mutated a row in place. The upstream pipeline does exactly that — `flushIncludesState` in [`packages/db/src/query/live/collection-config-builder.ts`](https://github.com/TanStack/db/blob/main/packages/db/src/query/live/collection-config-builder.ts) writes new `toArray` include values onto the parent row via `setIncludedValue(parentResult, …)` without changing its identity. When `syncDataFromCollection` then called

```ts
reconcile(Array.from(currentCollection.values()))(prev).filter(Boolean)
```

Solid's `reconcile` short-circuited on `next === target` at the parent-row level (and again on the include array reference if reached). The reactive `state` `ReactiveMap` still updated correctly — `subscribeChanges` hands `state.set` a fresh row reference per change event — but the `data` array fell out of sync.

The reporter @Tarmean documented this for `toArray` includes specifically, with `flushIncludesState` cloning rows for the events but mutating the rows the collection itself returns from `.values()`.

## How

`syncDataFromCollection` now shallow-clones each row (and any array-valued field on it) before handing the snapshot to `reconcile`. The wrapper is small and lives next to the sync helper so the rationale is local:

```ts
const cloneRowForReconcile = <T,>(v: T): T => {
  if (v === null || typeof v !== 'object' || Array.isArray(v)) return v
  const source = v as Record<string, unknown>
  const out: Record<string, unknown> = {}
  for (const key in source) {
    const val = source[key]
    out[key] = Array.isArray(val) ? [...val] : val
  }
  return out as T
}
```

Field-level diffing then actually runs and `toArray` include arrays propagate. Non-object values (primitives) pass through; arrays at the top level pass through (`reconcile` already iterates them); object values get a one-level copy with arrays inside them also copied so the include-array reference changes.

The `state` `ReactiveMap` path is untouched — `subscribeChanges` was already getting fresh row references per change event, so only the `Array.from(currentCollection.values())` sync needed help.

## Tests

Adds `toArray include reactivity (#1571)` describe block in `packages/solid-db/tests/useLiveQuery.test.tsx`:

- Set up two collections (`projects`, `issues`) and a `useLiveQuery` that projects `{ id, name, issueTitles: toArray(...) }`.
- Insert a new child issue tied to project Alpha.
- Assert the parent row's `issueTitles` array grows from `['Bug in Alpha']` to `['Bug in Alpha', 'Feature for Alpha']` in the rendered `data` array (not just in the `state` map).
- Asserts Beta's include is untouched.

Without the clone, the rendered `data` array stays at the original include content; the test would fail on `expect(alphaAfter.issueTitles.length).toBe(2)`. All 39 tests in `packages/solid-db/tests/useLiveQuery.test.tsx` pass. Lint clean (per AGENTS.md).

## Scope

This addresses the **clone-in-reconcile half** of #1571. @Tarmean's report flagged a second, related case:

> If an include is initially empty, the proxy is not attached to the parent. It remains null. This is why the initial child field can be null, breaking reactivity for collection includes as well.

That one lives at `packages/db/src/query/live/collection-config-builder.ts:1851-1867` and is a different code path (proxy attachment, not consumer-side reconcile). Happy to file it as a separate issue and take a look if maintainers agree it's the same bug class.

## Notes for maintainers

- The clone-on-sync trades a small allocation per row per sync for correctness. For typical UIs (hundreds of rows) this is far cheaper than the missed-update bug. If the perf trade-off is unwelcome, the alternative is to fix this upstream in `flushIncludesState` so include writes produce new parent row identities — happy to switch directions on guidance.
- I considered using `state.values()` instead of `currentCollection.values()` (since `state` already holds fresh references per `subscribeChanges` event), but `state`'s insertion order doesn't match the collection's sort order on inserts, so the `data` array would drift. Cloning preserved the sort order from `currentCollection.values()`.

## Changeset

`@tanstack/solid-db: patch`

## Bug Fixes
- Fixed live queries not updating correctly when included related data changed (such as after inserting a new child entry). Included fields now refresh reliably so the displayed arrays reflect the most recent matching records.
- Added regression coverage to ensure `toArray`-based includes update as expected, including cases where an initially empty include expands after the first matching child insert.

### DIFF
--- a/.changeset/solid-db-reconcile-clone-rows.md
+++ b/.changeset/solid-db-reconcile-clone-rows.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/solid-db": patch
+---
+
+Fix `useLiveQuery` not propagating `toArray` include changes. When the upstream collection mutated parent rows in place (as it does when flushing `toArray` include values), Solid's `reconcile` short-circuited on the stable parent reference and the rendered `data` array kept showing the old include content. `syncDataFromCollection` now shallow-clones each row (and any array-valued field on it) before handing it to `reconcile`, so field-level diffing actually runs.
--- a/packages/solid-db/src/useLiveQuery.ts
+++ b/packages/solid-db/src/useLiveQuery.ts
@@ -341,12 +341,35 @@ export function useLiveQuery(
     },
   )
 
-  // Helper to sync data array from collection in correct order
+  // Helper to sync data array from collection in correct order.
+  //
+  // `reconcile` short-circuits on `next === target` (both at the array-item
+  // level and at the field level), so when the collection mutates a row in
+  // place — as it does for `toArray` include values flushed by
+  // `flushIncludesState` in `packages/db/src/query/live/collection-config-builder.ts`
+  // — downstream consumers never see the change. Shallow-clone each row (and
+  // any array-valued field on it) so `reconcile` recurses into the field-level
+  // diff instead of bailing on the stable parent / include reference.
+  // Tracking issue: https://github.com/TanStack/db/issues/1571
+  const cloneRowForReconcile = <T,>(v: T): T => {
+    if (v === null || typeof v !== `object` || Array.isArray(v)) {
+      return v
+    }
+    const source = v as Record<string, unknown>
+    const out: Record<string, unknown> = {}
+    for (const key in source) {
+      const val = source[key]
+      out[key] = Array.isArray(val) ? [...val] : val
+    }
+    return out as T
+  }
   const syncDataFromCollection = (
     currentCollection: Collection<any, any, any>,
   ) => {
     setData((prev) =>
-      reconcile(Array.from(currentCollection.values()))(prev).filter(Boolean),
+      reconcile(
+        Array.from(currentCollection.values(), cloneRowForReconcile),
+      )(prev).filter(Boolean),
     )
   }
 
--- a/packages/solid-db/tests/useLiveQuery.test.tsx
+++ b/packages/solid-db/tests/useLiveQuery.test.tsx
@@ -2544,6 +2544,62 @@ describe(`Query Collections`, () => {
       }
       expect(beta.issueTitles).toEqual([`Bug in Beta`])
     })
+
+    it(`should update when a toArray include starts empty and gets first child`, async () => {
+      // Empty-include variant — the parent's `issueTitles` array starts at
+      // `[]` and must reflect a first child insert. The clone-on-sync path
+      // has to widen the existing empty array to a new one so reconcile sees
+      // the change.
+      const projects = createCollection(
+        mockSyncCollectionOptions<IncludeProject>({
+          id: `toarray-include-projects-empty`,
+          getKey: (p) => p.id,
+          initialData: [{ id: 3, name: `Gamma` }],
+        }),
+      )
+      const issues = createCollection(
+        mockSyncCollectionOptions<IncludeIssue>({
+          id: `toarray-include-issues-empty`,
+          getKey: (i) => i.id,
+          initialData: [],
+        }),
+      )
+
+      const rendered = renderHook(() => {
+        return useLiveQuery((q) =>
+          q.from({ p: projects }).select(({ p }) => ({
+            id: p.id,
+            issueTitles: toArray(
+              q
+                .from({ i: issues })
+                .where(({ i }) => eq(i.projectId, p.id))
+                .select(({ i }) => i.title),
+            ),
+          })),
+        )
+      })
+
+      await waitFor(() => {
+        const gamma = rendered.result().find((r: any) => r.id === 3) as
+          | { issueTitles: Array<string> }
+          | undefined
+        expect(gamma?.issueTitles).toEqual([])
+      })
+
+      issues.utils.begin()
+      issues.utils.write({
+        type: `insert`,
+        value: { id: 30, projectId: 3, title: `First Gamma issue` },
+      })
+      issues.utils.commit()
+
+      await waitFor(() => {
+        const gamma = rendered.result().find((r: any) => r.id === 3) as {
+          issueTitles: Array<string>
+        }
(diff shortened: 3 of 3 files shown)


tsushanth:
Thanks @coderabbitai — added the empty-include variant in [95d2d407](https://github.com/TanStack/db/pull/1604/commits/95d2d4079c0e2e9d8e75e7a16c9c40fcb45a4f81). Parent starts with `issueTitles: []`, child insert lands, parent's array updates. Both #1571 cases (populated-then-grow and empty-then-first-child) now covered.

Tarmean:
I think this change maybe should wait until this PR was merged? https://github.com/TanStack/db/pull/1598

I'm not super deep in solidjs or tanstack-db, but my understanding of reconcile is as follows:


- It matches (pointer- or id-) equal  common prefixes and suffixes
- Then, if we have an id key, match the rest by the id key. It defaults to 'id' and checks if the first new row has a non-null property with that name
- Otherwise, if we pass merge=true to reconcile (we don't), match the rest by index
- Otherwise, don't match

Problem: this change would intentionally remove all pointer equality. So if the row doesn't happen to have an 'id' field, we lose all incrementalization and solid would fully tear down+recreate For blocks.

> Behavior
> 
>     reconcile is for reconciling an incoming snapshot against existing store state.
>     With the default key, array items are matched by "id" where possible and otherwise fall back to positional or reference matching.
>     Non-wrappable values are replaced directly.
>     When merge is false, reconciliation prefers replacing non-matching branches. When merge is true, it pushes updates deeper into the structure.

https://docs.solidjs.com/reference/store-utilities/reconcile

Either way, thank you for your work on this!

tsushanth:
Thanks for the careful read @Tarmean — you're right that the clone gives up row-level pointer equality, and with the current default reconcile key (`'id'`) that's a real downside for tables whose rows don't have an `id` field.

The reason I went with the clone is that `reconcile`'s `next === target` short-circuit fires at the **row** level too: if the include flush mutates `parent.issueTitles` in place without changing the parent reference, the field never reaches subscribers no matter what reconcile does deeper. So fixing this in solid-db alone seems to force *some* identity break per row.

#1598's `key: '$key'` does address my concern about the clone (matching survives the new reference), so if it lands first this PR becomes mostly compatible. If you'd rather not have a window where both PRs disagree on key strategy, I'm happy to either (a) park this one until #1598 merges and rebase on top, or (b) instead chase the root cause in `flushIncludesState` so the collection produces a new row reference whenever an include changes — letting reconcile bail at the parent untouched. Let me know which fits the roadmap better.

tsushanth:
Thanks @Tarmean — your reading of reconcile's matching is right, and it's worth being precise about whether #1604 stands on its own, gates on #1598, or should be rethought.

The question this turns on: does #1598 (`key: '$key'`) alone fix the toArray-include problem in #1571?

I don't think it does, and that's the case for keeping both. The reproduction in #1571 is:

1. The live collection holds row `R = {..., issueTitles: [...]}`.
2. `flushIncludesState` (in `packages/db/src/query/live/collection-config-builder.ts`) calls `setIncludedValue(storedParent, resultPath, childValue)` — it mutates the stored parent row's include field in place (`storedParent !== parentResult` short-circuit at line 1622 only skips redundant work; the parent is mutated by the preceding `setIncludedValue` on `parentResult`, and the stored row is mutated here).
3. `syncDataFromCollection` reads `collection.values()` and gets back the same `R` reference it returned last time.
4. `reconcile([R])(prev)` sees prefix-match `next[0] === prev[0]` (same object), and the recursive `reconcile(R, R)` short-circuits on reference equality before the `key` option is ever consulted.

#1598 fixes the matching step (which row is "old R" and which is "new R" when references diverge), which matters for the cases #1524 surfaces. But the in-place-mutation case never reaches matching: reconcile bails at the recursive same-reference check first. So `key: '$key'` and shallow-clone-before-reconcile are independent failure modes.

That's why #1604 only forces a fresh shallow copy of each row and any array-valued field. After that, reconcile's normal logic — pointer equality on still-deep-equal scalars, key match on the rest — runs as advertised.

You're right that breaking row-level reference equality has a real cost. Two readings:

- After #1598 lands: matching falls through to `$key` (always present on tanstack-db rows), so we still find the right rows even with cloned references. The extra work is field-level diff inside matched rows; for `{name, id, scalar, scalar}`-shaped rows that's bounded.
- Without #1598: no-id user data falls back to positional + reference matching, and shallow-clone breaks reference matching. Stable-order lists still match positionally; reorders would tear down. That's the regression you're describing.

A few paths:

1. Wait for #1598, rebase #1604 on top of it. With `$key` matching in place the cost is bounded and predictable, and the two fixes compose.
2. Fix at the source: stop mutating stored parent rows in place in `flushIncludesState`. Heavier change, and you'd know better than me whether downstream consumers rely on stable refs from the collection map.
3. Tighten the shallow-clone to only clone fields that `flushIncludesState` mutates (the include values), not the whole row. That preserves row-level pointer equality for the prefix/suffix match path, at the cost of some coupling between solid-db and `collection-config-builder`'s internals.

My lean is (1), with (3) as a follow-up if field-level diff turns out to be hot. Happy to convert this PR to draft until #1598 merges, then rebase and confirm the test still demonstrates the fix.

One other note: the second sub-issue in #1571 (empty initial include → null proxy, never reactive) isn't addressed by either PR. That looks like it needs the `useLiveQuery` proxy plumbing to widen `null → array` to be reactive, which is independent of both fixes. Happy to take that on as a separate PR once we've sorted this one.

tsushanth:
Closing — this has been open without activity long enough that it's more useful to withdraw and let a fresh PR reflect current `main`/`master` than to keep pinging for review on stale state. Happy to reopen or resubmit if there's still interest.

Issue #1576, opened 2026-06-08
`preload` after `cleanup` on a collection hangs forever


willbowdler-cc:
- [x] I've validated the bug against the latest version of DB packages

**Describe the bug**

When I call `cleanup` on a collection, and then do a `preload` on that collection before refreshing the page, then the `preload` hangs forever if I await it

**To Reproduce**
see above

**Expected behavior**

I expect preload to resolve and load the data after the cleanup. Our use case is switching profiles if you are curious

**Desktop (please complete the following information):**

- OS: [e.g. iOS] Tahoe 26.5
[Brave 1.90.124 (Official Build) (arm64)](https://brave.com/latest/)
Chromium: 148.0.7778.179

**Smartphone (please complete the following information):**
N/A
**Additional context**

PR #1605, opened 2026-06-22
test: failing repro for preload hanging after cleanup (#1576)

kevin-dp:
## Reproduction for #1576

`preload` after `cleanup` on a collection hangs forever.

This PR adds a **failing** unit test that reproduces the issue, so CI captures the buggy behaviour. The fix is in a follow-up PR stacked on top of this branch.

### What it reproduces
When a source collection is cleaned up, any live query that depends on it transitions to an error state. Once everything is torn down and the live query is `preload()`-ed again (the "switching profiles without a page refresh" use case from the issue), the preload promise never resolves.

### Root cause
In `packages/db/src/query/live/collection-config-builder.ts`, `transitionToError()` latches `isInErrorState = true`, and that flag is never reset. The config-builder instance is reused across sync sessions, so when sync restarts via `preload()`, `updateLiveQueryStatus()` returns early and `markReady()` is never called — the preload hangs.

### Test
`packages/db/tests/live-query-preload-after-cleanup.test.ts` — fails fast (2s timeout guard) instead of hanging CI.

## Summary by CodeRabbit

* **Tests**
  * Added integration test coverage ensuring live queries can restart synchronization and return to a ready state with the expected data after both the source collection and the live query itself are cleaned up.

### DIFF
--- a/packages/db/tests/query/live-query-collection.test.ts
+++ b/packages/db/tests/query/live-query-collection.test.ts
@@ -594,6 +594,33 @@ describe(`createLiveQueryCollection`, () => {
     finalSubscription.unsubscribe()
   })
 
+  it(`loads its data again when preloaded after the live query and its source collection were cleaned up`, async () => {
+    const activeUsers = createLiveQueryCollection({
+      query: (q) =>
+        q
+          .from({ user: usersCollection })
+          .where(({ user }) => eq(user.active, true)),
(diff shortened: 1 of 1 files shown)


kevin-dp:
Closing this as reproduced. It's fixed in https://github.com/TanStack/db/pull/1606

PR #1606, opened 2026-06-22
fix: preload hangs forever after cleanup (#1576)

kevin-dp:
## Fix for #1576

Follow-up to #1605 (the failing-repro PR). **Stacked on top of that branch**, so this PR's diff currently contains both the regression test and the fix; once #1605 merges it reduces to just the fix.

### Problem
`preload` after `cleanup` on a collection hangs forever. When a source collection is cleaned up while a live query depends on it, `handleSourceStatusChange()` transitions the live query into an error state via `transitionToError()`, which latches `isInErrorState = true`.

The `LiveQueryCollectionConfigBuilder` instance is reused across sync sessions, and that flag was never reset. So when sync restarts (e.g. `preload()` again after cleanup — the "switching profiles without a page refresh" use case in the issue), `updateLiveQueryStatus()` returns early at the `isInErrorState` guard, `markReady()` is never called, and the preload promise never resolves.

### Fix
Reset `isInErrorState = false` at the start of each sync session in `syncFn()`. Re-subscribing to the source collections restarts them, the live query observes them becoming ready, and `markReady()` fires as expected.

`packages/db/src/query/live/collection-config-builder.ts`

### Verification
- The regression test from #1605 now passes.
- Full `packages/db` query suite passes (1587 tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where live query `preload()` could hang indefinitely after cleanup/error events.
  * Live queries now properly recover across subsequent preload attempts, returning to the expected ready state and resuming synchronization.

### DIFF
--- a/.changeset/preload-after-cleanup-1576.md
+++ b/.changeset/preload-after-cleanup-1576.md
@@ -1,0 +1,7 @@
+---
+'@tanstack/db': patch
+---
+
+Fix live query `preload()` hanging forever after a source collection was cleaned up (#1576)
+
+When a source collection is cleaned up while a live query depends on it, the live query transitions to an error state and latches an internal `isInErrorState` flag. That flag was never reset, so restarting sync (e.g. calling `preload()` again after cleanup when switching profiles) left the live query unable to become ready and the returned promise never resolved. The flag is now cleared at the start of each sync session so the live query can recover.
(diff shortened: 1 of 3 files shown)


willbowdler-cc:
Thanks for doing this!!!

KyleAMathews (approved):

PR #1607, opened 2026-06-22
fix(db): nested toArray includes drop children when sibling groups share a correlation key (#1501)

### LINKED ISSUE #1501: 3-level nested toArray: shared buffer in createPerEntryIncludesStates drops children when correlation keys overlap across parent groups

**Describe the bug**

3-level nested `toArray` includes return empty arrays for some children when sibling parent groups share the same correlation key at the deepest level.

`createPerEntryIncludesStates` assigns `nestedSetups` by reference (`state.nestedSetups = setup.nestedSetups`), so all per-parent-group child collection entries share the same pipeline buffer. `drainNestedBuffers` deletes buffer entries after routing them to the first matching parent group — subsequent groups with the same nested correlation key find nothing.

**To Reproduce**

Minimal repro — 3 collections, 3-level nesting, overlapping `regionId` across products:

```ts
import { createCollection, eq, localOnlyCollectionOptions, queryOnce, toArray } from "@tanstack/db"

const products = createCollection(localOnlyCollectionOptions({
  getKey: (p: any) => p.id,
  initialData: [
    { id: 1, title: "T-Shirt" },
    { id: 2, title: "Hoodie" },
  ],
}))

const priceRanges = createCollection(localOnlyCollectionOptions({
  getKey: (r: any) => r.id,
  initialData: [
    { id: 1, productId: 1, regionId: 1 },
    { id: 2, productId: 1, regionId: 2 },
    { id: 3, productId: 2, regionId: 1 }, // same regionId as priceRange 1
  ],
}))

const regions = createCollection(localOnlyCollectionOptions({
  getKey: (r: any) => r.id,
  initialData: [
    { id: 1, name: "Europe" },
    { id: 2, name: "North America" },
  ],
}))

const anchor = createCollection(localOnlyCollectionOptions({
  getKey: (r: any) => r.id,
  initialData: [{ id: 1 }],
}))

await Promise.all([products.preload(), priceRanges.preload(), regions.preload(), anchor.preload()])

const result = await queryOnce((qb) =>
  qb
    .from({ _: anchor })
    .select(({ _ }) => ({
      products: toArray(
        qb
          .from({ p: products })
          .where(({ p }) => eq(p.id, _.id)) // dummy correlation
          .select(({ p }) => ({
            id: p.id,
            title: p.title,
            priceRanges: toArray(
              qb
                .from({ pr: priceRanges })
                .where(({ pr }) => eq(pr.productId, p.id))
                .select(({ pr }) => ({
                  id: pr.id,
                  regionId: pr.regionId,
                  region: toArray(
                    qb
                      .from({ r: regions })
                      .where(({ r }) => eq(r.id, pr.regionId))
                      .select(({ r }) => ({ id: r.id, name: r.name })),
                  ),
                })),
            ),
          })),
      ),
    }))
    .findOne(),
)

const tshirt = result!.products.find((p: any) => p.title === "T-Shirt")
const pr1 = tshirt.priceRanges.find((r: any) => r.id === 1)
console.log(pr1.region) // [] — should be [{ id: 1, name: "Europe" }]
```

Control: changing priceRange 3's `regionId` to `3` (no overlap) makes all results correct.

**Expected behavior**

`pr1.region` should contain `[{ id: 1, name: "Europe" }]`. Each parent group should receive its own copy of nested pipeline results regardless of whether sibling groups share the same correlation key.

**Root cause**

[`createPerEntryIncludesStates`](https://github.com/TanStack/db/blob/%40tanstack/db%400.6.5/packages/db/src/query/live/collection-config-builder.ts#L1293-L1318) shares `nestedSetups` (and their `buffer`) by reference across all child collection entries:

```ts
// line 1309
state.nestedSetups = setup.nestedSetups
```

[`drainNestedBuffers`](https://github.com/TanStack/db/blob/%40tanstack/db%400.6.5/packages/db/src/query/live/collection-config-builder.ts#L1322-L1375) iterates `setup.buffer`, routes entries via `state.nestedRoutingIndex`, and deletes them from the shared buffer (line 1368-1372). When a second parent group drains the same buffer looking for the same nested correlation key, the entry is already gone.

**Workaround**

Adding a tautological parent-referencing condition at the 3rd level (e.g. `gte(pr.id, 0)`) forces `parentProjection`, which makes routing keys unique per parent row and avoids the buffer collision:

```ts
.where(({ r }) => and(eq(r.id, pr.regionId), gte(pr.id, 0)))
```

**Additional context**

- `@tanstack/db` `0.6.5`
- Related: #1454 (sibling alias collision is a similar data-routing issue at a different layer)

kevin-dp:
Fixes #1501.

## The bug
With three (or more) levels of nested `toArray` includes (e.g. `products → priceRanges → region`), when two children in **different** parent groups share the same deepest correlation key, only one of them receives the nested rows — the other comes back as an empty array:

```
T-Shirt → priceRange 1 (regionId 1) → region: []          ❌  expected [Europe]
T-Shirt → priceRange 2 (regionId 2) → region: [North America]   ✓
Hoodie  → priceRange 3 (regionId 1) → region: [Europe]          ✓
```

## Root cause
The nested pipeline writes results into a buffer shared across per-parent-group states, and routing assumes one parent per nested correlation key. Two compounding issues:

1. **1:1 routing + destructive drain.** `nestedRoutingIndex` mapped each nested correlation key to a single parent group (last-writer-wins), and `drainNestedBuffers` deleted the buffer entry after routing to the first match — so a sibling group sharing the key was dropped.
2. **No re-emit for late arrivals.** The nested pipeline does not re-emit already-materialized rows, so a parent group that starts referencing an existing correlation key *after* the rows were drained (e.g. a sibling inserted after the initial load) saw nothing.

## Fix
- `nestedRoutingIndex` now maps a nested correlation key to a **Set** of parent groups; `drainNestedBuffers` fans buffered grandchild changes out to every ready parent group before dropping the buffer entry. Routing-index inserts/deletes and parent-delete cleanup maintain the per-key parent sets.
- A per-level **cumulative snapshot** of net-present grandchild rows seeds late-arriving parent groups with the rows their siblings already received.

## Tests
Three tests under `nested toArray includes (depth 3+)` cover initial load, a sibling inserted after load, and deleting one of two siblings that share a correlation key. Full `tests/query/` suite (1596 tests) passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed nested `toArray` and live nested includes behavior when sibling parent groups share the same deepest correlation key, preventing dropped or mis-associated child results.
  * Improved fan-out routing so nested updates are applied to all relevant parent groups, including correct seeding for newly-associated parents without breaking sibling isolation during inserts and deletes.
* **Tests**
  * Added regression coverage for multi-level nested includes (`region` under `priceRanges` under `products`) for both live and materialized collection modes, including preload, post-load insert, updates, and multiple delete scenarios.

### DIFF
--- a/.changeset/nested-toarray-shared-buffer-overlap.md
+++ b/.changeset/nested-toarray-shared-buffer-overlap.md
@@ -1,0 +1,9 @@
+---
+'@tanstack/db': patch
+---
+
+fix(db): nested `toArray` includes dropping children when sibling parent groups share a correlation key
+
+With three (or more) levels of nested `toArray` includes, when two children in different parent groups shared the same deepest correlation key, only one of them received the nested rows and the other came back as an empty array. The nested-pipeline routing index mapped each nested correlation key to a single parent group and the shared buffer entry was deleted after routing to the first match, so sibling groups sharing the key were dropped.
+
+The routing index now maps a nested correlation key to all parent groups that reference it and fans buffered grandchild changes out to each. A per-level snapshot of already-materialized rows also seeds parent groups that start referencing an existing correlation key after the rows were drained (e.g. inserted after the initial load), since the pipeline does not re-emit them.
--- a/.changeset/nested-toarray-shared-buffer-overlap.md
+++ b/.changeset/nested-toarray-shared-buffer-overlap.md
@@ -2,8 +2,8 @@
 '@tanstack/db': patch
 ---
 
-fix(db): nested `toArray` includes dropping children when sibling parent groups share a correlation key
+fix(db): keep deeply nested includes in sync when sibling groups share nested correlation keys
 
-With three (or more) levels of nested `toArray` includes, when two children in different parent groups shared the same deepest correlation key, only one of them received the nested rows and the other came back as an empty array. The nested-pipeline routing index mapped each nested correlation key to a single parent group and the shared buffer entry was deleted after routing to the first match, so sibling groups sharing the key were dropped.
+Deeply nested includes could drop or stop updating nested rows when sibling parent groups shared the same nested correlation key, especially when one sibling group was inserted after the initial load. Shared nested pipeline buffers were being drained through route state that was scoped too narrowly, so one branch could consume a buffered update before other branches that referenced the same nested row received it.
 
-The routing index now maps a nested correlation key to all parent groups that reference it and fans buffered grandchild changes out to each. A per-level snapshot of already-materialized rows also seeds parent groups that start referencing an existing correlation key after the rows were drained (e.g. inserted after the initial load), since the pipeline does not re-emit them.
+Nested route state is now shared at the same scope as the nested buffer and routes updates to every concrete destination branch before clearing the buffer. Snapshot replay still seeds late-arriving sibling groups with already-materialized rows, and recursive pending-change detection ensures deeper routed updates are flushed back up through the result tree.
--- a/packages/db/src/query/live/collection-config-builder.ts
+++ b/packages/db/src/query/live/collection-config-builder.ts
@@ -907,10 +907,6 @@ export class CollectionConfigBuilder<
           entry.childCompilationResult.includes,
           syncState,
         )
-        state.nestedRoutingIndex = state.nestedSetups.map(() => new Map())
-        state.nestedRoutingReverseIndex = state.nestedSetups.map(
-          () => new Map(),
-        )
       }
 
       return state
@@ -1159,6 +1155,21 @@ type SnapshotRow = {
   count: number
 }
 
+type NestedRouteIndex = Map<
+  unknown,
+  Map<IncludesOutputState, Map<unknown, Set<unknown>>>
+>
+
+type NestedRouteReverseIndex = Map<
+  IncludesOutputState,
+  Map<unknown, Set<unknown>>
+>
+
+type NestedRouteChildToNested = Map<
+  IncludesOutputState,
+  Map<unknown, Map<unknown, unknown>>
+>
+
 /**
  * Shared buffer setup for a single nested includes level.
  * Pipeline output writes into the buffer; during flush the buffer is drained
@@ -1177,6 +1188,14 @@ type NestedIncludesSetup = {
    * groups be seeded with the rows their siblings already received.
    */
   snapshot: Map<unknown, Map<unknown, SnapshotRow>>
+  /**
+   * Shared route store for this shared buffer. Routes target concrete
+   * IncludesOutputState instances so one emitted child row can fan out across
+   * every per-entry state that references the same nested correlation key.
+   */
+  routingIndex: NestedRouteIndex
+  routingReverseIndex: NestedRouteReverseIndex
+  routingChildToNested: NestedRouteChildToNested
   /** For 3+ levels of nesting */
   nestedSetups?: Array<NestedIncludesSetup>
 }
@@ -1202,35 +1221,6 @@ type IncludesOutputState = {
   correlationToParentKeys: Map<unknown, Set<unknown>>
   /** Shared nested pipeline setups (one per nested includes level) */
   nestedSetups?: Array<NestedIncludesSetup>
-  /**
-   * Per nested setup: nestedCorrelationKey → (parentCorrelationKey → Set<childKey>).
-   * One nested correlation key can map to multiple parent groups when sibling
-   * parents share the same correlation value (e.g. two price ranges that
-   * reference the same region), so buffered grandchild changes must fan out to
-   * every parent group rather than a single one.
-   *
-   * Within a single parent group, multiple child rows can share the same nested
-   * correlation key (e.g. two price ranges in the same product both pointing at
-   * region 1). We track the referencing child keys so the parent group is only
-   * dropped from the route once its *last* referencing child row is removed —
-   * deleting one sibling must not strand the survivor.
-   *
-   * Keyed per-setup (one map per nested include field): two sibling includes can
-   * resolve the same correlation value (regionId === currencyId), so a shared
-   * store would let one include's cleanup drop the route the other still needs.
-   */
-  nestedRoutingIndex?: Array<Map<unknown, Map<unknown, Set<unknown>>>>
-  /** Per nested setup: parentCorrelationKey → Set<nestedCorrelationKeys>. */
-  nestedRoutingReverseIndex?: Array<Map<unknown, Set<unknown>>>
-  /**
-   * Per nested setup: parentCorrelationKey → (childKey → current nestedKey).
-   * Records which nested key each child row currently routes to, so an update
-   * that changes a child row's nested correlation key can drop its *previous*
-   * reference (the update change only carries the new key). Keyed per-setup so a
-   * change to one nested include never disturbs a different nested include on the
-   * same child row.
-   */
-  nestedRoutingChildToNested?: Array<Map<unknown, Map<unknown, unknown>>>
 }
 
 type ChildCollectionEntry = {
@@ -1342,6 +1332,9 @@ function setupNestedPipelines(
       compilationResult: entry,
       buffer,
       snapshot: new Map(),
+      routingIndex: new Map(),
+      routingReverseIndex: new Map(),
+      routingChildToNested: new Map(),
     }
 
     // Recursively set up deeper levels
@@ -1378,8 +1371,6 @@ function createPerEntryIncludesStates(
 
     if (setup.nestedSetups) {
       state.nestedSetups = setup.nestedSetups
-      state.nestedRoutingIndex = setup.nestedSetups.map(() => new Map())
-      state.nestedRoutingReverseIndex = setup.nestedSetups.map(() => new Map())
     }
 
     return state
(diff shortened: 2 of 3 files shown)


kevin-dp:
Here's a complete explanation of the bugs and the fix. This PR implements the snapshot accumulation fix but it means we store each distinct nested row in that snapshot. The explanation below also documents a potential alternative.

## The example

```
products → priceRanges → region
```

For each product you want its price ranges, and for each price range its region. Three levels deep. Our problem data:

- product 1 (T-Shirt) has priceRange 1 → **regionId 1** (Europe)
- product 2 (Hoodie) has priceRange 3 → **regionId 1** (Europe)  ← same region!

## How includes actually compute (the important part)

You might imagine the engine runs the "region" query fresh for each price range. It doesn't. That would be slow and wouldn't update reactively. Instead:

**There is ONE shared "region" pipeline.** It runs once over all regions and emits region rows tagged with a **correlation key** — the value that says "who do I belong to". For region, that key is `regionId`. So the region pipeline emits one thing:

```
regionId 1 → { Europe }
```

Just one entry. Even though *two* price ranges want region 1, the pipeline produces it once, keyed by `regionId 1`.

**Results land in a "buffer"** — a temporary mailbox: `regionId → the region rows`. Think of it as "here are freshly-computed region rows, waiting to be delivered."

**Then a "drain" step delivers buffered rows to the right parents.** This is the routing step. To deliver region 1 to the correct price range, the engine keeps a **routing index**:

```
routingIndex:  regionId 1  →  "which price-range group does this belong to?"
```

Each price-range group has its **own private copy** of region results (its own little child collection), so the final tree can show each one its own `region` array. The drain reads the buffer, looks up the parent in the routing index, copies the rows into that parent's private copy, and then **deletes the buffer entry** (it's been delivered — don't deliver twice).

## Why it broke

The whole design quietly assumed **one nested key belongs to one parent**. Two things follow from that assumption, and both are wrong when two siblings share `regionId 1`:

**Bug 1 — the routing index only remembered one parent.**

```
routingIndex.set(regionId 1, priceRange1)   // T-Shirt's
routingIndex.set(regionId 1, priceRange3)   // Hoodie's — OVERWRITES
```

It's a plain key→value map, so the second write clobbers the first. Now `regionId 1` points to only *one* group. Then drain delivers region 1 to that one group and **deletes the buffer entry**. The other group's lookup finds nothing → its `region` comes back `[]`. That's the empty array you saw.

**Bug 2 — late arrivals get nothing.**

Say T-Shirt's priceRange exists at load, but Hoodie's priceRange 3 is inserted *later*. By then the region pipeline already emitted region 1 (at load), it was drained, and the buffer entry was deleted. When priceRange 3 shows up, the pipeline **does not re-emit region 1** — from its point of view nothing about regions changed; it already produced that row. So the buffer is empty and there's nothing to deliver to Hoodie. Empty array again.

Both of these were confirmed with debug logging before writing the fix.

## The fix — two matching pieces

### Fix 1: let one nested key map to *many* parents

The routing index value changes from a single parent to a **Set of parents**:

```
routingIndex:  regionId 1  →  { priceRange1, priceRange3 }
```

- Adding a parent now *adds to the set* instead of overwriting (the `updateRoutingIndex` insert change).
- Removing a parent removes just that one from the set, leaving the others (the delete change and `cleanRoutingIndexOnDelete`). This is what keeps the *deletion* case working — deleting Hoodie mustn't wipe regionId 1 while T-Shirt still uses it.
- Drain now loops over **every** parent in the set and copies the rows into each one's private copy, and only then deletes the buffer entry. That's the `for (const parentCorrelationKey of parentCorrelationKeys)` loop. This fixes the initial-load case.

### Fix 2: remember delivered rows so late arrivals can be seeded

Because the pipeline won't re-emit region 1 for a price range added later, the engine needs its own memory of "what region rows currently exist for regionId 1". That's the new **snapshot**:

```
snapshot:  regionId 1  →  { Europe }   (kept around, not deleted on drain)
```

- Every time something is drained, it's also folded into the snapshot (`accumulateSnapshot`). The snapshot tracks a running count so that if a region is later deleted, it drops out of the snapshot too — it stays an accurate picture of "what currently exists", not a pile of stale rows.
- When a new parent group first references an existing key (`isNewParent` in `updateRoutingIndex`), we **seed** it straight from the snapshot (`seedParentFromSnapshot`) — hand it the rows its siblings already have, without waiting on the pipeline. This fixes the insert-after-load case.

### One-line summary

The nested-includes system assumed each nested key had exactly one parent and treated delivered rows as throwaway. The fix makes a nested key able to feed **multiple** parents (Set instead of single value + fan-out on drain), and keeps a **snapshot** of currently-existing rows so parents that show up late can be handed the same data their siblings already received.

---

## Is the snapshot a memory problem?

Partly a fair concern, but it's bounded, not unbounded:

- It stores **one entry per distinct (nestedKey, childKey)** — i.e. one copy of each distinct nested row, **not** one per parent. Siblings sharing a key share the single snapshot entry. So its size tracks the *distinct nested result set*, not parent count.
- It holds **references** to the same row `value` objects already materialized elsewhere, not deep copies. The extra cost is Map overhead + one slot per distinct row.
- It **shrinks**: `accumulateSnapshot` tracks a net count and deletes rows (and empty keys) when the count hits 0 on delete. So it's not an ever-growing log — it stays proportional to currently-live data.

So it's not a leak and not multiplied by parents. But it is a real added cost: roughly **O(distinct nested rows)** extra — essentially one more index over the deepest level, for the lifetime of the live query. For a query with a very large nested collection, that's non-trivial memory.

---

## The alternative: copy from an existing sibling instead of keeping a snapshot

### The key realization

When does a late-arriving parent need seeding at all? Only when the region rows were **already delivered and the buffer entry deleted in an earlier flush**, *and* the pipeline won't re-emit them. But "already delivered to whom?" — to a sibling. So whenever seeding is needed, **at least one sibling already has the rows materialized** in its own private copy.

That's the insight the alternative exploits: you don't need a separate stash of the data, because a sibling is already holding a perfectly good copy of it. Go ask the sibling.

### How "copy from sibling" works

Walk through Hoodie's priceRange 3 arriving late, referencing `regionId 1`:

1. **Look up the routing index** (which is now a Set thanks to Fix 1):
   ```
   regionId 1 → { priceRange1, priceRange3 }
   ```
   priceRange3 is the newcomer; priceRange1 is an existing sibling.

2. **Find a sibling that already has the data.** priceRange1's private region state has a `childRegistry` keyed by `regionId`, and `childRegistry[regionId 1]` is a little collection already holding `{ Europe }`.

3. **Copy those rows into priceRange3's pending changes**, then let the normal materialize step run. priceRange3 now shows `[Europe]`.

So instead of consulting a snapshot, step 2 reaches into a sibling's already-materialized collection.

### What it costs you in code

The snapshot stored exactly what the delivery machinery needs: for each row, its `childKey`, its `value`, and its `orderByIndex`. A live child collection only stores the rows themselves. To rebuild "changes" from it you have to look those extra bits back up:

- the sibling state's `resultKeys` map (row → childKey),
- its `orderByIndices` map (row → orderByIndex),

then iterate the sibling's collection and reconstruct insert-changes from each row. It's the same idea as `seedParentFromSnapshot`, just reading from a collection + two side-maps instead of from one tidy snapshot Map.

You also need a small guard: "pick a sibling that is *actually* populated right now." The routing-index Set gives you candidates, but you'd skip the newcomer itself and any sibling that happens to be empty, and copy from the first good one. (As argued above, a good one is essentially guaranteed to exist whenever seeding is needed — if none did, the buffer path would still be handling it.)

---

## The two approaches compared

| | Snapshot (current fix) | Copy-from-sibling |
|---|---|---|
| **Standing memory** | one extra index over distinct nested rows, kept for the query's life | none — reuses data already in siblings |
| **Work at steady state** | a little on every drain (fold into snapshot) | none |
| **Work on a late insert** | cheap: read tidy snapshot | a bit more: find a ready sibling, rebuild changes from its collection + side-maps |
| **Code complexity** | self-contained, one clean structure | more fiddly: sibling selection + reconstructing changes |
| **Correctness risk** | must keep the snapshot's net-count in sync with deletes/updates | sibling is always the live truth, so no parallel structure to drift — but "is a sibling ready?" is a subtler question |

### How to choose

The snapshot trades **memory** for **simplicity and predictability**: one clean structure, consulted in one place, no "go find a healthy donor" logic. Copy-from-sibling trades that memory away but pays in **code complexity** and a reconstruction step on each late insert.

For typical includes (modest nested result sets) the snapshot's overhead is negligible and the simpler code wins. If you expect very large nested collections where doubling an index matters, copy-from-sibling is the better fit — and it's a localized swap: only `accumulateSnapshot` + `seedParentFromSnapshot` change; the Set-based routing fan-out (Fix 1) stays exactly as-is either way.

kevin-dp:
Had a quick discussion with @samwillis on the 2 alternatives. We decided to go for correctness and simplicity so we pick the snapshot approach.

samwillis (changes requested):
I found one remaining routing hole in the new `Set<parentCorrelationKey>` approach.

A single parent correlation key can be backed by multiple child rows that share the same nested correlation key. The new route set dedupes those rows down to one parent entry, so when one of the child rows is deleted the delete path removes the whole parent route even though another child row in that same parent group still needs it. After that, later grandchild updates/deletes for the shared nested key no longer reach the surviving child.

The problematic path is around `packages/db/src/query/live/collection-config-builder.ts:1591`, where the delete path does `parents.delete(correlationKey)` without knowing whether another child row in that parent group still references the same nested key.

I pushed a minimal failing repro here:

- Branch: https://github.com/samwillis/db/tree/codex/pr-1607-same-parent-routing-repro
- Commit: https://github.com/samwillis/db/commit/21ac924c671233ad1dc42d8debdb7509285c6239

The repro adds one product with two price ranges sharing `regionId: 1`, deletes one price range, then updates region `1`. The surviving price range still shows `Europe` instead of `Renamed Europe`.

Targeted command used:

```sh
PATH=/Users/samwillis/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH pnpm exec vitest --run tests/query/includes.test.ts -t 'keeps routing when one of multiple same-parent siblings sharing a nested key is deleted'
```

This likely needs per-parent route multiplicity/refcounting, or tracking routes at the child-key level, so deleting one sibling does not remove routing for the surviving sibling.

kevin-dp:
Great catch @samwillis — confirmed and fixed in 33bfcaa5.

I reproduced your repro against the branch and it failed exactly as you described: deleting one of two same-parent siblings that share a nested correlation key dropped the route for the survivor, so later grandchild updates never reached it. Root cause was exactly what you identified at `collection-config-builder.ts:1591` — the `Set<parentCorrelationKey>` route had no notion of *how many* child rows in a parent group referenced a nested key, so the first sibling delete emptied the entry and dropped the whole route.

**Fix:** the routing index now tracks the referencing child keys per `(nestedKey, parentGroup)` — `Map<nestedKey, Map<parentCorrelationKey, Set<childKey>>>` — and a parent group is only dropped from a route once its *last* referencing child row is removed. Extracted a `removeChildKeyFromRoute` helper so the delete path and parent-delete cleanup share one consistent cascade.

While verifying, I found a second routing hole your report exposed: an **update that changes a child row's nested correlation key** (e.g. a price range's `regionId` changes) only carries the *new* key, so the row's stale reference under the *old* key was never released. With the new refcounting that stale child key kept the old route alive, and a later sibling delete then mis-routed grandchild changes (the updated row lost its grandchildren). Fixed with a per-setup `childKey → nestedKey` map that releases the previous reference on update — scoped per nested setup so a change to one nested include can't disturb another include on the same child row.

New regression tests cover all three: your same-parent-sibling-delete case, the update-then-sibling-delete case, and cross-include isolation. Full `tests/query/` suite passes (1601), no type errors, lint clean.

Separately — and **not** introduced by this PR — I noticed that an *atomic mutual swap* (two same-parent siblings exchanging nested keys in a single tick) produces wrong results; it reproduces on `main` too, so it predates this work. I'll open a separate issue with a repro rather than expand this PR's scope. Let me know if you'd prefer it folded in here.

samwillis (changes requested):
Thanks for the update. The previous same-parent repro is fixed now: the added test `keeps routing when one of multiple same-parent siblings sharing a nested key is deleted` passes on `ea83de909`.

I found one remaining routing hole in the new child-key refcounting approach, though. `nestedRoutingIndex` is still shared across all sibling nested includes at a level (`nestedCorrelationKey -> parent -> childKeys`), while `nestedRoutingChildToNested` is per nested setup. When two sibling nested includes use the same nested correlation value for the same child row, cleanup for one include can delete the shared route that the other include still needs.

Concrete failing case:

- A price range has `regionId: 1` and `currencyId: 1`.
- The child select defines nested includes in this order: `currency` first, `region` second.
- Updating only `regionId` from `1` to `2` makes the `currency` setup observe key `1` unchanged first, then the `region` setup removes child key `1` from the shared route for nested key `1`.
- Because the route is shared and only keyed by child key, the later removal deletes the route needed by `currency`.
- A subsequent currency update no longer reaches the surviving `currency` include, so the row still shows `EUR` instead of `USD`.

The problematic shape is around `packages/db/src/query/live/collection-config-builder.ts:1216` and the deletion at `packages/db/src/query/live/collection-config-builder.ts:1556`: the route refcount is not scoped by nested setup/include field, so equal correlation values from different nested includes collide.

Temporary repro I ran locally:

```ts
it(`keeps sibling include routing when an earlier include still references the old nested key`, async () => {
  // one price range: regionId: 1, currencyId: 1
  // select order: currency first, region second
  priceRanges.update(1, (draft) => {
    draft.regionId = 2
  })
  await new Promise((r) => setTimeout(r, 50))

  currencies.update(1, (draft) => {
    draft.code = `USD`
  })
  await new Promise((r) => setTimeout(r, 50))

  // Fails: currency remains EUR
})
```

Targeted command:

```sh
PATH=/Users/samwillis/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH pnpm exec vitest --run tests/query/includes.test.ts -t 'keeps sibling include routing when an earlier include still references the old nested key'
```

This likely needs the route reference tracking to be scoped per nested setup/include field as well as by parent and child key, or otherwise include setup identity in the routing key/refcount so sibling includes with equal correlation values cannot remove each other's routes.

kevin-dp:
Confirmed and fixed — thanks again @samwillis, this was exactly right.

Reproduced your case (added as `resolves two nested includes on the same child independently when they share a correlation value` in 1553c4ef) and traced it: with `regionId === currencyId === 1`, both the `currency` and `region` setups resolve the same `nestedRoutingKey` and collapse into a single `Set<childKey>` entry in the **shared** route store. Re-pointing `regionId 1→2` removes child key `1` from that shared set, it empties, and the whole route for key `1` is dropped — even though `currency` still needs it. A later currency change then finds no route (`[DRAIN] setup=0 nestedKey=1 routes=NONE`) and the row keeps `EUR`.

Root cause as you pinpointed: the child-key refcount tracked the *removal decision* per nested setup (`nestedRoutingChildToNested`), but the route store itself (`nestedRoutingIndex` / `nestedRoutingReverseIndex`) was shared across all sibling includes at a level, and the routing key doesn't encode the include field — so equal correlation values from different includes can remove each other's routes.

**Fix (492f0852):** scope the route store per nested setup — `nestedRoutingIndex` and `nestedRoutingReverseIndex` are now arrays indexed by setup, mirroring `nestedRoutingChildToNested`. Each include owns its route store, so a removal in one include can't touch another's. `removeChildKeyFromRoute` now takes the per-setup maps and parent-delete cleanup iterates every setup. Drain targeting was already per-setup (it routes into `entry.includesStates[i]`), so only the store needed to move to the same granularity.

Full `tests/query/` suite passes (1602), no type errors, lint clean. The earlier same-parent/sibling-parent guarantees are unchanged — those all live within a single setup, which is now just `index[0]`.

samwillis (approved):
Re-reviewed latest head `29b8fb16589d48a87342001a6779f14357b2d751` after the follow-up fixes.

The route stores are now scoped per nested setup, so the sibling nested includes that share the same correlation value no longer collide. I also checked the same-parent sibling route cleanup and the recursive deeper-buffer path added for depth 3+ fanout.

Verification:
- `pnpm exec vitest --run tests/query/includes.test.ts` (134 passed, no type errors)

I don't see any remaining blocking issues in this update.

PR #1614, opened 2026-06-24
fix(query): lazy-join loads through the collection the join key resolves to

### LINKED ISSUE #1494: Lazy-join `Join requires an index` warning blames the wrong collection when a subquery's select field traces across an inner join
- [x] Validated against `@tanstack/db@0.6.5`

**Describe the bug**

In `packages/db/src/query/compiler/joins.ts`, the lazy-join path emits:

```
[TanStack DB] [<collectionId>] Join requires an index on "<fieldPath>" for efficient loading. Falling back to loading all data. ...
```

When the joined side is a subquery, the `collectionId` in the warning comes from `followRefCollection.id` — the ultimate source of the join expression traced through the subquery's `select`. But the subscription that actually failed to optimize is looked up via `aliasRemapping[lazyAlias]`, which resolves to the subquery's innermost `.from` alias.

When these resolve to different collections (because the subquery's select field comes from a *joined* side rather than the `.from` side), the warning names a collection whose advice is typically already satisfied, and hides the real failing subscription.

Data still flows via the fallback full-load, so this is a warning-quality bug, not a correctness bug.

**To Reproduce**

```ts
// packages/db/tests/lazy-join-wrong-collection-id.test.ts
import { beforeEach, describe, expect, test, vi } from 'vitest'
import { createLiveQueryCollection, eq } from '../src/query/index.js'
import { BTreeIndex, createCollection } from '../src/collection/index.js'
import { flushPromises, mockSyncCollectionOptions } from './utils.js'

type A = { id: string }
type B = { id: string; aId: string }

describe('lazy-join warning misattributed when followRef crosses a subquery join', () => {
  let a: ReturnType<typeof createCollection<A>>
  let b: ReturnType<typeof createCollection<B>>
  let warnSpy: ReturnType<typeof vi.spyOn>

  beforeEach(() => {
    a = createCollection(mockSyncCollectionOptions<A>({
      id: 'a', getKey: (r) => r.id,
      autoIndex: 'eager', defaultIndexType: BTreeIndex,
      initialData: [{ id: 'a1' }],
    }))
    b = createCollection(mockSyncCollectionOptions<B>({
      id: 'b', getKey: (r) => r.id,
      initialData: [{ id: 'b1', aId: 'a1' }],
    }))
    warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {})
  })

  test('warning blames followRefCollection instead of the failing subscription', async () => {
    const live = createLiveQueryCollection({
      id: 'outer', startSync: true,
      query: (q) => {
        const sub = q
          .from({ innerB: b })
          .leftJoin({ innerA: a }, ({ innerA, innerB }) => eq(innerA.id, innerB.aId))
          .select(({ innerA }) => ({ aId: innerA?.id }))

        return q
          .from({ outerB: b })
          .leftJoin({ sub }, ({ outerB, sub }) => eq(sub.aId, outerB.aId))
          .select(({ outerB }) => ({ id: outerB.id }))
      },
    })

    await flushPromises()

    const warnings = warnSpy.mock.calls.map((c) => String(c[0]))
      .filter((m) => m.includes('Join requires an index'))

    expect(warnings).toHaveLength(1)
    // Warning names `[a]`. But `a` has an auto-indexed `id`; the advice
    // "create an index on a.id" is already satisfied. The actually failing
    // subscription is on `b` (aliasRemapping['sub'] = 'innerB').
    expect(warnings[0]).toContain('[a]')
    expect(warnings[0]).toContain('"id"')
  })
})
```

Passes against `@tanstack/db@0.6.5`.

**Expected behavior**

Either:
- No warning — the lazy-loader should reach the correct collection's subscription (whichever one can actually satisfy the index lookup).
- Or a warning whose collection ID names the subscription that actually failed — so the "consider creating an index" advice is actionable.

**Additional context**

Trace through the compiler for the outer `.leftJoin({ sub }, eq(sub.aId, outerB.aId))`:

- `followRef(rawQuery, sub.aId, lazySource)` walks `sub` → subquery → `select.aId = innerA.id` → `innerA` → collection A. Returns `{ collection: a, path: ['id'] }`. So `followRefCollection = a`.
- `subQueryResult.collectionId = b` (subquery's innermost `.from`), so `aliasRemapping['sub'] = 'innerB'`, and `subscriptions['innerB']` is the `b` subscription.
- The tap runs `bSubscription.requestSnapshot({ where: inArray(PropRef(['id']), joinKeys), optimizedOnly: true })`. `b` has an `id` field but no index → `findIndexForField(b, ['id'])` returns `undefined` → `canOptimize: false` → `currentStateAsChanges` returns `undefined` → `loaded = false` → warning with `collectionId = followRefCollection.id = 'a'`.

I instrumented the warning site to log directly and confirmed the mismatch:

```
[VIBL-VERIFY] subscription.collection.id=b resolvedAlias=innerB followRefCollection.id=a fieldPath=id
```

Workaround: swap the subquery's `.from` so the innermost alias maps to `followRefCollection`. E.g., `q.from({ innerA: a }).leftJoin({ innerB: b }, ...).select(({ innerA }) => ({ aId: innerA.id }))`. Tested — no warning in that shape.

kevin-dp:
## What

Fixes #1494.

When a subquery used in a `JOIN` clause selects its join key from a **joined** source rather than from its own `from` clause, the outer join key resolves to a different collection than the subquery's `from` source. The lazy-join loader diverged here:

- `followRef` traced the index requirement to the collection the key actually resolves to (e.g. `teams.id`), but
- the subscription it drove loading through was looked up via the subquery's `from` alias (e.g. `members`).

Because those two diverged, the loader queried the wrong subscription, fell back to a full load, and emitted a `Join requires an index` warning naming the **already-indexed** collection — so the advice ("create an index") was already satisfied and unactionable. Data was still correct via the fallback, so this was a diagnostic-/efficiency-quality bug, not a correctness one.

## Fix

`followRef` now also returns the alias of the source the ref ultimately resolves to. `getLazyLoadTargets` uses that alias as the subscription target (falling back to the existing from-clause remapping when the key resolves directly to the `from` source). Lazy loading then drives through the correct collection's index — no spurious warning, no unnecessary full-load.

## Tests

Adds a regression test in `packages/db/tests/query/join-subquery.test.ts`: a subquery whose join key resolves to an already-indexed collection must load through that index without emitting a `Join requires an index` warning that points at it. The test fails without the fix and passes with it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed lazy-join behavior so nested joins now resolve through the correct collection when the join key comes from a joined subquery.
  * Prevented incorrect “Join requires an index” warnings in cases where the resolved collection is already indexed.
  * Improved join loading to avoid unnecessary full-load fallback in these scenarios.

* **Tests**
  * Added coverage for lazy joins involving indexed and unindexed collections to verify the correct result and warning behavior.


samwillis (approved):
Reviewed locally in the PR worktree. The lazy-load target alias change looks correct for the subquery-join case, and I did not find blocking issues.

Verified:
- `pnpm --filter @tanstack/db exec vitest --run tests/query/join-subquery.test.ts`
- `pnpm --filter @tanstack/db test -- tests/query/join-subquery.test.ts`

PR #1616, opened 2026-06-25
feat(PowerSync): add attachments support

Chriztiaan:
## 🎯 Changes

Derived from Steven's efforts in https://github.com/powersync-ja/powersync-js/pull/983, and addresses https://github.com/TanStack/db/discussions/1563.

## Problem

PowerSync ships an attachment helper for syncing files (photos, documents) between local and remote storage. It's separate from regular synced tables: a local-only attachments table tracks each file's lifecycle (QUEUED_UPLOAD, SYNCED, QUEUED_DELETE), and an AttachmentQueue drives uploads/downloads in the background.

TanStackDB, on the other hand, gives you an optimistic, reactive, joinable view over synced data. For users who want to use the attachment helper alongside the PowerSync+TanstackDB integration there are blockers. Saving a file (in the local-only attachments table) and associating it with a record (e.g. setting user.photo_id) are two independent writes which could make data races and fatal errors a problem for data consistency.

The original POC ([powersync-js#983](https://github.com/powersync-ja/powersync-js/pull/983,)) proved this integration was viable. This PR productionises a a subset of it as reusable functionality.

## Solution

  A `TanStackDBAttachmentQueue` that extends the SDK's AttachmentQueue (for saving and deleting a file) and backs it with a TanStack DB collection.

The package owns the collection-backed saveFile/delete implementation and leaves the wiring to the application (covered in documentation).

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Added attachment support for PowerSync DB collections through a TanStack DB-backed attachment queue.
  * Supports atomic saves and deletes, custom attachment IDs, related-record updates, and local attachment URLs.
  * Added rollback and cleanup handling when attachment operations fail.
* **Documentation**
  * Added setup and usage guidance for attachment collections, queues, watching attachments, and local URLs.
* **Tests**
  * Added coverage for attachment lifecycle, synchronization, cleanup, custom IDs, concurrent saves, and transaction rollback.


Chriztiaan:
@samwillis  - could you review this one please?

tannerlinsley:
I ran the exact package suite locally and all 92 tests pass, but I still need a few correctness fixes before merging this:

- `saveFile()` runs before the collection transaction, so a failed insert, hook, or commit leaves an orphaned local file. Please clean it up on failure and add a regression test.
- The docs show async `updateHook` callbacks, but the hook is called inside synchronous `transaction.mutate()` and its promise is ignored. Any mutation after an `await` escapes the transaction. The contract and examples need to be explicitly synchronous, or the implementation needs an actually awaited atomic design.
- `watchAttachments` calls need to be serialized so an older async update can't win after a newer one.
- This adds public API and raises the PowerSync peer floor from 1.41 to 1.57, so the changeset should be `minor`, not `patch`.

The earlier review threads are marked resolved, but the current diff still has these behaviors.

Chriztiaan:
Thanks for the feedback!
> * `saveFile()` runs before the collection transaction, so a failed insert, hook, or commit leaves an orphaned local file. Please clean it up on failure and add a regression test.
Done.
> * The docs show async `updateHook` callbacks, but the hook is called inside synchronous `transaction.mutate()` and its promise is ignored. Any mutation after an `await` escapes the transaction. The contract and examples need to be explicitly synchronous, or the implementation needs an actually awaited atomic design.
Fixed outdated docs, this is no longer async.

> * `watchAttachments` calls need to be serialized so an older async update can't win after a newer one.
I don't believe this comment is correct/needed. They should already be serialized (at least by our SDK).

> * This adds public API and raises the PowerSync peer floor from 1.41 to 1.57, so the changeset should be `minor`, not `patch`.
Done.

packages/powersync-db-collection/src/attachments.ts
```diff
@@ -0,0 +1,225 @@
...
+  /**
+   * Called synchronously within the same TanStackDB transaction as the attachment write,
+   * so any mutations made to other collections are committed atomically with it.
+   *
+   * Must not be async. `Transaction.mutate` unregisters the ambient transaction as soon as
+   * this callback returns, so any mutation made after an `await` inside the hook escapes the
+   * transaction and is not committed atomically with the attachment. Do asynchronous work
+   * before calling `save` or `delete`.
+   */
+  updateHook?: (attachment: AttachmentQueueRow) => void
```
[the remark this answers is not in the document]
Chriztiaan: I think this is overkill?

packages/powersync-db-collection/src/attachments.ts
```diff
@@ -0,0 +1,225 @@
...
+    updateHook,
+  }: SaveOptions): Promise<AttachmentQueueRow> {
+    const resolvedId = id ?? (await this.generateAttachmentId())
+
+    /**
+     * Checked before the file is written. Writing first would overwrite the existing
+     * attachment's local file, and the cleanup below would then delete it — leaving the
+     * pre-existing record pointing at a file that no longer exists.
+     */
+    if (this.collection.get(resolvedId)) {
```
tannerlinsley: Concurrent saves with the same custom `id` can both pass this check because it runs before `withAttachmentContext`. Both then write the same URI; when the second insert fails, its catch deletes that URI, removing the first successful attachment file. Please serialize the duplicate check and file write with the attachment context, and add a concurrent-save regression test so a losing save cannot overwrite or delete the winner file.

Chriztiaan: Addressed, thank you for the feedback.

Chriztiaan:
Hey @tannerlinsley, friendly poke here :)
Are there other reviewers I can bother to rubberstamp this?

KyleAMathews:
@Chriztiaan @tannerlinsley I independently reproduced both reported bugs at `fc55bee66c05376e594aa7e377ab0a96f7087405`, using the locked `@powersync/common@1.57.0`, real SQLite, the real collection adapter, and actual SDK `syncStorage()` completion writes.

**I recommend holding merge for these two correctness fixes.** The original suite passes: 95 runtime tests plus one type test. Eight added boundary cases yield four failures and four passing controls; the native-SDK failure is a causal control for the same upload race, not a fourth distinct bug. The added tests typecheck and lint clean.

### 1. Reusing an ID before a fresh collection is ready destroys the existing local file

Save an attachment, construct another collection against the same table without waiting for readiness, then save different bytes under the same ID/extension. The collection lookup misses the existing SQL row. The new save rejects, the original SQL row survives, but its original file is gone. The otherwise identical ready-collection case preserves the file and bytes.

The guard must establish identity/ownership before a destructive file write. Waiting for collection readiness addresses this particular witness but is not an authoritative duplicate check for stale/partial caches. Failure cleanup must only remove resources owned by the failed operation.

An additional boundary probe uses two queue instances sharing the same DB/storage and ID: one save succeeds, the other rejects, and the loser removes the winner's file. This needs either shared-resource ownership/reservation or an explicit, enforced single-owner contract; the existing per-instance mutex does not cover it.

### 2. Delete during upload is overwritten by SDK completion

Hold the remote upload; call the TanStack queue's delete with an updateHook that detaches the related row; verify SQLite contains QUEUED_DELETE and a null foreign key; release the upload. Two further real sync passes make zero remote delete calls, leaving the uploaded remote file behind. Delete-before-upload and delete-after-upload controls pass.

The **native SDK AttachmentQueue alone also reproduces this**: saveFile → hold upload → deleteFile → release upload changes SQL state from QUEUED_DELETE (2) to SYNCED (3). The SDK captures a row before I/O and later saves the stale completed row unconditionally. Serializing the individual writes does not preserve a delete committed between them.

This belongs at the SDK completion boundary, rather than a second attachment state machine in TanStack. Completion must reconcile with current committed intent and preserve outstanding remote deletion. Watcher/archive behavior also needs to be exercised before choosing the exact merge/update rule.

### Next step / coordination

We're starting a bounded lifecycle oracle on top of this PR: independent desired intent and local/remote bytes; real SDK completion SQL; controlled transport timing; cold/warm caches; delete before/during/after upload; rollback and ownership assertions; fixed witnesses plus replayable generated histories. No production changes have been made by this review.

Could the SDK maintainers coordinate the completion fix and confirm which SDK line this integration should target alongside #1688? That PR upgrades to v2 and currently conflicts with main. A source check of SDK main `fb05b28bacdd291e4e768b0a9d96b7c877314887` still shows the unconditional completion save; the executable repro here is specifically against 1.57.0.

For completeness, the earlier concern about serializing successive watched snapshots does **not** reproduce: a controlled test confirms the SDK context already serializes those callbacks. The synchronous-hook documentation and minor changeset are also present.

<details>
<summary>Executable test-only patch against fc55bee66 (intentionally RED)</summary>

Run from `packages/powersync-db-collection` after building this checkout's core packages:
`pnpm exec vitest run tests/attachments.test.ts --coverage.enabled=false --typecheck.enabled=false`

```diff
diff --git a/packages/powersync-db-collection/tests/attachments.test.ts b/packages/powersync-db-collection/tests/attachments.test.ts
index 193f8fa0c..1cb5833e9 100644
--- a/packages/powersync-db-collection/tests/attachments.test.ts
+++ b/packages/powersync-db-collection/tests/attachments.test.ts
@@ -1,7 +1,9 @@
 import { randomUUID } from 'node:crypto'
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
+import pDefer from 'p-defer'
 import {
+  AttachmentQueue,
   AttachmentState,
   AttachmentTable,
   Schema,
@@ -212,6 +214,260 @@ describePowerSync(`PowerSync AttachmentQueue (TanStackDB)`, () => {
   }
 
   describe(`save`, () => {
+    it(`serializes successive watched snapshots through the SDK context`, async () => {
+      const fixture = await setup()
+      let update!: Parameters<WatchAttachments>[0]
+      const queue = fixture.createQueue({
+        archivedCacheLimit: 100,
+        watchAttachments: (callback) => {
+          update = callback
+        },
+      })
+      const record = await queue.save({
+        data: createMockJpegBuffer(),
+        fileExtension: `jpg`,
+      })
+      await queue.syncStorage()
+      await queue.startSync()
+      const entered = pDefer<void>()
+      const release = pDefer<void>()
+      const held = queue.withAttachmentContext(async () => {
+        entered.resolve()
+        await release.promise
+      })
+      await entered.promise
+      const completed: Array<number> = []
+      const first = update([]).then(() => {
+        completed.push(1)
+      })
+      const second = update([{ id: record.id, fileExtension: `jpg` }]).then(
+        () => {
+          completed.push(2)
+        },
+      )
+      try {
+        await Promise.resolve()
+        expect(completed).toEqual([])
+        release.resolve()
+        await Promise.all([held, first, second])
+        expect(completed).toEqual([1, 2])
+        expect(
+          await fixture.db.get(`SELECT state FROM attachments WHERE id = ?`, [
+            record.id,
+          ]),
+        ).toEqual({ state: AttachmentState.SYNCED })
+      } finally {
+        release.resolve()
+        await Promise.all([held, first, second])
+      }
+    })
+
+    it(`preserves the winner file across two queue instances sharing storage`, async () => {
+      const fixture = await setup()
+      const first = fixture.createQueue()
+      const second = fixture.createQueue()
+      const bothWritten = pDefer<void>()
+      const saveFile = fixture.localStorage.saveFile.bind(fixture.localStorage)
+      let writes = 0
+      vi.spyOn(fixture.localStorage, `saveFile`).mockImplementation(
+        async (...args) => {
+          const size = await saveFile(...args)
+          if (++writes === 2) bothWritten.resolve()
+          await bothWritten.promise
+          return size
+        },
+      )
+      const results = await Promise.allSettled([
+        first.save({
+          id: `shared-id`,
+          data: createMockJpegBuffer(),
+          fileExtension: `jpg`,
+        }),
+        second.save({
+          id: `shared-id`,
+          data: new Uint8Array([7, 8, 9]).buffer,
+          fileExtension: `jpg`,
+        }),
+      ])
+      expect(
+        results.filter((result) => result.status === `fulfilled`),
+      ).toHaveLength(1)
+      expect(
+        await fixture.localStorage.fileExists(
+          fixture.localStorage.getLocalUri(`shared-id.jpg`),
+        ),
+      ).toBe(true)
+    })
+
+    it.each([`ready`, `cold`] as const)(
+      `preserves an existing file when reused through a %s collection`,
+      async (phase) => {
+        const fixture = await setup()
+        const original = await fixture.createQueue().save({
+          id: `reused-after-reopen`,
+          data: createMockJpegBuffer(),
+          fileExtension: `jpg`,
+        })
+        const collection =
+          phase === `ready`
+            ? fixture.attachmentsCollection
+            : createCollection(
+                powerSyncCollectionOptions({
+                  database: fixture.db,
+                  table: APP_SCHEMA.props.attachments,
+                }),
+              )
+        onTestFinished(() => collection.cleanup())
+        const queue = new TanStackDBAttachmentQueue({
+          db: fixture.db,
+          attachmentsCollection: collection,
+          localStorage: fixture.localStorage,
+          remoteStorage: fixture.remoteStorage,
+          watchAttachments: () => {},
+        })
+        onTestFinished(() => queue.stopSync())
+        if (phase === `cold`)
+          expect(collection.get(original.id)).toBeUndefined()
+        await expect(
+          queue.save({
+            id: original.id,
+            data: new Uint8Array([7, 8, 9]).buffer,
+            fileExtension: `jpg`,
+          }),
+        ).rejects.toThrow()
+        expect(
+          await fixture.db.getOptional(
+            `SELECT id FROM attachments WHERE id = ?`,
+            [original.id],
+          ),
+        ).toEqual({ id: original.id })
+        expect(await fixture.localStorage.fileExists(original.local_uri!)).toBe(
+          true,
+        )
+        expect(
+          new Uint8Array(
+            await fixture.localStorage.readFile(original.local_uri!),
+          ),
+        ).toEqual(new Uint8Array(createMockJpegBuffer()))
+      },
+    )
+
+    it.each([`before`, `during`, `after`] as const)(
+      `preserves delete intent %s an SDK upload`,
+      async (timing) => {
+        const fixture = await setup()
+        const queue = fixture.createQueue()
+        const uploaded = pDefer<void>()
+        const release = pDefer<void>()
+        const remoteFiles = new Set<string>()
+        fixture.uploadFile.mockImplementation(async (_, attachment) => {
+          uploaded.resolve()
+          await release.promise
+          remoteFiles.add(attachment.id)
+        })
+        fixture.deleteFile.mockImplementation((attachment) => {
+          remoteFiles.delete(attachment.id)
+          return Promise.resolve()
+        })
+        const userId = randomUUID()
+        const record = await queue.save({
+          data: createMockJpegBuffer(),
+          fileExtension: `jpg`,
+          updateHook: (attachment) => {
+            fixture.usersCollection.insert({
+              id: userId,
+              name: `owner`,
+              email: null,
+              photo_id: attachment.id,
+            })
+          },
+        })
+        let sync: Promise<void> | undefined
+        try {
+          if (timing !== `before`) {
+            sync = queue.syncStorage()
+            await uploaded.promise
+            if (timing === `after`) {
+              release.resolve()
+              await sync
+            }
+          }
+          await queue.delete({
+            id: record.id,
+            updateHook: () => {
+              fixture.usersCollection.update(userId, (row) => {
+                row.photo_id = null
+              })
+            },
+          })
+          expect(
+            await fixture.db.get(`SELECT state FROM attachments WHERE id = ?`, [
+              record.id,
+            ]),
+          ).toEqual({ state: AttachmentState.QUEUED_DELETE })
+          expect(
+            await fixture.db.get(`SELECT photo_id FROM users WHERE id = ?`, [
+              userId,
+            ]),
+          ).toEqual({ photo_id: null })
+          release.resolve()
+          await sync
+          // Complete two real SDK passes, not a timeout waiting for a mock state.
+          await queue.syncStorage()
+          await queue.syncStorage()
+          expect(fixture.deleteFile).toHaveBeenCalledTimes(1)
+          expect(remoteFiles.has(record.id)).toBe(false)
+          expect(await fixture.localStorage.fileExists(record.local_uri!)).toBe(
+            false,
+          )
+        } finally {
+          release.resolve()
+          await sync
+        }
+      },
+    )
+
+    it(`keeps a native SDK delete queued when upload completion arrives`, async () => {
+      const fixture = await setup()
+      const queue = new AttachmentQueue({
+        db: fixture.db,
+        localStorage: fixture.localStorage,
+        remoteStorage: fixture.remoteStorage,
+        watchAttachments: () => {},
+      })
+      const uploaded = pDefer<void>()
+      const release = pDefer<void>()
+      fixture.uploadFile.mockImplementation(async () => {
+        uploaded.resolve()
+        await release.promise
+      })
+      const record = await queue.saveFile({
+        data: createMockJpegBuffer(),
+        fileExtension: `jpg`,
+      })
+      const sync = queue.syncStorage()
+      try {
+        await uploaded.promise
+        await queue.deleteFile({ id: record.id })
+        expect(
+          await fixture.db.get(`SELECT state FROM attachments WHERE id = ?`, [
+            record.id,
+          ]),
+        ).toEqual({ state: AttachmentState.QUEUED_DELETE })
+        release.resolve()
+        await sync
+        expect(
+          await fixture.db.get(`SELECT state FROM attachments WHERE id = ?`, [
+            record.id,
+          ]),
+        ).toEqual({ state: AttachmentState.QUEUED_DELETE })
+      } finally {
+        release.resolve()
+        await sync
+        await queue.stopSync()
+      }
+    })
+
     it(`writes the local file and inserts a QUEUED_UPLOAD row into the collection`, async () => {
       const { createQueue, attachmentsCollection, localStorage } = await setup()
       const queue = createQueue()

```

</details>

KyleAMathews:
Companion fixes and oracle: https://github.com/powersync-ja/tanstack-db/pull/10 (targets your feat/powersync-attachments branch). Tanner’s startup fix informed a targeted-ID live-query lease that also works with on-demand collections. It adds duplicate/file-ownership guards and preserves SDK filename-based storage relocation. Local gate: 121 tests pass, types/lint/build pass. The upload/delete completion race remains an upstream blocker: the companion includes five separately runnable failing native-SDK/integration checks for both success and retry completion, with no SDK workaround or expected-failure waiver. See the companion body and ATTACHMENT-ORACLE.md for commands and exact scope.

PR #1617, opened 2026-06-25
proposal: PostgreSQL float semantics for NaN (discussion)

kevin-dp:
## Summary

This is a **proposal / discussion PR**: it makes the query engine follow **PostgreSQL float semantics for `NaN`** instead of raw JavaScript semantics. I'd like maintainer input before considering it for merge.

### The problem

`NaN === NaN` is `false` in JavaScript, and the WHERE evaluator inherited that:

- `eq(row.value, NaN)` matches nothing, even for rows whose value is `NaN`.
- `NaN` has no consistent order (it's neither `<`, `>`, nor `=` anything), so it can't be sorted or stored in a tree-based index deterministically. A single stored `NaN` can corrupt range traversal and make ordering depend on insertion order.

Critically, this is also **internally inconsistent today**: joins, `groupBy`, and `distinct` already treat `NaN = NaN` (they match/group keys via a hash/`Map` index, and JS `Map` treats `NaN` as equal to `NaN`), while `WHERE eq(x, NaN)` returns nothing. The same value behaves differently depending on where it appears.

### The change

Following PostgreSQL — which deliberately overrides IEEE-754 *precisely so floats can be sorted and indexed* — `NaN` (and invalid `Date` values, whose timestamp is `NaN`) is now treated as:

- **equal to itself**, and
- **greater than every other non-null value**.

Concretely:

- `ascComparator` places `NaN`/invalid Dates last (greatest non-null), giving a valid total order.
- The evaluator's `eq` / `gt` / `gte` / `lt` / `lte` / `in` implement the same: `eq(x, NaN)` matches `NaN` rows, range ops treat `NaN` as greatest, `IN` matches a `NaN` member.
- `null` / `undefined` are unchanged — still three-valued logic (a comparison with `null` is `UNKNOWN`), still ordered by `NULLS FIRST` / `NULLS LAST`.

Because the index ordering and the evaluator now **agree** on `NaN`, index-served and full-scan queries return identical results with **no re-filtering** — the `nan-semantics.test.ts` suite asserts this equivalence for `eq`/`gt`/`lt`/`in`.

### End-to-end consistency (verified)

Every place `NaN` can be compared or grouped now agrees on Postgres semantics. I verified each:

| Site | Behavior with this PR |
|---|---|
| **WHERE** (`eq`/`gt`/`gte`/`lt`/`lte`/`in`) | matches `NaN` as a value greater than all numbers (this PR) |
| **Ordering** (`currentStateAsChanges` and live-query `orderBy`) | `NaN` sorts last |
| **Joins** (equi-join on a `NaN` key) | `NaN` rows join — already true on `main`, via the hash/`Map` index |
| **`groupBy`** | `NaN` rows form a single group — already true on `main` |
| **`distinct`** | `NaN` dedups to one value — already true on `main` |

Joins/`groupBy`/`distinct` already used `NaN = NaN` (SameValueZero in the incremental layer). On `main`, WHERE/ordering disagreed with them; this PR makes WHERE/ordering match, so it **removes** the existing inconsistency rather than introducing a gap.

### Why this matters / context

This came out of a review thread on #1582 (index-optimization correctness). That PR keeps JavaScript `NaN` semantics (`NaN != NaN`) and handles the index/ordering problem by giving `NaN` a stable position *for sorting only* while the evaluator still rejects it — correct, but a JS/SQL hybrid that leaves the join/group-vs-filter mismatch in place. This PR is the alternative: pick one coherent, SQL-faithful model end-to-end. If the team prefers this direction, it would supersede the `NaN`-specific bits of #1582.

### The one open question

This is fundamentally a **semantics decision**: do we want Postgres float semantics (`NaN = NaN`, `NaN` greatest) as the documented contract? It's the most SQL-faithful and internally consistent option — and it's what the join/group/distinct machinery already does — but it is a user-facing change for anyone relying on JS `NaN` semantics, so it deserves an explicit, documented call. `null`/`undefined` are intentionally left as three-valued logic (absence / `UNKNOWN`), distinct from `NaN` (a concrete greatest value), matching Postgres.

### Testing & docs

- New `tests/comparison.test.ts` (comparator: `NaN` sorts last, invalid Dates, null still first).
- New `tests/nan-semantics.test.ts` (collection-level `eq`/`gt`/`lt`/`in`/`orderBy`, asserting indexed == full-scan).
- Extended `tests/query/compiler/evaluators.test.ts` with a `NaN` block.
- Documented under "Comparison semantics" in the live-queries guide; changeset included (minor).
- Full `@tanstack/db` suite green (2414 tests). No existing test asserted `NaN` behavior (the property tests explicitly excluded it because the old comparator broke on it).

### DIFF
--- a/.changeset/nan-postgres-semantics.md
+++ b/.changeset/nan-postgres-semantics.md
@@ -1,5 +1,5 @@
 ---
-"@tanstack/db": minor
+'@tanstack/db': minor
 ---
 
 Adopt PostgreSQL float semantics for `NaN` in `where` clauses and ordering.
--- a/.changeset/nan-postgres-semantics.md
+++ b/.changeset/nan-postgres-semantics.md
@@ -1,0 +1,15 @@
+---
+"@tanstack/db": minor
+---
+
+Adopt PostgreSQL float semantics for `NaN` in `where` clauses and ordering.
+
+`NaN` (and invalid `Date` values, whose timestamp is `NaN`) previously had no consistent order — `NaN === NaN` is `false` in JavaScript, so `NaN` compared unequal to everything and could not be sorted or indexed deterministically. Following PostgreSQL, `NaN` is now treated as **equal to itself** and **greater than every other non-null value**:
+
+- `eq(row.value, NaN)` matches rows whose value is `NaN`; `inArray(row.value, [NaN, ...])` matches them too.
+- Range comparisons treat `NaN` as the greatest value: `gt`/`gte` include it, `lt`/`lte` exclude it.
+- Ordering by a field containing `NaN` is now deterministic, with `NaN` sorting last (and `null` still ordered by `NULLS FIRST`/`NULLS LAST`).
+
+`null`/`undefined` are unaffected: they continue to use three-valued logic (a comparison with `null` yields `UNKNOWN`).
+
+This makes results independent of whether a query is served from an index or a full scan.
--- a/docs/guides/live-queries.md
+++ b/docs/guides/live-queries.md
@@ -728,6 +728,13 @@ not(condition)
 
 For a complete reference of all available functions, see the [Expression Functions Reference](#expression-functions-reference) section.
 
+### Comparison semantics
+
+Comparisons follow SQL/PostgreSQL conventions rather than raw JavaScript:
+
+- **`null` / `undefined` use three-valued logic.** Any comparison involving `null` or `undefined` evaluates to `UNKNOWN`, so the row is not matched. For example `eq(user.score, null)` matches nothing — use a dedicated null check (e.g. `isUndefined`) to match missing values.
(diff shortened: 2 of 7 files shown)


kevin-dp:
Superseded by #1582.

The PostgreSQL `NaN` semantics proposed here have been folded directly into #1582, so the index-optimization correctness work is built on the final, coherent contract rather than the interim JS-semantics handling. The integrated change carries the full set:

- `comparison.ts` — `ascComparator` orders `NaN`/invalid Dates as the greatest non-null value; `isUnorderable` exported.
- `evaluators.ts` — `eq`/`gt`/`gte`/`lt`/`lte`/`in` follow PostgreSQL float semantics.
- `index-optimization.ts` — the integration this proposal didn't cover: because the index and evaluator now agree on `NaN`/invalid Dates, the optimizer treats them as **exact** (no re-filter) and stops treating invalid Dates as range-divergent. This also resolves the invalid-Date `eq`/`IN` finding from the #1582 review thread.
- Tests (`nan-semantics.test.ts`, PG-semantics `comparison.test.ts`, evaluator `NaN` block), the live-queries doc note, and a minor changeset all came across; the NaN tests that encoded the old JS behavior were updated.

Closing in favor of #1582. Thanks for the discussion that shaped the direction.

PR #1619, opened 2026-06-25
fix(db-sqlite-persistence-core): rows marked with metadata prior to insert get reset causing stale data

### LINKED ISSUE #1618: ExpoSQLitePersistence + queryCollectionOptions + offline-transactions - stale items after insert

**Describe the bug**
With persistedCollectionOptions (ExpoSQLitePersistence) + queryCollectionOptions + offline-transactions (although I imagine it probably occurs without offline-transactions given the diagnosis below)

I've done a little bit of debugging and it seems to point to the owners metadata not getting set properly when inserting, so when the query responds on app launch it doesn't clean up the stale data.

**To Reproduce**

1. Insert item on app (sync occurs via the offlineExecutor mutationFn) - item appears in db + api response
2. close app
3. delete the inserted item from the db (no longer appears in list items API response)
4. open app to list items

**Expected behavior**
I would expect in step 4, it would realise that the API response doesn't contain the inserted item anymore, and clean up the stale item. Instead, the stale item remains, despite the API response returning [].

**Screenshots**

- Device: iPhone 13 Pro
- OS: iOS 26.5
- Expo Development Build SDK 56

**Additional context**
Note there is another bug related to the typing of `SQLite.SQLiteDatabase` -> `ExpoSQLiteDatabaseLike` so I initialise like this: `const persistence = createExpoSQLitePersistence({
    database: database as unknown as ExpoSQLiteDatabaseLike,
});` - but this shouldn't cause this issue. There is a PR addressing this here (I haven't tested - but it has been raised): https://github.com/TanStack/db/pull/1560

tombryden:
## 🎯 Changes

Fixes https://github.com/TanStack/db/issues/1618 - rows in the same transaction marked with metadata get their metadata reset on insert, causing bugs like stale data when used with `queryCollectionOptions`

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where stale item metadata could remain after inserts, causing outdated items to persist.
  * Preserved existing row metadata when an insert occurs in the same transaction as an explicit metadata update.
  * Cleared leftover metadata only when no newer metadata update was already queued, improving sync accuracy.

### DIFF
--- a/.changeset/small-tables-listen.md
+++ b/.changeset/small-tables-listen.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/db-sqlite-persistence-core': patch
+---
+
+Fixed bug where internal metadata would get reset on insert causing stale items to remain
--- a/packages/db-sqlite-persistence-core/src/persisted.ts
+++ b/packages/db-sqlite-persistence-core/src/persisted.ts
@@ -2290,9 +2290,21 @@ function createWrappedSyncConfig<
             message.type === `insert` &&
             normalization.operation.metadata === undefined
           ) {
-            openTransaction.rowMetadataWrites.set(normalization.operation.key, {
-              type: `delete`,
-            })
+            // Reset stale metadata for a fresh insert, but don't clobber an
+            // explicit metadata write already queued for this key in the same
+            // transaction (e.g. query reconcile stamps owners, then inserts).
+            if (
+              !openTransaction.rowMetadataWrites.has(
+                normalization.operation.key,
+              )
+            ) {
+              openTransaction.rowMetadataWrites.set(
+                normalization.operation.key,
+                {
+                  type: `delete`,
+                },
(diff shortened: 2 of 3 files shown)


kevin-dp:
Hi @tombryden, thanks for addressing this issue!
The fix seems to be spot on! 🥇 

I'm closing this PR in favor of https://github.com/TanStack/db/pull/1626, i added a higher-level more end-to-end reproduction and cherry picked your fix such that attribution is correct.

tombryden:
> Hi @tombryden, thanks for addressing this issue! The fix seems to be spot on! 🥇
> 
> I'm closing this PR in favor of #1626, i added a higher-level more end-to-end reproduction and cherry picked your fix such that attribution is correct.

Great, thanks!

PR #1627, opened 2026-06-30
Fix unused vars

Jaime02:
Fix CI. Unused vars should use underscore prefix or have an `// eslint-disable-next-line @typescript-eslint/no-unused-vars`

## Summary by CodeRabbit

* **Tests**
  * Updated a type-level test to better verify that omitting an optional field through destructuring does not bring that field back into the resulting type.


KyleAMathews (approved):
Thanks!

PR #1630, opened 2026-06-30
RFC: Mutation log reconciliation for optimistic writes

KyleAMathews:
## Summary

Implements Phase 1 of Mutation Log reconciliation for optimistic writes. Committed sync/base updates now apply immediately while unsettled optimistic mutations remain projected in visible collection state, improving reconciliation stability for collections, `subscribeChanges`, and live-query consumers.

## References

- RFC: [`docs/rfcs/2026-06-25-mutation-log-reconciliation.md`](https://github.com/TanStack/db/blob/rfc-transactions-sync-state/docs/rfcs/2026-06-25-mutation-log-reconciliation.md)

## Root Cause

TanStack DB previously treated persisting local transactions as a reason to delay normal committed sync application. That avoided some flicker, but it also meant synced/base state could lag behind while optimistic mutations were active, making materialized visible state and change events depend on delayed-sync branches instead of a single reconciliation model.

## Approach

This PR shifts collection state toward the invariant:

```txt
authoritative synced/base state
+ unsettled optimistic mutations owned by the collection
= visible collection state
```

Key implementation details:

- Apply committed sync/base transactions immediately in `CollectionStateManager.commitPendingTransactions()`.
- Capture previous visible state before sync/base mutation, then diff against final visible state after active optimistic mutations are projected.
- Keep transaction state as the lifecycle source of truth; mutations derive from their owning transaction.
- Preserve virtual-property transitions like `$synced: false -> true` when they are part of subscribed visible values.
- Use keyed authoritative sync payloads in tests; this intentionally does not infer keys or server identity from matching row contents.

## Key Invariants

- A persisting transaction must not block unrelated committed sync/base rows from becoming visible.
- An active optimistic mutation must remain visible over newer base state until the owning transaction settles and reconciliation removes it.
- Change events are emitted from previous visible state to next visible state, not from raw synced/base deltas alone.
- Identical sync confirmation should not produce duplicate user-data insert/update events, but virtual-only confirmation updates should still be emitted where materialized.
- Sync payloads must provide normal keys; this PR does not add key-inference fallbacks or server-key matching heuristics.

## Non-goals

This is intentionally limited to the Phase 1 internal reconciliation slice. It does not add:

- public write-status APIs like `$hasPendingWrites`, `$writeStatus`, `tx.when(...)`, or `db.mutations`
- transport confirmation states like `accepted` or `observed`
- operation targeting, base-only query modes, or sync batch API redesigns
- server-key/stable-identity matching heuristics
- offline durability changes

## Trade-offs

The implementation keeps the existing mutation/transaction primitives instead of introducing a new operation lifecycle. That keeps Phase 1 smaller and aligned with current public vocabulary, but it leaves future public mutation-log/query APIs as follow-up work rather than mixing them into this reconciliation change.

## Verification

```bash
pnpm --filter @tanstack/db exec tsc --noEmit
pnpm --filter @tanstack/db exec vitest run tests/collection-subscribe-changes.test.ts tests/collection.test.ts --pool-options.threads.maxThreads=2
pnpm --filter @tanstack/db test
```

Local validation during prep:

- focused collection / subscribeChanges tests passed
- full `@tanstack/db` suite passed before final prep-review fixes

## Files changed

- `packages/db/src/collection/state.ts` — immediate sync/base reconciliation, optimistic projection, visible-state diffing, cleanup/confirmation handling.
- `packages/db/src/collection/sync.ts` — small sync write adjustment for keyed authoritative payload handling.
- `packages/db/tests/collection.test.ts` — direct collection reconciliation regressions.
- `packages/db/tests/collection-subscribe-changes.test.ts` — stable change-event, metadata, virtual prop, and confirmation regressions.
- `packages/db/tests/query/live-query-collection.test.ts` and `packages/db/tests/query/query-while-syncing.test.ts` — live-query behavior under immediate-base reconciliation.
- `docs/rfcs/2026-06-25-mutation-log-reconciliation.md` — RFC for the mutation-log reconciliation direction and future/non-Phase-1 boundaries.
- `docs/superpowers/plans/2026-06-29-mutation-log-reconciliation.md` — implementation plan/history for this slice.
- `.changeset/mutation-log-reconciliation.md` — patch changeset for `@tanstack/db`.


KyleAMathews:
Closing this iteration after RFC clarification: generic immediate sync while mutationFns are pending is not sound without explicit temp/server identity mapping. We'll start a smaller follow-up PR scoped to the remaining valid work.

Issue #1623, opened 2026-06-26
RFC: Live Query Framework Bindings Platform


KyleAMathews:
# Live Query Framework Bindings Platform

Date: 2026-06-25

## Summary

TanStack DB needs an official internal framework-binding platform, not more framework-by-framework ports of React hooks.

The current issue cluster spans React, Vue, Svelte, Solid, Angular, infinite queries, and SSR. The repeated root problem is that each adapter owns too much of the same semantic state machine: input normalization, collection ownership, sync activation, status/readiness/error tracking, result cardinality, initial snapshots, subscription timing, ordered materialization, pagination windows, SSR identity, and stale async protection.

This RFC proposes a full binding platform:

```text
DbClient / scope / identity / hydration
        │
        ▼
Internal LiveQueryObserver contract
        │
        ├── stable snapshots
        ├── lifecycle + ownership
        ├── status/error/readiness
        ├── cardinality
        ├── non-reentrant subscriptions
        └── stale async protection
        │
        ▼
Optional shared controllers
        ├── layout revisions for ordered snapshots
        └── infinite/windowed query controller
        │
        ▼
Framework-native materializers
        ├── React
        ├── Vue
        ├── Svelte
        ├── Solid
        └── Angular
        │
        ▼
Shared adapter conformance suite
```

The key principle is:

> Share DB/query semantics and lifecycle; keep renderer state management framework-native.

React should not be the reference implementation. Solid should not be the reference implementation. Vue should not be asked to match React or Solid. All official adapters should match the same internal observable contract while using their own native reactivity models.

`getSnapshot()` is the baseline semantic contract, not the whole materialization strategy. It gives every adapter a canonical, synchronous, revisioned view of a live query. It does not require Vue, Svelte, Solid, or Angular to become React-shaped `useSyncExternalStore` wrappers. Those adapters can still materialize the snapshot into native reactive objects, signals, stores, proxies, resources, or row-level subscriptions. Fine-grained materialization is a follow-up workstream, not a prerequisite for centralizing correctness.

## Implementation progress

- [x] **Phase 1 — shared contract tests.** #1636 added one conformance suite for React, Vue, Svelte, Solid, and Angular. It established the cross-adapter safety net and tracked known gaps explicitly.
- [x] **Phase 2 — shared adapter helpers.** #1641 extracted the reactivity-free input, cardinality, and status helpers and adopted them across all five adapters.
[...]

PR #1636, opened 2026-07-02
test: useLiveQuery conformance suite across all five framework adapters (RFC #1623)

kevin-dp:
Draft — first slice of RFC #1623 (framework-binding platform). Discussion: https://github.com/TanStack/db/issues/1623

## Scope — this is test infrastructure only

**This PR is entirely test-only — no `src/` changes, no shared production code.** It does **not** extract the duplicated adapter logic into a shared observer/module; that's the *next* milestone (RFC step 2+). This PR builds the **safety net** that must exist *before* that extraction, so we can rip shared logic out of five adapters and prove behavior is unchanged.

The per-framework files here (`packages/*/tests/conformance.test.*`) are small **test harnesses**, not new production abstractions. Each just wraps its framework's existing `useLiveQuery` (mount → read → flush → unmount) so one shared spec can run against all five. (In the code the harness interface is named `LiveQueryDriver` — "driver" in the JUnit/conformance-testing sense of "the thing that drives the tests," not a runtime adapter.)

## What this is

A shared, framework-agnostic **conformance suite** for the `useLiveQuery` adapters. One behavioral spec (`packages/db/tests/conformance/suite.ts`, 24 scenarios) runs against **all five adapters** through a thin per-framework test harness implementing a common contract (`packages/db/tests/conformance/contract.ts`).

This is deliberately the lowest-risk, highest-consensus piece of #1623: it documents current behavior, commits to no architecture, and becomes the safety net for the later observer/adapter refactor.

## How it's sourced

Bottom-up from the **union of the existing adapter test suites** — every fixed bug already left a regression test behind, so those tests are the real contract. The diff between adapters' suites *is* the drift #1623 is about, so folding them into one spec surfaces it. The RFC's aspirational list contributes only the small additive tail (the #1601 order-only-move case, a universal expected-fail).

## Design notes

- **Realm-safe injection.** The per-framework harness — not the scenarios — creates collections and supplies query operators, both from the *adapter's* `@tanstack/db`. Scenarios never import `@tanstack/db`, avoiding the dual-package `instanceof CollectionImpl` mismatch.
- **Per-adapter `knownGaps`, populated empirically.** Each scenario runs as `it` or `it.fails` per adapter based on `knownGaps`. A key is added only when it *actually* fails on a run — the matrix says where to look, the run says what's broken. When a gap closes, `it.fails` errors ("expected to fail but passed") prompting removal. So drift and bugs are documented without the suite going red, and the `knownGaps` lists double as the follow-up to-do list.
- **`apply(fn)` handle primitive.** A cross-framework "run a mutation inside the framework's update scope, then settle" (React `act` / Vue `nextTick` / Svelte `flushSync` / Solid `batch`) — needed for the optimistic-mutation scenario.

## Coverage (24 scenarios)

Query/where/select, live insert/update/delete, orderBy, join, groupBy/aggregate, nested aggregates, `.includes` subqueries, findOne cardinality (+ reactive update/delete), disabled + enable/disable transitions, deferred readiness / eager-visible-while-loading / ready-with-no-data, param recompilation, optimistic mutation, pre-created & config-object inputs, error status, and the #1601 order-only-move tail.

## Results — all five adapters covered

Each passes 24/24 (documented `knownGaps` run as expected-fail). The suite surfaced **three real cross-adapter divergences/bugs**:

| Adapter | knownGaps | Finding | Fix |
|---|---|---|---|
| **React** | — | reference harness, clean | — |
| **Vue** | — | clean — and closes Vue's **`findOne` coverage gap** (vue-db had zero findOne tests; it works, was just untested) | — |
| **Svelte** | `disabled-explicit`, `disabled-transition` | **Bug:** `toValue()` (added for the reactive `() => collection` form) calls a `() => null` disabled query as a getter, unwraps it to null, and passes `{...null}` into `createLiveQueryCollection` → crash in `getQueryIR`. The disabled short-circuit is unreachable; svelte-db has no disabled tests. | #1637 |
| **Solid** | `error-status` | **Divergence:** errors are surfaced by throwing (`CollectionStateError`) through the `createResource`/Suspense path for an `<ErrorBoundary>`, not via a readable `isError` flag like the others. | #1639 |
| **Angular** | `config-object-input` | **Divergence:** the plain `{ query }` config path calls `createLiveQueryCollection(opts)` without injecting `startSync: true` (its query-fn path does), so a bare config never syncs. The others auto-start; Angular requires explicit `startSync: true`. | #1638 |

Also confirmed **not** drift: every adapter eagerly starts a pre-created collection on mount (an earlier scenario wrongly tested this as React-specific; corrected to build over a not-ready source so it measures the actual contract — isReady stays false only when the *source* never readies).

## Out of scope for this suite (by design)

Framework-specific tests (Solid proxy identity, Svelte runes, React overloads, Angular DI lifecycle) stay per-adapter. Separate hook families (`useLiveInfiniteQuery`, `useLiveSuspenseQuery`, `useLiveQueryEffect`) get their own suites later.

## Follow-up PRs (stacked on this branch)

Each clears one `knownGap` and is stacked on this branch (re-target to `main` once this lands):

- **#1637** — `fix(svelte-db)`: the disabled-query crash (the only outright bug). Clears `disabled-explicit` + `disabled-transition`.
- **#1638** — `fix(angular-db)`: config-object input now syncs (defaults `startSync: true`). Clears `config-object-input`.
- **#1639** — `test(conformance)`: parametrizes error surfacing (`errorSurface: 'flag' | 'throw'`) rather than "fixing" Solid's idiomatic ErrorBoundary model. Clears `error-status`.

Each removes its `knownGap` entry, so the previously-`it.fails` scenario runs as a normal passing test — the suite itself proves the fix.

## Next milestone (NOT this PR)

The actual duplication cleanup — extracting the ~7 items each adapter copies (input normalization, disabled handling, cardinality, status derivation, ready-race, etc.) into a shared observer/module in `@tanstack/db`, then slimming each adapter's `src/useLiveQuery` down to it — is a **separate follow-up**. That's the PR that will show `src/` changes and shrink duplication. This suite exists to keep it honest: it must stay green through that refactor.

## Summary by CodeRabbit

* **New Features**
  * Added a shared live-query conformance test suite.
  * Added adapter drivers for Angular, React, Solid, Svelte, and Vue so the same query scenarios run consistently across frameworks.
  * Expanded coverage for filtering, inserts/updates/deletes, ordering, joins, grouping, deferred loading, disabled queries, errors, and optimistic updates.
* **Bug Fixes**
  * Added expected-fail tracking to keep known framework differences from breaking the suite.
  * Improved cleanup so mounted queries are torn down after each scenario.

### DIFF
--- a/packages/angular-db/tests/conformance.test.ts
+++ b/packages/angular-db/tests/conformance.test.ts
@@ -71,7 +71,9 @@ function makeSource<T extends { id: string }>(
   }
 }
 
-function makeDeferredSource<T extends { id: string }>(): DeferredSourceHandle<T> {
+function makeDeferredSource<
+  T extends { id: string },
+>(): DeferredSourceHandle<T> {
   const collection = createCollection(
     mockSyncCollectionOptionsNoInitialState<T>({
       id: `conformance-angular-${sourceSeq++}`,
--- a/packages/angular-db/tests/conformance.test.ts
+++ b/packages/angular-db/tests/conformance.test.ts
@@ -136,6 +136,7 @@ function makeHandle(result: any, destroy: () => void): LiveQueryHandle {
         status: result.status(),
         isReady: Boolean(result.isReady()),
         isError: Boolean(result.isError()),
+        // angular-db exposes no `isEnabled`; derive it from status (status-derived).
         isEnabled: result.status() !== `disabled`,
       }
     },
--- a/packages/angular-db/tests/conformance.test.ts
+++ b/packages/angular-db/tests/conformance.test.ts
@@ -1,0 +1,224 @@
+/**
+ * Angular driver for the shared live-query conformance suite (#1623).
+ *
+ * `injectLiveQuery` needs an injection context, so each mount runs inside a
+ * child `EnvironmentInjector` created off TestBed's; `unmount` calls
+ * `injector.destroy()`, firing the `DestroyRef` cleanup. Result signals are read
+ * after settling. Controllable inputs use Angular's reactive `{ params, query }`
+ * form driven by a signal.
+ *
+ * `knownGaps` is populated empirically from the run below.
+ */
+import {
+  EnvironmentInjector,
+  createEnvironmentInjector,
+  runInInjectionContext,
+  signal,
+} from '@angular/core'
+import { TestBed } from '@angular/core/testing'
+import {
+  coalesce,
+  count,
+  createCollection,
+  createLiveQueryCollection,
+  createOptimisticAction,
+  eq,
+  gt,
+  sum,
+} from '@tanstack/db'
+import {
+  mockSyncCollectionOptions,
+  mockSyncCollectionOptionsNoInitialState,
+} from '../../db/tests/utils'
+import { injectLiveQuery } from '../src/index'
+import { runSuite } from '../../db/tests/conformance/suite'
+import type {
+  ConformanceResult,
+  ControllableHandle,
+  DeferredSourceHandle,
+  LiveQueryDriver,
+  LiveQueryHandle,
+  QueryBuild,
+  SourceHandle,
+} from '../../db/tests/conformance/contract'
+
+let sourceSeq = 0
+
+function writer<T extends { id: string }>(collection: any) {
+  return (type: `insert` | `update` | `delete`, value: T) => {
+    collection.utils.begin()
+    collection.utils.write({ type, value })
+    collection.utils.commit()
+  }
+}
+
+function makeSource<T extends { id: string }>(
+  initialData: ReadonlyArray<T>,
+): SourceHandle<T> {
+  const collection = createCollection(
+    mockSyncCollectionOptions<T>({
+      id: `conformance-angular-${sourceSeq++}`,
+      getKey: (r) => r.id,
+      initialData: [...initialData],
+    }),
+  )
+  const write = writer<T>(collection)
+  return {
+    collection,
+    insert: (row) => write(`insert`, row),
+    update: (row) => write(`update`, row),
+    remove: (row) => write(`delete`, row),
+  }
+}
+
+function makeDeferredSource<T extends { id: string }>(): DeferredSourceHandle<T> {
+  const collection = createCollection(
+    mockSyncCollectionOptionsNoInitialState<T>({
+      id: `conformance-angular-${sourceSeq++}`,
+      getKey: (r) => r.id,
+    }),
+  )
+  collection.startSyncImmediate()
+  const write = writer<T>(collection)
+  return {
+    collection,
+    insert: (row) => write(`insert`, row),
+    update: (row) => write(`update`, row),
+    remove: (row) => write(`delete`, row),
+    emit: (rows) => {
+      collection.utils.begin()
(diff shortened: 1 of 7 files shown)


KyleAMathews (approved):
I think the conformance-suite shape is good, and I’m approving this direction. I did a pass over an external review and validated a few test-infrastructure issues that seem worth addressing, either in this PR or a quick follow-up:

1. The React conformance driver currently derives normalized `isEnabled` from `status !== 'disabled'`, so the suite doesn’t actually catch a broken React `isEnabled` value. I verified by intentionally returning `isEnabled: true` for disabled React queries: the conformance test still passed. Mapping React’s normalized field from `r.isEnabled` makes the disabled conformance scenarios catch it. For adapters that don’t expose a real `isEnabled` field/signal, deriving from status is fine, but the normalized field should probably be named/treated honestly as status-derived.

2. Expected-fail scenarios can leak mounted framework roots because `h.unmount()` is after assertions. I instrumented Solid roots and saw leaks from the expected-fail `error-status` and universal `order-only-move` scenarios. Wrapping mounted handles in `try/finally { h.unmount() }` fixes it.

3. `knownGaps` should be validated against registered scenario keys. I added a bogus React gap and the suite still passed. A small `scenarioKeys` set plus validation test catches stale/misspelled gaps.

4. `includes-subquery` is under-asserted. The suggested stronger assertion is directionally right, though `johnSmith.issues` is a child collection rather than `[]`, so the strengthened assertion should read child collection contents through the collection API.

5. The `suite.ts` header is stale: it says engine-heavy scenarios are “ported next,” but this PR already includes them.

I also checked for bidi/Trojan Source controls in the added conformance files and didn’t find any, so I wouldn’t block on that.

kevin-dp:
Thanks for the careful pass, Kyle — all five were legit. Addressed each as its own commit on this branch:

1. **React `isEnabled` was status-derived, masking a broken value** (`e5c131fb`) — the React driver now maps the normalized field from `r.isEnabled`, so the disabled scenarios actually catch it. The other four adapters don't expose an `isEnabled`, so they keep the status-derived value with an explicit comment noting that.

2. **Expected-fail scenarios leaked mounted roots** (`457d1099`) — the `mount*` methods are now wrapped to track every handle, and each scenario tears them all down in a `finally`, so teardown runs even when a scenario throws before its own `h.unmount()`. This covers both the universal `order-only-move` and any per-adapter `knownGap` (e.g. Solid `error-status`).

3. **`knownGaps` weren't validated** (`596bba2c`) — every scenario key is now registered, and a test asserts each driver's `knownGaps` (and `UNIVERSAL_EXPECTED_FAIL`) reference a real key. I confirmed it catches a bogus/misspelled gap. Handy since gaps are churning across the follow-up fix PRs.

4. **`includes-subquery` was under-asserted** (`4c117967`) — you're right that `john.issues` is a child collection, so it now reads the contents through the collection API and asserts John (id 1) has exactly `i1` and `i3`.

5. **Stale header** (`bed14e62`) — replaced the "ported next" STATUS note with the actual coverage list.

All five suites stay green at 25/25 (the +1 is the new `knownGaps` validation). No changes needed for the bidi/Trojan-source note. The rest of the stack (#1641/#1642 and the three fix PRs) has been rebased onto this and re-verified.

PR #1640, opened 2026-07-02
fix(db): avoid full row origin snapshots

samwillis:
## Summary

Fixes two collection/transaction performance hot paths:

- Avoids source-sized `rowOrigins` snapshots during incremental collection updates.
- Makes `Transaction.applyMutations` merge bulk mutations in O(n) instead of O(n²).

## Root Cause

Single-row incremental writes were cloning the full `rowOrigins` map before propagating changes. Since `rowOrigins` grows with the source collection, normal incremental writes paid O(source collection size) work before the query graph could process the actual changed keys.

Kevin also identified a separate bulk mutation issue: `Transaction.applyMutations` used `findIndex` for every incoming mutation, making large bulk operations quadratic.

## Changes

- Reuse the existing `rowOrigins` map during optimistic recompute, where confirmed origins are not mutated.
- Snapshot row origins only for keys affected by a synced commit, plus pending direct optimistic keys that may be cleaned up by sync confirmation.
- Merge transaction mutations through a `globalKey`-keyed `Map`, then rebuild the public mutations array in place to preserve identity.
- Keep the PR branch free of tracing or benchmark instrumentation.

## Validation

- `pnpm --dir packages/db exec tsc --noEmit`
- `pnpm --dir packages/db exec vitest --run tests/transactions.test.ts tests/collection-subscribe-changes.test.ts tests/collection-change-events.test.ts tests/collection-truncate.test.ts`

### DIFF
--- a/.changeset/quiet-row-origins.md
+++ b/.changeset/quiet-row-origins.md
@@ -2,4 +2,4 @@
 '@tanstack/db': patch
 ---
 
-Avoid full row origin snapshots during incremental collection updates.
+Avoid full row origin snapshots during incremental collection updates and make bulk mutation merging linear.
--- a/.changeset/quiet-row-origins.md
+++ b/.changeset/quiet-row-origins.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/db': patch
+---
+
+Avoid full row origin snapshots during incremental collection updates.
--- a/packages/db/src/collection/state.ts
+++ b/packages/db/src/collection/state.ts
@@ -226,6 +226,21 @@ export class CollectionStateManager<
     })
   }
 
+  private snapshotRowOriginsForKeys(
+    keys: Iterable<TKey>,
+  ): Map<TKey, VirtualOrigin> {
+    const rowOrigins = new Map<TKey, VirtualOrigin>()
+
+    for (const key of keys) {
+      const origin = this.rowOrigins.get(key)
+      if (origin !== undefined) {
+        rowOrigins.set(key, origin)
+      }
+    }
+
+    return rowOrigins
+  }
+
   private enrichWithVirtualPropsSnapshot(
     row: TOutput,
     virtualProps: VirtualRowProps<TKey>,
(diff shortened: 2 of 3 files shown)


kevin-dp:
We hit the same root cause in a parallel investigation (profiled the two `new Map(this.rowOrigins)` clones at >80% of per-write CPU at 50k rows) and ended up with a functionally equivalent fix — copy-on-write overlay instead of an eager bounded snapshot. Benchmarked both head-to-head: within noise, single-row update on a 50k-row collection goes ~9.5ms → ~0.18ms either way. So 👍 on this fix as-is.

One additional, independent hotspot we found along the way that this PR doesn't cover: **`Transaction.applyMutations` is O(n²) for bulk mutations.**

**Problem/cause:** `applyMutations` merges each incoming mutation via `this.mutations.findIndex((m) => m.globalKey === newMutation.globalKey)` — a linear scan of the already-accepted mutations for every incoming one. A bulk `collection.insert(rows)` with 50k rows is a single call with 50k mutations, so the k-th mutation scans ~k existing ones: ≈1.25 billion key comparisons total. In our profiling, seeding a 50k-row collection through the optimistic path took tens of seconds, with the majority of CPU inside `applyMutations`. (This doesn't affect the per-write numbers in the external benchmark — single-row transactions only ever scan a 1-element array — it bites bulk inserts/updates and initial data loads.)

**Fix:** merge through a `globalKey`-keyed `Map` instead of scanning. `Map` preserves insertion order and `set()` on an existing key keeps its original position, so the existing replace-in-place / cancel-out / append semantics are unchanged; the `mutations` array is rebuilt in place to preserve its identity since it's a public field. O(n²) → O(n) per call.

Commit: kevin-dp/db@5b900cd411fb42c711aeaf54fae478a12453b1f9 (on our investigation branch, `perf/avoid-per-write-row-origins-clone`). Happy to open it as a separate PR — probably cleaner than piggybacking on this one.

kevin-dp (approved):

PR #1641, opened 2026-07-02
refactor(db): extract shared live-query adapter helpers (RFC #1623 step 2)

kevin-dp:
The first real **deduplication** slice of RFC #1623 — the first PR that touches adapter `src/` and shrinks copy-paste. Follows the conformance suite (#1636), which guards this refactor.

**Base:** stacked on `test/live-query-conformance-suite` (#1636) so the conformance suite verifies the refactored adapters. Re-target to `main` once #1636 lands.

## What's extracted

Three helpers added to `@tanstack/db` (`packages/db/src/live-query-adapter.ts`), and all five adapters migrated to them:

- **`isCollection(value)`** — one structural, multi-realm-safe collection guard. Replaces the duck-typed check copy-pasted in React/Vue/Svelte/Angular **and** Solid's `instanceof CollectionImpl`. `instanceof` gives false negatives across dual-package/multi-realm boundaries — the exact hazard the conformance suite hit — so consolidating on duck typing is also a correctness win.
- **`isSingleResultCollection(collection)`** — shared `findOne` cardinality check (was `config.singleResult`, five times).
- **`getLiveQueryStatusFlags(status)`** — `status → { isLoading, isReady, isIdle, isError, isCleanedUp }`.

## Honest scope note

This slice is smaller than the RFC's "7 duplicated items" might suggest — and that's a real finding. Only the **reactivity-free** logic lifts cleanly into plain functions:

- `getLiveQueryStatusFlags` fits React's snapshot path, but the reactive adapters (Vue/Svelte/Solid/Angular) each expose every flag as its *own* `computed`/signal, so an object-returning helper doesn't fit them — forcing it would *add* code.
- Input/disabled classification, status derivation, and change→state projection are all **coupled to each framework's reactivity**. They can't be lifted into plain helpers without the observer.

So this PR takes the clean, universal wins now (notably unifying `isCollection` and killing the `instanceof` divergence). The larger dedup is exactly what the **observer (step 3)** exists to do — and per the design decision, that observer will carry both a snapshot *and* the `ChangeMessage[]` so granular adapters (Vue/Svelte) keep fine-grained updates while others consume the snapshot.

## Verification

No behavior change. All five adapter suites green (including the conformance suite that guards this), and `@tanstack/db` (2456 tests) green. Minor changeset for `@tanstack/db` (new exports) + patch for the adapters.

## Summary by CodeRabbit

* **New Features**
  * Added shared live-query helpers for detecting collections, handling single-result collections, and mapping query status flags.
  * Framework adapters now use the same shared logic for live queries, improving consistency across React, Vue, Svelte, Solid, and Angular.

* **Bug Fixes**
  * Improved collection detection in cases where objects come from different runtime contexts.
  * Single-result live queries now return the expected item more reliably across adapters.


KyleAMathews (approved):
:shipit:

PR #1648, opened 2026-07-07
perf(db): explore incremental update hot paths

samwillis:
## Draft scope

This is the original exploratory branch from the incremental update benchmark investigation. It is intentionally broad and should be reviewed as a bundle of instrumentation, benchmark tooling, and candidate performance changes rather than as a narrow production fix.

The branch currently contains 32 commits over `main` across 43 files.

### Instrumentation and benchmark tooling

- Adds opt-in perf tracing helpers for DB live queries and db-ivm operators.
- Adds trace aggregation coverage for the incremental query path, including hashing/operator timings.
- Adds `scripts/bench/incremental-update.ts` for reproducing the issue-style benchmark shapes and source-size experiments.
- Adds the investigation docs:
  - `INCREMENTAL_UPDATE_INSTRUMENTATION_PLAN.md`
  - `INCREMENTAL_UPDATE_FINDINGS.md`

### DB collection and sync hot paths

- Avoids full row origin snapshots and avoids storing implicit remote row origins.
- Adds faster single-op and two-row sync commit paths.
- Reduces redundant synced-data/index map lookups during common insert/update/delete paths.
- Trims virtual prop enrichment and simple sync update checks when they are not needed.
- Reduces incremental change delivery allocations and skips empty filtered subscription batches.
- Makes `Transaction.applyMutations` linear rather than repeatedly shifting work from the front of an array.

### Query/live graph hot paths

- Runs immediate live query graph work directly when there is no transaction context.
- Skips redundant pending graph checks.
- Avoids full nested include scans in the benchmarked include shapes.
- Trims live query update allocations.

### Index/sorted map/hash/db-ivm changes

- Adds fast paths for key-sorted `SortedMap` updates and duplicate lookup avoidance.
- Caches compiled index expression evaluators and reduces index timestamp allocation overhead.
- Reduces group aggregate hashing cost.
- Reuses group aggregate entries and specializes simple group aggregates.
- Adds fast paths for single-row joins.

## Why

The benchmark shared during the investigation showed very slow steady-state incremental updates for query shapes that look like filtered lists plus joins/includes/counts. The main question was whether the time was in source collection update application, hashing, live query scheduling, or the db-ivm query graph.

This branch keeps the instrumentation used to answer that question and the candidate optimizations that were explored from those traces.

## Review notes

- This is a draft PR for review/discussion and not currently shaped as a minimal merge candidate.
- Some production-ready fixes from this investigation have already been split out elsewhere. This branch preserves the full exploratory thread for comparison and review.
- The diff includes tracing and benchmark docs/tooling by design.

## Validation performed during the investigation

- Ran the local incremental benchmark repeatedly across multiple source sizes, including the larger issue/user/comment datasets used to approximate the shared benchmark.
- Compared baseline, this branch, the external performance PR branch, and the combined branch.
- Added targeted tests for trace aggregation and db-ivm groupBy incremental behavior.

No fresh full CI run was performed immediately before opening this draft PR.

### DIFF
--- a/INCREMENTAL_UPDATE_FINDINGS.md
+++ b/INCREMENTAL_UPDATE_FINDINGS.md
@@ -1,0 +1,295 @@
+# Incremental Update Performance Findings
+
+Date: 2026-07-02
+
+Branch: https://github.com/samwillis/db/tree/codex/incremental-update-instrumentation-plan
+
+Instrumentation commit: https://github.com/samwillis/db/commit/a13fd382a
+
+Source-index benchmark commit: https://github.com/samwillis/db/commit/1ee5188cf
+
+## Executive Summary
+
+We have identified the likely source of the slow incremental update results seen in the shared benchmark table.
+
+The incremental hot path is doing O(source-collection-size) snapshot work on every write. In particular, both synced and optimistic paths clone `rowOrigins` for the mutated collection:
+
+- synced path: `collection.commitPendingTransactions.snapshotState`
+- optimistic path: `collection.recomputeOptimisticState.snapshotState`
+
+At 100k source rows this snapshot work is about 95% of traced source-to-result time. At the estimated shared benchmark shape of `100k issues`, `100k users`, and `400k comments`, it is about 98% of traced source-to-result time.
+
+This makes the reported shared TanStack DB numbers plausible if their benchmark used roughly:
+
+- `100k` issues
+- `100k` users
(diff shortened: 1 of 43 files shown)


KyleAMathews:
Closing this exploratory draft to reduce the active queue. The branch and discussion remain available as an investigation record, while the shaped performance work remains in #1645 and split-out fixes.

Issue #1526, opened 2026-05-12
[PowerSync]: $synced reflects local SQLite persistence, not backend upload status


vitorcamachoo:
## Description

When using `@tanstack/powersync-db-collection`, the `$synced` virtual property on rows becomes `true` as soon as the optimistic mutation is resolved against the local SQLite database (via the diff trigger in `PowerSyncTransactor.applyTransaction`). It does **not** reflect whether the row has actually been uploaded to the backend via PowerSync's `uploadData` connector method.

This makes `$synced` unsuitable for determining whether a row has pending changes that need to be sent to the server -- which is the primary use case when building offline-first UIs (e.g. showing a "pending sync" indicator, or allowing edits only on unsynced rows).

## Current behaviour

1. `collection.insert(row)` -> row enters `optimisticUpserts` -> `$synced = false`
2. `onInsert` calls `transactor.applyTransaction()` -> writes to local SQLite
3. SQLite diff trigger fires -> TanStack DB sync handler picks up the change
4. `optimisticUpserts` is cleared -> `$synced = true`

This entire flow is local. Step 4 happens almost immediately after step 1, regardless of whether PowerSync is connected or whether `uploadData` has been called.

## Expected behaviour

For the PowerSync integration, `$synced` should remain `false` until the row has been confirmed by the backend (i.e. the CRUD entry has been removed from `ps_crud` and the data has been synced back via the PowerSync sync stream).

Alternatively, a separate virtual property (e.g. `$uploaded` or `$serverConfirmed`) could indicate whether the row's mutations have been acknowledged by the server, distinct from the local optimistic resolution.

## Workaround

The only reliable way to determine pending upload status is to query the `ps_crud` table directly:

```ts
const rows = await powerSyncDb.getAll<{ data: string }>('SELECT data FROM ps_crud');
const pendingIds = new Set<string>();
for (const row of rows) {
  const parsed = JSON.parse(row.data);
  if (parsed.type === 'my_table' && parsed.id) pendingIds.add(parsed.id);
}
```

This works but requires polling and is not reactive.

## Context

- `@tanstack/db` v0.6.5
- `@tanstack/powersync-db-collection`
- `@powersync/react-native`

Related: #900 (also about `$synced` resolving before server confirmation with PowerSync)

stevensJourney:
The concept of an "online mode" for PowerSync transactions has been on our radar for a while. 

[...]

PR #1219, opened 2026-02-06
docs: Add transaction submission status tracking guide

KyleAMathews:
## 🎯 Changes

Added comprehensive documentation for tracking submission status of offline transactions in the offline-transactions package README. This new section covers:

- **Using the Transaction Object**: Explains how to check transaction state and wait for server confirmation using `tx.state` and `tx.isPersisted.promise`
- **Tracking Per-Item Status**: Provides a practical example of maintaining a map of pending transactions to show sync indicators in UI
- **Monitoring the Offline Outbox**: Documents executor methods (`getPendingCount()`, `getRunningCount()`, `peekOutbox()`) for visibility into queued transactions and retry information
- **Using Inside mutationFns**: Shows how the transaction lifecycle works within mutation functions, including automatic state transitions and rollback behavior

This documentation addresses a common need in offline-first applications and provides developers with clear patterns and examples for implementing submission status tracking.

## ✅ Checklist

- [x] I have followed the steps in the [Contributing guide](https://github.com/TanStack/db/blob/main/CONTRIBUTING.md).
- [x] I have tested this code locally with `pnpm test:pr`.

## 🚀 Release Impact

- [x] This change is docs/CI/dev-only (no release).

https://claude.ai/code/session_01UHyeQv9yEn4FKK49mK9Vvw

PR #1215, opened 2026-02-04
docs: Add guide for multiple pending transactions

KyleAMathews:
## 🎯 Changes

Added documentation explaining how multiple pending transactions work in the mutations guide. This section clarifies the layering behavior when users make rapid sequential mutations to the same item before earlier ones have synced back from the server.

The new section includes:
- An explanation of how transactions are layered with the most recent taking priority
- A concrete code example showing rapid updates (A → B → C)
- A detailed walkthrough of how the optimistic state is recomputed as each transaction syncs back
- Context on why this behavior provides a good user experience for rapid sequential updates

## ✅ Checklist

- [x] I have followed the steps in the [Contributing guide](https://github.com/TanStack/db/blob/main/CONTRIBUTING.md).
- [x] I have tested this code locally with `pnpm test:pr`.

## 🚀 Release Impact

- [x] This change is docs/CI/dev-only (no release).

https://claude.ai/code/session_01HruHi63EH2HhdQy86tWteX

PR #1652, opened 2026-07-07
Clarify local write status docs and pending sync contract

KyleAMathews:
## Summary

Clarifies TanStack DB's existing local write status docs and tightens the core pending-sync contract around optimistic temp-key inserts. User-visible docs now distinguish local optimistic write state from backend upload/read-path confirmation, and the core test makes explicit that ambiguous server-key sync is not exposed while the local mutation is still pending.

## Root cause

`$synced` and `tx.isPersisted.promise` were easy to read as backend confirmation APIs. In practice, they describe TanStack DB's local transaction/projection state: a row may have no pending local optimistic mutation without proving backend read-back, and a transaction may settle without representing a generic sync-observation guarantee.

## Approach

- Reword `$synced` JSDoc to describe whether a row currently has pending local optimistic mutations in the collection's visible projection.
- Reword `isPersisted.promise` JSDoc to describe transaction settlement, including rollback/failure, while noting that non-empty commits normally settle through the mutation function.
- Tighten the core collection test for an optimistic insert using a temporary key followed by synced data under a different server key. While the mutation is pending, the possible server-key echo stays queued and the collection does not expose both rows.
- Add a changeset for `@tanstack/db`.

## Key invariants

- `$synced` is local write/projection status; it is not backend upload, confirmation, or read-path observation.
- `tx.isPersisted.promise` is the promise to await for transaction settlement; callers should not await the deferred object itself.
- Without explicit temp-key to server-key mapping, core must not expose both an optimistic temp row and a possible server-key echo while the mutation function is still pending.

## Non-goals

- No new public write status APIs in this PR.
- No mutation log rewrite.
- No generic immediate sync application while mutation functions are pending.
- No temp-key/server-key mapping or mutation receipts.
- No generated reference docs updates; this PR updates the source JSDoc that feeds them.

## Verification

```bash
pnpm --filter @tanstack/db exec vitest run tests/collection.test.ts --pool-options.threads.maxThreads=2
pnpm --filter @tanstack/db exec tsc --noEmit
```

## Files changed

- `.changeset/clarify-local-write-status.md` — records the patch-level docs/test clarification for `@tanstack/db`.
- `packages/db/src/virtual-props.ts` — clarifies `$synced` as local pending optimistic write state.
- `packages/db/src/transactions.ts` — clarifies `isPersisted.promise` transaction settlement semantics.
- `packages/db/src/collection/state.ts` — aligns an internal `$synced` helper comment with the new wording.
- `packages/db/tests/collection.test.ts` — makes the ambiguous temp/server-key pending sync contract explicit.

---

Refs #20, #661, #1017, #1166, #1215, #1219, #1322, #1431, #1526.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that local sync indicators (for `$synced` and row sync status) reflect pending local optimistic writes, not backend-confirmed state.
  * Expanded guidance on `isPersisted.promise` lifecycle semantics.
* **Bug Fixes**
  * Ensured server-generated keys from ambiguous sync are not exposed until the related optimistic insert has finished persisting.
* **Tests**
  * Updated and added coverage for the queued server-key sync scenario while a temp-key optimistic insert is still pending.


packages/db/src/collection/state.ts
```diff
@@ -154,7 +154,7 @@ export class CollectionStateManager<
   }
 
   /**
-   * Checks if a row has pending optimistic mutations (not yet confirmed by sync).
```
kevin-dp, line 157: I find "(not yet confirmed by sync)" clearer than "in the visible projection"

kevin-dp (approved):

packages/db/src/collection/state.ts
KyleAMathews: One of the clarifications this PR is trying to make is that not all "confirmations by the collection" means that the write has been persisted to the backend e.g. powersync just confirms when the write is in the local sqlite. I pushed this commit which hopefully is clearer https://github.com/TanStack/db/pull/1652/commits/4bbc2689995ec9719237acefb79db1c38caf6fe1

Issue #1643, opened 2026-07-02
RFC: roadmap for query-db-collection / TanStack Query integration


KyleAMathews:
### Outcome

The staged RFC work is complete on `main`. It clarified and hardened the boundary without adding a parallel Query API, a nested Query-options bag, or a new runtime binding abstraction:

- TanStack Query owns fetching, retries, staleness, invalidation, focus/reconnect behavior, cancellation, hydration, persistence, and response caching.
- TanStack DB owns row identity, materialization, local queries, optimistic transactions, and normalized row state.
- The adapter owns projection between Query responses and DB rows, plus the ownership bookkeeping that connects Query results to materialized rows.

### Completed, merged work

- **Persistence-safe on-demand metadata** — [#1644](https://github.com/TanStack/db/pull/1644) keeps runtime subscription state out of dehydrated Query metadata and locks the contract with `structuredClone(dehydrate(queryClient))` coverage.
- **Invalidation behavior matrix** — [#1655](https://github.com/TanStack/db/pull/1655) covers eager/on-demand exact and prefix invalidation, overlapping subsets, failed refetch retention, inactive cached queries, and removed queries. TanStack Query remains the invalidation authority.
- **Option and projection contract** — [#1653](https://github.com/TanStack/db/pull/1653), [#1654](https://github.com/TanStack/db/pull/1654), and [#1665](https://github.com/TanStack/db/pull/1665) document and extend the supported flat top-level observer options, preserve `QueryClient.defaultOptions`, and define adapter `select` as row extraction while Query retains the original response envelope.
- **Runtime and business scoping guidance** — [#1667](https://github.com/TanStack/db/pull/1667) and [#1671](https://github.com/TanStack/db/pull/1671) document plain factories and stable memoization by runtime `QueryClient` and business scope. No binding API was needed.
- **Ownership and lifecycle hardening** — [#1664](https://github.com/TanStack/db/pull/1664), [#1672](https://github.com/TanStack/db/pull/1672), [#1673](https://github.com/TanStack/db/pull/1673), and [#1681](https://github.com/TanStack/db/pull/1681) simplify row/query ownership, preserve authoritative empty results, release temporary readiness listeners, and cover late success and rejection after subset cleanup. #1672 resolved its requested simplification before merge.
[...]

Issue #346, opened 2025-07-31
Add Initial Data and Query Options Support to query-db-collection


KyleAMathews:
### Background

TanStack Query provides several options for data initialization and transformation that are not currently exposed in `query-db-collection`. While TanStack DB has its own patterns (like live queries with `.select()`), some Query options would enhance the developer experience and provide feature parity.

### Current State

- **initialData**: Only supported in `localOnlyCollectionOptions`, not in `queryCollectionOptions`
- **placeholderData**: Not supported
- **structuralSharing**: Hard-coded to `true` in query.ts:320
- **refetchOnWindowFocus/refetchOnReconnect**: Not configurable

#### 1. Initial Data Support

Add `initialData` option to `QueryCollectionConfig` for pre-populating collections:

```typescript
export interface QueryCollectionConfig<TItem, TError, TQueryKey> {
  // ... existing options ...
  
  /**
   * Initial data to populate the collection before the first fetch
   * Useful for SSR hydration or cached data from previous sessions
   */
  initialData?: Array<TItem> | (() => Array<TItem>)
  
  /**
   * When the initial data was last updated (for staleness calculations)
   */
  initialDataUpdatedAt?: number | (() => number)
}
```

Implementation approach:
- Apply initial data before starting the query observer
- Mark as synced data so it appears immediately in live queries
- Let TanStack Query's staleness logic determine if refetch is needed

#### 2. Structural Sharing Configuration

Allow customizing how data updates are merged:

```typescript
export interface QueryCollectionConfig<TItem, TError, TQueryKey> {
  // ... existing options ...
  
  /**
   * How to structurally share data between updates
   * - true: Use TanStack Query's default structural sharing
   * - false: Always use new references
   * - function: Custom sharing logic
   */
  structuralSharing?: boolean | ((oldData: Array<TItem>, newData: Array<TItem>) => Array<TItem>)
}
```

#### 3. Refetch Triggers

Expose window focus and reconnect options:

```typescript
export interface QueryCollectionConfig<TItem, TError, TQueryKey> {
  // ... existing options ...
  
  /**
   * Refetch on window focus
   * @default true (when data is stale)
   */
  refetchOnWindowFocus?: boolean | ((query: Query) => boolean)
  
  /**
   * Refetch on reconnect
   * @default true (when data is stale)
   */
  refetchOnReconnect?: boolean | ((query: Query) => boolean)
  
  /**
   * Network mode for offline behavior
   * @default 'online'
   */
[...]

Issue #183, opened 2025-06-17
Deepen the integration with Query


KyleAMathews:
There's a number of options we need to evaluate how to support e.g. refreshing on mount/window focus, etc.

`queryCollectionOptions` should take the same options & behave the same as Query.

A follow-up to #195

KyleAMathews:
We need more help from people deep in doing a Query -> DB conversion

jmarbutt:
I would be interested on this one.

KyleAMathews:
Good deep research — https://chatgpt.com/share/688b6070-a930-800c-875e-9eadbf635b6c

Going to turn this into a set of detailed issues to start knocking off

KyleAMathews:
Completed through #1653, #1665, and #1683. Query Collections now document and support the approved flat observer-option surface while preserving QueryClient defaults and keeping adapter-owned semantics explicit. Full QueryObserver option parity was considered and rejected where it would conflict with normalized-row synchronization. The final outcome is recorded in RFC #1643.

PR #1653, opened 2026-07-07
docs(query-db-collection): document query option compatibility

KyleAMathews:
Documents the current TanStack Query option compatibility surface for Query Collections. Users can now see which options are forwarded, how `QueryClient.defaultOptions` interact with omitted options, which fields are adapter-owned, and which common Query options are not currently exposed.

## Approach

This keeps the change docs-only and focused in the existing Query Collection guide:

- Adds `gcTime` to the documented forwarded Query options.
- Clarifies that forwarded options other than adapter-extended `meta` are only passed when explicitly defined, so `QueryClient.defaultOptions` can still apply.
- Separates adapter-owned fields such as `queryKey`, `queryFn`, `select`, `queryClient`, `syncMode`, `getKey`, and mutation handlers from ordinary Query option pass-through.
- Lists common TanStack Query options that are not currently exposed through `queryCollectionOptions`.

## Key invariants

- This PR describes current behavior only.
- It does not change runtime behavior, TypeScript types, tests, or generated reference docs.
- It does not promise future support for unsupported Query options.

## Non-goals

- No option pass-through implementation.
- No generated reference docs update.
- No changes to Query Collection APIs.

## Verification

```bash
pnpm prettier --write docs/collections/query-collection.md
pnpm prettier --check docs/collections/query-collection.md
```

## Files changed

- `.changeset/document-query-option-compatibility.md` — records the user-facing Query Collection docs update.
- `docs/collections/query-collection.md` — documents the current Query option compatibility categories.

## Related

- Addresses #183
- Addresses #346
- Related RFC: #1643

## Summary by CodeRabbit

* **Documentation**
  * Updated “Query Options” guidance for collection-based queries to clarify that `queryCollectionOptions` is not a full `QueryObserverOptions` pass-through.
  * Reworked the explanation of which options are forwarded to the query engine versus handled by the collection layer.
  * Noted additional common options that are currently not exposed (including several initialization, refetch, error-handling, and `structuralSharing` configurations).


kevin-dp (approved):
I don't think we need a changeset since this PR only touches docs.

PR #1656, opened 2026-07-08
Fix/live query record drop on subset unmount

jdday:
## 🎯 Changes

When a row was first inserted by subset query A, `write({ type: 'insert' })` cleared the row's pending metadata. `setPersistedOwners` was called *before* `write`, so the ownership it set was immediately overwritten by the insert's metadata-delete.

When a later live query called `getHydratedOwnedRowsForQueryBaseline`, it read the stale persisted state and overwrote the in-memory `rowToQueries` map, dropping query A's hash from the owner set. When that query was GC'd, it found zero remaining owners and deleted the row — removing it from unrelated list views.

## ✅ Checklist

- [ x ] I have tested this code locally with `pnpm test`.

## 🚀 Release Impact

## Notes

Currently facing the below issue with react-db and I believe the below re-ordering of lines fixes it for myself but need to understand the impact elsewhere or if there is a better place to add this fix.

- Live query A contains a list of rows
- Live query B contains a subset of those rows with a different where clause
- When Live Query B unmounts then the rows are removed from the collection completely even with Live Query A still mounted and requires those rows., 

Any new live queries seem to take ownership of existing rows when mounting. Meaning that when the new live query unmounts it causes that record to be removed from the collection even if other useLiveQueries still should have that row existing.

I am unsure if this is a similar impact across the other vue/svelte db packages I am currently only using react-db so any advise if this change would negatively impact other areas would be helpful. 

One thing noticed - my team have observed that the row will re-appear in the correct live query if we force a re-render during local development

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected record syncing to ensure existing rows receive the appropriate update behavior rather than being skipped.
  - Fixed ownership tracking to re-apply ownership after writes, improving consistency and preventing unintended row deletion during sibling live query garbage collection.
- **Tests**
  - Added coverage confirming that unmounting a narrowly filtered `useLiveQuery` does not remove records from another still-mounted, unfiltered list `useLiveQuery`.

### DIFF
--- a/.changeset/evil-maps-shine.md
+++ b/.changeset/evil-maps-shine.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/query-db-collection': minor
+---
+
+fix: move setPersistedOwners after write to prevent row deletion when sibling live query GCs
--- a/packages/query-db-collection/src/query.ts
+++ b/packages/query-db-collection/src/query.ts
@@ -1318,14 +1318,19 @@ export function queryCollectionOptions(
       })
 
       newItemsMap.forEach((newItem, key) => {
+        addRow(key, hashedQueryKey)
+        const existingItem = currentSyncedItems.get(key)
+        if (!existingItem) {
+          write({ type: `insert`, value: newItem })
+        } else if (!previouslyOwnedRows.has(key)) {
+          write({ type: `update`, value: newItem })
+        }
+        // Read owners AFTER write: insert clears pending metadata, so reading here
+        // ensures we see the post-insert baseline and correctly re-apply ownership.
         const owners = getPersistedOwners(key)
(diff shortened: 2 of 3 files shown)


KyleAMathews:
Closing as superseded after RFC #1657 Phase 0 review. The immediate persisted-metadata clobber described here was addressed more fundamentally by #1626, and #1664 has since refactored the ownership paths touched by this branch. The added React test uses mockSyncCollectionOptions and does not exercise query-db ownership. Broader ownership, provider-lifetime, and destructive-diff semantics should be handled by RFC #1657 PR-5's coverage registry rather than landing this partial ownership expansion. A genuine query-db-level red/green regression test can still be harvested separately if it exposes remaining behavior. Thank you for the investigation and contribution.

PR #1660, opened 2026-07-08
fix(db): preserve explicit gcTime of 0 on live query collections

spokodev:
The live query config builder built its collection options with `gcTime: this.config.gcTime || 5000`. A `gcTime` of `0` disables garbage collection (the core lifecycle treats `gcTime <= 0` as "never GC"), but `0 || 5000` evaluates to `5000`, so an explicit `gcTime: 0` on a live query collection is silently replaced by the 5s default and the collection is garbage collected 5s after its last subscriber unsubscribes.

Use `??` so only `undefined` falls back to the default; an explicit `0` is forwarded and GC is disabled as documented.

Repro: `liveQueryCollectionOptions({ query, gcTime: 0 }).gcTime` returns `5000` instead of `0`. Added a regression test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed live query settings so an explicitly set `gcTime: 0` is preserved instead of being replaced with the default timeout.
  * Live query behavior now respects disabled garbage collection as intended.
  * Added coverage to verify that `gcTime: 0` continues to work correctly.


KyleAMathews (approved):
thanks!

Issue #1657, opened 2026-07-08
RFC: Fixing the loadSubset / pagination core (no new features)


KyleAMathews:
# Current disposition — 2026-09-14

**The implementation and four final follow-ups have merged. This RFC remains open for the post-merge correction in #1822 and final combined verification.**

The implemented design is exact-request reuse, applied settlement, separate demand/acquisition lifetimes, and private rebuilding behind a coherent public result. The earlier coverage-registry, general subset-algebra, and outcome-envelope plans are withdrawn—not unfinished work to restore.

## Merged work

| Work | Status |
| --- | --- |
| Consolidated implementation #1797 and review follow-up #1800 | Merged |
| Collection/adapter lifecycle hardening #1785 | Merged; Electric E2E passed on its reviewed branch |
| TrailBase stream lifetime fixes #1805 | Merged |
| Ignored pagination-callback removal and server-page guidance #1806 | Merged |
| Independent optimistic composition/attribution fixes #1807 | Merged; queue timing unchanged |
| Oracle audit, guide, DBSP laws, and exposed boundary fixes #1816 | Merged; bounded oracle initiative #1808 is closed |

## Post-merge correction: #1822

Review against `origin/main 150bde99ceb45b764f1b125753d7d31ef0cc733f` reproduced three remaining boundaries. Existing oracles were expanded before the fixes:

- A delayed outbox scan can return a transaction after successful durable deletion. Reread scans that cross an acknowledgment; unfinished peers must still run. This is not the separate failed-deletion policy in #1659.
- Top-K equality must retain sparse-array length and RegExp source/flags/position; ordinary replacements must also work without a global File constructor.
- An accepted unsynced insert hidden by a later optimistic delete must return if that delete rolls back, including when truncate happened while it was hidden. Preserve the accepted whole-row snapshot, not a merge with later synced fields.

The fixes are in #1822 at `6385f2968`, not yet merged. The extended families were red before repair and green afterward; the original retention seed/path also passes unchanged.

Local combined verification on that branch, after rebuilding IVM/core: core **6,155**, IVM **549**, offline **140**, Query DB **454**, TrailBase **67**, Electric **820**, PowerSync **136**, persistence core **176** reported passing tests. Standalone core/IVM/offline types and enabled package typechecks pass. PowerSync includes native SQLite/SDK execution. These are package gates, **not a fresh Electric service E2E run or mobile-device certification**. #1822's CI is still a separate gate.

## Remaining boundaries and owners

[...]

Issue #1017, opened 2025-12-14
Synced data does not appear in a derived (live query) collection while there is a pending optimistic mutation.


thruflo:
- [x] I've validated the bug against the latest version of DB packages

## Describe the bug

Synced data from the source collection does not appear in a derived (live query) collection while there is a pending optimistic mutation on that derived collection.

When using `createOptimisticAction` to insert into a `createLiveQueryCollection`, any new items that sync into the source collection while the mutation is pending are invisible in the derived collection. The items are processed (the pipeline runs), but they don't appear in the collection's output.

This affects all derived collections, regardless of query complexity (simple passthrough, orderBy, fn.select, multi-stage pipelines).

## To Reproduce

1. Create a source collection with sync
2. Create a derived collection via `createLiveQueryCollection` that queries the source
3. Use `createOptimisticAction` to insert an item into the derived collection
4. While the mutation is pending (mutationFn hasn't resolved), sync a NEW item into the source collection
5. Observe that the synced item is invisible in the derived collection

### Minimal Reproduction Code

```typescript
import {
  createCollection,
  createLiveQueryCollection,
  createOptimisticAction,
} from '@tanstack/db'

interface Item {
  id: string
  value: string
}

// 1. Create source collection with controllable sync
let begin, write, commit, markReady
const source = createCollection<Item>({
  id: 'source',
  getKey: (item) => item.id,
  startSync: true,
  sync: {
    sync: (params) => {
      begin = params.begin
      write = params.write
      commit = params.commit
      markReady = params.markReady
    },
  },
})

// 2. Create derived collection (simple passthrough)
const derived = createLiveQueryCollection<Item>({
  query: (q) => q.from({ item: source }),
  getKey: (item) => item.id,
  startSync: true,
})

// Mark source ready
markReady()

// 3. Create optimistic action
let resolveOptimistic: () => void
const optimisticInsert = createOptimisticAction<Item>({
  onMutate: (item) => {
    derived.insert(item)
  },
  mutationFn: async () => {
    // Keep mutation pending until we explicitly resolve
    await new Promise<void>((resolve) => {
      resolveOptimistic = resolve
    })
  },
})

// 4. Insert optimistic item (mutation will be PENDING)
optimisticInsert({ id: 'optimistic-1', value: 'optimistic' })

console.log('After optimistic insert:')
console.log('  derived.size:', derived.size)           // 1 ✓
console.log('  has optimistic-1:', derived.has('optimistic-1'))  // true ✓

[...]

PR #1662, opened 2026-07-10
Failing repro: loadSubset resolves & live query marks ready while synced rows are parked behind a persisting transaction

TomasGonzalez:
## What this is

A **failing test** (no fix) reproducing an on-demand `loadSubset` readiness bug in `@tanstack/db`. Opened as **draft** so the repro runs in CI — happy to turn it into a fix PR once maintainers weigh in on the approach.

- Adds `packages/db/tests/query/load-subset-persisting-deferral.test.ts`
- Control test passes; the persisting-transaction case fails (`received []`). Verified failing on current `main` (`expected [] to deeply equal [ 'r1', 'r2' ]`).

## The bug

In on-demand sync mode, `await loadSubset(...)`, `collection.status === 'ready'`, and `toArrayWhenReady()` all resolve **before the rows `loadSubset` fetched are readable**, whenever any user transaction on the collection is in state `persisting` (an in-flight optimistic mutation).

`commit()` is fire-and-forget: `commitPendingTransactions()` in `packages/db/src/collection/state.ts` skips applying committed synced transactions while `hasPersistingTransaction` (guard: `!hasPersistingTransaction || hasTruncateSync || hasImmediateSync`), leaving the rows parked in `pendingSyncedTransactions`. Nothing downstream is signalled, so `loadSubset` resolves, the live query calls `markReady()`, and the read returns `[]`. The rows become readable only when the mutation settles.

The test's `loadSubset` begins/writes/commits synchronously then resolves — exactly what `@tanstack/electric-db-collection` does inside its subscribe handler on a `subset-end` control message before `requestSnapshot` resolves. `@electric-sql/client` is not involved and not at fault (it delivers to subscribers before resolving since 1.5.20).

## Why it matters

Downstream (offline-first apps using `@tanstack/offline-transactions`) this surfaces as silently empty/partial reads whenever a background write is in flight — e.g. duplicating a template's child rows into a new record returns nothing. Settle-window/timer workarounds cannot recover the rows, because a parked commit emits no change events until the transaction settles.

## Possible fixes (happy to implement)

- `loadSubset` shouldn't resolve until its synced transaction(s) are applied (e.g. `commit()` returns a promise / a `whenApplied()` signal). This also fixes the `DeduplicatedLoadSubset` coverage-poisoning, since tracking runs on resolution.
- Or commit subset-snapshot data with `begin({ immediate: true })` (already bypasses the guard; used today for resume-metadata resets). Authoritative server rows have nothing to reconcile against.
- Or make live-query readiness account for committed-but-parked synced transactions.

## Related

- #1017 — same underlying deferral (fix attempt #1167 closed unmerged); this adds the readiness/`loadSubset` composite on top.
- #1657 — loadSubset / pagination RFC; this failure mode isn't listed there yet.

### DIFF
--- a/packages/db/tests/query/load-subset-persisting-deferral.test.ts
+++ b/packages/db/tests/query/load-subset-persisting-deferral.test.ts
@@ -1,0 +1,118 @@
+import { describe, expect, it } from 'vitest'
+import { createCollection } from '../../src/collection/index.js'
+import { createLiveQueryCollection, eq } from '../../src/query/index.js'
+import { createTransaction } from '../../src/transactions'
+
+// Reproduction for the on-demand `loadSubset` readiness bug.
+//
+// When any user transaction on a collection is in state `persisting`, a synced
+// transaction committed by `loadSubset` is parked by `commitPendingTransactions`
+// (src/collection/state.ts skips applying while `hasPersistingTransaction`), yet
+// the `loadSubset` promise still resolves and the live query still marks ready.
+// So `toArrayWhenReady()` resolves with an empty/stale result for rows that were
+// actually fetched and committed.
+//
+// The `loadSubset` below begins/writes/commits synchronously and then resolves —
+// exactly like @tanstack/electric-db-collection's subscribe handler does on a
+// `subset-end` control message, before its `requestSnapshot` promise resolves.
+
+interface Row {
+  id: string
+  projectId: string
+  name: string
+}
+
+const row1: Row = { id: `r1`, projectId: `p1`, name: `a` }
+const row2: Row = { id: `r2`, projectId: `p1`, name: `b` }
+
+function createOnDemandCollection() {
+  let loadSubsetCalls = 0
+  const collection = createCollection<Row>({
+    id: `loadsubset-persisting-repro`,
+    getKey: (row) => row.id,
+    syncMode: `on-demand`,
+    sync: {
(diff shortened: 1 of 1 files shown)


KyleAMathews:
Thanks for the excellent reproduction — this is exactly the kind of report that's easy to act on.

We verified it against `main`: the control case passes and the persisting case fails precisely as you described (`expected [] to deeply equal ['r1','r2']`), and the mechanism checks out — the `commitPendingTransactions` guard silently parks committed synced transactions with no downstream signal, `commit()` returns void, and `DeduplicatedLoadSubset` records the predicate as covered on resolution, so retries can't recover. We also confirmed your `begin({ immediate: true })` direction makes the repro pass, though with the caveat that an immediate commit flushes *all* parked sync transactions.

This is now tracked in the loadSubset/pagination RFC (#1657) as evidence item E13 with a dedicated work item (§3.8): **loadSubset resolution must mean the rows are actually applied** — your test is the red test for it. A maintainer will be picking up the RFC work soon, and we'd rather land this as a proper fix at the contract level within that effort than as a one-off patch — that's the whole point of the cluster analysis. We'll keep this PR open for its repro until then.

KyleAMathews:
Resolved by #1769. This reproduction became the red test for applied settlement: subset loads now stay on the normal causal queue and settle only after their sync commits are visible. The merged oracle retains the parked-transaction, readiness, cancellation, and coverage-publication cases.

PR #1669, opened 2026-07-13
feat(db): ordered snapshot / layout-revision contract (RFC #1623 phase 4)

kevin-dp:
Phase 4 of the live-query platform RFC (#1623): the **ordered snapshot / layout contract**. Stacked on #1642 (observer migration) — review/merge that first; this PR's base is `refactor/live-query-observer` so the diff is Phase 4 only.

## Problem

An `orderBy` live query that reorders its rows *without changing any projected row value* (an "order-only move") is swallowed by the collection's value-diff: `.values()`/`.entries()` re-sort internally, but no change event fires, so `useLiveQuery` keeps rendering the stale order. This is the last universal expected-fail in the cross-adapter conformance suite (tracked as #1601).

## Approach

The RFC is explicit that this should be **"an explicit layout-revision requirement, not a hidden forced-update path"**. So instead of forging a row `update`:

- The live-query flush captures the retracted side of each change and, after `commit()`, detects an order-only move (projected value deep-equal, `orderByIndex` moved) and publishes a **first-class empty layout-change notification** via a new `CollectionChangesManager.emitLayoutChangeEvent()`. It reuses the existing empty-batch delivery already used for the ready signal — subscribers re-read, nothing is faked.
- The shared observer snapshot gains **`layoutRevision`**, which increments on any visible membership, ordering, or order-only-move change. This is the canonical contract the RFC wants for future fine-grained materializers; adapters currently pick up the reorder through their existing wholesale re-read.

## Result

`order-only-move` is removed from `UNIVERSAL_EXPECTED_FAIL` and now passes on **all five adapters** (React, Vue, Svelte, Solid, Angular) — 26/26 conformance each. Full `@tanstack/db` suite green (2464 tests), all five adapter suites green.

## Coverage

- `packages/db/tests/live-query-order-only-move.test.ts` — core mechanism: republish + `layoutRevision` bump on an order-only move; no bump when order is unchanged (no spurious notification); bump on membership change.
- Conformance `order-only-move` scenario now a real pass across all adapters.

## Relationship to #1601

@v-anton's #1601 fixes the same bug via a forced row `update`. This PR takes the RFC-sanctioned layout-revision approach instead (a distinct, first-class notification + `layoutRevision`), so it supersedes rather than duplicates that path. Happy to coordinate on which lands — flagging for maintainer decision.

## Summary by CodeRabbit

* **New Features**
  * Added a framework-agnostic live-query observer with granular or whole-snapshot updates.
  * Live-query snapshots now expose data, status, ordering, and revision information.
* **Bug Fixes**
  * Ordered live queries now republish when items move without changing their values.
  * Improved handling of nested includes, filtering, truncation, refetching, and stale keys.
* **Documentation**
  * Updated release notes to clarify live-query synchronization behavior.


KyleAMathews (changes requested):
### Code review

Found 2 issues, both reproduced with failing regression tests against `d7628551f219cab27ddd542bffda40c8ee7f5aad`:

1. A graph flush containing both an ordinary projected-value update and an order-only move publishes twice. `commit()` synchronously emits the ordinary row batch, then `emitLayoutChangeEvent()` emits a second empty batch even though listeners already observe the final values and order. My regression expected one observer callback and received two. Please coalesce layout dirtiness into the commit publication (or otherwise suppress the separate layout event when that commit already publishes), and cover this with an exact-one-callback mixed-batch test.

https://github.com/TanStack/db/blob/d7628551f219cab27ddd542bffda40c8ee7f5aad/packages/db/src/query/live/collection-config-builder.ts#L815-L830

2. The layout fix does not cover ordered child collections produced by `includes`. On retract-then-insert, the insertion side replaces `existing.value` but leaves the retracted `orderByIndex`; later the child collection commits without any layout-only signal. I reproduced this with a child projection that omits its sort field: after moving `c1` behind `c2`, the child collection remained ordered `[c1, c2]` instead of `[c2, c1]`. Please preserve the insertion-side order metadata, publish child layout-only changes through the same atomic mechanism, and add an ordered-includes regression.

https://github.com/TanStack/db/blob/d7628551f219cab27ddd542bffda40c8ee7f5aad/packages/db/src/query/live/collection-config-builder.ts#L886-L918

https://github.com/TanStack/db/blob/d7628551f219cab27ddd542bffda40c8ee7f5aad/packages/db/src/query/live/collection-config-builder.ts#L1977-L2005

Generated with Claude Code

KyleAMathews:
### Reproductions for the requested changes

I verified both findings against head `d7628551f219cab27ddd542bffda40c8ee7f5aad` with focused tests. Here are the essential regressions to add.

#### 1. Mixed value update + order-only move must publish exactly once

Using the existing ordered query whose projection omits `age`:

```ts
it(`publishes a mixed value update and order-only move exactly once`, async () => {
  const source = makeSource()
  const lq = await makeOrderedByAge(source)
  const observer = createLiveQueryObserver<
    { id: string; name: string },
    string
  >(lq as any)

  let notifications = 0
  observer.subscribe(() => notifications++)
  notifications = 0 // exclude subscribeChanges' initial-state publication

  source.utils.begin()
  source.utils.write({
    type: `update`,
    value: { id: `1`, name: `Alicia`, age: 30 },
  })
  source.utils.write({
    type: `update`,
    value: { id: `2`, name: `Bob`, age: 99 },
  })
  source.utils.commit()

  const after = observer.getSnapshot()
  expect((after.data as Array<Person>).map(({ id, name }) => [id, name])).toEqual([
    [`1`, `Alicia`],
    [`3`, `Carol`],
    [`2`, `Bob`],
  ])
  expect(notifications).toBe(1)
})
```

Result on the PR head:

```text
AssertionError: expected 2 to be 1
```

The first callback comes from `commit()`'s ordinary row update. The collection already has its final value and ordering at that point. The post-commit `emitLayoutChangeEvent()` then produces a redundant second callback. Layout dirtiness needs to participate in the same publication as row changes; a layout-only flush should publish only when no ordinary publication represents that flush.

Relevant code: https://github.com/TanStack/db/blob/d7628551f219cab27ddd542bffda40c8ee7f5aad/packages/db/src/query/live/collection-config-builder.ts#L815-L830

#### 2. Ordered included child must consume the new order metadata and publish its move

```ts
it(`publishes an ordered included child move exactly once`, async () => {
  const parents = createCollection(
    mockSyncCollectionOptions<{ id: string }>({
      id: `parents`,
      getKey: ({ id }) => id,
      initialData: [{ id: `p1` }],
    }),
  )
  const children = createCollection(
    mockSyncCollectionOptions<{
      id: string
      parentId: string
      name: string
      position: number
    }>({
      id: `children`,
      getKey: ({ id }) => id,
      initialData: [
        { id: `c1`, parentId: `p1`, name: `One`, position: 1 },
        { id: `c2`, parentId: `p1`, name: `Two`, position: 2 },
      ],
    }),
  )
  const lq = createLiveQueryCollection((q) =>
    q.from({ parent: parents }).select(({ parent }) => ({
      id: parent.id,
      children: q
        .from({ child: children })
        .where(({ child }) => eq(child.parentId, parent.id))
        .orderBy(({ child }) => child.position)
        .select(({ child }) => ({ id: child.id, name: child.name })),
    })),
  )
  await lq.preload()

  const childCollection = lq.get(`p1`)!.children
  let notifications = 0
  const subscription = childCollection.subscribeChanges(
    () => notifications++,
    { includeInitialState: false },
  )

  children.utils.begin()
  children.utils.write({
    type: `update`,
    value: { id: `c1`, parentId: `p1`, name: `One`, position: 3 },
  })
  children.utils.commit()

  expect([...childCollection.values()].map(({ id }) => id)).toEqual([
    `c2`,
    `c1`,
  ])
  expect(notifications).toBe(1)
  subscription.unsubscribe()
})
```

Result on the PR head:

```text
AssertionError: expected [ 'c1', 'c2' ] to deeply equal [ 'c2', 'c1' ]
```

There are two gaps in this path:

1. On retract-then-insert, `existing.value` is replaced by the insertion-side value, but `existing.orderByIndex` is not replaced by the insertion-side index.
2. The child collection calls `commit()` but has no equivalent layout-only publication when the projected child value is unchanged.

Relevant code:

- https://github.com/TanStack/db/blob/d7628551f219cab27ddd542bffda40c8ee7f5aad/packages/db/src/query/live/collection-config-builder.ts#L886-L918
- https://github.com/TanStack/db/blob/d7628551f219cab27ddd542bffda40c8ee7f5aad/packages/db/src/query/live/collection-config-builder.ts#L1977-L2005

I also tested the stable-rank case (`age: 20 → 21`, still first). After excluding the expected initial subscription publication, it already produces zero callback delta, preserves snapshot identity, and keeps `layoutRevision` stable. No change is needed for that case.

kevin-dp:
Thanks for the thorough review, Kyle — both findings were spot on. Fixed in c9ec751d, with the two regressions you specified added first (43f1a7d6) so the failure is on record.

**1. Mixed value update + order-only move now publishes exactly once.** Replaced `hasOrderOnlyMove` with `needsLayoutOnlyPublication`, which fires the layout event only when the commit published nothing else. Any real insert/delete/value-changed update already notifies subscribers (who re-read the re-sorted collection), so the separate layout event is skipped in that case. Covered by the exact-one-callback mixed-batch test.

**2. Ordered `includes` children now consume the insertion-side order metadata and publish their move.** Two gaps, both closed:
- The child accumulate replaced `existing.value` on the insert side but left the retracted `orderByIndex`. It now updates `orderByIndex` on insert and captures `previousValue`/`previousOrderByIndex` on retract (both the single-level output and the nested-includes buffer accumulate).
- The child flush committed without a layout-only publication when the projected child value was unchanged. It now runs the same `needsLayoutOnlyPublication` check after the child `commit()` and publishes through a shared `emitLayoutChange` helper.

Covered by the ordered-includes regression.

**Depth.** Since `flushIncludesState` is recursive and the child layout-emit runs at every level, I added a two-level nested-includes regression (`org → teams → members`, moving a grandchild) to guard that this holds beyond one level — verified it goes red if the child-flush publication is removed. An independent pass also confirmed the exactly-once contract and re-sort at 1/2/3 levels, plus the insert/delete-combined-with-move and multiple-moves-in-one-commit cases.

**Stable-rank case** (sort field changes but position doesn't, value unchanged) needs no change, as you noted — it produces zero layout events; there's a test asserting that too.

Full `@tanstack/db` suite and all five adapter conformance suites are green.

KyleAMathews (changes requested):
## [P1] Tie the layout revision to the sync transaction that changes the layout

This is reproducible on the current head, `df9403a`.

`markLayoutChange()` currently increments the collection-wide `layoutRevision` and sets `pendingLayoutChange` before the live-query result collection calls `commit()`. A normal sync commit is not always applied at that point: `commitPendingTransactions()` parks committed sync transactions while a user transaction on the same collection is `persisting`.

That lets the layout clock describe state that is not visible yet.

### Deterministic reproduction

1. Create a live query ordered by `age`, but project only `{ id, name }`.
2. Read a detached observer snapshot. Its initial order is `[2, 1, 3]`.
3. Start a non-optimistic mutation on the result collection and keep its `onUpdate` handler unresolved. The user transaction is now `persisting`.
4. Update source row `2` so its age moves it to the end. The projected row value remains deep-equal.
5. The query pipeline calls `markLayoutChange()`, advancing `layoutRevision`, then commits its result-collection sync transaction.
6. The sync transaction is parked behind the persisting mutation, so the visible order is still `[2, 1, 3]`.
7. A detached `getSnapshot()` sees the new layout revision and caches the old entries against it.
8. Resolve the mutation. The parked sync transaction applies and the collection order becomes `[1, 3, 2]`.
9. Because the projected values are deep-equal, no row event or state-revision increment occurs. The layout revision also does not advance again.
10. The next detached snapshot sees neither revision change and remains stuck on `[2, 1, 3]`.

I added that regression locally. Before the fix it fails at the final assertion:

```text
expected [ '2', '1', '3' ] to deeply equal [ '1', '3', '2' ]
```

There is a related attached-observer hazard: while the sync transaction is parked, an unrelated forced publication can consume the collection-wide `pendingLayoutChange` flag. The eventual order-only commit then has neither row events nor a layout marker, so it cannot send the corrective publication.

### Requested fix

Please store layout dirtiness on the exact pending sync transaction that produced it, rather than on the collection:

```ts
interface PendingSyncedTransaction<...> {
  // existing fields
  layoutChanged: boolean
}
```

`_markLayoutChange()` should mark the active, uncommitted sync transaction. When `commitPendingTransactions()` actually applies committed transactions, it should OR their `layoutChanged` flags and advance/publish the collection layout revision after applying state and immediately before emitting the final batch.

This preserves the required properties:

- The revision advances only when the matching reordered state becomes visible.
- The layout signal and row-event batch publish atomically.
- Multiple transactions applied in one flush coalesce into one layout publication.
- Unrelated optimistic recomputations cannot steal the marker.
- Cleanup naturally discards the marker with the pending transaction.

I verified this design locally: the new regression turns green, all 10 order-only layout tests pass, and the broader collection transaction suites pass (149 tests total, with no type errors). ESLint, Prettier, and `git diff --check` also pass.

The mixed membership/order batching and nested ordered-`includes` fixes on this head look correct. This transaction-boundary race is the remaining blocker.

KyleAMathews (approved):
Verified the transaction-scoped layout revision fix on 619c3a37. The parked detached-observer regression now passes, the layout revision remains unchanged until the sync transaction is applied, and the broader collection/order-only suites pass (149 tests, no type errors). The implementation also removes the collection-wide pending marker, so unrelated publications cannot consume it. Ready from my review.

KyleAMathews (approved):
Re-reviewed current head c27d5721 after the main merge. The follow-up cleanly promotes the two includes-oracle cases to positive regressions, reuses the existing active-sync-transaction guard, and corrects the observer changeset wording. The equivalent merged tree passes the full @tanstack/db suite (2,534 tests, 5 skipped, no type errors). Ready.

tannerlinsley:
@KyleAMathews Thanks for re-reviewing. #1669 is green and mergeable now—please merge when ready: https://github.com/TanStack/db/pull/1669

#1675 is clean and stacked behind it.

PR #1671, opened 2026-07-13
docs(query-db-collection): document scoped collection factories

KyleAMathews:
## Summary

Document how to create Query Collections for tenant, project, account, and route-level business scopes. The new guidance helps users preserve collection identity and lifecycle correctly without confusing business scope with relational subset loading.

## Approach

- Add a typed project-scoped factory that accepts both a runtime `QueryClient` and a scope parameter.
- Include the scope in the flat top-level `queryKey` and `queryFn`.
- Show stable memoization keyed by both `QueryClient` and scope.
- Explain eviction and `collection.cleanup()` for unbounded, long-lived scope caches.
- Keep relational predicates, ordering, limits, and offsets in `LoadSubsetOptions` rather than creating collections per subset.
- Link back to the existing runtime `QueryClient` and predicate push-down guidance instead of duplicating it.

## Key invariants

- A collection is reused for the same `QueryClient` and business scope.
- Distinct server resources use distinct business-scoped collections.
- Relational subsets do not become collection identity and do not cause collection recreation.
- Examples use the current flat `queryCollectionOptions` API.

## Non-goals

- No runtime or type changes.
- No `.bind(...)` API or nested query configuration.
- No recommendation to recreate collections per render or relational subset.

## Trade-offs

The example uses explicit `WeakMap<QueryClient, Map<scope, Collection>>` memoization. This is more verbose than an implicit binding API, but keeps dependencies and lifecycle ownership visible while requiring no new runtime abstraction.

## Verification

```bash
pnpm prettier docs/collections/query-collection.md --check
pnpm exec tsx scripts/verify-links.ts
git diff origin/main --check
```

All commands pass. The repository's `pnpm test:docs` wrapper cannot directly execute `scripts/verify-links.ts` under the current Node environment, so the same verifier was run through the installed `tsx` runner.

## Files changed

- `docs/collections/query-collection.md` — adds business-scoped factory, memoization, lifecycle, and relational-subset guidance.
- `.changeset/document-scoped-query-factories.md` — records the documentation update for `@tanstack/query-db-collection`.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for creating business-scoped query collections.
  * Clarified how scope identifiers should affect collection queries.
  * Added memoization and lifecycle recommendations for reusing scoped collections.
  * Explained the difference between business-scoped collections and relational subsets.
  * Updated examples to discourage creating separate collections for every filter, sort, or limit.

### DIFF
--- a/.changeset/document-scoped-query-factories.md
+++ b/.changeset/document-scoped-query-factories.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/query-db-collection': patch
+---
+
+Document business-scoped Query Collection factories, including stable memoization and lifecycle guidance, while distinguishing business scope from relational subset loading.
--- a/.changeset/document-scoped-query-factories.md
+++ b/.changeset/document-scoped-query-factories.md
@@ -1,5 +1,0 @@
----
-'@tanstack/query-db-collection': patch
----
-
-Document business-scoped Query Collection factories, including stable memoization and lifecycle guidance, while distinguishing business scope from relational subset loading.
--- a/docs/collections/query-collection.md
+++ b/docs/collections/query-collection.md
@@ -112,25 +112,79 @@ export function getTodosCollection(
 
 Avoid calling `createCollection(todoCollectionOptions(queryClient))` independently during render or in each consumer. Share the stable collection instance for the lifetime of that `QueryClient` scope.
 
-The same pattern works for scoped or parameterized collections. Pass route params, a tenant ID, or filters into the factory alongside the `QueryClient`:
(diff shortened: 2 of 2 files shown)


.changeset/document-scoped-query-factories.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tanstack/query-db-collection': patch
```
kevin-dp: No need for a changeset on a docs only PR

kevin-dp (approved):

PR #1672, opened 2026-07-13
fix(query-db): clean up empty ownership sets

KyleAMathews:
## Summary

Cleans up Query DB ownership bookkeeping so empty row/query ownership sets are not retained, while preserving the semantic distinction between an unresolved query and a resolved query with an empty result.

This keeps ownership state bounded across subset unloads, cache expiry, persisted-row hydration, and empty revalidation without deleting rows that are still shared by another query.

### Root cause

`rowToQueries` and `queryToRows` used empty `Set`s both as stale bookkeeping and as a marker that a query's ownership had been resolved. Removing those empty entries directly would lose the resolved-empty baseline and could cause hydration or revalidation to infer ownership again incorrectly.

### Approach

- Delete reverse and forward ownership-map entries as soon as their sets become empty.
- Track authoritative query ownership separately in `resolvedOwnershipQueries`, including valid empty results.
- Clear that marker when retained ownership expires or the query/collection is cleaned up.
- Add test-only ownership inspection and lifecycle coverage for overlapping subsets, cache GC, persisted hydration, empty revalidation, and retained ownership expiry.

### Invariants

- Ownership maps contain only non-empty sets.
- A row is removed only after its final query owner is removed.
- A resolved empty result remains authoritative until its query lifecycle ends.
- Persisted retention markers and in-memory ownership markers expire together.

### Non-goals and tradeoffs

- No public API or persistence-format changes.
- The separate resolved-query set adds a small amount of state, but avoids overloading empty map entries and is explicitly bounded by existing cleanup paths.
- Internal ownership-map inspection is exposed only in test environments.

## Verification

- `pnpm --filter @tanstack/query-db-collection test -- query.test.ts`
  - 232 tests passed and type assertions reported no errors in the target suite.
  - Command exits non-zero because Vitest also type-checks `db-collection-e2e` sources that cannot resolve `@tanstack/electric-db-collection` in this worktree (3 unrelated source errors).

## Files

- `packages/query-db-collection/src/query.ts` — compact ownership maps and preserve resolved-empty semantics with bounded lifecycle tracking.
- `packages/query-db-collection/tests/query.test.ts` — characterize and verify ownership cleanup, hydration, shared rows, cache expiry, and retained-query cleanup.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup and synchronization of query↔row ownership when data is unloaded, expires, revalidates, or hydrates from persisted metadata.
  * Empty query results now establish an authoritative ownership baseline without incorrectly deleting shared/retained rows.
  * Retained-row ownership is cleared reliably after retention ends and when an explicit collection cleanup runs.

* **Tests**
  * Added deeper lifecycle coverage for ownership mapping across overlapping subsets, cache expiry, persisted hydration, empty-result revalidation, and retention cleanup.


kevin-dp (changes requested):
The lifecycle fix (deleting stale entries at the right points) looks right to me. I'd like to suggest a simplification before this lands, though: I don't think we need `resolvedOwnershipQueries` as a separate structure.

The marker is only ever consulted as `knownRows || resolvedOwnershipQueries.has(k)`,
i.e. a non-empty `queryToRows` entry already counts as authoritative on its own. The
new set only carries information when the row set is empty — which means we can encode "resolved" as entry **presence** in `queryToRows` itself, with an empty `Set` allowed:

- present entry (possibly empty) = ownership resolved
- absent entry = unresolved

Concretely:

- At the resolution point, replace `resolvedOwnershipQueries.add(k)` with `queryToRows.set(k, queryToRows.get(k) ?? new Set())`.
- In `removeRowOwner`, keep the drained `queryToRows` entry instead of deleting it — a query whose rows are diffed away during revalidation is still resolved and now legitimately owns nothing. (The delete-when-empty stays correct for `rowToQueries`, since a row with no owners has no "resolved" meaning.)
- The two teardown paths (`cleanupPersistedPlaceholder` and query cleanup) then just delete the one entry instead of clearing two structures.

The read side reverts to the simpler pre-PR shape (`if (knownRows)`), and the compound `(knownRows || resolvedOwnershipQueries.has(k)) && Array.from(knownRows ?? [])` check goes away. More importantly, there's one source of truth: right now three separate mutation sites have to keep `resolvedOwnershipQueries` in sync with the map's lifecycle, and a missed site in a future change becomes a silent drift bug.

The trade-off is that the invariant becomes implicit — an accidental get-or-create on
`queryToRows` (like the existing `addRowOwner` pattern) would silently mark an unresolved query as resolved-empty. I think that's addressable by funneling writes through a couple of small helpers, documenting the presence-means-resolved invariant on the map declaration, and asserting it via the test-only inspection hook this PR already adds.

Memory-wise it's equivalent, so this is purely about keeping the bookkeeping in one place. WDYT?

kevin-dp (approved):

PR #1553, opened 2026-05-25
fix(react-db): let setWindow control limit/offset in useLiveInfiniteQuery

v-anton:
## 🎯 Changes

`useLiveInfiniteQuery` baked `.limit(pageSize + 1).offset(0)` into the initial query builder. This capped results even after `setWindow()` tried to expand the window for subsequent pages, making pagination beyond page 1 return no additional data.

The fix removes the hardcoded limit/offset from the initial query and lets `setWindow()` (called in the `useEffect` below) be the sole authority on the query window.

## ✅ Checklist

- [x] I have tested this code locally with `pnpm test:pr`.
- [x] I have followed the steps in the [Contributing guide](https://github.com/TanStack/db/blob/main/CONTRIBUTING.md).

## 🚀 Release Impact

- [x] This change affects published code, and I have generated a changeset.
- [ ] This change is docs/CI/dev-only (no release).

## Bug Fixes
* Fixed infinite query pagination to properly manage dynamic pagination windows across multiple pages without hardcoded initial limits, ensuring correct data retrieval for all pages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/db/pull/1553?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #1513, opened 2026-05-04
feat(vue-db): implement useLiveInfiniteQuery for vue 3

miguelrk:
## 🎯 Changes

This PR implements `useLiveInfiniteQuery` for Vue 3. This reference implementation followed is React (https://github.com/TanStack/db/pull/666) and the open PR implementing `useLiveInfiniteQuery` for Svelte 5 (https://github.com/TanStack/db/pull/1447).

## ✅ Checklist

- [x] I have tested this code locally with `pnpm test`.

## 🚀 Release Impact

- [x] This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).
- [ ] This change is docs/CI/dev-only (no release).

PR #1447, opened 2026-04-03
feat(svelte-db): implement useLiveInfiniteQuery for svelte 5

devcsrj:
## 🎯 Changes

This PR implements `useLiveInfiniteQuery` for Svelte 5. This reference implementation followed is React (https://github.com/TanStack/db/pull/666).

## ✅ Checklist

- [x] I have tested this code locally with `pnpm test`.

## 🚀 Release Impact

- [x] This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).
- [ ] This change is docs/CI/dev-only (no release).

PR #1675, opened 2026-07-15
feat(db): shared live-query window controller (RFC #1623 phase 5)

kevin-dp:
This implements Phase 5 of the live-query platform RFC (#1623): an internal, framework-neutral window controller now owns infinite-query pagination. React's `useLiveInfiniteQuery` is a thin binding over that controller, with no public API change.

Users keep the same forward-pagination behavior, while concurrent consumers, failed subset loads, cleanup/restart cycles, runtime page-size changes, and order-only updates now produce coherent results.

## Root cause

The React hook previously owned page state and mutated a live-query collection's physical window directly. That coupled pagination to React and left no shared owner when several consumers used the same collection.

The first extraction exposed several distinctions that the old hook did not need to model explicitly:

- a requested window is not committed until its async subset load succeeds;
- several controllers may hold leases on one physical query window;
- preload, subscribe, reset, and page expansion may overlap;
- layout-only changes still need to reach both observer and public collection subscribers;
- a released runtime window must not survive collection cleanup and recompilation.

Without those distinctions, overlapping operations could shrink a pending page request, failures could leave an inflated shared lease, observer updates could disappear while a load was pending, and a stale window could return after restart.

## Approach

- Add `createLiveQueryWindowController` to `@tanstack/db`. It owns page slicing, peek-ahead `hasNextPage`, loading/error state, reset, preload, subscription, and disposal.
- Coordinate physical windows per collection with leases. The largest active lease wins, failed requests roll back transactionally, and stale async completions cannot commit superseded state.
- Have `setWindow` return the real subset-load promise so page commits and errors follow the underlying load outcome.
- Preserve source and status notifications during pending pagination, including public empty change batches for order-only moves.
- Expose the query's compiled initial window and clear runtime window state on sync cleanup, so later consumers restart from the query definition.
- Rebuild `useLiveInfiniteQuery` around `useSyncExternalStore`, preserving loaded pages across page-shape changes and structurally equal dependency values.

## Key invariants

- The physical window covers the largest active controller lease.
- A page count changes only after its requested window succeeds.
- A failed or released request cannot inflate, shrink, or resurrect another consumer's window.
- `status` and its boolean flags always agree.
- Data and status updates remain observable while pagination is pending.
- The last released lease does not override the query's compiled window after cleanup.
- The React hook retains its existing public return type and forward-only semantics.

## Non-goals

- No bidirectional or backward pagination.
- No new `getNextPageParam` behavior; it remains accepted but unused.
- No stabilization of the controller as a public extension API. It remains internal and unstable while the RFC is in progress.
- No redesign of the publication queue; its target snapshots preserve reentrant subscribe/unsubscribe semantics.

## Trade-offs

The coordinator adds generation and lease bookkeeping, but keeps arbitration in one collection-scoped place instead of spreading race handling across framework hooks. React still lets `useSyncExternalStore` unsubscribe replaced controllers rather than disposing them during render, which avoids breaking abandoned or concurrent renders.

## Verification

```bash
cd packages/db
pnpm vitest run tests/live-query-window-controller.test.ts tests/live-query-order-only-move.test.ts --pool-options.threads.maxThreads=2 --coverage.enabled=false

cd ../react-db
pnpm vitest run tests/useLiveInfiniteQuery.test.tsx --pool-options.threads.maxThreads=2 --coverage.enabled=false

cd ../..
pnpm --filter @tanstack/db build
pnpm --filter @tanstack/react-db build
```

Focused results:

- `@tanstack/db`: 47 tests passed
- `@tanstack/react-db`: 42 tests passed
- ESLint passed for all changed TypeScript files
- Both production builds passed
- The implementation commit passed Test, E2E, Preview, example build, security, and autofix checks

## Files changed

- `packages/db/src/live-query-window-controller.ts`: shared controller, lease coordinator, transactional commits, and snapshot state.
- `packages/db/src/query/live/collection-config-builder.ts`: physical-window inspection, real subset-load completion, and restart cleanup.
- `packages/db/src/collection/*` and `packages/db/src/query/live/collection-subscriber.ts`: layout publication and rejection-safe load tracking.
- `packages/db/tests/live-query-window-controller.test.ts` and `live-query-order-only-move.test.ts`: controller races, failures, lifecycle, and public layout regressions.
- `packages/react-db/src/useLiveInfiniteQuery.ts`: thin React binding with stable pagination across compatible rerenders.
- `packages/react-db/tests/useLiveInfiniteQuery.test.tsx`: hook lifecycle, dependency, error, warning, and pagination coverage.
- `.changeset/live-query-window-controller.md`: patch release note for `@tanstack/db` and `@tanstack/react-db`.

## Related work

- RFC: #1623
- Window/limit-offset semantics: #1553
- Prior observer and ordered-layout phases: #1642, #1669
- Community adapter work: #1447, #1513 (cc @devcsrj @miguelrk)


KyleAMathews:
I reproduced the reported failures against head `1ca1a07` with focused tests. The extraction is the right direction, but I think the following correctness issues should be fixed before merge.

## 1. Render-time activation and callback generation

`useLiveInfiniteQuery` calls `collection.startSyncImmediate()` for a supplied collection and creates query-function collections with `startSync: true` during render. A render that never commits therefore activates resources with no committed subscription cleanup.

This reproduction fails with `source.subscriberCount === 2` instead of `0`:

```tsx
const never = new Promise<void>(() => {})

function AbandonedQuery(): ReactNode {
  useLiveInfiniteQuery(
    (q) => q.from({ row: source }).orderBy(({ row }) => row.n, `asc`),
    { pageSize: 3 },
  )
  throw never
}

render(
  <Suspense fallback={null}>
    <AbandonedQuery />
  </Suspense>,
)

await new Promise((resolve) => setTimeout(resolve, 0))
expect(source.subscriberCount).toBe(0)
```

The stable pagination callback has a related generation problem:

```ts
const fetchNextPage = useCallback(() => {
  controllerRef.current?.fetchNextPage()
}, [])
```

A callback returned for controller A follows the mutable ref after a later render replaces it with controller B. Retaining A's callback, replacing the collection, and then invoking A's callback advances B to two pages.

Please make collection construction inert and let the first committed `subscribe()` activate synchronization. Bind the callback to the controller from its render:

```ts
const fetchNextPage = useCallback(() => {
  void controller.fetchNextPage()
}, [controller])
```

One clarification: a normal Strict Mode mount/unmount sequence did return subscriber counts to baseline. The verified lifecycle failure is an abandoned render that never commits.

## 2. Dependency identity collisions

`JSON.stringify(deps)` is now the sole query-generation key. Distinct dependencies can serialize identically. For example, all `Map` instances serialize as `{}`.

This test remains on group `a` after replacing the dependency with a map for group `b`:

```tsx
const { result, rerender } = renderHook(
  ({ filter }: { filter: Map<string, string> }) =>
    useLiveInfiniteQuery(
      (q) =>
        q
          .from({ row: source })
          .where(({ row }) => eq(row.group, filter.get(`group`)))
          .orderBy(({ row }) => row.n, `asc`),
      { pageSize: 3 },
      [filter],
    ),
  { initialProps: { filter: new Map([[`group`, `a`]]) } },
)

await waitFor(() =>
  expect(result.current.data.every((row) => row.group === `a`)).toBe(true),
)

rerender({ filter: new Map([[`group`, `b`]]) })

await waitFor(() =>
  expect(result.current.data.every((row) => row.group === `b`)).toBe(true),
)
```

Functions, `Set`, `RegExp`, and objects with omitted `undefined` fields have similar collisions. Circular values also throw even though React dependency arrays need not be serializable.

Please match `useLiveQuery` with elementwise identity comparison and retain a shallow copy of the accepted dependency array.

## 3. Changing page shape discards loaded pages

Recreating the controller when `pageSize` or `initialPageParam` changes resets `loadedPageCount` to one. Before this extraction, page count lived independently and a page-size change re-windowed the same number of loaded pages.

I loaded three 3-row pages, changed `pageSize` to 5, and expected three 5-row pages. On this head, the result resets to one page and five rows:

```tsx
await waitFor(() => expect(result.current.isReady).toBe(true))
act(() => result.current.fetchNextPage())
await waitFor(() => expect(result.current.pages).toHaveLength(2))
act(() => result.current.fetchNextPage())
await waitFor(() => expect(result.current.pages).toHaveLength(3))

rerender({ pageSize: 5 })

await waitFor(() => expect(result.current.pages).toHaveLength(3))
expect(result.current.data).toHaveLength(15)
```

If this is meant to preserve public behavior, carry the committed page count into the replacement controller or add a reconfiguration operation. If reset semantics are intentional, please document the behavior change and test it after multiple pages have loaded. The current runtime page-size test changes the size while only one page exists, so it cannot distinguish these semantics.

## 4. Failed window updates commit pagination and poison retries

`fetchNextPage()` increments `loadedPageCount` before `setWindow()` succeeds. `applyWindow()` assigns `appliedLimit` before invoking `setWindow()` and swallows promise rejection.

A synchronous failure poisons the next attempt:

```ts
const setWindow = vi
  .spyOn(collection.utils, `setWindow`)
  .mockImplementationOnce(() => {
    throw new Error(`window failed`)
  })
  .mockReturnValue(true)

expect(() => controller.subscribe(() => {})).toThrow(`window failed`)
controller.subscribe(() => {})

expect(setWindow).toHaveBeenCalledTimes(2)
```

Actual result: `setWindow` is called once because the failed first call leaves `appliedLimit` set.

An asynchronous failure also commits the next page:

```ts
expect(controller.getSnapshot().hasNextPage).toBe(true)

vi.spyOn(collection.utils, `setWindow`).mockRejectedValueOnce(
  new Error(`load failed`),
)

await expect(controller.fetchNextPage()).rejects.toThrow(`load failed`)

expect(controller.getSnapshot().pages).toHaveLength(1)
expect(controller.getSnapshot().hasNextPage).toBe(true)
```

On this head, the controller exposes two pages after rejection. It can also lose `hasNextPage`, removing the normal retry path.

Please separate committed page count, requested page count, pending generation/limit, accepted limit, and pagination error. Keep slicing at the committed count until the current-generation window request succeeds. On failure, retain the old pages and `hasNextPage`, expose the error as `unknown`, and leave the limit retryable. `fetchNextPage()` should return the load promise, as required by RFC #1623.

## 5. Async expansion publishes duplicate and incoherent snapshots

For an asynchronous window expansion, `setFetching(true)` notifies and `fetchNextPage()` immediately calls `notify()` again. Both callbacks expose the same state, and both expose the new page count before the load succeeds.

This sequence test receives two identical immediate `{ pages: 2, fetching: true }` snapshots:

```ts
const snapshots: Array<{ pages: number; fetching: boolean }> = []
controller.subscribe(() => {
  const snapshot = controller.getSnapshot()
  snapshots.push({
    pages: snapshot.pages.length,
    fetching: snapshot.isFetchingNextPage,
  })
})

let resolveWindow!: () => void
vi.spyOn(collection.utils, `setWindow`).mockReturnValueOnce(
  new Promise<void>((resolve) => {
    resolveWindow = resolve
  }),
)

const fetch = controller.fetchNextPage()
expect(snapshots).toEqual([{ pages: 1, fetching: true }])

resolveWindow()
await fetch
expect(snapshots).toEqual([
  { pages: 1, fetching: true },
  { pages: 2, fetching: false },
])
```

Please place observer delivery behind a publication barrier while requested window, fetching state, and committed page count change. Emit one coherent loading snapshot and one success or failure snapshot. Synchronous observer notifications raised by `setWindow()` should join the same transition rather than expose intermediate state.

## 6. Preload and pipeline restart lose the desired window

The controller's `appliedLimit` outlives the compiled operator it was applied to. Collection cleanup clears the graph and pipeline caches. A later synchronization compiles a fresh top-K operator from the original AST limit, while the controller still believes its larger limit is installed.

Reproduction:

```ts
const unsubscribe = controller.subscribe(() => {})
await collection.preload()

controller.fetchNextPage()
await flush()
expect(controller.getSnapshot().data).toHaveLength(4)

unsubscribe()
await collection.cleanup()

controller.subscribe(() => {})
await collection.preload()
await flush()

expect(controller.getSnapshot().data).toHaveLength(4)
expect(controller.getSnapshot().hasNextPage).toBe(true)
```

With an original AST limit of 3 and a controller-expanded limit of 5, the restarted result contains only 3 rows. The controller skips reapplication because `appliedLimit` still equals 5.

`controller.preload()` has the same gap. Starting with an AST limit of 2 and `pageSize: 2`, this fails because preload never installs the required limit of 3:

```ts
await controller.preload()
expect(controller.getSnapshot().hasNextPage).toBe(true)
expect(collection.utils.getWindow()).toEqual({ offset: 0, limit: 3 })
```

Please store the desired window as durable builder state and replay it whenever `compileBasePipeline()` installs a new `windowFn`. The controller should establish that desired window before preload and on each zero-to-one attachment. A cache of the last applied limit cannot outlive the pipeline generation it mutated.

## 7. Multiple controllers overwrite one shared collection window

Each controller stores a private page count, but `setWindow()` mutates one collection-global operator. Two consumers of the same pre-created collection therefore overwrite each other.

```ts
const larger = createLiveQueryWindowController(collection, {
  pageSize: 2,
  mode: `wholesale`,
})
const smaller = createLiveQueryWindowController(collection, {
  pageSize: 1,
  mode: `wholesale`,
})

larger.subscribe(() => {})
smaller.subscribe(() => {})
await collection.preload()
await flush()

expect(larger.getSnapshot().data).toHaveLength(2)
expect(larger.getSnapshot().hasNextPage).toBe(true)
```

The second assertion fails. The smaller controller applies limit 2 after the larger controller applies limit 3. The larger controller then loses its peek row and reports `hasNextPage === false`.

For the current forward-only, offset-zero model, a collection-owned lease coordinator can apply the maximum active requested limit while each controller slices its own snapshot. Updating or releasing a lease recomputes the physical window. This also gives pipeline recompilation one authoritative desired window to reinstall. If shared ownership is out of scope, please enforce exclusive ownership and throw when a second controller attaches; silent last-writer-wins behavior corrupts pagination state.

## 8. The default observer mode does not match the controller contract

The controller inherits the observer's default granular mode, but its listener is only `() => void`; it discards the granular `ChangeMessage[]` payload.

With default options, this test fails because the listener fires synchronously inside `subscribe()`:

```ts
await collection.preload()
const controller = createLiveQueryWindowController(collection, {
  pageSize: 2,
})

let subscribing = true
let notifiedDuringSubscribe = false

controller.subscribe(() => {
  if (subscribing) notifiedDuringSubscribe = true
})
subscribing = false

expect(notifiedDuringSubscribe).toBe(false)
```

The existing no-synchronous-notify test passes `mode: 'wholesale'`, so it does not cover the factory default.

Please make the window controller wholesale and non-reentrant by construction and remove `mode` from its options. If a granular controller is added later, it should expose a delta-bearing listener contract.

## API boundary

RFC #1623 describes this as a shared internal controller and explicitly requires promise/error semantics, non-reentrant subscriptions, inert rendering, one atomic revision per observable change, and single ownership of the active window. This PR root-exports the controller and advertises it in a changeset without an `@internal`/unstable marker.

Please either complete that contract before presenting the controller as a new framework-neutral feature, or mark the factory, controller, options, and snapshot as internal/unstable for this phase and update the changeset accordingly.

## Verification

I implemented proof fixes locally and reran the focused regressions plus both full package suites:

- `@tanstack/db`: 2,554 passed, 5 skipped, no type errors
- `@tanstack/react-db`: 129 passed, no type errors
- both package builds and focused lint passed

I recommend keeping the extraction. The main design change is to treat requested pagination state, committed pagination state, and the collection-owned physical window as separate concepts.

tannerlinsley:
@KyleAMathews All eight reproduced blockers are fixed, the follow-up review pass is clean, and every check is green. Please re-review and merge when ready: https://github.com/TanStack/db/pull/1675

KyleAMathews:
I rechecked the follow-up changes on current head `d704b954`. The earlier window-controller blockers are fixed, but two additional correctness issues still reproduce.

## 1. `useLiveInfiniteQuery` reuses the wrong controller after an input-kind round trip

`needsNewCollection` checks `configRef` only on the collection path and `depsRef` only on the query-function path:

```ts
const needsNewCollection =
  !collectionRef.current ||
  (isCollection && configRef.current !== queryFnOrCollection) ||
  dependenciesChanged
```

Those refs retain values from the last visit to their respective path. This makes the first kind switch work, but switching back to the same prior input makes `needsNewCollection` false and leaves the hook bound to the intervening collection.

### Collection -> query function -> same collection

```tsx
const { result, rerender } = renderHook(
  ({ useCollection }) =>
    useLiveInfiniteQuery(
      useCollection ? collectionInput : queryInput,
      { pageSize: 3 },
      ...(useCollection ? [] : [[]]),
    ),
  { initialProps: { useCollection: true } },
)

await waitFor(() => expect(result.current.data[0].id).toBe(`1`))

rerender({ useCollection: false })
await waitFor(() => expect(result.current.data[0].id).toBe(`query-1`))

rerender({ useCollection: true })
await waitFor(() => expect(result.current.data[0].id).toBe(`1`))
```

The final assertion fails: the hook still returns `query-1`.

The inverse round trip also fails:

```tsx
// query function -> collection -> same query function
expect(result.current.data[0].id).toBe(`1`)
rerender({ useCollection: true })
await waitFor(() => expect(result.current.data[0].id).toBe(`collection-1`))
rerender({ useCollection: false })
await waitFor(() => expect(result.current.data[0].id).toBe(`1`))
```

The final value remains `collection-1`.

Please track the active input kind as part of collection identity. For example, include `inputKindRef.current !== isCollection` in `needsNewCollection`, update that ref whenever a new collection is accepted, and clear or overwrite the inactive branch's identity state. Add regressions for both round-trip directions; a one-way switch does not expose this bug.

## 2. Order-only moves escape through the public change API as an empty batch

`CollectionChangesManager.emitEvents()` now allows `rawEvents.length === 0` when `layoutChanged` is true, then calls every public `subscribeChanges` callback with `[]`. That reuses the existing empty-batch/ready sentinel as a new layout signal. A public consumer that treats a post-ready callback as containing at least one `ChangeMessage` can now misfire or crash.

This focused reproduction fails on the current head:

```ts
const source = makeSource([
  { id: `1`, name: `Alice`, age: 30 },
  { id: `2`, name: `Bob`, age: 20 },
  { id: `3`, name: `Carol`, age: 40 },
])
const lq = createLiveQueryCollection((q) =>
  q
    .from({ p: source })
    .orderBy(({ p }) => p.age, `asc`)
    .select(({ p }) => ({ id: p.id, name: p.name })),
)
await lq.preload()

const publications: Array<Array<unknown>> = []
const subscription = lq.subscribeChanges(
  (changes) => publications.push(changes),
  { includeInitialState: false },
)

source.utils.begin()
source.utils.write({
  type: `update`,
  value: { id: `2`, name: `Bob`, age: 99 },
})
source.utils.commit()

expect([...lq.values()].map((row) => row.id)).toEqual([`1`, `3`, `2`])
expect(publications).toEqual([])
```

Actual `publications` is `[[]]`. The same behavior is encoded in the current `publishes a mixed batch to a subscriber that filters out the row update` test, which expects `[[]]`.

The observer does need an explicit layout-only notification so wholesale consumers can re-read the ordered snapshot. Please carry that through a separate internal layout event/channel rather than overloading `ChangeMessage[]`. Public `subscribeChanges` should remain a stream of actual changes (apart from its documented bootstrap/ready behavior), while `LiveQueryObserver` can subscribe to both value changes and layout changes.

Verification on `d704b954`:

- both input-kind round-trip tests fail, with no type errors;
- the layout-only public-subscription test fails with `expected [ [] ] to deeply equal []`.

PR #1678, opened 2026-07-16
fix(db): restore query typechecking for generic collection row types

### LINKED ISSUE #1677: Since 0.6.6, queries don't typecheck when the collection's row type is a generic type parameter

**Describe the bug**

If you write a function that works on any `Collection<T>` (where `T` is a type parameter rather than a concrete type), the query builder's types break: inside `where`/`join`/`select` callbacks, the refs have no properties at all. Even properties that `T`'s constraint guarantees (like `id` in `T extends { id: string }`) can't be accessed. This worked fine up to 0.6.5 and broke in 0.6.6.

The cause seems to be this type, added in 0.6.6 as part of `unionAll` (#1537):

```ts
type RefForContextValue<T, Nullable extends boolean = false> =
  IsPlainObject<T> extends true ? Ref<T, Nullable> : RefLeaf<T, Nullable>;
```

When `T` is still an unresolved type parameter, TypeScript can't pick a side of that conditional (same for the `IsNonExactOptional`/`IsNonExactNullable` conditionals in `RefsForContext`), so the ref ends up as a union of every possible branch. One of those branches is `RefLeaf`, which has no row properties — and a union only exposes what all its members share, so the ref exposes nothing. Before 0.6.6, refs were always `Ref<...>`, so generic row types still worked.

**To Reproduce**

Typecheck this with `tsc --strict` against `@tanstack/db` 0.6.6 or later:

```ts
import { createLiveQueryCollection, eq, type Collection } from "@tanstack/db";

export function findById<T extends { id: string }>(items: Collection<T, string>, id: string) {
  return createLiveQueryCollection((q) => q.from({ items }).where(({ items }) => eq(items.id, id)));
}
```

```
error TS2339: Property 'id' does not exist on type
  'Ref<NonNullable<WithVirtualProps<T, string>>, true> | RefLeaf<NonNullable<WithVirtualProps<T, string>>, true> | ...'.
  Property 'id' does not exist on type 'RefLeaf<NonNullable<WithVirtualProps<T, string>>, true>'.
```

The same root cause also breaks joins, though the error looks unrelated at first: using a subquery over a generic collection as a join source fails because its `QueryBuilder<ContextFromSource<...>>` no longer satisfies `leftJoin`'s `QueryBuilder<Context>` parameter (the `RefLeaf` branch is missing the `$synced`/`$origin`/`$key`/`$collectionId` properties that every ref is required to have):

```ts
export function withDiff<T extends { id: string }>(a: Collection<T, string>, b: Collection<{ id: string }, string>) {
  return createLiveQueryCollection((q) => {
    const sub = q.from({ a });
    return q.from({ b }).leftJoin({ sub }, ({ b, sub }) => eq(b.id, sub.id)); // TS2322 on `{ sub }`
  });
}
```

**Expected behavior**

Both snippets compile, as they did on 0.6.5. A ref for a generic row type should expose at least the properties its constraint guarantees.

- `@tanstack/db`: broken in 0.6.6 through 0.6.14 (latest), works in 0.6.5
- TypeScript: 5.9.3 (`--strict`)
- No browser/OS involved — this is compile-time only

**Additional context**

I tested every release from 0.6.5 through 0.6.14: 0.6.5 compiles, and every version from 0.6.6 onward fails. The `Ref<T>` distributivity change in 0.6.13 (#1597) changes the wording of the error but isn't where it started.

obeattie:
This restores TypeScript query-builder support for collections whose row type is an unresolved generic parameter. Generic `where`, `select`, and subquery join-source callbacks once again expose fields guaranteed by the generic constraint, fixing the regression introduced in v0.6.6.

## Root cause

The query builder decides whether a callback value should be represented as a property-bearing `Ref` or an opaque `RefLeaf` through a conditional type. For an unresolved generic such as `T extends { id: string }`, the conditional was deferred into a union containing both possibilities. Because `RefLeaf` has no row properties, TypeScript rejected access to fields such as `id` even though the generic constraint guarantees them.

Subquery sources had an additional mapped `Prettify` layer that prevented TypeScript from resolving the generic constraint through the ref conditional.

## Approach

- Make `RefForContextValue` distributive so TypeScript can evaluate the ref shape through the generic constraint.
- Use the unwrapped `GetRawResult` for internal subquery schemas, while retaining `GetResult` for final user-facing result types.
- Add type regressions for direct `where`, direct `select`, and generic subqueries used as join sources.

## Key invariants

- A constraint such as `T extends { id: string }` must expose `id` inside supported query callbacks.
- Concrete result types retain their existing IDE-friendly presentation.
- The internal schema change does not alter runtime query behavior.

## Non-goals

This PR restores the three cases reported in #1677. Broader generic-constraint propagation through nullable joined aliases and both forms of `unionAll` is tracked separately in #1679; resolving those cases safely requires additional type-system work and must preserve existing heterogeneous union and join inference.

## Trade-offs

`GetRawResult` is used only on the internal subquery schema path. Final results continue to use the prettified representation, avoiding a user-facing type-display regression.

## Verification

The new type tests were added first and verified RED against the regressed implementation, then GREEN after the fix.

```bash
pnpm --filter @tanstack/db test
```

## Files changed

- `packages/db/src/query/builder/types.ts` — preserves generic constraints through callback ref and subquery schema construction.
- `packages/db/tests/query/generic-collection.test-d.ts` — covers generic filtering, selection, and subquery join sources.
- `.changeset/generic-row-type-refs.md` — records the patch release impact.

---

Fixes #1677

Follow-up: #1679

### DIFF
--- a/.changeset/generic-row-type-refs.md
+++ b/.changeset/generic-row-type-refs.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/db': patch
+---
+
+Fix queries failing to typecheck when the collection's row type is a generic type parameter. Refs inside where/join/select callbacks now expose the properties guaranteed by the type parameter's constraint, and subqueries over generic collections can be used as join sources again (regression introduced in 0.6.6).
--- a/packages/db/src/query/builder/types.ts
+++ b/packages/db/src/query/builder/types.ts
@@ -117,7 +117,7 @@ export type SchemaFromSource<T extends Source> = Prettify<{
   [K in keyof T]: T[K] extends CollectionImpl<any, any, any, any, any>
     ? InferCollectionType<T[K]>
     : T[K] extends QueryBuilder<infer TContext>
-      ? GetResult<TContext>
+      ? GetRawResult<TContext>
       : never
 }>
 
@@ -662,8 +662,11 @@ type ValueOfUnion<T, K extends PropertyKey> = T extends unknown
     ? T[K]
     : never
   : never
-type RefForContextValue<T, Nullable extends boolean = false> =
-  IsPlainObject<T> extends true ? Ref<T, Nullable> : RefLeaf<T, Nullable>
+type RefForContextValue<T, Nullable extends boolean = false> = T extends unknown
+  ? IsPlainObject<T> extends true
+    ? Ref<T, Nullable>
+    : RefLeaf<T, Nullable>
+  : never
 type RefsSchemaForContext<TContext extends Context> =
   IsExactlyUndefined<TContext[`refsSchema`]> extends true
     ? TContext[`schema`]
--- a/packages/db/tests/query/generic-collection.test-d.ts
+++ b/packages/db/tests/query/generic-collection.test-d.ts
@@ -1,0 +1,53 @@
+import { describe, test } from 'vitest'
+import { createLiveQueryCollection, eq } from '../../src/query/index.js'
+import type { Collection } from '../../src/collection/index.js'
+
+// Regression tests for https://github.com/TanStack/db/issues/1677
+// Queries over a Collection<T> where T is an unresolved generic type parameter
+// must still expose the properties guaranteed by T's constraint inside
+// where/join/select callbacks. This worked in 0.6.5 and broke in 0.6.6.
+
+describe(`queries over generic collection row types`, () => {
+  test(`where callback can access properties guaranteed by the type constraint`, () => {
+    function findById<T extends { id: string }>(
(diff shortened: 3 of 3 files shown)


KyleAMathews (approved):
Verified the reported #1677 regression and the PR's three type-test cases: generic where callbacks, select callbacks, and generic subqueries used as join sources. The implementation is appropriately scoped and the broader nullable/union constraint-propagation work is now documented in #1679. Approving this as a valid incremental regression fix.

PR #1682, opened 2026-07-16
docs(query-db): design initial data semantics

KyleAMathews:
## Summary

- define Query cache, normalized row, and server-result authority for initial and placeholder data
- recommend flat eager-only `initialData` and `initialDataUpdatedAt` fields while preserving existing cache seeding and hydration behavior
- keep `placeholderData` observer-local and out of DB materialization, ownership, and persistence
- specify select, direct-write, shared QueryClient, same-key, lifecycle, ownership, and persistence semantics
- propose a test-first sequence of small implementation PRs

## Key decisions

- Initial data is ordinary successful Query cache data and materializes through the existing row extraction and ownership pipeline.
- Existing cached or hydrated data wins; initial data only initializes a missing Query document.
- Collections sharing an exact Query key share one Query document, even when they share a QueryClient with many other collections.
- The first public API is eager-only because one initializer cannot establish membership for arbitrary on-demand subsets.
- Placeholder data is UI presentation state and must never become collection-wide normalized rows.
- Wrapped responses require the existing `select`; direct writes must not fabricate a reverse projection.

## Scope

Design and documentation only. No runtime behavior, public option forwarding, persistence format, or tests are changed.

Relates to #1643 and #346.

## Verification

- `pnpm exec prettier --check docs/collections/query-initial-placeholder-data-design.md`
- `git diff --check`

## Summary by CodeRabbit

* **Documentation**
  * Added design documentation for `initialData` and `placeholderData` behavior in query-backed collections.
  * Clarified that `initialData` is supported for eager collections and treated as stored query data.
  * Documented that placeholder data is not supported in the initial implementation.
  * Defined initialization, freshness, persistence, ownership, projection, and write behavior.

### DIFF
--- a/docs/collections/query-initial-placeholder-data-design.md
+++ b/docs/collections/query-initial-placeholder-data-design.md
@@ -1,0 +1,248 @@
+# Query Collection initial and placeholder data semantics
+
+## Status and scope
+
+This document proposes the semantics for TanStack Query `initialData` and
+`placeholderData` at the `@tanstack/query-db-collection` boundary. It is the
+design follow-up for [RFC #1643](https://github.com/TanStack/db/issues/1643)
+and [issue #346](https://github.com/TanStack/db/issues/346). It does not change
+runtime behavior or the persistence format.
+
+The adapter connects two different models:
+
+- TanStack Query owns a document cache and remote-query lifecycle.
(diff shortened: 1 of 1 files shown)


KyleAMathews:
Closing the design-only PR so this branch can proceed with the approved implementation and tests.

PR #1683, opened 2026-07-16
feat(query-db): support eager initial data

### LINKED ISSUE #346: Add Initial Data and Query Options Support to query-db-collection

KyleAMathews:
Adds typed, collection-local `initialData` and `initialDataUpdatedAt` support to eager Query Collections. Initial Query responses now materialize immediately as collection rows while retaining TanStack Query's cache identity, freshness, and wrapped-response semantics.

## Root Cause

Query Collections already materialized data preloaded into a `QueryClient`, but they could not declare initial data per collection. Applications sharing one client across many collections therefore had to choose between overly broad client defaults and imperative cache seeding. Forwarding defaults unchanged was also unsafe for on-demand subsets and placeholder data: neither has collection-wide row-membership semantics.

## Approach

- Add typed top-level `initialData` and `initialDataUpdatedAt` options using the original Query response type.
- Forward those options only for eager observers, then reuse the existing `select`, materialization, and row-ownership pipeline.
- Reject explicit collection-level initial data in on-demand mode and suppress client-wide initial-data defaults for derived subset observers.
- Suppress `placeholderData` defaults because placeholder results are observer-local UI state, not Query cache data suitable for normalized rows.
- Document the authority, lifecycle, wrapped-response, write, ownership, and persistence semantics.

## Key Invariants

- The Query cache remains authoritative; existing cached or hydrated data wins over a later initializer.
- Exact Query keys identify shared documents, even when collections share a `QueryClient`.
- `select` materializes rows without replacing the original wrapped response in the Query cache.
- Stale initial rows remain available during refetch and reconcile through normal ownership rules.
- Placeholder data and collection-wide initializers never become rows for arbitrary on-demand subsets.
- Persistence needs no new format or temporary-row metadata.

## Non-goals

- Materializing `placeholderData` as DB rows.
- Defining a subset-aware initializer for on-demand collections.
- Giving each collection a private cache document when it uses the same exact Query key.
- Changing direct-write or persistence formats.

## Trade-offs

On-demand mode rejects explicit initial-data options rather than guessing subset membership. Callers that know an exact derived key can still seed or hydrate that Query cache entry. Eager mode suppresses placeholder defaults to keep observer presentation state out of the collection's shared normalized state.

## Verification

```bash
pnpm vitest run packages/query-db-collection/tests/query.test.ts --maxWorkers=1
pnpm --filter @tanstack/query-db-collection lint
git diff --check
```

The focused Query Collection suite covers fresh and stale initial data, function initializers, wrapped response projection, shared clients and exact keys, existing-cache precedence, on-demand rejection/default suppression, and placeholder-default suppression. Lint passes with existing warnings.

## Files changed

- `packages/query-db-collection/src/query.ts` — adds the public options and observer initialization policy.
- `packages/query-db-collection/src/errors.ts` — adds the on-demand configuration error.
- `packages/query-db-collection/tests/query.test.ts` — adds initialization and isolation coverage.
- `docs/collections/query-collection.md` — documents usage and user-visible semantics.
- `docs/collections/query-initial-placeholder-data-design.md` — records authority, ownership, transitions, writes, and persistence decisions.
- `.changeset/young-cats-initialize.md` — records the minor package feature.

---

Closes #346. Advances #1643.

## Summary by CodeRabbit

- **New Features**
  - Added eager collection support for TanStack Query’s `initialData` and `initialDataUpdatedAt`.
  - Initial data is immediately available as collection rows while preserving wrapped response data.
  - Fresh initial data can prevent an unnecessary fetch; stale data is refreshed and reconciled with server results.
  - Added safeguards so placeholder data is not materialized as collection rows.
  - Added clear errors when initial data is used with on-demand collections.

- **Documentation**
  - Documented initial data behavior, cache precedence, wrapped responses, and placeholder data limitations.

### DIFF
--- a/.changeset/young-cats-initialize.md
+++ b/.changeset/young-cats-initialize.md
@@ -3,3 +3,5 @@
 ---
 
 Add eager collection support for TanStack Query `initialData` and `initialDataUpdatedAt`, including wrapped response projection and collection-local initialization on shared QueryClient instances.
+
+QueryClient-default `placeholderData` no longer materializes as collection rows, and QueryClient-default `initialData` no longer seeds on-demand subset observers.
--- a/.changeset/young-cats-initialize.md
+++ b/.changeset/young-cats-initialize.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/query-db-collection': minor
+---
+
+Add eager collection support for TanStack Query `initialData` and `initialDataUpdatedAt`, including wrapped response projection and collection-local initialization on shared QueryClient instances.
--- a/docs/collections/query-collection.md
+++ b/docs/collections/query-collection.md
@@ -200,6 +200,8 @@ The following top-level Query Collection options are forwarded to the underlying
 - `refetchOnReconnect`: Whether to refetch when the network reconnects
 - `refetchOnMount`: Whether to refetch when the observer mounts
 - `networkMode`: Query network mode
+- `initialData`: Initial Query response for eager collections
+- `initialDataUpdatedAt`: Timestamp used by TanStack Query to determine initial data freshness
 - `meta`: Metadata passed to the query function context. Query Collections may add `loadSubsetOptions` for on-demand queries.
 
 ```ts
@@ -237,7 +239,7 @@ Some TanStack Query fields are owned or reinterpreted by Query Collection and ar
 - `subscribed` (Query Collection owns the observer subscription lifecycle)
 - `structuralSharing` and `notifyOnChangeProps` (managed by Query Collection synchronization)
 
-Other semantic options, such as `initialData`, `placeholderData`, and TanStack Query observer-level `select`, are intentionally deferred until their Query Collection behavior is explicitly designed.
+`placeholderData` is intentionally unsupported. TanStack Query treats placeholder data as observer-local presentation state rather than cached Query data. Materializing it would expose temporary UI data as collection-wide normalized rows. Render placeholders in the consuming UI instead.
 
 ### Using with `queryOptions(...)`
 
@@ -271,6 +273,51 @@ const todosCollection = createCollection(
 
 If `queryFn` is missing at runtime, `queryCollectionOptions` throws `QueryFnRequiredError`.
 
+### Initial Data
+
+Eager Query Collections support TanStack Query's `initialData` and
+`initialDataUpdatedAt` options. Initial data has the original Query response
+shape, is stored in the Query cache, and is immediately materialized as
+normalized collection rows. TanStack Query uses `initialDataUpdatedAt` together
+with `staleTime` to decide whether to fetch.
+
+```typescript
+const serverRenderedAt = Date.now()
+const initialTodos = [
+  { id: "1", title: "Write documentation" },
+  { id: "2", title: "Ship initial data support" },
+]
+
+const todosCollection = createCollection(
+  queryCollectionOptions({
+    queryKey: ["todos"],
+    queryFn: fetchTodos,
+    queryClient,
+    getKey: (todo) => todo.id,
+    initialData: initialTodos,
+    initialDataUpdatedAt: serverRenderedAt,
+    staleTime: 60_000,
+  }),
+)
+```
+
+An existing cached or hydrated Query response takes precedence over
+`initialData`. Query keys remain the cache identity: two collections using the
+same QueryClient and exact Query key observe one shared Query document, and a
+later collection's initializer does not replace it. Use distinct Query keys for
+independent documents.
+
+Initial data is supported only for eager collections. A collection-wide value
+cannot establish row membership for arbitrary on-demand predicates, ordering,
+limits, and offsets. For `syncMode: "on-demand"`, seed or hydrate the exact
+derived Query cache entries instead.
+
+If a stale initial response triggers a fetch, the initial rows remain available
+while it is in flight. A successful response reconciles them through the normal
+row ownership pipeline; an error retains the initial rows. Direct writes use the
+same Query cache-patching rules as fetched data, and a later successful server
+response may reconcile or replace those writes.
+
 ### Selecting Rows from Wrapped Responses
 
 Many APIs return rows inside a response envelope that also contains metadata such as pagination cursors, totals, or request information. Use `select` to extract the row array that TanStack DB should materialize:
@@ -289,6 +336,11 @@ const todosCollection = createCollection(
       const response = await fetch("/api/todos")
       return response.json()
     },
+    initialData: {
+      items: [{ id: "1", title: "Initial todo" }],
+      nextCursor: undefined,
+      total: 1,
+    },
     select: (response) => response.items,
     queryClient,
     getKey: (item) => item.id,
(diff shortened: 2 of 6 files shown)


kevin-dp:
@KyleAMathews i had Fable review the PR and chatted through it with Fable. This is what we landed on:

**Verdict: approve the architecture; a few things worth addressing before merge — one real risk (persistence interplay), one changeset omission, and some test gaps relative to the PR's own design doc.** CI is fully green (unit, E2E, example build, lint/autofix).

### Issue 1 — persistence interplay can clobber newer persisted rows (main risk)

On restart with SQLite persistence, the Query cache is empty but persisted rows exist under `until-revalidated` retention (`retainedQueriesPendingRevalidation`). With `initialData` configured:

1. Observer creation seeds the cache; `subscribeToQuery` immediately fires a success result.
2. `handleQueryResult` sees the pending-revalidation flag and calls `reconcileSuccessfulResult` (`query.ts:1534`), which reconciles the **persisted baseline against the static seed** — deleting persisted rows the seed omits and clearing the retention flag (`query.ts:1494`).
3. Crucially, if `initialDataUpdatedAt` is omitted, query-core stamps `dataUpdatedAt = Date.now()`, so the seed always looks fresh and **no refetch follows** for the duration of `staleTime`.

So a retention mechanism meaning "keep these rows until the server revalidates them" is now satisfiable by static config data with no network involved, silently regressing the user's last-seen state. Mechanically it's "same as a stale server response," but there's no server in the loop. Suggestions:

- At minimum: a persistence + `initialData` test (the design doc's own step 5 promises exactly this) and a docs warning to always set `initialDataUpdatedAt` when persistence is enabled.
- Worth considering: when a retained query's success result was never fetched (`query.state.dataUpdateCount === 0` / `isFetchedAfterMount === false`), trigger a revalidating refetch instead of treating the seed as revalidation.

### Issue 2 — changeset omits two silent behavior changes

The changeset only advertises the new feature, but the PR also changes existing behavior for collections that never opt in:

- Client-default `placeholderData` no longer materializes (for **all** query collections).
- Client-default `initialData` no longer seeds on-demand subset observers.

Both are deliberate correctness fixes, but they remove rows an app may currently be (mis)relying on. They should be called out in the changeset so the release notes warn anyone affected.

### Issue 3 — test coverage vs. the PR's own claims

Covered well: fresh/stale seeds, function-initializer-evaluated-once, wrapped envelope preserved in cache, shared-client scoping, existing-cache precedence, on-demand rejection and default suppression, placeholder suppression.

Missing relative to the design doc's behavior table and its own implementation sequence (steps 3–5):

- **Refetch error retains initial rows** — explicitly promised in the docs ("an error retains the initial rows") but untested; cheap to add.
- Direct writes (`writeInsert`/`writeUpdate`) against a wrapped initial envelope.
- Cleanup/late-notification and the persistence scenarios above.

Given that, "Closes #346" is a stretch by the design doc's own sequencing ("close or narrow #346 only after these contracts ship" — step 6). Suggest "Advances #346" unless the error-retention and persistence tests land in this PR.

### Minor

- `docs/collections/query-initial-placeholder-data-design.md` is internal design material in the published docs folder; it's not registered in `docs/config.json` so it won't render in the site nav. Consider moving it out of `docs/` (e.g., an `rfcs/` location) or linking it from RFC #1643 instead.

PR #1688, opened 2026-07-21
PowerSync collection: Upgrade PowerSync SDK to v2

simolus3:
## 🎯 Changes

This upgrades `@powersync/` dependencies to their latest version, which includes breaking changes affecting `@tanstack/powersync-db-collection`:

1. `AbstractPowerSyncDatabase` is a deprecated type-alias, we should use `CommonPowerSyncDatabase` instead.
2. The internal logger on databases is based on a single `log()` function instead of one per severity now.
3. `Table.columnMap` has been removed in favor of a type parameter, this changes a few helper types to `infer` the column structure instead of extracting it from a field.
4. The `CrudEntry` class is an implementation detail and no longer exposed. The integration used that in tests, this replaces usages by directly querying from the underlying internal table instead. 

This simplifies some types around `serializeForSQLite`, but nothing that affects compiled JS.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Added compatibility with PowerSync version 2.
  * Improved type inference for table records, optional fields, and SQLite serialization.
  * Schema transformations are now reflected in collection item types and comparison callbacks.
  * Updated transaction handling for the current PowerSync database interface.

* **Bug Fixes**
  * Standardized logging for synchronization, serialization, and mutation-processing errors.
  * Simplified warnings when mutation metadata is unavailable.

* **Documentation**
  * Updated backend connector examples and batching guidance for the current database API.

### DIFF
--- a/.changeset/olive-guests-bathe.md
+++ b/.changeset/olive-guests-bathe.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/powersync-db-collection': minor
+---
+
+Upgrade PowerSync to version 2.
--- a/docs/collections/powersync-collection.md
+++ b/docs/collections/powersync-collection.md
@@ -58,7 +58,7 @@ const db = new PowerSyncDatabase({
 
 ```ts
 import {
-  AbstractPowerSyncDatabase,
+  CommonPowerSyncDatabase,
   PowerSyncBackendConnector,
   PowerSyncCredentials,
 } from "@powersync/web"
@@ -69,11 +69,11 @@ class Connector implements PowerSyncBackendConnector {
 
   /** Upload local changes to the app backend.
    *
-   * Use {@link AbstractPowerSyncDatabase.getCrudBatch} to get a batch of changes to upload.
+   * Use {@link CommonPowerSyncDatabase.getCrudBatch} to get a batch of changes to upload.
    *
    * Any thrown errors will result in a retry after the configured wait period (default: 5 seconds).
    */
-  uploadData: (database: AbstractPowerSyncDatabase) => Promise<void>
+  uploadData: (database: CommonPowerSyncDatabase) => Promise<void>
 }
 
 // Configure the client to connect to a PowerSync service and your backend
--- a/packages/powersync-db-collection/package.json
+++ b/packages/powersync-db-collection/package.json
@@ -59,11 +59,11 @@
     "p-defer": "^4.0.1"
   },
   "peerDependencies": {
-    "@powersync/common": "^1.41.0"
+    "@powersync/common": "^2.0.0"
   },
   "devDependencies": {
-    "@powersync/common": "1.49.0",
-    "@powersync/node": "0.18.1",
+    "@powersync/common": "^2.0.0",
+    "@powersync/node": "^0.20.0",
     "@types/debug": "^4.1.12",
     "@vitest/coverage-istanbul": "^3.2.4",
     "better-sqlite3": "^12.6.2"
--- a/packages/powersync-db-collection/src/PowerSyncTransactor.ts
+++ b/packages/powersync-db-collection/src/PowerSyncTransactor.ts
@@ -1,8 +1,8 @@
-import { sanitizeSQL } from '@powersync/common'
+import { LogLevels, sanitizeSQL } from '@powersync/common'
 import DebugModule from 'debug'
 import { PendingOperationStore } from './PendingOperationStore'
 import { asPowerSyncRecord, mapOperationToPowerSync } from './helpers'
-import type { AbstractPowerSyncDatabase, LockContext } from '@powersync/common'
+import type { CommonPowerSyncDatabase, LockContext } from '@powersync/common'
 import type { PendingMutation, Transaction } from '@tanstack/db'
 import type { PendingOperation } from './PendingOperationStore'
 import type {
@@ -13,7 +13,7 @@ import type {
 const debug = DebugModule.debug(`ts/db:powersync`)
 
 export type TransactorOptions = {
-  database: AbstractPowerSyncDatabase
+  database: CommonPowerSyncDatabase
 }
 
 /**
@@ -52,7 +52,7 @@ export type TransactorOptions = {
  * @returns A promise that resolves when the mutations have been persisted to PowerSync
  */
 export class PowerSyncTransactor {
-  database: AbstractPowerSyncDatabase
+  database: CommonPowerSyncDatabase
   pendingOperationStore: PendingOperationStore
 
   constructor(options: TransactorOptions) {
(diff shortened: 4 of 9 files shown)


tannerlinsley:
I tested this branch by itself and stacked after #1585, #1616, and #1649. The package tests and type checks pass, and the attachments code compiles against PowerSync 2, but I can't merge this version yet.

After the preceding PRs land, this needs a rebase. The stack currently conflicts in `package.json`, `powersync.test.ts`, and `pnpm-lock.yaml`. More importantly, the v2 test run repeatedly logs tracking-table flushes after disposal and `cleaned-up` to `ready` lifecycle errors when trigger setup finishes after cleanup. Those races are caught internally, so the tests stay green while the integration is still logging real failures. Please fix those paths and add regression coverage before merge.

There are also two small cleanups in the current diff: the metadata path says it logs a warning but uses `LogLevels.info`, and `getAll<{ op: 'string' }>` should use `string`. The PowerSync 2 peer requirement is intentionally breaking, and the `minor` changeset is appropriate for the current 0.x package.

jaltin:
A data point from running this combination in anger, in case it's useful while this waits on its rebase.

We've shipped `@powersync/common` 2.1.0 / `@powersync/web` 2.2.0 / `@powersync/react-native` 2.1.0 against the published `@tanstack/powersync-db-collection` **0.1.59**, in a React + React Native app with 16 collections, by wrapping the database in a `Proxy` that presents a 1.x-shaped logger:

```ts
new Proxy(database, {
  get(target, property) {
    if (property === 'logger') return legacyLoggerFacade  // .info/.warn/.error -> log({ level, message, error })
    const value = Reflect.get(target, property, target)
    return typeof value === 'function' ? value.bind(target) : value
  },
})
```

Two things that might save you time:

**1. The logger really is the only incompatibility.** We enumerated every `database.*` access in 0.1.59's `dist` — `logger.{info,warn,error}`, `triggers.createDiffTrigger`, `writeTransaction`, `writeLock`, `onChangeWithCallback`, `getAll`, plus the `sanitizeSQL` / `DiffTriggerOperation` / `TriggerDiffRecord` / `ColumnType` / `Table` imports — and everything except the logger is unchanged in 2.x. That matches this PR's diff exactly. With the logger substituted, the adapter works unmodified: `instanceof` holds, `triggers` returns an identical reference, and a full `createDiffTrigger` + `beforeCreate` cycle completes against a real `@powersync/node` 0.21.0 database.

**2. Without it the failure is total, not cosmetic** — worth flagging for anyone who tries 2.x before this lands. The first thing each collection does on start is log `Sync is ready for <table>`, so *every* collection's sync start throws `TypeError: database.logger.info is not a function`, and the `.catch()` that would report it throws again on `logger.error`. We saw 32 `TypeError`s and zero rows before adding the shim.

On the lifecycle races @tannerlinsley mentions — tracking-table flushes after disposal, and `cleaned-up` → `ready` when trigger setup finishes after cleanup — we see the disposal half of that in our RN app too: two generations of `__<table>_tracking_<hash>` tables for the same collections within one session, with `Sync has been stopped for <table> into __<table>_tracking_<hash>` for both. In our case the trigger is app-side (we recreate collections on a tenant switch, and separately connect twice at startup), so it may be our churn rather than an adapter bug — but it does mean the adapter is being disposed and recreated in normal use, which is presumably the same window your test run is hitting.

Happy to test a rebased build against our app if that's useful — we have both web and React Native surfaces on 2.x, plus a `@powersync/node` load-test harness.

mattriese:
> A data point from running this combination in anger

To my fellow Americans who have never heard the phrase "in anger" used this way before: its a [british idiom](https://dictionary.cambridge.org/dictionary/english/in-anger) meaning "for real / in a real world setting". At first I thought @jaltin was saying he was rage-testing this branch 😅

simolus3:
I have rebased this ontop of the latest `main` containing two of the other PowerSync-related PRs you mentioned. I don't think it makes sense to stack it ontop of the attachment integration as independent to that.

> More importantly, the v2 test run repeatedly logs tracking-table flushes after disposal and `cleaned-up` to `ready` lifecycle errors when trigger setup finishes after cleanup. Those races are caught internally, so the tests stay green while the integration is still logging real failures. Please fix those paths and add regression coverage before merge.

This doesn't appear to be a regression from the v2 update, and I see there's [an existing branch](https://github.com/TanStack/db/compare/main...codex/powersync-startup-barrier) to improve that.

simolus3:
@tannerlinsley @KyleAMathews Could I get another review here? Since this is blocking our users from upgrading, we'd like to land this if possible instead of stacking it on another PR. It also seems like the reported "repeatedly logs tracking-table flushes after disposal" issue from the first review is not a regression here.

KyleAMathews:
@simolus3 what's the plan with the powersync prs? I thought https://github.com/TanStack/db/pull/1616 was the urgent one? Is this stacked on that then or?

Chriztiaan:
Hey @KyleAMathews, this has become a bit more important than the attachment PR seeing as more users are asking for the V2 version bump and you had feedback on the attachment PR which might take a minute to be resolved.

I am not expecting any blocking feedback on this PR so it feels like reducing the PowerSync review surface by getting this in should be reasonable.

We wanted to maintain chronological order for simplicity, but these PRs are no longer stacked.

KyleAMathews (approved):
I found two issues worth following up on. I'm approving so this doesn't block the PowerSync v2 upgrade, but I'd like these addressed before merge if practical or tracked explicitly as follow-up work.

**1. The `restConfig` cast hides a current transformed-schema bug.**

`BasePowerSyncCollectionConfig` types `compare` against the SQLite `RowType<TTable>`, but the resulting collection invokes it with the schema's transformed output type. For example, a schema that transforms `title: string` into a `Holder` still allows:

```ts
compare: (left, right) =>
  left.title!.localeCompare(right.title!)
```

That configuration type-checks, but the comparator receives `Holder` instances at runtime and throws `TypeError: left.title.localeCompare is not a function`.

Removing the cast at `powersync.ts:953` exposes this mismatch as TS2322. Please align output-sensitive configuration fields—at least `compare`—with the schema output type, narrow/remove the cast, and add a transformed-schema regression test.

**2. Some v2 lifecycle tests are false-green.**

At least these tests pass while logging `no such table: __products_tracking_*` errors:

- `does not repeat release work started by a reentrant cleanup`
- `rechecks active demand before evicting released rows`

They spy on `db.getAll`, but the v2 path flushes through the transaction context's `getAll`, so the mock no longer covers the operation under test. The same focused tests on the parent commit with PowerSync v1 do not log these failures. Please update the test seam—or use a real trigger—capture logger errors, and await cleanup before asserting.

Two smaller cleanups are also valid:

- Remove the unused `MapBaseColumnType` helper and its `BaseColumnType` import.
- Decide whether intentionally to retain the metadata previously passed to `logger.warn`. Using `error: mutation.metadata` would surface it, although `error` is semantically an awkward metadata channel.

The serializer callback types themselves appear sound at the public configuration boundary: transformed callback inputs remain typed, typoed property access is rejected, and serializers must return the corresponding SQLite column type.

Keeping this migration separate from the attachment work makes sense. Approved so it can move forward when ready; please address or track the comparator and false-green test issues.

simolus3:
@KyleAMathews  I have added type and runtime tests for mapped schemas, removed stubbed triggers from new tests to avoid them logging exceptions and `MapBaseColumnType`.

> Decide whether intentionally to retain the metadata previously passed to `logger.warn`. Using `error: mutation.metadata` would surface it, although `error` is semantically an awkward metadata channel.

We don't want to pass untyped metadata through loggers on the PowerSync database, I think including the metadata object is not required for that message which just complains about metadata being set at all (instead of it e.g. having invalid values).

simolus3:
Merged as part of https://github.com/TanStack/db/pull/1831, closing this.

Issue #1625, opened 2026-06-26
RFC: Mutation log reconciliation for optimistic writes


KyleAMathews:
# RFC: Mutation log reconciliation for optimistic writes

Status: draft for maintainer review  
Date: 2026-06-25

## Summary

TanStack DB should make collection-owned optimistic mutations explicit by introducing an internal **Mutation Log**. The log/index becomes the single projection source for unsettled optimistic mutations and the queryable surface for local write state and write errors.

The core state model should become:

```txt
authoritative synced/base state
+ unsettled optimistic mutations owned by the collection
= visible collection state
```

This lets collections reconcile authoritative sync updates and optimistic writes through one explicit projection model. However, committed sync cannot always be applied immediately while optimistic writes are pending. In particular, optimistic inserts may use temporary client keys while the authoritative sync stream later contains different server-generated keys. Without explicit identity mapping or a mutation receipt, core cannot know whether the server row replaces the optimistic row or is an unrelated row from another client.

Therefore, the existing contract remains: sync that could create ambiguous overlap with unsettled optimistic mutations must stay queued until the potentially overlapping mutation functions settle. Once a mutation function settles successfully, core drops that transaction's optimistic projection and can safely apply queued authoritative sync without displaying both the temp row and a possible server-key echo. Immediate sync application while mutations are pending is only sound when core can prove the sync cannot be the authoritative representation of any unsettled optimistic mutation.

This RFC focuses on the core design needed for 1.0:

- Introduce a Mutation Log as a refinement of current transaction/mutation state.
- Apply provably safe committed authoritative sync/base changes without waiting for unrelated mutation functions, while preserving blocking for sync that could ambiguously overlap with unsettled optimistic identity.
- Reproject unsettled optimistic mutations over the latest exposed base state; committed-but-queued authoritative sync remains pending reconciliation work until it can be applied without ambiguous temp/server identity overlap.
- Preserve a stable materialized change stream for subscriptions and live queries.
- Replace ambiguous `$synced` / `isPersisted` concepts with clearer local write state.
- Expose queryable logged mutations joined with transaction state for write status and write errors.
- Retain failed transaction/mutation records with bounded automatic GC.
- Slim `@tanstack/offline-transactions` into durability/execution over the log.

[...]

PR #1701, opened 2026-07-27
fix(db): compose optimistic updates from changes instead of whole-row snapshots

marbemac:
## 🎯 Changes

Small fix to how the optimistic overlay composes rows.

Today an optimistic insert/update stores `mutation.modified` — the whole row as it looked at `mutate()` time — into `optimisticUpserts`. Since `modified` is built from the *visible* row (`Object.assign({}, visibleRow, payload)` in `mutations.ts`), whichever transaction sorts last dictates **every** field, including ones it never touched.

The clearest symptom: rolling back one of several in-flight transactions doesn't actually remove its change.

It doesn't change **when** committed sync gets applied. Sync still stays queued while a transaction is `persisting`, exactly as today. That's the half of #1630 that turned out to be unsound without temp/server key mapping, and it's not needed for any of this.

So this is only the projection half — same base as before, just composed per-field instead of replaced wholesale.

Related to RFC #1625.

```ts
// three transactions in flight on the same row
t1.mutate(() => collection.update(1, (d) => { d.a = `a1` }))
t2.mutate(() => collection.update(1, (d) => { d.b = `b2` }))
t3.mutate(() => collection.update(1, (d) => { d.c = `c3` }))
t1.commit(); t2.commit(); t3.commit()

t2.rollback()
// expected: { a: 'a1', b: 'b0', c: 'c3' }
// actual:   { a: 'a1', b: 'b2', c: 'c3' }   <- b2 is back
```

t3's snapshot was taken when `b` was already `b2`, so once t2 is gone t3 puts it right back. I don't think there's a reading of that which is correct.

The fix: `update` mutations now compose `mutation.changes` over the value they were layered on, so a mutation only owns the fields it actually changed. Inserts keep using `modified`, since an insert's `changes` deliberately omit schema defaults (there's a comment in `mutations.ts` saying as much).

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed concurrent optimistic updates so each transaction changes only the fields it modified.
  - Prevented rollbacks from overwriting unrelated updates or leaving stale values visible.
  - Preserved optimistic insert defaults when subsequent updates are applied.
  - Prevented redundant change notifications when an optimistic overlay produces no effective change.

### DIFF
--- a/.changeset/soft-donkeys-shave.md
+++ b/.changeset/soft-donkeys-shave.md
@@ -1,0 +1,9 @@
+---
+'@tanstack/db': patch
+---
+
+Fix concurrent optimistic updates to the same row overwriting each other's fields.
+
+Each pending update was stored as a whole-row snapshot taken at `mutate()` time, so whichever transaction applied last set every field — including fields it never touched. The clearest symptom: rolling back one of several in-flight transactions left its change visible, restored from a sibling transaction's snapshot.
+
+Updates now apply only the top-level fields they actually changed. Inserts are unaffected — they still apply the whole row, because an insert's `changes` omit schema defaults.
--- a/packages/db/src/collection/state.ts
+++ b/packages/db/src/collection/state.ts
@@ -13,6 +13,7 @@ import type {
   ChangeMessage,
   CollectionConfig,
   OptimisticChangeMessage,
+  PendingMutation,
 } from '../types'
 import type { CollectionImpl } from './index.js'
 import type { CollectionLifecycleManager } from './lifecycle'
@@ -102,6 +103,11 @@ export class CollectionStateManager<
    */
   public pendingLocalChanges = new Set<TKey>()
   public pendingLocalOrigins = new Set<TKey>()
+
+  private composedUpserts = new WeakMap<
+    object,
+    { base: TOutput; composed: TOutput }
+  >()
 
   private virtualPropsCache = new WeakMap<
     object,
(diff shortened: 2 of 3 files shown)


MILLERMARRU (commented):
I checked collection/state.ts and confirmed both places that seed `optimisticUpserts` for an `update` mutation just did `mutation.modified as TOutput`, a full snapshot, not a diff. That's the exact bug: two transactions mutating different fields on the same row would fight over which snapshot wins the overlay, and rolling back one would blow away fields a sibling transaction owns, since there's no per-field ownership tracked anywhere. `resolveOptimisticUpsert` composing `{...base, ...mutation.changes}` fixes this cleanly. The caching by the `changes` object reference plus a `base === cached.base` check is a nice detail, it keeps object identity stable for an unchanged overlay so downstream consumers relying on reference equality (e.g. to suppress redundant update events) don't get spurious churn, which the last test in the suite explicitly checks. I'd double check the WeakMap doesn't leak in long-lived collections with high mutation churn, but since it's keyed by the `changes` object (per-mutation and short-lived), that seems fine.

KyleAMathews:
Closing as superseded by #1807. The current optimistic-update contract intentionally uses captured whole-row snapshots, with oracle coverage for that behavior; applying this PR now would create hybrid rows that neither the mutation nor the server produced. RFC #1625 remains the place for any future redesign of reconciliation semantics. Thanks for working through this.

PR #1705, opened 2026-07-29
fix: correlated subquery resolves empty when correlation targets a joined alias

### LINKED ISSUE #1704: `materialize()` subquery resolves to `[]` when its correlation predicate targets a joined alias
- [x] I've validated the bug against the latest version of DB packages (`@tanstack/db@0.6.17`)

**Describe the bug**

A `materialize()` correlated subquery silently resolves to an empty array when its correlation
predicate references a **joined** alias instead of the subquery's own `from` alias.

```js
// ❌ always [] — correlation reaches `part.id` through the joined `order` alias
materialize(
  q.from({ production: productions })
   .innerJoin({ order: orders }, ({ production, order }) => eq(production.orderId, order.id))
   .where(({ order }) => eq(order.partId, part.id))
)

// ✅ works — correlation targets the subquery's own `from` alias
materialize(
  q.from({ order: orders }).where(({ order }) => eq(order.partId, part.id))
)
```

There is no error or warning — the field is just permanently empty, so the failure is silent data
loss. In our app this quietly dropped two whole relations out of a stock ledger, and it went
unnoticed because the surrounding relations populated correctly.

**To Reproduce**

1. `npm i @tanstack/db@0.6.17`
2. Save as `repro.cjs`:

```js
const {
  createCollection, createLiveQueryCollection, eq, localOnlyCollectionOptions, materialize,
} = require("@tanstack/db");

const mk = (id, initialData) =>
  createCollection(localOnlyCollectionOptions({ id, getKey: (r) => r.id, initialData }));

const parts       = mk("part",       [{ id: 1 }]);
const orders      = mk("order",      [{ id: 1, partId: 1 }]);
const productions = mk("production", [{ id: 1, orderId: 1 }]);

const live = createLiveQueryCollection({
  startSync: true,
  query: (q) =>
    q.from({ part: parts })
      .select(({ part }) => ({
        ...part,

        // CONTROL — correlation references this subquery's own `from` alias.
        control: materialize(
          q.from({ order: orders }).where(({ order }) => eq(order.partId, part.id)),
        ),

        // BUG — identical correlation, but reached through a joined alias.
        broken: materialize(
          q.from({ production: productions })
            .innerJoin({ order: orders }, ({ production, order }) =>
              eq(production.orderId, order.id))
            .where(({ order }) => eq(order.partId, part.id))
            .select(({ production }) => ({ ...production })),
        ),
      }))
      .findOne(),
});

setTimeout(() => {
  const row = Array.from(live.entries(), ([, v]) => v)[0];
  console.log("control (expect 1):", row?.control?.length ?? 0);
  console.log("broken  (expect 1):", row?.broken?.length ?? 0);
}, 100);
```

3. `node repro.cjs`

```
control (expect 1): 1
broken  (expect 1): 0
```

**Expected behavior**

`broken` should contain the single `production` row, the same as running that join + `where`
outside `materialize()`.

**Additional context**

- **The join and predicate are correct in isolation.** The exact same
  `from(production).innerJoin(order).where(order.partId === 1)` as a plain, uncorrelated live query
  returns 1 row. Only the correlated `materialize()` form returns `[]`.
- **Not an indexing artifact.** Adding the indexes suggested by the
  `"Join requires an index on \"orderId\""` warning (`production.orderId`, `order.partId`,
  `order.id`, with `defaultIndexType: BasicIndex`) produces the identical `broken: 0`.
- **Not a sync-timing artifact.** All rows are present before the query is built, `localOnly`
  collections, and the result is stable indefinitely — it never converges.
- Reproduced first against Electric-backed collections in a real app, then reduced to the
  `localOnly` case above.
- Possibly related to the lazy-target/alias resolution that RFC #1658 (PR 2) restructures — the
  correlation target here is resolved through a joined alias — but this failure mode isn't among
  the issues that RFC enumerates.

**Environment**

- `@tanstack/db@0.6.17` (current latest), Node v24.14.0
- Plain Node script, no framework adapter or browser involved

Mohith26:
Fixes #1704, reported by @Jaime02.

In `compileQuery`, the includes-child path re-keys the raw main-source input by `childCorrelationField.path.slice(1)`, assuming the correlation ref lives on the subquery's own `from` alias. When it references a joined alias instead (`eq(order.partId, part.id)` with `from(production).innerJoin(order)`), the nested lookup returns `undefined`, so the inner join with parent keys drops every row: the subquery resolves to `[]` permanently, with no error.

The fix defers the parent-key filter when the correlation alias is not the main source: the identical inner join is applied after `processJoins` against the namespaced pipeline using the compiled correlation expression, tagging `__correlationKey` on the main-source namespace and merging parent context so group-by, order-by, and output routing consume it unchanged. Lazy loading needed no change (`followRef` already resolves joined aliases). Patch changeset included.

Two regression tests cover initial resolution, incremental insert/delete propagation, the issue's exact spread-select shape, and a from-alias-with-join control; both fail without the fix. Package suite: 2485 passed, zero new failures; eslint, build, and vitest typecheck clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `materialize()` returning empty results when correlation conditions referenced fields from joined data.
  * Ensured correlated results remain accurate after incremental inserts and deletes.
  * Improved support for queries that select fields from joined sources while preserving parent-child relationships.

### DIFF
--- a/.changeset/plenty-hoops-shake.md
+++ b/.changeset/plenty-hoops-shake.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/db': patch
+---
+
+Fix: `materialize()` correlated subquery silently resolves to an empty array when the correlation predicate references a joined alias instead of the subquery's own `from` alias. The parent-key filter is now deferred until after joins when the correlation field lives on a joined source.
--- a/packages/db/src/query/compiler/index.ts
+++ b/packages/db/src/query/compiler/index.ts
@@ -243,7 +243,20 @@ export function compileQuery(
   // The inner join happens BEFORE namespace wrapping / WHERE / SELECT / ORDER BY,
   // so the child pipeline only processes rows that match parents.
   let pipeline: NamespacedAndKeyedStream = initialPipeline
-  if (!isUnionFrom && parentKeyStream && childCorrelationField) {
+  // When the correlation field references a joined alias (not the main from
+  // alias), the correlation value isn't available on the raw input rows, so
+  // the parent-key filter must be deferred until after joins are processed.
+  const correlationOnJoinedAlias: boolean =
+    !isUnionFrom &&
+    !!parentKeyStream &&
+    !!childCorrelationField &&
+    childCorrelationField.path[0] !== mainSource
+  if (
+    !isUnionFrom &&
+    parentKeyStream &&
+    childCorrelationField &&
+    !correlationOnJoinedAlias
+  ) {
     const mainInput = sources[mainSource]!
     let filteredMainInput = mainInput
     // Re-key child input by correlation field: [correlationValue, [childKey, childRow]]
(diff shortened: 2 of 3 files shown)


Jaime02:
Thank you so much for the contribution, but I guess that this PR is outdated and should be closed, see #1740

KyleAMathews:
Yup, thanks @Mohith26 !

Issue #1658, opened 2026-07-08
RFC: Stabilizing includes / nested materialization


KyleAMathews:
# RFC: Stabilizing includes / nested materialization

**Status**: production implementation complete in PR #1740; awaiting merge
**Scope**: bug fixes and internal refactors only — no new public API surface, no behavior changes beyond fixing verified bugs.
**Working model**: one coherent D2 graph implementation, backed by direct oracle assertions and narrow boundary adapters.

**Progress**:

- [x] Base recompute-oracle harness merged in #1716.
- [x] Oracle hardening merged in #1717: assertion-specific known failures, stable converted puts, and repeated optimistic rollback coverage.
- [x] Reusable trace runner merged in #1718: driver/projection split, same-turn checkpoints, cleanup-error preservation, synchronous assertion typing, and a full-row batch driver.
- [x] The expanded scalar-materialization driver found a new deterministic stale nested-reference failure; its reduced expected-failure seed landed in #1718.
- [x] State-equivalence coverage expansion merged in #1719: generated full-row batches at depths 1–4, guaranteed connected paths, reorder and single-row rekey/reparent coverage, flat array/concat projections, and three reduced expected-failure classes.
- [x] Controlled readiness/timing drivers merged in #1720, including exact expected-failure checkpoints and obsolete-demand coverage.
- [x] State-aware relationship-transition fuzzing merged in #1722: transition-only and stateful paired cases across every depth/target level, connected scalar activity, explicit key-separation invariants, and reduced expected-failure traces for every observed rekey and reparented-subtree boundary.
- [x] Audited every issue and PR opened since this RFC on 2026-07-08; added the relevant reports below and recorded adjacent work owned by other RFCs.
- [x] Shared observer #1642, layout revision #1669, and window controller #1675 merged in the parallel live-query platform track; use those contracts for publication/revision coverage rather than adding adapter-local state.
- [x] Query-shape oracle coverage merged in #1725: joined-alias correlation (#1704), initially-null correlation plus transition controls (#1706), and duplicate-key join multiplicity (#1703).
- [x] Transition-history oracle merged in #1733: ordered relationship histories, retired-route reuse, shared-route subscriber lifetime, moved-child replacement, delivery-order mirrors, atomic controls, and exact structural failure classifiers.
[...]

PR #1707, opened 2026-07-30
fix: resolve materialize(findOne()) to undefined when the correlation key is null

### LINKED ISSUE #1706: `materialize(q…findOne())` resolves to `null`, not `undefined`, when the correlation key is itself null
**Describe the bug**

`materialize(q…findOne())` is documented and typed as resolving to `T | undefined`. It does — but only when the correlated FK holds a value that matches no child row.

When the FK column is itself `null`, the field resolves to **`null`** instead. That value is outside the inferred type, so every `=== undefined` / `!== undefined` guard silently falls through and the next property access throws.

The FK-set-but-unmatched path was tested when `materialize()` landed (#1504 lists "missing-match → `undefined`" among its runtime tests). The FK-is-null path never reaches `materializeIncludedValue` at all, so it returns the raw projection placeholder.

**To Reproduce**

Two collections, one live query, no framework adapter. Three posts isolate the three paths:

```ts
import {
  createCollection,
  createLiveQueryCollection,
  eq,
  localOnlyCollectionOptions,
  materialize,
} from "@tanstack/db";

type TAuthor = { id: number; name: string };
type TPost = { id: number; title: string; authorId: number | null };

const authorCollection = createCollection(
  localOnlyCollectionOptions<TAuthor>({
    getKey: (author) => author.id,
    initialData: [{ id: 1, name: "Ada" }],
  }),
);

const postCollection = createCollection(
  localOnlyCollectionOptions<TPost>({
    getKey: (post) => post.id,
    initialData: [
      { id: 1, title: "fk matches an author", authorId: 1 },
      { id: 2, title: "fk set, but no author matches", authorId: 999 },
      { id: 3, title: "fk is null", authorId: null },
    ],
  }),
);

const postsWithAuthor = createLiveQueryCollection((q) =>
  q.from({ post: postCollection }).select(({ post }) => ({
    ...post,
    author: materialize(
      q
        .from({ author: authorCollection })
        .where(({ author }) => eq(author.id, post.authorId))
        .findOne(),
    ),
  })),
);

await postsWithAuthor.preload();

for (const row of postsWithAuthor.toArray) {
  console.log(`post ${row.id} (${row.title})`);
  console.log(`  author = ${JSON.stringify(row.author)}`);
  console.log(`  typeof = ${row.author === null ? "null" : typeof row.author}`);  // This null check shows a type error, but during runtime, row.author may be null
}

// Inferred type is `TAuthor | undefined`. `null` is not part of it, so this
// annotation compiles cleanly even though row 3 makes it false at runtime.
const nullableAuthor: TAuthor | undefined = postsWithAuthor.toArray[2]?.author;
console.log(`row 3 typed \`TAuthor | undefined\`, actual: ${JSON.stringify(nullableAuthor)}`);
```

Output:

```
post 1 (fk matches an author)
  author = {"id":1,"name":"Ada","$synced":true,"$origin":"local","$key":1,"$collectionId":"…"}
  typeof = object
post 2 (fk set, but no author matches)
  author = undefined
  typeof = undefined
post 3 (fk is null)
  author = null          <-- type says this is impossible
  typeof = null
row 3 typed `TAuthor | undefined`, actual: null
```

Post 2 confirms the documented behaviour still works, so this is specifically the null-correlation-key path and not a general "empty result" problem.

**Expected behavior**

Post 3 resolves to `undefined`, matching both the JSDoc on `materialize()` ("the parent receives a single `T | undefined` value — `undefined` when no child matches") and the inferred `ResultTypeFromSelect` output.

**Notes / related**

- Belongs to the includes cluster tracked by RFC #1658. Worth noting that its table lists #1571 part 2 ("initially-empty include starts `null` and never becomes reactive") as **Not reproduced** — that verification used a bare subquery-in-select, which yields an empty child `Collection`, rather than `materialize(…findOne())` with a null FK. The `null` symptom is real on this path.
- #1467 is the same failure shape one branch over: `ResultTypeFromSelect` typing a `leftJoin` whole-object `Ref` as non-nullable when the runtime value is `undefined`. There the fix is in the types; here the types are already correct.

**Environment**

- `@tanstack/db` 0.6.17
- `@tanstack/react-db` 0.1.95 (repro above uses `@tanstack/db` directly; no adapter involved)
- TypeScript 7.0.2, `strictNullChecks: true`
- Bun, Windows 11

Mohith26:
Fixes #1706, reported by @Jaime02.

`materialize(q...findOne())` returned `null` instead of the typed `undefined` whenever the correlated FK column was itself null: the compiler replaces the include in select with a literal `null` placeholder, and Phase 1 of the includes flush only handled `correlationKey != null`, so null-FK parents never reached `materializeIncludedValue` and the raw placeholder leaked out, violating the `T | undefined` contract in the types.

The fix resolves null-correlation-key fields eagerly to their empty materialized value at parent-insert time (`undefined` for findOne, `[]` for arrays, `""` for concat), with a `skipped` routing flag so conditional-select branches that do not own the field are untouched. Bare non-materialized includes with null FK intentionally keep the placeholder pending RFC #1658.

Independent of #1705 (different region of the compiler plus the live-collection routing; reproduced on plain main; no overlapping hunks). Patch changeset included. Three regression tests (the issue's three-post repro with an FK-set-but-unmatched control, array shape, live FK-update-to-null transition) fail without the fix; full suite 2486 passed, zero new failures; eslint/prettier/typecheck clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed materialized query results when correlation keys are `null`.
  * `findOne()` now consistently returns `undefined`, array queries return `[]`, and concatenated results return an empty value.
  * Corrected updates when a correlation key changes from a valid value to `null`.
  * Prevented placeholder values from appearing in included results.

### DIFF
--- a/.changeset/shiny-donkeys-repeat.md
+++ b/.changeset/shiny-donkeys-repeat.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/db': patch
+---
+
+Fix: `materialize(q…findOne())` resolved to `null` instead of `undefined` when the correlation key itself was null. A null correlation key never reached the includes materialization step, so the compiler's `null` select placeholder leaked into the result, violating the typed `T | undefined` contract. Null correlation keys now resolve to the empty materialized value (`undefined` for `findOne()`, `[]` for arrays, `` for `concat`).
--- a/packages/db/src/query/compiler/index.ts
+++ b/packages/db/src/query/compiler/index.ts
@@ -346,6 +346,12 @@ export function compileQuery(
     getRouting: (nsRow: any) => {
       correlationKey: unknown
       parentContext: Record<string, any> | null
+      /**
+       * Set when this row does not participate in the include at all
+       * (conditional-select guards failed or no routing info was found),
+       * as opposed to participating with a null correlation key.
+       */
+      skipped?: boolean
     }
   }> = []
   for (const { sourceAlias, include } of sourceIncludes) {
@@ -388,12 +394,13 @@ export function compileQuery(
         fieldName,
         getRouting: (nsRow: any) => {
           if (!matchesConditionalSelectGuards(compiledGuards, nsRow)) {
-            return { correlationKey: null, parentContext: null }
+            return { correlationKey: null, parentContext: null, skipped: true }
           }
           return (
             nsRow[sourceAlias]?.[INCLUDES_ROUTING]?.[include.fieldName] ?? {
               correlationKey: null,
               parentContext: null,
+              skipped: true,
             }
           )
         },
@@ -427,12 +434,17 @@ export function compileQuery(
           fieldName,
           getRouting: (nsRow: any) => {
             if (!matchesConditionalSelectGuards(compiledGuards, nsRow)) {
-              return { correlationKey: null, parentContext: null }
+              return {
+                correlationKey: null,
+                parentContext: null,
+                skipped: true,
+              }
             }
             return (
               nsRow[INCLUDES_ROUTING]?.[include.fieldName] ?? {
                 correlationKey: null,
                 parentContext: null,
+                skipped: true,
               }
             )
           },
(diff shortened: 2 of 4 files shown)


Jaime02:
Thank you so much for the contribution, but I guess that this PR is outdated and should be closed, see https://github.com/TanStack/db/pull/1740

PR #1721, opened 2026-08-12
fix(react-db): idiomatic `useLiveQuery` & `useLiveSuspenseQuery` hooks

zeorin:
## 🎯 Changes

These hooks where using refs to track previous versions of certain variables, and reading those during render, which is an anti-pattern that breaks the Rules of Hooks and may lead to subtle bugs, especially in concurrent mode.

Using state instead is more idiomatic and ensures there will be no state tearing, even during concurrent mode updates.

I discovered this while trying to work around an issue where an error thrown by the `queryFn` in `QueryCollection` would lead to infinite retries, and eventually, the browser running out of memory.

TODO: I haven't yet checked or adjusted the following hooks:
 - `useLiveInfiniteQuery`
 - `useLiveQueryEffect`
 - `usePacedMutations`

## ✅ Checklist

- [x] I have tested this code locally with `pnpm test`. There are failing tests, but they were not introduced by my changes.

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved live query state handling for more reliable updates during concurrent rendering.
  * Fixed collection changes and Suspense readiness transitions to update consistently.
  * Improved promise reuse during Suspense preloading.
  * Stabilized live query subscriptions and returned results to reduce unnecessary updates.
  * Improved consistency when live query inputs change, helping prevent stale or mismatched results.

### DIFF
--- a/.changeset/eleven-gifts-shine.md
+++ b/.changeset/eleven-gifts-shine.md
@@ -1,0 +1,12 @@
+---
+'@tanstack/react-db': patch
+---
+
+fix(react-db): Avoid using refs to track changes in React hooks.
+
+Using refs to track previous versions of variables, and reading those refs
+during render, is an anti-pattern that breaks the Rules of Hooks and may lead to
+subtle bugs, especially in concurrent mode.
+
+Using state instead is more idiomatic and ensures there will be no state
+tearing, even during concurrent mode updates.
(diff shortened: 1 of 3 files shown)


KyleAMathews:
Closing this stale implementation because it predates the current DbClient, query identity, SSR hydration, deferred-source, and external-store architecture. The general render-purity concern can remain a design topic, but any implementation now needs a current-main reproduction and the modern server-snapshot and lifecycle contracts.

PR #1734, opened 2026-08-14
test(db): fuzz includes route lifecycles

KyleAMathews:
This extends the includes recompute oracle with validated route-lifecycle histories across topology, destination reuse, unsubscribe, and resubscribe cases. It also pins two root-only bugs: entering an already-live shared route can miss its existing ordered snapshot, while restoring a retired shared route can leak its updated snapshot to a root that remains departed.

## Approach

The scenario builder uses explicit transition descriptors. Rekeys may target `fresh`, `restore`, `merge`, `split`, or `retired` routes; reparents are restricted to moving a contribution into a different live route. The builder replays model state while assembling each trace and validates every lifecycle label against live routes, seen routes, retired owners, and current subscribers.

Two bounded FastCheck matrices cover:

- **Transition topology:** ancestor→descendant, descendant→ancestor, true siblings under one parent, cross-branch, and root transitions.
- **Destination lifecycle:** fresh, restore, merge→split, and retired-route reuse.

Every transition must change the recomputed relationship projection. Negative tests also reject mislabeled descriptors, including subscriber-lifecycle labels on reparent transitions.

## Newly classified failure

The merge-into-live-route matrix tests both entering rows at parent levels 0, 1, and 2. Only roots currently fail; nested subscribers receive the existing snapshot and serve as green controls.

The expected-failure classifier compares the complete actual result with recomputation minus exactly the missing subtree. It therefore rejects unrelated scalar, ordering, descendant, or sibling corruption. A separate ordered two-child case proves the root misses the entire route snapshot, not merely one child.

## Lifecycle controls

- The last subscriber can leave a shared route without corrupting later child updates at parent levels 0–2.
- A row can retire a route, allow its child to update while unsubscribed, and restore the current snapshot at parent levels 0–2. Root restoration is an exact expected failure because it also publishes that snapshot to the departed root; nested levels are green controls.
- The topology fixtures prove that sibling targets share a parent and that the descendant remains attached before a descendant→ancestor sequence moves its ancestor.

## Key invariants

- Descriptor labels match the route state at the transition.
- Subscriber lifecycle strategies cannot be applied to contribution-only reparent transitions.
- Generated transitions visibly change the recomputed relationship tree.
- Expected failures accept only the exact missing-snapshot shape at the exact checkpoint.
- Source rows and the recompute model remain independent.
- Fixed seeds and bounded values keep failures reproducible and CI cost small.

## Non-goals

- Fixing the newly found runtime bug.
- Changing query execution, subscription ownership, or publication code.
- Expanding destination histories across every parent level or adding the broader batch-shape matrix in this PR.

## Trade-offs

The descriptor builder tracks more model metadata than hand-written traces, but it makes lifecycle labels enforceable and exposes invalid matrix cells early. The property runs stay small because the explicit matrices provide structural coverage while fixed seeds preserve reproducibility.

## Verification

```bash
pnpm exec vitest run packages/db/tests/query/includes-oracle.property.test.ts --maxWorkers=2
pnpm exec tsc --noEmit -p packages/db/tsconfig.json
pnpm exec eslint packages/db/tests/query/includes-oracle.property.test.ts
pnpm exec prettier --check packages/db/tests/query/includes-oracle.property.test.ts
git diff --check origin/main...HEAD
```

The merged targeted oracle suite passes all 136 tests, including exact expected-failure classifiers.

## Files changed

- `packages/db/tests/query/includes-oracle.property.test.ts` — adds route descriptors and validation, topology and lifecycle matrices, exact snapshot-failure classifiers, and adjacent green controls.

---

Refs #1658

## Summary by CodeRabbit

- **Tests**
  - Expanded automated coverage for relationship queries and shared-route behavior.
  - Added scenarios for route creation, merging, splitting, retirement, reuse, resubscription, and subscriber departure.
  - Added validation for topology changes, ordering, materialization, transition visibility, and invalid strategies.
  - Added regression checks to detect corrupted or unexpected query results and verify route lifecycle consistency.

### DIFF
--- a/packages/db/tests/query/includes-oracle.property.test.ts
+++ b/packages/db/tests/query/includes-oracle.property.test.ts
@@ -101,17 +101,22 @@ function hasDirectChild(value: unknown, parentId: number, childId: number) {
   )
 }
 
-function classifyUnexpectedSharedChild(
+function classifyUnexpectedSharedRoute(
   { actual, expected }: AssertionDifference,
   unexpectedParentId: number,
   expectedParentId: number,
   childId: number,
 ): boolean {
+  const expectedParent = findRelationshipNode(expected, expectedParentId)
+  const expectedWithSharedRoute = replaceDirectChildren(
+    expected,
+    unexpectedParentId,
+    Array.isArray(expectedParent?.children) ? expectedParent.children : [],
+  )
   return (
-    hasDirectChild(actual, unexpectedParentId, childId) &&
     !hasDirectChild(expected, unexpectedParentId, childId) &&
-    hasDirectChild(actual, expectedParentId, childId) &&
-    hasDirectChild(expected, expectedParentId, childId)
+    hasDirectChild(expected, expectedParentId, childId) &&
+    isDeepStrictEqual(actual, expectedWithSharedRoute)
   )
 }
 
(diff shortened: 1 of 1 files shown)


tannerlinsley (approved):

PR #1735, opened 2026-08-14
test(db): cover optimistic includes relationships

KyleAMathews:
This adds a recompute oracle for optimistic `includes` relationship transitions. It proves rollback and confirmation histories converge to authoritative state, and it pins the existing deep-rekey defect when the same transition is produced optimistically.

## Approach

The driver maintains two independent views of state:

- cloned collection rows feed the live incremental query;
- an authoritative model plus pending optimistic overlays feeds fresh recomputation.

Trace actions start optimistic reparent/rekey updates, confirm them with controlled sync delivery, reject them for rollback, and deliver later descendant sync changes. Every checkpoint compares the live nested result with recomputation.

The histories cover:

- immediate optimistic rekey;
- rollback after rekey;
- an optimistic ancestor reparent plus descendant update and both rollback orders while the other mutation remains pending;
- same-route and different-route authoritative confirmation;
- a sibling entering the old route before rollback;
- repeated rollback and confirmation cycles;
- descendant updates after settlement to prove routing remains reactive.

## Known failure

An optimistic level-1 rekey with deeper descendants reproduces the existing deep-rekey defect: row `11` retains child `21` even though recomputation makes that child globally unreachable. The expected-failure classifier reconstructs the complete defective output—row `21` retained with an empty descendant set—and rejects any collateral scalar, ordering, sibling, or subtree corruption.

This is another producer path for an already cataloged bug, not a new defect class.

## Key invariants

- Source and oracle rows never share object identity.
- Recompute overlays pending optimistic rows on authoritative state without reading the live query.
- A confirmation must use the pending row's ID.
- The sync mock allows at most one pending mutation per collection level; duplicate handles and same-level overlap are rejected explicitly.
- Cleanup rejects any transaction left pending by an expected-failure checkpoint.
- Post-settlement descendant updates must still follow the settled route.

## Non-goals

- Fixing the deep-rekey runtime defect.
- Modeling concurrent same-level mutations, which the shared sync mock cannot settle independently.
- Changing production query, transaction, or synchronization code.

## Trade-offs

The same-level guard narrows the generated histories to what the current mock can represent faithfully. Different levels can still remain pending together, which covers compound ancestor/descendant rollback without introducing harness-made settlement behavior.

## Verification

```bash
pnpm exec vitest run packages/db/tests/query/includes-optimistic-oracle.property.test.ts packages/db/tests/expected-failure.test.ts --maxWorkers=2
pnpm exec tsc --noEmit -p packages/db/tsconfig.json
pnpm exec eslint packages/db/tests/query/includes-optimistic-oracle.property.test.ts
pnpm exec prettier --check packages/db/tests/query/includes-optimistic-oracle.property.test.ts
git diff --check origin/main...HEAD
```

The focused suites pass all 17 tests.

## Files changed

- `packages/db/tests/query/includes-optimistic-oracle.property.test.ts` — adds the optimistic trace driver, independent overlay recomputation, classified known failure, settlement histories, and harness guards.

---

Refs #1658

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive property-based coverage for optimistic relationship updates across nested live queries.
  * Validated rollback and confirmation behavior, including reparenting, rekeying, descendant updates, sibling insertion, and repeated operation histories.
  * Added checks for invalid operation handles and confirmations across matching and differing routes.

### DIFF
--- a/packages/db/tests/query/includes-optimistic-oracle.property.test.ts
+++ b/packages/db/tests/query/includes-optimistic-oracle.property.test.ts
@@ -74,7 +74,7 @@ type OracleNode = RootRow & {
   children?: Array<OracleNode>
 }
 
-type RelationshipNode = {
+type RelationshipNode = Record<string, unknown> & {
   id: number
   children?: unknown
 }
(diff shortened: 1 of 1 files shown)


tannerlinsley (approved):
Reviewed locally; focused oracle suites and type checks pass.

Issue #1713, opened 2026-08-04
Updating a parent source row empties `materialize()` fields in live queries


samvantoever:
- [x] I've validated the bug against the latest version of DB packages (`@tanstack/db@0.6.17`)

**Describe the bug**

Setup:

1. A live query collection (**Q1**) selects a `materialize()` include, correlated on a field of its source rows — e.g. `subItems: materialize(q.from({ sub }).where(({ sub }) => eq(sub.itemId, item.id)))`.
2. A second live query (**Q2**) is created *from* Q1 — `q.from({ row: Q1 })` — for example to add a `.where`/`.orderBy` per component.

Bug: when a row in Q1's **parent source** (`items`) is updated — any field, even one unrelated to the correlation — Q2's copy of that row comes out with the materialized field set to `null`. The child rows still exist and still match the correlation; nothing about `subItems` changed. Q1 itself keeps reporting the correct array for the same row.

The `null` is persisted in Q2's collection state, so it is not a render-timing issue — it stays wrong indefinitely. It only recovers when one of that row's *child* rows actually changes later.

When it triggers:

- any update of a parent source row: plain synced write, optimistic `collection.update()`, or the confirm write after an optimistic transaction
- with or without joins in Q1, with or without `.where`/`.orderBy`/`select` in Q2
- all `materialize()` fields of the affected row null out together (e.g. two sibling includes)

Practical impact: in our app, a shared pre-joined view with two `materialize()` fields is consumed by per-section `useLiveQuery(q => q.from({ row: view }).where(...).orderBy(...))` components. Dragging an item (which just patches an ordering field on the parent row) blanks all of that item's materialized children in the UI, and they stay blank until the user happens to edit one of them.

**To Reproduce**

Self-contained — local-only collections, no backend, no joins, no optimistic transaction. Paste into any React app and render `<MaterializeIncludeDropRepro />`, then click **Rename parent item** once.

```tsx
import {
  createCollection,
  eq,
  localOnlyCollectionOptions,
  materialize,
} from '@tanstack/db'
import { createLiveQueryCollection, useLiveQuery } from '@tanstack/react-db'

type Item = { id: string; name: string }
type SubItem = { id: string; itemId: string; label: string }

const items = createCollection(
  localOnlyCollectionOptions<Item, string>({ id: 'items', getKey: (r) => r.id })
)
const subItems = createCollection(
  localOnlyCollectionOptions<SubItem, string>({
    id: 'subItems',
    getKey: (r) => r.id,
[...]

PR #1736, opened 2026-08-14
test(db): cover layered includes publication

KyleAMathews:
## Summary

Adds a layered-query oracle for the publication bug in #1713. The driver compares Q1 and a Q2 built from Q1 against an independent recomputation after each controlled source change.

The expected-failure classifier is narrow: Q1 must be exactly correct, while Q2 must be exactly correct except that both sibling `materialize()` fields are `null`.

## Coverage

- two sibling materializations, so the oracle proves that all includes drop together
- Q1 with and without an inner join
- Q2 pass-through, `where`, `orderBy`, and `select` pipelines
- synced parent updates across the full Q1 × Q2 shape matrix
- optimistic publication before confirmation and after confirmation across the full shape matrix
- the documented parent-update failure followed by a child update that repairs Q2
- optimistic rollback, same-route atomic replacement, and route-changing replacement histories
- green controls for child-only updates and parent route changes

This catalogs the existing #1713 publication defect; it did not find a separate bug class.

## Verification

- `pnpm exec vitest run packages/db/tests/query/includes-publication-oracle.test.ts packages/db/tests/expected-failure.test.ts --maxWorkers=2` — 43/43 tests pass
- `pnpm exec tsc --noEmit -p packages/db/tsconfig.json`
- `pnpm exec eslint packages/db/tests/query/includes-publication-oracle.test.ts`
- `pnpm exec prettier --check packages/db/tests/query/includes-publication-oracle.test.ts`
- `git diff --check`

Test-only change; no changeset is needed.

Refs #1658
Refs #1713

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for layered live queries across parent, child, secondary-child, and metadata relationships.
  * Validated results for direct and joined queries through scalar updates, related-record changes, route changes, replacements, optimistic confirmations, and rollbacks.
  * Added property-based scenarios covering expected outcomes and known materialization failures to improve confidence in live query behavior.

### DIFF
--- a/packages/db/tests/query/includes-publication-oracle.test.ts
+++ b/packages/db/tests/query/includes-publication-oracle.test.ts
@@ -258,7 +258,12 @@ const publicationProjection: TraceProjection<
 }
 
 function sameValue(left: unknown, right: unknown): boolean {
-  return JSON.stringify(left) === JSON.stringify(right)
+  try {
+    expect(left).toEqual(right)
+    return true
+  } catch {
+    return false
+  }
 }
 
 // #1713 is specifically publication of Q1's compiled null placeholder to Q2:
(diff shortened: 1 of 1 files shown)


tannerlinsley (approved):

Issue #1631, opened 2026-06-30
Ref count misassumption


willbowdler-cc:
- [x] I've validated the bug against the latest version of DB packages

**Describe the bug**

We are occasionally seeing our eagerly loaded collections cleared, but they will retain a `ready` state. 

**To Reproduce**
I'm new to the codebase, so I don't have a comprehensive understanding of how everything works. This has been intermittent for us, but I believe what is happening is:

1. Subscribers to the underlying collection query unsub to where there are no subs
2. `cleanupQueryIfIdle` runs and sees that there is still a `refcount` of `1` since when you create an `eager` collection this code runs. 
    ```
        queryRefCounts.set(
          hashedQueryKey,
          (queryRefCounts.get(hashedQueryKey) || 0) + 1,
        )
    ```
3. Since there is a ref count and no listeners, db is removing the data from the collection under the assumption that this will prevent memory leaks. 
> Even if refcount > 0, we treat hasListeners as authoritative to prevent leaks. This can happen if subscriptions are GC'd without calling unloadSubset.

But refcount can also be 1 if you created the collection as an eager collection

4. But for us this somehow leads to `collection.status = 'ready'` and `collection.size = 0`

**Expected behavior**

I don't expect the collection to clear like this after sitting idle

**Screenshots**

OS: [e.g. iOS] Tahoe 26.5
[Brave 1.90.124 (Official Build) (arm64)](https://brave.com/latest/)
Chromium: 148.0.7778.179
Smartphone (please complete the following information):
N/A
Additional context

If it would be helpful, I had AI write up an explanation as well of what the suspected issue is. This may make more sense
https://gist.github.com/willbowdler-cc/3448aba0e8fd554f6c33c5fdd7cd80e1

codecraf8:
Confirming this against `@tanstack/query-db-collection@1.2.7`, with a deterministic repro — it isn't intermittent, it's guaranteed for every `eager` collection. The only reason it looks intermittent is that it needs the `gcTime` timer to fire, and the default `gcTime` is 5 minutes.

The chain is exactly as described above:

1. `syncMode` defaults to `eager` (`query.ts:698`).
2. Eager creates the base query at startup via `createQueryFromOpts({})`, which sets **refcount = 1** (`query.ts:1389`).
[...]

Issue #1488, opened 2026-04-21
on-demand store entries are removed while queries still use them when reusing an existing observer


flybayer:
- [x] I've validated the bug against the latest version of DB packages

## Describe the bug

With `@tanstack/query-db-collection` in `syncMode: \"on-demand\"`, reusing an existing observer/query can return cached or success data without re-registering row ownership for that query.

Later, when another query for the same row goes inactive and is cleaned up, row-level cleanup can delete the row from the collection even though the reused query is still active and still reading that row.

This causes rows to disappear from the store incorrectly.

## Relevant source

The bug appears to be in the existing-observer reuse path in [`packages/query-db-collection/src/query.ts`](https://github.com/TanStack/db/blob/main/packages/query-db-collection/src/query.ts).

Relevant areas:

- [`createQueryFromOpts`](https://github.com/TanStack/db/blob/main/packages/query-db-collection/src/query.ts)
- [`applySuccessfulResult`](https://github.com/TanStack/db/blob/main/packages/query-db-collection/src/query.ts)
- [`cleanupQueryInternal`](https://github.com/TanStack/db/blob/main/packages/query-db-collection/src/query.ts)

## Why this seems to happen

In `createQueryFromOpts`, when an observer already exists for a `hashedQueryKey`, the code increments the refcount and returns early based on:

- `observer.getCurrentResult().isSuccess`
- or `queryClient.getQueryData(key) !== undefined`

But in those early-return paths, it does not run the normal success handling path that calls `applySuccessfulResult(...)`, so row ownership is not re-added for that query.

That means the query is active from the app's perspective, but `rowToQueries` still does not include it.

When a different query that previously owned the same row is cleaned up, `cleanupQueryInternal(...)` sees no remaining owners and deletes the row.

## Reproduction pattern

I hit this consistently with an on-demand `moduleVersions` collection **with `useLiveQuery` with `gcTime` set to a value. This bug happens when that `gcTime` is hit.**

Two different on-demand queries can reference the same row:

1. detail query
```ts
where(id == \"mver_a...\")
```

2. list query
```ts
where(id in [\"mver_a\", \"mver_b\", \"mver_c\"])
```

The same module version row can appear in both.

### Navigation sequence that reproduces it

1. Start on a detail/overview page that queries a specific row
2. Reload there
3. Navigate to a list page that loads a broader query including that same row
4. Navigate back to the detail/overview page

At this point, the detail query is hitting the existing-observer reuse path.

Then:

[...]

PR #1737, opened 2026-08-14
test(query-db): cover ownership lifecycles

KyleAMathews:
## Summary

Adds a controlled ownership/lifecycle oracle for `@tanstack/query-db-collection`. It observes public rows alongside the production query-to-row ownership maps and transactional synced metadata.

## Findings and coverage

- **#1631 remains reproducible.** When an eager collection loses its last listener and the backing query-cache entry is removed, the collection stays `ready` but loses its row and owner. The expected failure is pinned to the exact checkpoint, result shape, and invariant warning.
- **PR #1656's persisted-owner bug remains reproducible.** The real transactional metadata path loses the first inserted row's owner, then retains only the later query for the shared row while omitting the later inserted row's owner. Both checkpoints are classified as two surfaces of the same defect; later live ownership cleanup remains green.
- **#1488 does not reproduce on current `main`.** The oracle proves why: the final acquisition retires its observer and ownership atomically while the query cache stays warm, and reacquisition creates a cached observer that re-registers ownership.
- Green controls cover repeated acquisition refcounts, overlapping query owners, staggered cleanup, and live ownership after persisted metadata divergence.

The test delegates to production transaction metadata rather than replacing it, so metadata writes and row inserts share the real commit boundary.

## Verification

- `pnpm exec vitest run tests/ownership-lifecycle.oracle.test.ts --maxWorkers=2` from `packages/query-db-collection`
- package TypeScript check through Vitest
- `pnpm exec eslint packages/query-db-collection/tests/ownership-lifecycle.oracle.test.ts`
- `pnpm exec prettier --check packages/query-db-collection/tests/ownership-lifecycle.oracle.test.ts`
- `git diff --check`

Test-only change; no changeset is needed.

Refs #1488
Refs #1631
Refs #1656
Refs #1658

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for query collection ownership lifecycles.
  * Added scenarios for reused subsets, observer retirement, cached-data reacquisition, overlapping ownership boundaries, listener removal, and row insertion.
  * Added validation for persisted ownership, checkpoints, metadata writes, and classified lifecycle failures.

### DIFF
--- a/packages/query-db-collection/tests/ownership-lifecycle.oracle.test.ts
+++ b/packages/query-db-collection/tests/ownership-lifecycle.oracle.test.ts
@@ -1,0 +1,701 @@
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { QueryClient } from '@tanstack/query-core'
+import { createCollection, eq } from '@tanstack/db'
+import { expectAssertionFailure } from '../../db/tests/expected-failure.js'
+import { TraceAssertionError } from '../../db/tests/trace-runner.js'
+import { queryCollectionOptions } from '../src/query.js'
+import type { SyncMetadataApi } from '@tanstack/db'
+
+type Item = {
+  id: string
+  category: string
+  name: string
+}
+
+type OwnershipMaps = {
+  rowToQueries: Map<string | number, Set<string>>
+  queryToRows: Map<string, Set<string | number>>
+}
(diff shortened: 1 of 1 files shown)


tannerlinsley (approved):

PR #1739, opened 2026-08-14
test(db): fuzz includes relationship batch shapes

KyleAMathews:
This adds a recompute-oracle matrix for relationship replacements around moved include subtrees. It changes tests only; runtime behavior and public APIs stay unchanged.

## Approach

The property tests generate every valid combination of:

- split or atomic delivery
- delete-then-insert or insert-then-delete order
- reused or new public row IDs
- handed-off or fresh relationship routes
- route-only ancestor moves or compound route-and-position moves

Each candidate has an adjacent control that keeps the replacement history and position change but omits the route move. The matrix also proves that equivalent delivery boundaries and orderings converge on the same recomputed final state.

The expected-failure boundary is narrow: only split, delete-then-insert replacements with a new public ID on a handed-off route may omit the existing grandchild. The classifier compares the full result structurally after removing that exact child, so no second difference can hide behind the known defect.

## Key invariants

- Every valid matrix cell reaches the same recomputed final state for one fixture.
- Controls must always match the incremental engine.
- Expected failures must contain only the known missing grandchild.
- Generated row and route keys remain disjoint.
- Replacing a row and moving its ancestor may differ in delivery shape, but not in final semantics.

## Non-goals

This PR does not change runtime code or fix the known missing-grandchild defect. It also excludes same-ID insert-before-delete cases: inserting a duplicate public key before retiring the existing row is invalid input, not another delivery of the same final state.

## Trade-offs

The test spells out the matrix axes instead of sampling them from one opaque arbitrary. This adds some scaffolding, but gives each semantic dimension a stable name, makes exclusions explicit, and keeps the expected-failure classifier auditable.

## Verification

```bash
pnpm exec vitest run packages/db/tests/query/includes-oracle.property.test.ts --maxWorkers=2
pnpm exec tsc -p packages/db/tsconfig.json --noEmit
pnpm exec eslint packages/db/tests/query/includes-oracle.property.test.ts
pnpm exec prettier --check packages/db/tests/query/includes-oracle.property.test.ts
```

The focused suite passes all 103 tests. An extended campaign exercised 800 generated fixture bundles across the full matrix and found no additional bug class.

## Files changed

- `packages/db/tests/query/includes-oracle.property.test.ts` adds the batch-shape generators, adjacent controls, exact classifier, validity exclusion, and property matrix.

---
Refs #1658

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for relationship replacement scenarios.
  * Validated split and atomic updates, change ordering, reused and new identifiers, handoffs, and ancestor changes.
  * Added recursive direct-child removal checks and stricter mismatch classification.
  * Confirmed that missing replacement children are accepted only in the expected split handoff case.
  * Verified final recomputed states and controls across multiple scenario combinations.

### DIFF
--- a/packages/db/tests/query/includes-oracle.property.test.ts
+++ b/packages/db/tests/query/includes-oracle.property.test.ts
@@ -3253,6 +3253,7 @@ describe(`includes recompute oracle`, () => {
 
             // Delivery boundaries and change order must not alter the final
             // recompute semantics for one generated fixture.
+            expect(finalStates.length).toBeGreaterThan(1)
             for (const finalState of finalStates.slice(1)) {
               expect(finalState).toEqual(finalStates[0])
             }
--- a/packages/db/tests/query/includes-oracle.property.test.ts
+++ b/packages/db/tests/query/includes-oracle.property.test.ts
@@ -126,6 +126,55 @@ function classifyMissingReplacementChild(
   )
 }
 
+function removeDirectRelationshipChild(
+  value: unknown,
+  parentId: number,
+  childId: number,
+): unknown {
+  if (!Array.isArray(value)) return value
+
+  return value.map((entry) => {
+    if (!isRelationshipNode(entry) || !Array.isArray(entry.children)) {
+      return entry
+    }
+
+    return {
+      ...entry,
+      children:
+        entry.id === parentId
+          ? entry.children.filter(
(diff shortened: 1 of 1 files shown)


tannerlinsley (approved):
Reviewed locally; focused oracle suites and type checks pass.

PR #1744, opened 2026-08-18
fix(db): reclaim collections that sync before anything subscribes

### LINKED ISSUE #1178: react-db's useLiveQuery seems to leave an extra subscriber in StrictMode

edzis:
Fixes #1178. Root cause and cost measurements: [this comment](https://github.com/TanStack/db/issues/1178#issuecomment-5326592699).

In short: `startGCTimer()` is only called on the 1 → 0 subscriber edge, so a collection that starts syncing and never gains a subscriber never arms it. `useLiveQuery` starts sync from its own render body, so every discarded render strands one — no misuse required, and no way for an app to opt out. `preload()` in a route loader reaches the same state.

### Approach

Arm at `CollectionSyncManager.startSync()`. Every route into sync passes through it: the `startSync: true` constructor (`collection/index.ts:405`), `startSyncImmediate()` (`:493`), `preload()` (`sync.ts:429`), lifecycle restart (`lifecycle.ts:123`), `addSubscriber()` (`changes.ts:227`). One call site, every adapter.

- `sync.ts` calls a new `lifecycle.startGCTimerIfUnsubscribed()` once sync has started. No-op when a subscriber exists.
- `addSubscriber()` cancels a pending arm before starting sync, so a late subscriber keeps the collection.
- `cleanup-queue.ts`: `cancel()` clears the shared root timer when the queue drains. Pre-existing bug, invisible until many collections register.

### The render → commit gap

Adapters pass `gcTime: 1` for immediate unmount teardown. Reusing that for the unsubscribed arm would reclaim a collection before the commit that subscribes to it.

So a 50 ms floor (`UNSUBSCRIBED_GC_FLOOR_MS`) applies to that arm only. The 1 → 0 arm still fires at exactly `gcTime`, so unmount teardown is unchanged. In browsers the destructive step defers further through `safeRequestIdleCallback(…, { timeout: 1000 })`, which React's scheduling starves until the render burst ends.

Matches what `types.ts:546-557` documents: collected "when it has no active subscribers".

### Results

1000 never-subscribed live queries, forced GC:

| | before | after |
| --- | --- | --- |
| retained heap | 18.6 MB | 0.4 MB |
| source `subscriberCount` after `gcTime` | 1000 | 0 |
| 50 writes with 200 orphans, vs none | 253.1x | 0.6x |

App-level numbers are in the issue.

### Tests

- `db/tests/orphaned-live-query-gc.test.ts` — no React; 4 of 5 fail before.
- `react-db/tests/useLiveQuery.uncommitted-render.test.tsx` — suspend and throw arms; 2 of 3 fail before.
- `db/tests/unsubscribed-sync-gc.test.ts` — pins the design: a late subscriber still attaches, the 1 → 0 arm keeps `gcTime` unmodified, `preload()`ed collections get reclaimed.

Each guard checked against a deliberately broken build — floor at 0, floor leaked into the 1 → 0 arm, fix removed — and each fails a distinct test. Workspace green, 20 packages.

### Your call

Two behaviour changes:

- `preload()` without a subscriber is now reclaimed after `gcTime` (default 5 min) instead of living forever. Deliberate and asserted. Matches the contract, but affects `loader: () => collection.preload()` with a long gap before navigation.
- `gcTime: 0` opts out of the new arming, since `startGCTimer` returns early on `gcTime <= 0 || !Number.isFinite()` before the floor applies. Consistent with "GC disabled", and the reverse of what someone writing it to mean "don't retain this" expects.

Both go away if you'd rather narrow this to `createLiveQueryCollection` only. Happy to cut it that way.

### Docs (`de2c2c95`)

Two lines that describe the behaviour this PR changes:

- `meta-framework/SKILL.md` said an unpreloaded collection starts syncing on mount. It starts on first render, and mount is only where teardown is keyed. That gap is where this defect lives, so the line is worth correcting whether or not you take the code change.
- `live-queries/SKILL.md` said `gcTime: 0` disables collection for a derived collection. Complete before this PR, not after: `gcTime: 0` now also opts out of the arming above.

The second line also notes that `gcTime: 0` means the opposite in TanStack Query, which collects as soon as the query goes inactive. That inversion produced both of our occurrences — an internal note assumed the Query meaning and recommended `gcTime: 0` for one-shot reads, and two call sites followed it. Unlike a wrong `staleTime`, it fails silently and permanently.

Happy to split `de2c2c95` into its own PR to keep this one to the behaviour change. It's a single commit, so it cherry-picks cleanly.

Unrelated: `db-core`'s skill frontmatter reads `library_version: '0.6.17'` on a package publishing 0.7.2. Left it alone in case that belongs to your release tooling.

### Also in the diff

`react-db` unchanged. Moving construction off the render path would close the race the floor only mitigates, but it doesn't cover `preload()` (called from `useLiveSuspenseQuery`'s render body), `startSyncImmediate()`, direct `createLiveQueryCollection({ startSync: true })` or the vue/svelte/solid/angular adapters. Tried it: `react-db` goes 128/128 → 13 failures, all in `useLiveSuspenseQuery`.

Three assertions rescoped in `electric-db-collection/tests/electric.test.ts`. They asserted `vi.getTimerCount() === 0` while a `beforeEach` fixture holds an eagerly-syncing, never-subscribed collection, so they measured that ambient timer rather than the refresh timeout they name. Now they diff against a baseline captured before the operation. Fixture untouched, and removing the `clearTimeout` under test still fails one of the three, as before.

Node now keeps one pending `setTimeout` for up to `gcTime` per eagerly-synced collection. Single shared root timer, cleared when the queue drains. No `unref()`, since it isn't portable to browsers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of collections that begin syncing before any subscribers attach.
  * Unused collections are now garbage-collected after the configured retention period, with a 50 ms minimum grace period.
  * Garbage collection remains disabled when configured with `gcTime: 0`.
  * Cleanup timers are canceled correctly when collections become active again.
  * Collections can resynchronize correctly after being reclaimed.

* **Documentation**
  * Clarified live-query garbage-collection behavior and preload guidance.

### DIFF
--- a/.changeset/lucky-donkeys-repeat.md
+++ b/.changeset/lucky-donkeys-repeat.md
@@ -1,0 +1,11 @@
+---
+'@tanstack/db': patch
+---
+
+Garbage collect collections that start syncing before anything subscribes to them.
+
+`startGCTimer` only ran when the last subscriber left, so a collection whose subscriber count never rose above zero never armed it. Sync started from `startSync: true`, `preload()` or `startSyncImmediate()` therefore ran forever, and a live query started that way held a subscription on every collection it read from for the lifetime of the page — regardless of how short its `gcTime` was.
+
+Framework adapters build their live query collection while rendering and subscribe once that render commits, so every render React, Vue, Svelte or Solid discards before committing — a suspended subtree, a render that throws, a time-sliced render restarted by an interleaved update — stranded a fully compiled query graph rooted at a long-lived source collection. Under a route that repeatedly re-rendered without committing this exhausted the renderer's heap.
+
+Sync starting without a subscriber now arms the same GC timer, floored at 50ms so that a subscriber arriving with the commit is never beaten to it. Collections with `gcTime: 0` still opt out of GC entirely, and the timer armed when the last subscriber leaves still fires on `gcTime` exactly.
--- a/packages/db/skills/db-core/live-queries/SKILL.md
+++ b/packages/db/skills/db-core/live-queries/SKILL.md
@@ -203,7 +203,11 @@ const activeUserPosts = createLiveQueryCollection((q) =>
 Create derived collections once at module scope and reuse them. Do not recreate on every render or navigation.
 
 Live query collections default to `gcTime: 5_000`. An explicit `gcTime: 0` is
-preserved and disables garbage collection for that derived collection.
+preserved and disables garbage collection for that derived collection --
+including the reclamation of a collection that started syncing and never gained
+a subscriber. Note this is the opposite of `gcTime: 0` in TanStack Query, where
+it collects as soon as the query goes inactive; use a small positive value if
+you want prompt collection here.
 
 ## Virtual Properties
 
--- a/packages/db/skills/meta-framework/SKILL.md
+++ b/packages/db/skills/meta-framework/SKILL.md
@@ -373,7 +373,7 @@ export const Route = createFileRoute('/todos')({
 })
 ```
 
-Without preloading, the collection starts syncing only when the component mounts, causing a loading flash. Preloading in the route loader starts sync during navigation, making data available immediately when the component renders.
+Without preloading, the collection starts syncing when the component first renders, causing a loading flash. Preloading in the route loader starts sync during navigation, so the data is already there on that first render.
 
 ### MEDIUM Creating separate collection instances in one scope
 
--- a/packages/db/src/collection/cleanup-queue.ts
+++ b/packages/db/src/collection/cleanup-queue.ts
@@ -43,8 +43,8 @@ export class CleanupQueue {
   public cancel(key: unknown): void {
     this.tasks.delete(key)
 
-    // Draining the queue retires the root timer with it. Cancelling the
-    // earliest of several tasks leaves the timer where it is: it wakes early,
+    // Retire the root timer with the last task. A non-empty queue keeps its
+    // timer even when the cancelled task was the earliest: it wakes early,
     // finds nothing due and reschedules, which costs less than rescanning
     // every task on each cancellation.
     if (this.tasks.size === 0 && this.timeoutId !== null) {
--- a/packages/db/src/collection/cleanup-queue.ts
+++ b/packages/db/src/collection/cleanup-queue.ts
@@ -42,6 +42,15 @@ export class CleanupQueue {
 
   public cancel(key: unknown): void {
     this.tasks.delete(key)
+
+    // Draining the queue retires the root timer with it. Cancelling the
+    // earliest of several tasks leaves the timer where it is: it wakes early,
+    // finds nothing due and reschedules, which costs less than rescanning
+    // every task on each cancellation.
+    if (this.tasks.size === 0 && this.timeoutId !== null) {
+      clearTimeout(this.timeoutId)
+      this.timeoutId = null
+    }
   }
 
   /**
--- a/packages/db/src/collection/lifecycle.ts
+++ b/packages/db/src/collection/lifecycle.ts
@@ -18,15 +18,12 @@ import type { CollectionSyncManager } from './sync'
 import type { CollectionStateManager } from './state'
 
 /**
- * Floor applied to the GC delay of a collection that started syncing without
- * ever having gained a subscriber.
- *
- * Framework adapters build their live query collection while rendering and
- * subscribe to it only once that render commits, so the timer has to outlive
- * the render-to-commit gap. `gcTime` on its own does not: adapters pass a
- * near-zero `gcTime` so that teardown on unmount is immediate, and reusing it
- * here would race the commit. The floor never applies to the timer armed when
- * the last subscriber leaves, which still honours `gcTime` exactly.
+ * Floor applied to the GC delay of a collection that started syncing before
+ * anything subscribed. Adapters build their live query while rendering and
+ * subscribe when that render commits, so the delay has to outlive that gap;
+ * `gcTime` cannot, because adapters pass a near-zero one to make teardown on
+ * unmount immediate. Does not apply to the timer armed when the last
+ * subscriber leaves, which still honours `gcTime` exactly.
  */
 const UNSUBSCRIBED_GC_FLOOR_MS = 50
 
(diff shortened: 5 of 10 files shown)


edzis:
One documentation note that belongs with this change, since it encodes the assumption the patch corrects.

`packages/db/skills/meta-framework/SKILL.md:376`:

> Without preloading, the collection starts syncing only when the component **mounts**, causing a loading flash.

Sync starts during **render**, not mount — `useLiveQuery` constructs with `startSync: true` from its render body (`packages/react-db/src/useLiveQuery.ts:345`, `:383-408`). That gap between "starts on mount" and "starts on render" is exactly where this defect lives: a render that never commits still started sync, and mount is what the teardown path is keyed to. The statement stays inaccurate with this patch applied — the patch reclaims the collection afterwards, it does not move the sync start — so the line is worth correcting either way.

Worth saying that the rest of the shipped skills get this right, and would have prevented our two occurrences had we been reading them instead of our own local notes:

- `db-core/live-queries/SKILL.md:203` — "Create derived collections once at module scope and reuse them. Do not recreate on every render or navigation."
- `:205-206` — "An explicit `gcTime: 0` is preserved and disables garbage collection for that derived collection."

That second one is worth a docs emphasis independently of this PR. `gcTime: 0` means the opposite thing in TanStack Query, where it collects as soon as the query goes inactive (`query-core/removable.js` → `updateGcTime`, then a zero-delay timeout), whereas in DB it disables collection entirely (`collection/lifecycle.ts` → `if (gcTime <= 0 || !Number.isFinite(gcTime)) return`). In a codebase using both libraries side by side, a developer reaching for the idiom they know gets the inverse of what they intend — and, unlike a wrong `staleTime`, it fails silently and permanently. We had two instances of exactly that, both written by following our own internal guidance that had made the same inference.

Also note `library_version: '0.6.17'` in the db-core skill frontmatter while the package publishes 0.7.2, if keeping those in step is intended.

edzis:
Pushed both doc corrections rather than leaving them as notes — `de2c2c95`, now part of this PR.

- `meta-framework/SKILL.md:376` — "starts syncing only when the component mounts" → starts on the component's first render, with the reason stated (mount is where teardown is keyed, which is the asymmetry this PR addresses).
- `db-core/live-queries/SKILL.md:205` — `gcTime: 0` now also opts out of the reclamation this PR adds, so the existing sentence needed extending. Also states that `gcTime: 0` means the opposite in TanStack Query, since that inversion is what produced both of our occurrences.

Happy to split these into a separate docs PR if you would rather keep this one to the behaviour change — they are one commit, so it is a clean cherry-pick either way.

I left `library_version: '0.6.17'` in the db-core frontmatter alone, since that looks like it belongs to your release tooling rather than something to hand-edit.

edzis:
`de2c2c95` adds two documentation corrections, since both describe the behaviour this PR changes.

**`meta-framework/SKILL.md`** said an unpreloaded collection starts syncing when the component *mounts*. It starts on the component's first render — `useLiveQuery` constructs with `startSync: true` from the render body — and mount is only where teardown is keyed. That gap is where this defect lives, so the line is worth correcting whether or not you take the code change: a render that never commits has already started sync.

**`live-queries/SKILL.md`** documented that an explicit `gcTime: 0` disables collection for a derived collection. That was complete before this PR and isn't now, because `gcTime: 0` also opts out of the reclamation added here — `startGCTimer` returns early on its `gcTime <= 0 || !Number.isFinite()` guard before the floor applies. The same line now also notes that `gcTime: 0` means the opposite in TanStack Query, where it collects as soon as the query goes inactive (`query-core` `updateGcTime` → zero-delay timeout, versus `collection/lifecycle.ts`'s early return).

That inversion is worth documenting on its own merits. It is what produced both of our occurrences: an internal note had inferred the Query meaning and recommended `gcTime: 0` for ephemeral one-shot reads, and two call sites were written by following it. Unlike a wrong `staleTime`, it fails silently and permanently, and it is easy to reach for in a codebase using both libraries side by side.

Happy to split the docs commit into its own PR if you would rather keep this one to the behaviour change — it is a single commit, so it cherry-picks cleanly.

One aside: the `db-core` skill frontmatter reads `library_version: '0.6.17'` on a package publishing 0.7.2. I left it alone assuming it belongs to your release tooling, but flagging it in case keeping those in step is intended.

KyleAMathews:
Thanks for the work on this. The orphan-GC behavior was reimplemented in https://github.com/TanStack/db/pull/1810 (https://github.com/TanStack/db/commit/ddc129eeab84d7eca4f2972c3dcc37506202a43d), including active-preload protection and deterministic core and React lifecycle coverage. Closing this conflicting PR as superseded. Its heap and timing observations remain useful diagnostics.

PR #1772, opened 2026-08-25
feat(db): preserve applied loadSubset outcomes

KyleAMathews:
`loadSubset` adapters can now report whether more source rows exist, and core preserves that fact as an applied, exact-demand outcome through live-query demand and window coordination. This PR only adds the result channel; pagination still uses its current peek behavior.

## Root cause

The subset boundary reduced every successful asynchronous load to `Promise<void>`. That discarded source extent before core could bind it to the request whose writes established it. Later layers therefore had no safe way to distinguish `continues`, `exhausted`, and `unknown`, especially when requests were deduplicated, deferred, or routed through lazy sources.

## Approach

- Let adapters optionally return an explicit `{ hasMore: boolean | undefined }` result while existing `Promise<void>` adapters keep working.
- Normalize it after applied settlement into an internal outcome containing an immutable demand snapshot, generation, collection, optional lexical source, and tri-state extent.
- Preserve outcomes through subscriptions, persistence, lazy-demand aggregation, imperative window operations, and the window controller's internal boundary.
- Retain acquisition provenance across `DeduplicatedLoadSubset` without changing its shared-promise or abort-lease behavior. A narrower caller may await covering work but receives `unknown`, not the covering acquisition's raw extent.
- Keep same-generation outcomes from different live-query sources distinct, and release deferred acquisitions with the exact adapter options that acquired them.
- Keep canonical `DemandKey` construction in the query layer so collection sync does not depend on high-level query identity code.

## Key invariants

- A reported extent exists only after its establishing writes are visible.
- Extent belongs to one exact demand and attempt generation.
- Obsolete generations cannot replace newer outcomes.
- Ownership fields such as `AbortSignal` and `Subscription` are not retained as demand data.
- Child and lazy-source outcomes remain readiness inputs; this PR does not let them decide root pagination.
- Existing error, cleanup, cancellation, and synchronous-`true` behavior is unchanged.

## Non-goals

- Do not change `hasNextPage` or remove the current peek fallback.
- Do not add adapter authority rules, total-order semantics, or the coverage registry.
- Do not expose the internal outcome envelope as a new user-facing API.

## Trade-offs

`DeduplicatedLoadSubset` keeps returning the same promise to covered callers for compatibility. A small internal promise-provenance map records which exact demand produced the source result, avoiding a new public result envelope or a collection-to-query dependency cycle.

## Verification

```bash
cd packages/db
pnpm exec vitest run --coverage.enabled=false --reporter=dot
pnpm lint
cd ../..
pnpm run build
```

Results: 132 DB test files passed; 3,335 tests passed and 6 skipped. The SQLite persistence regression passed, all 28 workspace packages built, and ESLint reported no errors.

## Files changed

- `types.ts`: adds the optional adapter result and internal applied-outcome types.
- `load-subset-options.ts`, `load-subset-outcome.ts`: snapshots exact demand and normalizes/protects source extent.
- `sync.ts`, `subscription.ts`, `subset-dedupe.ts`: carry applied outcomes across deferred, deduplicated, canceled, and imperative load paths.
- `collection-subscriber.ts`, `subset-demand-controller.ts`, `collection-config-builder.ts`, `internal.ts`: preserve generation- and source-scoped outcomes through live-query demand.
- `live-query-window-controller.ts`, `effect.ts`: retain outcomes at the internal window boundary without changing pagination.
- `db-sqlite-persistence-core`: preserves authoritative source extent across the persistence wrapper.
- `ARCHITECTURE.md`: records the exact-demand source-extent contract.
- `load-subset-outcome.test.ts`, `load-subset-oracle.property.test.ts`: cover normalization, dedupe projection, deferred start, generations, lazy demand, and window transport.
- `.changeset/add-load-subset-outcomes.md`: documents the upgrade-safe API addition.

---

Part of #1657

## Summary by CodeRabbit

* **New Features**
  * Subset loading now reports whether more data is available, exhausted, or unknown.
  * Load outcomes are preserved across live queries, pagination, persistence, and window coordination.
  * Concurrent and overlapping subset requests are tracked and scoped more accurately.
  * Persistence-backed loading now propagates source results, including exhausted states.

* **Bug Fixes**
  * Improved handling of deferred, deduplicated, cancelled, and asynchronous subset-load requests.
  * Preserved load request details for reliable demand tracking.

* **Tests**
  * Added coverage for result normalization, demand scoping, generations, deduplication, window propagation, and persistence behavior.


KyleAMathews:
Closing as superseded by merged #1797. We chose exact-request deduplication and existing applied-settlement receipts rather than shipping this outcome-channel design. This closes the implementation proposal, not the remaining source-exhaustion work tracked in RFC #1657.

PR #1773, opened 2026-08-25
fix(powersync): serialize tracking startup

KyleAMathews:
PowerSync now serializes change observation, tracking setup, flush, and cleanup. This prevents startup changes from being dropped and prevents released queries or load hooks from creating work after cleanup.

## Root cause

PowerSync can notify that its diff table changed after creating the table but before `createDiffTrigger()` publishes the disposer. The adapter treated the missing disposer as “nothing to flush,” so that notification could be lost.

Startup also had two cancellation gaps: an on-demand subset could be released while waiting for observation to start and then reacquire tracking, and an eager `onLoad` hook could resolve after collection cleanup without being released.

## Approach

- Keep the active trigger setup promise and make flush/disposal wait until its handle is published.
- Start on-demand tracking only after change observation succeeds.
- Recheck subset ownership after the startup boundary, before acquiring a hook or predicate.
- Release an eager load hook that resolves after cleanup and skip trigger creation.

## Key invariants

- A diff notification received during setup is flushed after setup; it is not discarded.
- Cleanup waits for in-flight trigger creation and disposes each published trigger once.
- Released subset demand never reacquires tracking after an async startup boundary.
- A failed change observer cannot leave an unobserved diff trigger running.

## Non-goals

This does not change PowerSync predicate combination, row eviction, or mutation persistence semantics. The startup tracker stays local to this adapter rather than introducing a shared lifecycle abstraction.

## Verification

```bash
pnpm exec vitest run --pool-options.threads.maxThreads=2
pnpm --filter @tanstack/powersync-db-collection build
```

The package suite passes 98 tests across 7 files with no type errors. Regression tests cover notification-before-handle publication, observer startup failure, cleanup during trigger creation, subset release during startup, and a late eager load hook.

## Files changed

- `packages/powersync-db-collection/src/powersync.ts` — serialize startup and make cancellation ownership-aware.
- `packages/powersync-db-collection/tests/on-demand-sync.test.ts` — cover tracking startup, notification, release, and cleanup races.
- `packages/powersync-db-collection/tests/load-hooks.test.ts` — cover a late eager hook after cleanup.
- `.changeset/fix-powersync-tracking-startup.md` — release note.

---
Part of #1657

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sync tracking during startup, cancellation, cleanup, and retry scenarios.
  * Prevented stale subset or load-hook work from being applied after demand changes or collection cleanup.
  * Ensured deferred load-hook cleanup completes safely without creating unnecessary change triggers.
  * Improved recovery when tracking setup, change observation, or subset eviction fails.
* **Tests**
  * Added comprehensive coverage for eager and on-demand sync lifecycle edge cases.


KyleAMathews:
Closing as superseded by merged #1797, which incorporates and tests the PowerSync tracking-startup fixes. This branch no longer needs a separate merge.

PR #1776, opened 2026-08-25
fix(electric): cancel bounded refresh waits

KyleAMathews:
# Summary

Cancel an on-demand Electric refresh wait as soon as its collection or requesting demand is aborted. This prevents a subset load from remaining pending until the 250 ms bound and stops it from starting a snapshot after teardown.

## Root cause

PR #1575 bounded `forceDisconnectAndRefresh()` with a timer, but the race did not include either lifecycle signal. Cleanup and demand cancellation were only noticed after the refresh or timer settled, and collection cleanup was not checked before `requestSnapshot()`.

The existing tests covered early refresh success and failure, timeout, late settlement, and timer cleanup. They did not cancel the collection or demand while the pre-snapshot refresh was still pending.

## Approach

- Race the refresh and 250 ms timer against both the collection and request abort signals.
- Remove abort listeners and clear the timer on every settlement path.
- Recheck both signals before starting a snapshot.
- Keep an aborted demand out of dedup coverage so a later demand can retry.

## Key invariants

- A refresh that finishes or times out without cancellation still requests one snapshot.
- Cleanup and demand cancellation settle the waiting load without requesting a snapshot.
- Late refresh settlement cannot restart canceled work.
- An aborted demand remains retryable.

## Non-goals and upstream limit

This does not change the 250 ms bound from #1575. It also cannot cancel a snapshot after `ShapeStream.requestSnapshot()` has started: Electric exposes neither a request signal nor request identity for the rows it publishes. The existing source comment documents that boundary.

## Verification

```bash
pnpm exec vitest run packages/electric-db-collection/tests/electric.test.ts -t 'refresh wait' --pool-options.threads.maxThreads=2
pnpm --filter @tanstack/electric-db-collection test
pnpm exec eslint packages/electric-db-collection/src/electric.ts packages/electric-db-collection/tests/electric.test.ts
```

- Electric package: 493 tests passed; type checks passed.
- Focused cancellation/retry cases: 2 passed.
- Build passed.
- Lint passed with no errors; 13 pre-existing `no-shadow` warnings remain in the test file.

## Files changed

- `packages/electric-db-collection/src/electric.ts`: make the bounded refresh wait cancellation-aware.
- `packages/electric-db-collection/tests/electric.test.ts`: cover cleanup, demand abort, timer cleanup, late settlement, and retry.
- `.changeset/cancel-electric-refresh-wait.md`: patch release note.

---

Follow-up to #1575. Part of #1657.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved on-demand refresh cancellation when a request or collection is cleaned up.
  - Prevented unnecessary snapshot requests after cancellation or teardown.
  - Ensured interrupted refreshes can be retried successfully when a new request is made.
  - Improved cleanup of pending refresh operations to avoid lingering waits and unexpected activity after shutdown.
  - Prevented late snapshot data from being applied after cleanup.

### DIFF
--- a/.changeset/cancel-electric-refresh-wait.md
+++ b/.changeset/cancel-electric-refresh-wait.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/electric-db-collection': patch
+---
+
+Cancel an on-demand refresh wait when its request or collection is cleaned up, preventing snapshots from starting after teardown.
--- a/packages/electric-db-collection/src/electric.ts
+++ b/packages/electric-db-collection/src/electric.ts
@@ -625,10 +625,34 @@ function createLoadSubsetDedupe<T extends Row<unknown>>({
     // long-poll requests promptly. Bound the wait so on-demand live queries don't
     // remain loading until the long-poll naturally times out.
     // If the refresh fails or times out, we fall through to requestSnapshot which
-    // still works.
+    // still works. Cleanup or request cancellation ends the wait without starting
+    // a snapshot that no current demand can use.
     if (stream.isUpToDate) {
       let timeoutId: ReturnType<typeof setTimeout> | undefined
+      let removeAbortListeners = () => {}
       try {
+        const abortSignals = new Set(
+          [signal, opts.signal].filter(
+            (candidate): candidate is AbortSignal => candidate !== undefined,
+          ),
+        )
+        const aborted = new Promise<void>((resolve) => {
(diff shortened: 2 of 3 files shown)


KyleAMathews:
Closing as superseded by merged #1797, which incorporates and tests the Electric bounded-refresh cancellation fixes. This branch no longer needs a separate merge.

PR #1777, opened 2026-08-25
fix(db): terminate nested predicate subtraction

KyleAMathews:
Fix predicate subtraction so nested conjunctions cannot recurse forever. Queries whose requested and loaded predicates share a nested condition now compute their remaining demand instead of overflowing the stack.

## Root cause

The common-condition path flattened nested `AND` expressions when it found shared terms, but removed terms only from the top-level `AND`. A shared nested term could therefore trigger a recursive call with an unchanged operand. The same call then repeated until JavaScript raised `RangeError: Maximum call stack size exceeded`.

Normal tests missed this because their shared conditions were direct children of the top-level conjunction. A 100× oracle campaign found the asymmetric nested shape at seed `801267634`, path `2185:4:2:12:9:5:4:3`.

## Approach

- Flatten conjunctions consistently when removing shared conditions.
- Return whether each removal made progress.
- Recurse only when both operands strictly lose a flattened `AND` term.
- Add a deterministic regression plus a small Cartesian semantic oracle spanning nested layouts, equality, range, and `NOT` terms.

## Key invariant

Every recursive common-condition simplification must strictly reduce both operands. If it cannot prove that decrease, predicate subtraction falls through to the existing non-recursive rules.

## Non-goals

This does not redesign predicate normalization or add new subtraction laws. It only makes the existing common-condition reduction complete for nested conjunctions and explicitly terminating.

## Verification

Red on `origin/main`:

- The 100× load-subset oracle campaign failed with `RangeError: Maximum call stack size exceeded` at seed `801267634`, path `2185:4:2:12:9:5:4:3`.

Green on this branch:

```bash
pnpm vitest run packages/db/tests/query/predicate-utils.test.ts --pool-options.threads.maxThreads=2
```

- 145 predicate utility tests pass.
- An independent Cartesian probe checked 980,100 predicate pairs without a stack overflow or semantic mismatch.
- `pnpm exec changeset status --since=origin/main` validates the `@tanstack/db` patch changeset.

## Files changed

- `packages/db/src/query/predicate-utils.ts`: require strict recursive progress and remove nested shared conjunction terms.
- `packages/db/tests/query/predicate-utils.test.ts`: pin the failing shape and exhaust related nested predicate layouts over a small value domain.
- `.changeset/fix-nested-predicate-subtraction.md`: record the patch release note.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed predicate subtraction for queries with nested `AND` conditions.
  - Prevented predicate processing from getting stuck when conditions overlap across nested expressions.
  - Improved handling of equality, range, and negated conditions during predicate subtraction.

- **Tests**
  - Added coverage for nested predicate subtraction across multiple values and condition types.
  - Added property-based validation to verify correct results across varied predicate combinations and boolean arrangements.

### DIFF
--- a/.changeset/fix-nested-predicate-subtraction.md
+++ b/.changeset/fix-nested-predicate-subtraction.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/db': patch
+---
+
+Ensure predicate subtraction terminates when common conditions are nested inside conjunctions.
--- a/packages/db/src/query/predicate-utils.ts
+++ b/packages/db/src/query/predicate-utils.ts
@@ -404,21 +404,25 @@ export function minusWherePredicates(
   )
   if (commonConditions.length > 0) {
     // Extract predicates without common conditions
-    const fromWithoutCommon = removeConditions(fromPredicate, commonConditions)
-    const subtractWithoutCommon = removeConditions(
+    const fromRemoval = removeConditions(fromPredicate, commonConditions)
+    const subtractRemoval = removeConditions(
       subtractPredicate,
       commonConditions,
     )
 
-    // Recursively compute difference on simplified predicates
-    const simplifiedDifference = minusWherePredicates(
-      fromWithoutCommon,
-      subtractWithoutCommon,
-    )
+    // Recurse only when both operands lost at least one flattened AND term.
+    // This strict decrease is the termination measure for common-condition
+    // simplification.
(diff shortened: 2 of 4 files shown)


KyleAMathews:
Closing as obsolete after merged #1797 removed predicate subtraction and its public API. We did not ship a replacement subtraction implementation, so this fix no longer has a production target.

PR #1774, opened 2026-08-25
feat(db): track applied loadSubset coverage

KyleAMathews:
This adds a source-side coverage registry that publishes only exact, authoritative `loadSubset` facts backed by rows that have reached the collection. It also keeps cleanup retry-safe and restores Query DB's eager observer after TanStack Query removes its cache entry.

## Root cause

Request intent and a resolved adapter Promise do not prove that a source has established reusable coverage. Coverage also cannot own adapter resources or rows implicitly: compaction, stale results, failed release, and cache removal each have different lifetimes.

## Approach

- Track demand leases, physical acquisitions, applied coverage, and row provenance as separate records.
- Bind publication tokens to the exact collection, source, demand, and current generation.
- Publish finite-prefix coverage only from applied row keys: either at least the requested limit, or fewer rows plus exact exhaustion.
- Integrate publication and release with `_sync.loadSubset` and `_sync.unloadSubset`.
- Keep release atomic. A throwing adapter callback leaves ownership intact, surfaces the original error, and can be retried without repeating callbacks that already succeeded.
- Recreate Query DB's eager lifetime observer after cache removal without retiring its owned rows.

## Key invariants

- Requests and unknown extents establish no coverage.
- Stale or mismatched outcomes establish neither rows nor coverage.
- Public coverage facts are defensive immutable snapshots.
- Coverage compaction never erases leases, acquisitions, or row ownership.
- Logical retirement and GC happen only after every required release callback succeeds.
- Query cache removal does not release eager collection ownership.

## Non-goals

- This PR does not yet use coverage to suppress or combine future load requests.
- It does not change adapter query planning or public query syntax.

## Verification

```bash
pnpm vitest run packages/db/tests/load-subset-outcome.test.ts packages/db/tests/query/coverage-registry-oracle.property.test.ts packages/db/tests/query/load-subset-oracle.property.test.ts packages/query-db-collection/tests/ownership-lifecycle.oracle.test.ts packages/query-db-collection/tests/query.test.ts --maxWorkers=1
pnpm --filter @tanstack/db test
pnpm --filter @tanstack/query-db-collection test
pnpm --filter @tanstack/db build
pnpm --filter @tanstack/query-db-collection build
pnpm --filter @tanstack/db lint
pnpm --filter @tanstack/query-db-collection lint
```

The coverage-registry command model also passes a 100× campaign of 10,000 generated histories.

## Files changed

- `packages/db/src/query/coverage-registry.ts`: coverage, ownership, generation, and retry-safe release state.
[...]

PR #1778, opened 2026-08-26
fix(db): unify ordered window semantics

KyleAMathews:
## Summary

- define one `TotalOrder` for query execution, cursor boundaries, source refinement, and public-key ties
- retain applied row provenance in `WindowState` so offset/limit changes and SSE updates cannot widen from unproven rows
- fetch and refine the full filtered region when locale, reference, or invalid-Date order cannot be expressed safely as a cursor
- align DB-IVM, BTree, and live-query ordering for NaN keys and descending-index ties

## Why

Lazy ordered queries previously used different order and coverage rules at each layer. A source page could look complete to one layer while the live query still lacked a tied row, or a later window could reuse a boundary that did not prove its prefix. This made results depend on request shape, event timing, and index direction.

This PR makes the applied row receipt the source of boundary provenance and uses the same total order from acquisition through publication.

## Validation

- DB: 135 files, 3,450 passed, 6 skipped, no type errors
- Query DB: 8 files, 334 passed, no type errors
- focused TotalOrder/coverage suites: 417 passed
- 100x pagination and replay oracles: 130 passed
- DB, DB-IVM, and Query DB builds and lints
- independent adversarial runtime review: approved

Stacked on #1774. Advances the TotalOrder/window-state step of #1657.


KyleAMathews:
Closing as superseded by merged #1797, which replaces this ordered-window/lifecycle stack with the consolidated implementation. The earlier source-outcome and general-coverage completion claims do not carry over; remaining work stays tracked in RFC #1657.

PR #1787, opened 2026-08-31
fix(db): release synced keys after delete

### LINKED ISSUE #1786: Synced deletes retain keys indefinitely in `CollectionStateManager.syncedKeys`

## Describe the bug

Collections keep the keys of deleted rows in `CollectionStateManager.syncedKeys`, causing the set—and memory usage—to grow over time.

Every synced operation currently adds its key to `syncedKeys`:

https://github.com/TanStack/db/blob/bda2303a339a8e16d2f861775c448149ef7e986f/packages/db/src/collection/state.ts#L1015-L1018

When processing a delete, TanStack DB removes the row from `syncedData` and the other per-row tracking structures, but does not remove it from `syncedKeys`:

https://github.com/TanStack/db/blob/bda2303a339a8e16d2f861775c448149ef7e986f/packages/db/src/collection/state.ts#L1063-L1074

This means a long-running collection that receives and deletes many uniquely keyed rows can retain every key it has processed, even when the collection itself is empty.

I reproduced this with `@tanstack/db` 0.8.6 and the current `main` branch at commit `bda2303a339a8e16d2f861775c448149ef7e986f`.

## To reproduce

Create a collection, sync 50,000 uniquely keyed rows, and then sync-delete all of them:

```js
import { createCollection } from '@tanstack/db'

let controls

  id: 'synced-keys-reproduction',
  getKey: (row) => row.id,
  startSync: true,
  sync: {
    sync: (params) => {
      controls = params
    },
  },
})

controls.begin()

for (let id = 0; id < 50_000; id++) {
  controls.write({
    type: 'insert',
    value: { id },
  })
}

controls.commit()
controls.markReady()

await collection.stateWhenReady()

console.log('after insert', {
  rows: collection.state.size,
  syncedData: collection._state.syncedData.size,
  syncedKeys: collection._state.syncedKeys.size,
})

controls.begin()

for (let id = 0; id < 50_000; id++) {
  controls.write({
    type: 'delete',
    key: id,
  })
}

controls.commit()

console.log('after delete', {
  rows: collection.state.size,
  syncedData: collection._state.syncedData.size,
  syncedKeys: collection._state.syncedKeys.size,
})
```

Actual result:

```text
after insert { rows: 50000, syncedData: 50000, syncedKeys: 50000 }
after delete { rows: 0, syncedData: 0, syncedKeys: 50000 }
```

The problem can also be reproduced with only two rows:

1. Sync-insert keys `1` and `2`.
2. Sync-delete key `1`.
3. `syncedKeys` still contains `{1, 2}` instead of only `{2}`.

## Expected behavior

A synced delete should remove that row's key from `syncedKeys`, just as it removes the row from `syncedData` and the other per-row tracking structures.

After deleting all 50,000 rows, the result should be:

```text
after delete { rows: 0, syncedData: 0, syncedKeys: 0 }
```

### Current behavior

Image

The collection contains no rows after the deletes, but `syncedKeys` still contains all 50,000 keys.

### Behavior with the proposed fix

Image

After adding `this.syncedKeys.delete(key)` to the delete branch, `syncedKeys` returns to zero.

### Chrome heap snapshot

Image

Chrome identifies the retained set directly as:

```text
syncedKeys in CollectionStateManager
```

In this isolated 50,000-key reproduction:

- The selected `Set` retained approximately 1.8 MB.
- Its `CollectionStateManager` retained approximately 3.9 MB in total.
- The retainer chain continued through the collection's `_state`.

## Desktop

- OS: macOS
- Browser: Chrome
- Package version: `@tanstack/db` 0.8.6
- Tested against `main`: `bda2303a339a8e16d2f861775c448149ef7e986f`

## Smartphone

Not applicable.

## Additional context

A minimal fix is to remove the key in the synced delete branch:

```ts
case `delete`:
  this.syncedKeys.delete(key)
  this.syncedData.delete(key)
  // ...
```

I also prepared a focused regression test that:

1. Sync-inserts keys `1` and `2`.
2. Confirms `syncedKeys` contains both keys.
3. Sync-deletes key `1`.
4. Confirms `syncedKeys` contains only key `2`.

The test fails against the current implementation and passes with the proposed fix.

I can submit the fix, regression test, and patch changeset as a PR.

IdoZ14:
## 🎯 Changes

- Remove deleted row keys from `CollectionStateManager.syncedKeys` during synced deletes.
- Add a focused two-key regression test that verifies only the deleted key is removed.
- Add an `@tanstack/db` patch changeset.

The regression test inserts two rows and deletes one. Before this fix, the deleted row’s key remains in syncedKeys; after the fix, only the surviving row’s key remains.

Closes #1786

### Testing

- `pnpm build` — passed
- `pnpm test` — passed across all 28 workspace packages
- `@tanstack/db` — 3,324 tests passed, 6 skipped; no type errors
- Focused regression test — confirmed failing without the fix and passing with it
- ESLint on the changed TypeScript files — no errors

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Deleted rows are now correctly removed from sync tracking, preventing stale deleted keys from being retained.

* **Tests**
  * Added coverage verifying that sync tracking includes inserted rows and removes them after deletion.

* **Release**
  * Prepared a patch release for the database package.

### DIFF
--- a/.changeset/release-deleted-sync-keys.md
+++ b/.changeset/release-deleted-sync-keys.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/db': patch
+---
+
+Release synced key tracking when rows are deleted.
--- a/packages/db/src/collection/state.ts
+++ b/packages/db/src/collection/state.ts
@@ -1061,6 +1061,7 @@ export class CollectionStateManager<
               break
             }
             case `delete`:
+              this.syncedKeys.delete(key)
               this.syncedData.delete(key)
               this.syncedMetadata.delete(key)
               // Clean up origin and pending tracking for deleted rows
--- a/packages/db/tests/collection.test.ts
+++ b/packages/db/tests/collection.test.ts
@@ -1656,6 +1656,40 @@ describe(`Collection`, () => {
     })
   })
 
+  it(`should remove deleted keys from synced key tracking`, async () => {
+    const collection = createCollection(
+      mockSyncCollectionOptionsNoInitialState<{
+        id: number
+        value: string
+      }>({
+        id: `sync-key-delete-test`,
+        getKey: (item) => item.id,
+        startSync: true,
+      }),
+    )
+
+    collection.utils.begin()
+    collection.utils.write({
+      type: `insert`,
+      value: { id: 1, value: `one` },
+    })
+    collection.utils.write({
+      type: `insert`,
+      value: { id: 2, value: `two` },
+    })
+    collection.utils.commit()
+    collection.utils.markReady()
+
+    await collection.stateWhenReady()
+    expect(collection._state.syncedKeys).toEqual(new Set([1, 2]))
+
+    collection.utils.begin()
+    collection.utils.write({ type: `delete`, key: 1 })
+    collection.utils.commit()
+
+    expect(collection._state.syncedKeys).toEqual(new Set([2]))
+  })
+
   it(`should delete row metadata when sync deletes the row`, async () => {
     let testSyncFunctions: any = null
 


hazanid:
@KyleAMathews Could you take a look please? We need it for prduction issue 🙏

KyleAMathews:
Closing as superseded by #1797. Current origin/main at 3b991173 no longer has the persistent CollectionStateManager.syncedKeys set. I also ran two disjoint rounds of 2,000 unique sync inserts followed by deletes; after draining microtasks, every Set, Map, and SortedMap owned by CollectionStateManager returned to size 0. The existing collection-state retention oracle's 1,000-key churn case is green. Thanks for the original report and fix.

IdoZ14:
Thank you!

PR #1794, opened 2026-09-01
fix(db): preserve exact D2 rows during sync reconciliation

### LINKED ISSUE #1783: Calling writeInsert inside a createOptimisticAction throws error

**Describe the bug**

Calling `collection.utils.writeInsert()` inside a `createOptimisticAction` `mutationFn`, when that action's `onMutate` optimistically inserted the same key, corrupts the live-query pipeline. The create itself succeeds and reads back correctly; the **next** mutation of that row throws `Query contributors with the same row key are not congruent`, synchronously out of `onMutate`.

The create — reconciles the POST response with `writeInsert` before returning, as the `createOptimisticAction` docs instruct:

```ts
const addTodo = createOptimisticAction<Todo>({
  onMutate: (todo) => {
    todoCollection.insert(todo);
  },
  mutationFn: async (todo) => {
    const saved = await api.post(todo);

    todoCollection.utils.writeInsert(saved); // <-- corrupts the pipeline
  },
});
```

The update — same shape, `writeUpdate` instead:

```ts
const renameTodo = createOptimisticAction<{ id: string; text: string }>({
  onMutate: ({ id, text }) => {
    todoCollection.update(id, (draft) => {
      draft.text = text;
    });
  },
  mutationFn: async ({ id, text }) => {
    const saved = await api.patch({ id, text });

    todoCollection.utils.writeUpdate(saved);
  },
});
```

```ts
await addTodo({ id: 'todo-1', text: 'walk the dog' }).isPersisted.promise;
// fine: tx.state === 'completed', tx.error undefined, collection and live query both correct

renameTodo({ id: 'todo-1', text: 'walk the cat' });
// Uncaught Error: Query contributors with the same row key are not congruent
```

Full stack:

```
Uncaught Error: Query contributors with the same row key are not congruent
    at ReduceOperator (@tanstack/db/src/query/compiler/index.ts:1270)
    at ReduceOperator.run (@tanstack/db-ivm/src/operators/reduce.ts:42)
    at D2.step (@tanstack/db-ivm/src/d2.ts:51)
    at D2.run (@tanstack/db-ivm/src/d2.ts:61)
    at CollectionConfigBuilder.maybeRunGraph (@tanstack/db/src/query/live/collection-config-builder.ts:452)
    at CollectionConfigBuilder.executeGraphRun (@tanstack/db/src/query/live/collection-config-builder.ts:657)
    at Scheduler.flush (@tanstack/db/src/scheduler.ts:158)
    at CollectionChangesManager.publishEvents (@tanstack/db/src/collection/changes.ts:196)
```

Because it throws out of `onMutate` it is not a rejected transaction that rollback can absorb — it is an uncaught error inside the caller's event handler, and every live query over the collection stays broken from that point on.

Reconciling through a direct write API is what the `createOptimisticAction` docblock tells you to do:

> **Important:** Inside your `mutationFn`, you must ensure that your server writes have synced back before you return, as the optimistic state is dropped when you return from the mutation function. You generally use collection-specific helpers to do this, such as Query's `utils.refetch()`, **direct write APIs**, or Electric's `utils.awaitTxId()`.

That contract holds for `writeUpdate` over an optimistic *update* — the second action above can run repeatedly against a row that arrived from sync. It breaks for `writeInsert` over an optimistic *insert*.

**To Reproduce**

Self-contained script (no DOM, no test runner, no network) attached below. `npx tsx tanstack-db-writeinsert-repro.ts` on Node 20+.

Steps:

1. Create a `queryCollectionOptions` collection whose `queryFn` returns `[]`.
2. Create a live query over it (`q.from({ todo }).select(...)`) and attach a subscriber — the D2 graph has to actually be running.
3. Create an optimistic action whose `onMutate` calls `collection.insert(row)` and whose `mutationFn` awaits a POST and then calls `collection.utils.writeInsert(saved)`.
4. Run it and await `tx.isPersisted.promise`. It resolves; everything reads correctly.
5. Run a second optimistic action that calls `collection.update(key, ...)` in `onMutate`.
6. See the error thrown synchronously out of step 5.

Output of the attached script:

```
[broken] create isPersisted: resolved
[broken] create state: completed
[broken] create error: none
[broken] collection row: walk the dog
[broken] live query row: walk the dog
[broken] THREW: Query contributors with the same row key are not congruent
[control] rename 1 isPersisted: resolved
[control] rename 2 isPersisted: resolved
[control] final text: water plants thrice
[workaround] rename isPersisted: resolved
[workaround] final text: buy oat milk
```

**Expected behavior**

Reconciling an optimistically inserted row with the server's response via `utils.writeInsert()` before returning from `mutationFn` should leave the collection and its live queries in a consistent state, exactly as `utils.writeUpdate()` does for an optimistically updated row. Subsequent mutations of that key should behave normally.

If mixing a direct write with an in-flight optimistic insert is not supported, the direct write API should reject the call at the point it is made rather than silently corrupting the pipeline and surfacing an unrelated-looking error on a later mutation — and the `createOptimisticAction` docs should say so.

**Screenshots**

n/a — console output above.

- OS: macOS 26.5.1 (Darwin 25.5.0)
- Browser: originally hit in a Chromium browser in a Vite + React app; the attached script reproduces it headless, so it is not browser specific
- Version: Node 22.12.0 via `tsx`; package versions below

- n/a — not device specific.

**Additional context**

Package versions:

```
@tanstack/db                   0.8.5   (latest published)
@tanstack/db-ivm               0.1.19
@tanstack/query-db-collection  1.2.10
@tanstack/query-core           5.90.20
@tanstack/react-db             0.3.5
```

<details>
<summary><code>tanstack-db-writeinsert-repro.ts</code></summary>

```ts
/**
 * Repro: `utils.writeInsert()` inside a `createOptimisticAction` `mutationFn`
 * corrupts the live-query pipeline when that action's `onMutate` optimistically
 * inserted the same key.
 *
 *   @tanstack/db                 0.8.5
 *   @tanstack/query-db-collection 1.2.10
 *   @tanstack/query-core          5.x
 *
 * Run it (Node 20+ — `@tanstack/db-ivm` hashing needs global `File`, and
 * transaction IDs need global `crypto`):
 *
 *   npx tsx tanstack-db-writeinsert-repro.ts
 *
 * or import it from an app entry point and call `runRepro()`. No DOM, no test
 * runner, no network.
 *
 * Expected output:
 *
 *   [broken] create isPersisted: resolved
 *   [broken] create state: completed
 *   [broken] create error: none
 *   [broken] collection row: walk the dog
 *   [broken] live query row: walk the dog
 *   [broken] THREW: Query contributors with the same row key are not congruent
 *   [control] rename 1 isPersisted: resolved
 *   [control] rename 2 isPersisted: resolved
 *   [control] final text: water plants thrice
 *   [workaround] rename isPersisted: resolved
 *   [workaround] final text: buy oat milk
 */

import {
  createCollection,
  createLiveQueryCollection,
  createOptimisticAction,
} from '@tanstack/db';
import { QueryClient } from '@tanstack/query-core';
import { queryCollectionOptions } from '@tanstack/query-db-collection';

type Todo = {
  id: string;
  text: string;
};

const flush = () => new Promise((resolve) => setTimeout(resolve, 25));

/**
 * Reports how `tx.isPersisted.promise` settles without hanging on it, so the
 * script still prints its findings if a transaction never resolves.
 */
async function settle(tx: { isPersisted: { promise: Promise<unknown> } }) {
  return Promise.race([
    tx.isPersisted.promise.then(
      () => 'resolved',
      () => 'rejected'
    ),
    flush().then(() => 'still pending'),
  ]);
}

/** Stands in for the backend. IDs are client-generated, so POST echoes the row. */
const server = {
  post: async (todo: Todo): Promise<Todo> => ({ ...todo }),
  patch: async (todo: Todo): Promise<Todo> => ({ ...todo }),
};

async function setup() {
  const queryClient = new QueryClient({
    defaultOptions: { queries: { retry: false, staleTime: Infinity } },
  });

  // Starts empty. Every row arrives through the direct write API below.
  const todos = createCollection(
    queryCollectionOptions<Todo>({
      id: 'todos',
      queryClient,
      queryKey: ['todos'],
      getKey: (todo) => todo.id,
      queryFn: async () => [],
    })
  );

  const visibleTodos = createLiveQueryCollection({
    id: 'visible-todos',
    query: (q) => q.from({ todo: todos }).select(({ todo }) => ({ ...todo })),
    getKey: (row) => row.id,
  });

  await todos.preload();

  // An active subscriber is required — the D2 graph has to actually be running.
  const subscription = visibleTodos.subscribeChanges(() => {}, {
    includeInitialState: true,
  });

  await visibleTodos.preload();

  const addTodo = createOptimisticAction<Todo>({
    onMutate: (todo) => {
      todos.insert(todo);
    },
    mutationFn: async (todo) => {
      const saved = await server.post(todo);

      // Reconcile the server response before returning, as the
      // `createOptimisticAction` docblock instructs. This is the call that
      // corrupts the pipeline.
      todos.utils.writeInsert(saved);
    },
  });

  const renameTodo = createOptimisticAction<{ id: string; text: string }>({
    onMutate: ({ id, text }) => {
      todos.update(id, (draft) => {
        draft.text = text;
      });
    },
    mutationFn: async ({ id, text }) => {
      const saved = await server.patch({ id, text });

      todos.utils.writeUpdate(saved);
    },
  });

  return {
    todos,
    visibleTodos,
    addTodo,
    renameTodo,
    dispose: () => subscription.unsubscribe(),
  };
}

/** The bug: optimistic insert reconciled with `writeInsert`. */
async function brokenCase() {
  const { todos, visibleTodos, addTodo, renameTodo, dispose } = await setup();

  const tx = addTodo({ id: 'todo-1', text: 'walk the dog' });

  // The create itself looks completely healthy.
  console.log('[broken] create isPersisted:', await settle(tx));
  console.log('[broken] create state:', tx.state);
  console.log('[broken] create error:', tx.error ?? 'none');
  console.log('[broken] collection row:', todos.get('todo-1')?.text);
  console.log('[broken] live query row:', visibleTodos.get('todo-1')?.text);

  // The next mutation of that row throws synchronously, out of `onMutate`.
  try {
    renameTodo({ id: 'todo-1', text: 'walk the cat' });
    console.log('[broken] NO ERROR — bug did not reproduce');
  } catch (error) {
    console.log('[broken] THREW:', (error as Error).message);
  }

  dispose();
}

export async function runRepro() {
  await brokenCase();
}

runRepro()
  .catch((error) => {
    console.error('unexpected failure', error);
  })
  .finally(() => {
    // The QueryClient keeps timers alive, so a plain script would hang here.
    if (typeof process !== 'undefined') process.exit(0);
  });
```

</details>

obeattie:
## 🎯 Changes

Fixes #1783.

D2 applies an update by removing the old row and adding the new one. The row being removed must _exactly_ match the row previously added. The state diverged like this:

1. An optimistic insert added one snapshot of the row to D2.
2. Sync confirmation emitted another insert with the same key. This was correctly skipped so D2 would not count the row twice.
3. D2 therefore kept the optimistic snapshot, while the collection later described the row using its sync-confirmed snapshot.
4. A later update tried to remove the collection’s snapshot, but D2 held the optimistic snapshot.
5. The snapshots represented the same key but were not identical, so D2 threw `Query contributors with the same row key are not congruent`.

Any differing property can cause this; it is not specific to `$synced` or `$origin`. A focused test demonstrates the same mismatch with a normal `status` property. Changing the hashing would only hide the mismatch by making different rows appear equal. The correct fix is to remove the exact row that was previously added.

I bisected the regression to #1740. That PR added the congruence check which exposed the incorrect removal. This check is correct; the source bookkeeping was not.

The fix remembers the exact row sent to D2 for each key and uses it for later updates and deletes. The same bookkeeping also prevents duplicate inserts.

The regression test covers the original optimistic insert and synchronous sync confirmation, then verifies that the collection and live query each contain exactly one updated row.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed live queries throwing when optimistically inserted rows are updated after synchronous sync confirmation.
  - Improved handling of row updates, deletions, and duplicate inserts to keep query and collection results consistent.
  - Ensured updates correctly replace previously received row values without stale or duplicate data.

- **Tests**
  - Added coverage for optimistic insert updates and accurate row reconciliation.

### DIFF
--- a/.changeset/sixty-nights-refuse.md
+++ b/.changeset/sixty-nights-refuse.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/db': patch
+---
+
+Fix live queries throwing when updating an optimistically inserted row after synchronous sync confirmation.
--- a/packages/db/src/query/effect.ts
+++ b/packages/db/src/query/effect.ts
@@ -17,7 +17,7 @@ import {
   computeSubscriptionOrderByHints,
   extractCollectionSources,
   extractCollectionsFromQuery,
-  filterDuplicateInserts,
+  prepareChangesForD2,
   sendChangesToInput,
   splitUpdates,
   trackBiggestSentValue,
@@ -388,10 +388,10 @@ class EffectPipelineRunner<TRow extends object, TKey extends string | number> {
 
   // Subscription management
   private readonly unsubscribeCallbacks = new Set<() => void>()
-  // Duplicate insert prevention per lexical source
-  private readonly sentToD2KeysBySource = new Map<
+  // Exact rows contributed to D2 per lexical source
+  private readonly sentToD2RowsBySource = new Map<
     string,
-    Set<string | number>
+    Map<string | number, Record<string, unknown>>
   >()
 
   // Output accumulator
@@ -513,8 +513,8 @@ class EffectPipelineRunner<TRow extends object, TKey extends string | number> {
       const { sourceId, alias, collection } = source
       const collectionId = collection.id
 
-      // Initialise per-source duplicate tracking
-      this.sentToD2KeysBySource.set(sourceId, new Set())
+      // Initialise per-source D2 contribution tracking
+      this.sentToD2RowsBySource.set(sourceId, new Map())
 
       // Discover dependencies: if source collection is itself a live query
       // collection, its builder must run first during transaction flushes.
@@ -801,11 +801,10 @@ class EffectPipelineRunner<TRow extends object, TKey extends string | number> {
     const input = this.inputs[sourceId]
     if (!input) return 0
 
-    // Filter duplicates per lexical source
-    const sentKeys = this.sentToD2KeysBySource.get(sourceId)!
-    const filtered = filterDuplicateInserts(changes, sentKeys)
+    const sentRows = this.sentToD2RowsBySource.get(sourceId)!
+    const reconciled = prepareChangesForD2(changes, sentRows)
 
-    return sendChangesToInput(input, filtered)
+    return sendChangesToInput(input, reconciled)
   }
 
   /**
@@ -1040,11 +1039,11 @@ class EffectPipelineRunner<TRow extends object, TKey extends string | number> {
     changes: Array<ChangeMessage<any, string | number>>,
     comparator: (a: any, b: any) => number,
   ): void {
-    const sentKeys = this.sentToD2KeysBySource.get(sourceId) ?? new Set()
+    const sentRows = this.sentToD2RowsBySource.get(sourceId) ?? new Map()
     const result = trackBiggestSentValue(
       changes,
       this.biggestSentValue.get(sourceId),
-      sentKeys,
+      sentRows,
       comparator,
     )
     this.biggestSentValue.set(sourceId, result.biggest)
@@ -1069,7 +1068,7 @@ class EffectPipelineRunner<TRow extends object, TKey extends string | number> {
       }
     }
     this.unsubscribeCallbacks.clear()
-    this.sentToD2KeysBySource.clear()
+    this.sentToD2RowsBySource.clear()
     this.pendingChanges.clear()
     this.lazySources.clear()
     this.demand.clear()
--- a/packages/db/src/query/live/collection-subscriber.ts
+++ b/packages/db/src/query/live/collection-subscriber.ts
@@ -5,7 +5,7 @@ import {
 import {
   computeOrderedLoadCursor,
   computeSubscriptionOrderByHints,
-  filterDuplicateInserts,
+  prepareChangesForD2,
   sendChangesToInput,
   splitUpdates,
   trackBiggestSentValue,
@@ -46,10 +46,9 @@ export class CollectionSubscriber<
     { resolve: () => void }
   >()
 
-  // Track keys that have been sent to the D2 pipeline to prevent duplicate inserts
-  // This is necessary because different code paths (initial load, change events)
-  // can potentially send the same item to D2 multiple times.
-  private sentToD2Keys = new Set<string | number>()
+  // Keep the exact value contributed for every source key. This both prevents
+  // duplicate inserts and ensures later updates retract the value currently in D2.
+  private sentToD2Rows = new Map<string | number, Record<string, unknown>>()
 
   // Direct load tracking callback for ordered path (set during subscribeToOrderedChanges,
   // used by loadNextItems for subsequent requestLimitedSnapshot calls)
@@ -237,16 +236,16 @@ export class CollectionSubscriber<
     callback?: () => boolean,
   ) {
     const changesArray = Array.isArray(changes) ? changes : [...changes]
-    const filteredChanges = filterDuplicateInserts(
+    const reconciledChanges = prepareChangesForD2(
       changesArray,
-      this.sentToD2Keys,
+      this.sentToD2Rows,
     )
 
     // currentSyncState and input are always defined when this method is called
     // (only called from active subscriptions during a sync session)
     const input =
       this.collectionConfigBuilder.currentSyncState!.inputs[this.sourceId]!
-    const sentChanges = sendChangesToInput(input, filteredChanges)
+    const sentChanges = sendChangesToInput(input, reconciledChanges)
 
     // Do not provide the callback that loads more data
     // if there's no more data to load
@@ -351,14 +350,14 @@ export class CollectionSubscriber<
     subscriptionHolder.current = subscription
     this.registerSubscriptionCleanup(subscription)
 
-    // Listen for truncate events to reset cursor tracking state and sentToD2Keys
+    // Listen for truncate events to reset cursor and D2 source tracking state.
     // This ensures that after a must-refetch/truncate, we don't use stale cursor data
     // and allow re-inserts of previously sent keys
     const truncateUnsubscribe = this.collection.on(`truncate`, () => {
       this.biggest = undefined
       this.lastLoadRequestKey = undefined
       this.pendingOrderedLoadPromise = undefined
-      this.sentToD2Keys.clear()
+      this.sentToD2Rows.clear()
     })
 
     // Clean up truncate listener when subscription is unsubscribed
@@ -542,7 +541,7 @@ export class CollectionSubscriber<
     const result = trackBiggestSentValue(
       changes,
       this.biggest,
-      this.sentToD2Keys,
+      this.sentToD2Rows,
       comparator,
     )
     this.biggest = result.biggest
--- a/packages/db/src/query/live/utils.ts
+++ b/packages/db/src/query/live/utils.ts
@@ -166,21 +166,57 @@ export function filterDuplicateInserts(
 }
 
 /**
+ * Prevents duplicate source inserts and makes update/delete retractions use
+ * the exact row previously contributed to D2 for the key.
+ */
+export function prepareChangesForD2(
+  changes: Array<ChangeMessage<Record<string, unknown>, string | number>>,
+  sentRows: Map<string | number, Record<string, unknown>>,
+): Array<ChangeMessage<Record<string, unknown>, string | number>> {
+  return changes.flatMap((change) => {
+    if (change.type === `insert`) {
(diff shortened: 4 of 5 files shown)


InfinityBowman:
We hit this in production on 0.8.7 through a path that does not involve an optimistic insert or `writeInsert`: plain `collection.update` calls on an existing key, confirmed by ordinary sync commits. The trigger is a new update on a key starting while a sync commit for that key is already deferred behind an earlier in-flight mutation. Claude reduced it to the script below against bare `@tanstack/db@0.8.7` (main is at the 0.8.7 commit as of 2026-09-06). Applying this PR's diff to 0.8.7 makes it pass, so posting here in case the extra case is useful for the regression suite.

```ts
import { createCollection, createLiveQueryCollection, eq } from '@tanstack/db'

let hooks
const releases = new Map()
const col = createCollection({
  id: 'rows',
  getKey: (r) => r.id,
  startSync: true,
  sync: {
    rowUpdateMode: 'full',
    sync: (p) => {
      hooks = p
      p.begin()
      p.write({ type: 'insert', value: { id: 'c1', n: 0 } })
      p.commit()
      p.markReady()
    },
  },
  // Persist resolves when the test releases it, so confirmations can be ordered explicitly.
  onUpdate: ({ transaction }) =>
    new Promise((resolve) => releases.set(transaction.mutations[0].modified.n, resolve)),
})
const lq = createLiveQueryCollection({
  startSync: true,
  query: (q) => q.from({ c: col }).where(({ c }) => eq(c.id, 'c1')),
})
lq.subscribeChanges(() => {})
col.subscribeChanges((changes) => {
  for (const ch of changes) console.log('   event', ch.type, ch.value.n, ch.value.$synced, 'prev', ch.previousValue?.n, ch.previousValue?.$synced)
})

const mutate = (n) => col.update('c1', (d) => { d.n = n }).isPersisted.promise.catch(() => {})
const confirm = (n) => {
  hooks.begin()
  hooks.write({ type: 'update', value: { id: 'c1', n } })
  hooks.commit()
  releases.get(n)()
}
const tick = () => new Promise((r) => setTimeout(r, 0))

// m = collection.update, c = server confirmation for that update.
// c1 arrives while m2 is in flight, so its commit is deferred; m3 starts before it applies.
for (const step of ['m1', 'm2', 'c1', 'm3', 'c2', 'c3', 'm4']) {
  const n = Number(step[1])
  if (step[0] === 'm') mutate(n)
  else confirm(n)
  await tick()
  console.log(step, 'collection.$synced =', col.get('c1').$synced, '| liveQuery.$synced =', lq.toArray[0]?.$synced, 'n =', lq.toArray[0]?.n)
}
```

Output on 0.8.7 (Node 24, last steps only):

```
   event update 3 false prev 2 false
m3 collection.$synced = false | liveQuery.$synced = false n = 3
c2 collection.$synced = false | liveQuery.$synced = false n = 3
c3 collection.$synced = true | liveQuery.$synced = false n = 3      <- row flipped to synced, no change event
   event update 4 false prev 3 true                                   <- previousValue is a row the live query never received
m4 Error: Query contributors with the same row key are not congruent
```

After that, every further write on the key throws.

Two sequences that do not throw, for contrast: `m1 m2 c1 c2 m3` (the deferred commits apply before the next update starts, and the `$synced` flip is published as an event) and `m1 m2 m3 c1 c2 c3 m4` (all three in flight, then all confirmations).

So the collection-side defect in this path is that the last deferred commit updates the visible row without publishing the change, and the subscriber then retracts the collection's `previousValue` instead of the row it actually contributed. This PR's exact-row bookkeeping in `CollectionSubscriber` handles the second half, which is enough to keep the live query consistent. For completeness, #1795 alone does not change the outcome here. The missing publication on the state side may still be worth a cell in the publication oracle's overlap matrix mentioned on #1783.

This was investigated and reduced with Claude Code, and the script and output above are from actual runs.

KyleAMathews:
Closing as superseded by merged #1797, which incorporates and tests exact D2 before-images during source reconciliation. The consolidated PR also includes direct Query DB optimistic write-back regression traces. This branch no longer needs a separate merge.

PR #1795, opened 2026-09-01
fix(db): publish one event per key for metadata-only sync commits

### LINKED ISSUE #1767: transaction.isPersisted.promise never settles for the second consecutive insert (0.8.x)

**Describe the bug**

When two inserts happen one after another on the same collection, the second
one's `transaction.isPersisted.promise` never settles. It neither resolves nor
rejects — it just hangs forever.

Both writes are fine. Both rows reach the server and are saved. Only the
promise is stuck.

We log around the await. First insert:

```
[probe] awaiting isPersisted  createScene g1eudlenw7dh9q7nkaa9c703
[probe] isPersisted RESOLVED  createScene g1eudlenw7dh9q7nkaa9c703
[probe] navigating to /studio/$projectId/manuscript/$contentNodeId
```

Second insert, immediately after:

```
[probe] awaiting isPersisted  createScene mna5k3arqjl2cv4b9ewwjv6p
```

...and nothing follows. No resolve, no reject, no error in the console. The
server received both calls and the database has both rows.

**Why it matters**

We use that promise to know a write is durable before navigating the user to
the row they just created:

```ts
await tx.isPersisted.promise
navigate(newRow)
```

Because the promise never settles, the navigation never happens. The user
creates something and is left sitting on the previous item, with no error
shown. Any code that awaits durability before doing something will hang the
same way.

**To Reproduce**

1. Set up a collection using `persistedQueryCollectionOptions`, read by a live
   query that has a `leftJoin` and a correlated `toArray` subquery:

   ```ts
   q.from({ nodes: contentNodesCollection })
     .where(({ nodes }) => eq(nodes.id, contentNodeId))
     .leftJoin({ user: usersCollection }, ({ nodes, user }) =>
       eq(nodes.updatedBy, user.id),
     )
     .select(({ nodes, user }) => ({
       ...nodes,
       updatedByUser: user,
       tags: toArray(
         q.from({ nodeTag: contentNodeTagsCollection })
           .where(({ nodeTag }) => eq(nodeTag.contentNodeId, nodes.id))
           .innerJoin({ tag: tagsCollection }, ({ nodeTag, tag }) =>
             eq(nodeTag.tagId, tag.id),
           )
           .select(({ tag }) => ({ id: tag.id, name: tag.name, color: tag.color })),
       ),
     }))
   ```

2. Insert a row into `contentNodesCollection` via an optimistic action, and
   await the transaction's `isPersisted.promise`. This one resolves.
3. Without reloading, insert a second row the same way and await its
   `isPersisted.promise`.
4. The second promise never settles. Both rows are saved on the server.

**Expected behavior**

The second insert's `isPersisted.promise` resolves, the same as the first.

**Screenshots**

n/a

**Desktop:**

- OS: macOS 26.5.2
- Browser: Chromium 148.0.7778.280 (Electron 42.9.2)
- Version: `@tanstack/db` 0.8.4, `@tanstack/react-db` 0.3.4,
  `@tanstack/query-db-collection` 1.2.9,
  `@tanstack/offline-transactions` 1.0.50

**Smartphone:**

n/a — not tested on mobile.

**Additional context**

Versions we tried:

| Versions | Result |
|---|---|
| `db` 0.6.13, `react-db` 0.1.91 | fine |
| `db` 0.7.2, `react-db` 0.2.1 | fine |
| `db` 0.8.3, `react-db` 0.3.3 | broken, different symptom (below) |
| `db` 0.8.4, `react-db` 0.3.4 | broken as described above |

On **0.8.3** the same scenario failed differently: instead of hanging, the
promise *rejected*, with

```
Error: Query contributors with the same row key are not congruent
    at ReduceOperator.<anonymous>
    at ReduceOperator.run
    at D2.step
    at D2.run
    at CollectionConfigBuilder.maybeRunGraph
    at CollectionConfigBuilder.executeGraphRun
    at run
    at Scheduler.flush
    at withPublicationContext
    at CollectionChangesManager.publishEvents
```

0.8.4 no longer throws that — we assume #1761 addressed it — but the promise
now hangs instead. We can't tell from outside whether these are the same
underlying problem, so both are recorded here.

How we narrowed it: we hit this while upgrading several unrelated packages at
once, and went commit by commit. Everything passed until `db` 0.8.3 landed,
and passed again on 0.7.2 with nothing else changed. Our end-to-end suite
catches it reliably (2 of 4 tests in one file), and it reproduces by hand.

Rest of the stack: React 19, TanStack Start 1.168.49, Node 22.18.

One caveat on our own testing: an earlier round of manual probing was done in
a browser profile left dirty by unrelated offline testing of ours, which
produced a confusing third symptom. The behaviour reported above was
re-confirmed after clearing IndexedDB and local storage, so it isn't that.

We don't have a small standalone reproduction — this is from a real app.
Happy to build one if that would help.

elylucas:
## 🎯 Changes

Fixes #1767.

Two optimistic writes in a row on the same key blow up on the second one:

```
Error: Query contributors with the same row key are not congruent
```

### What's going on

When a collection publishes a batch of changes, that batch is a diff, so each key should appear in it at most once. Here it appeared twice, with both events saying the same thing.

The sequence, using a query collection as the example:

1. You make an optimistic update, and your `mutationFn` acknowledges it with `utils.writeUpsert`.
2. `writeUpsert` refreshes the query cache, and applying that result records row ownership metadata. That metadata write goes into its own sync transaction — one with **no row operations at all**, just a metadata entry.
3. That transaction stays pending while your transaction is persisting.
4. When your transaction completes, two different code paths each notice the row switching from optimistic to synced (`$synced: false → true`), and both emit an event for it. Both land in the same published batch.

The guard meant to prevent that double-emit is `recentlySyncedKeys`, seeded by `capturePreSyncVisibleState`. But it derived its keys from a sync transaction's row *operations* only, while `commitPendingTransactions` derives its keys from operations **plus** row metadata writes. A metadata-only transaction fell straight through the gap between those two sets.

Downstream, the duplicated diff leaves the live query counting the old row at -2 and the new one at +2. The next optimistic write on that key then leaves two positive contributors with different values, which is exactly what the keyed reduction rejects.

That also explains the shape of the bug report: round one is what corrupts the bookkeeping, and round two is what notices.

### The fix

Pull the "which keys does this sync commit touch" derivation into a single helper, `collectAffectedKeys`, and use it in all three places that ask the question — `commitPendingTransactions`, `capturePreSyncVisibleState`, and `cancelPendingSyncedTransaction`. The bug was those key sets drifting apart, so this keeps them from drifting again rather than just patching the one that was behind.

Behaviour change is limited to the two sites that were missing metadata keys. `commitPendingTransactions` already had them.

### Tests

Two regression tests, both of which fail with the exact error before the fix:

- `packages/db/tests/collection-metadata-only-sync-commit.test.ts` — core-level. Drives a metadata-only sync transaction directly through the sync API while an optimistic mutation is in flight, and asserts no published batch ever names one key twice.
- `packages/query-db-collection/tests/optimistic-writeback.test.ts` — the reported path, via `utils.writeUpsert`.

### Why the existing tests missed it

The suite covers sync transactions that carry row operations, and covers row metadata, but never a transaction that writes metadata with *no* row operations while an optimistic mutation is pending — the one shape where the two key sets disagree. The core test above closes that gap, and its "a batch never names a key twice" assertion is the more general form of the invariant, so it should catch other ways of producing a malformed batch too.

### Verification

`db` 3071 · `query-db-collection` 170 + 114 e2e · `electric` 242 · `react-db` 161 · `offline-transactions` 65 · `powersync` 92 · plus trailbase, rxdb, solid, svelte, vue and angular — all passing, typecheck and lint clean.

One note: `collection-events.test.ts > should emit events with correct structure` times out under full-suite load on my machine. It does the same on an unmodified `main` (3 runs out of 3), so it looks pre-existing and unrelated to this change.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed duplicate change events when synchronizing metadata-only updates.
  * Prevented live queries from becoming inconsistent after repeated optimistic writes to the same record.
  * Ensured synchronized metadata updates correctly participate in change tracking.

* **Tests**
  * Added regression coverage for metadata-only synchronization and consecutive optimistic updates.

### DIFF
--- a/.changeset/fix-metadata-only-sync-duplicate-events.md
+++ b/.changeset/fix-metadata-only-sync-duplicate-events.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/db': patch
+---
+
+Capture pre-sync visible state for keys that a pending sync transaction only writes metadata for, so retiring an optimistic mutation publishes one change event per key instead of two. The duplicated event corrupted multiplicity bookkeeping in live queries, making a later optimistic write on the same key fail with "Query contributors with the same row key are not congruent".
--- a/packages/db/src/collection/state.ts
+++ b/packages/db/src/collection/state.ts
@@ -825,6 +825,32 @@ export class CollectionStateManager<
   }
 
   /**
+   * Collects every key the given sync transactions can publish a change event
+   * for.
+   *
+   * Row metadata writes count even when the transaction carries no row
+   * operation: committing one still retires a completed optimistic mutation for
+   * that key, which flips virtual props such as `$synced`. Every site that asks
+   * "which keys does this sync commit touch" must agree on this set. When they
+   * disagree, `recomputeOptimisticState` and `commitPendingTransactions` each
+   * emit the same transition and the published batch names one key twice.
+   */
+  private collectAffectedKeys(
+    transactions: Iterable<PendingSyncedTransaction<TOutput, TKey>>,
+  ): Set<TKey> {
+    const keys = new Set<TKey>()
+    for (const transaction of transactions) {
+      for (const operation of transaction.operations) {
+        keys.add(operation.key as TKey)
+      }
+      for (const key of transaction.rowMetadataWrites.keys()) {
+        keys.add(key)
+      }
+    }
+    return keys
+  }
+
+  /**
    * Attempts to commit pending synced transactions if there are no active transactions
    * This method processes operations from pending transactions and applies them to the synced data
    */
@@ -905,15 +931,7 @@ export class CollectionStateManager<
       let truncatePendingLocalOrigins: Set<TKey> | undefined
 
       // First collect all keys that will be affected by sync operations
-      const changedKeys = new Set<TKey>()
-      for (const transaction of committedSyncedTransactions) {
-        for (const operation of transaction.operations) {
-          changedKeys.add(operation.key as TKey)
-        }
-        for (const [key] of transaction.rowMetadataWrites) {
-          changedKeys.add(key)
-        }
-      }
+      const changedKeys = this.collectAffectedKeys(committedSyncedTransactions)
 
       const virtualSnapshotKeys = new Set(changedKeys)
       for (const key of this.pendingOptimisticDirectUpserts) {
@@ -1391,14 +1409,10 @@ export class CollectionStateManager<
     this.pendingSyncedTransactions.splice(index, 1)
     transaction.applied.reject(new SyncTransactionAbortedError())
 
-    const remainingPendingKeys = new Set<TKey>()
-    for (const pending of this.pendingSyncedTransactions) {
-      for (const operation of pending.operations) {
-        remainingPendingKeys.add(operation.key as TKey)
-      }
-    }
-    for (const operation of transaction.operations) {
-      const key = operation.key as TKey
+    const remainingPendingKeys = this.collectAffectedKeys(
+      this.pendingSyncedTransactions,
+    )
+    for (const key of this.collectAffectedKeys([transaction])) {
       if (!remainingPendingKeys.has(key)) {
         this.recentlySyncedKeys.delete(key)
         this.preSyncVisibleState.delete(key)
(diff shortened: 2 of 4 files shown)


KyleAMathews:
Closing as superseded by merged #1797, which incorporates and tests one-event-per-key metadata-only sync publication. This branch no longer needs a separate merge.

Issue #968, opened 2025-12-04
useLiveInfiniteQuery doesn't fetch more data when collection has no more items


believer:
- [x] I've validated the bug against the latest version of DB packages

**Describe the bug**

I've converted a `useInfiniteQuery` from TanStack Query to `useLiveInfiniteQuery`. Initially it loads data and also loads more data as long as there is more data in the collection.

I expected that the `queryFn` would re-trigger when the collection had no more data to fetch more from the source, but once it displays all the items in the collection it does nothing.

I noticed that `queryFn` has a `pageParam`, but it's always `undefined` and is only logged on initial load. I might be misunderstanding how it should work?

Might be related to #872 

(Note: I know that my `queryFn` would overwrite existing items, but that's the next thing to fix 😄)

```tsx 
export const moviesCollection = createCollection(
  queryCollectionOptions({
    getKey: (item) => item.id,
    queryClient,
    queryFn: async ({ pageParam }) => {
      console.log('MOVIES', { pageParam })
      const url = new URL('https://movies.willcodefor.beer') // Returns 20 items

      url.searchParams.set('page', '1')

      const response = await fetch(url.toString(), {
        headers: { Accept: 'application/json' },
      })
      return response.json()
    },
    queryKey: ['movies'],
    schema,
  })
)

export function Movies() {
  const {
    data: movies,
    fetchNextPage,
    isFetchingNextPage,
  } = useLiveInfiniteQuery(
    (q) =>
      q.from({ alerts: moviesCollection }).orderBy(({ alerts }) => alerts.watchedAt.Time, 'desc'),
    {
      getNextPageParam: (_lastPage, _allPages, lastPageParam) => lastPageParam + 1,
      initialPageParam: 0,
    }
  )

  return (
    <List.Infinite
      data={movies}
	  // This is called when the list reaches a threshold
      fetchNextPage={fetchNextPage}
      id="id"
      isFetchingNextPage={isFetchingNextPage}
      refetch={moviesCollection.utils.refetch}
      renderItem={({ item }) => (
        <View style={{ padding: 16, rowGap: 8 }}>
          <Text>{item.title}</Text>
          <Text numberOfLines={3}>{item.overview}</Text>
        </View>
      )}
    />
  )
}
```

**To Reproduce**

1. All items in collection are loaded in list
2. Keep scrolling down
3. List never updates

**Smartphone:**

- Device: iPhone 17 Pro
- OS: iOS 26
- React Native 0.82.1

[...]

PR #1806, opened 2026-09-11
fix: reject ignored infinite-query page callbacks

KyleAMathews:
## Summary

Remove the silently ignored `getNextPageParam` option from React, Vue, and Svelte's `useLiveInfiniteQuery`. Untyped callers now get a clear error before query construction. Document how an on-demand Query Collection loads server pages through `meta.loadSubsetOptions`.

Addresses #968 and the server-pagination documentation gap in #1657. This is a migration to the existing loading protocol, not a new server-pagination engine.

## Why and approach

The live hook expands an ordered Collection window. Query DB uses `QueryObserver`, not `InfiniteQueryObserver`: it never supplies a transport `pageParam` or invokes `getNextPageParam`. An eager source therefore stopped when its cached rows ran out, despite a config that looked like TanStack Query's infinite-query API.

- Remove the callback from the public types and reject its presence at runtime in all three hooks. Keep generic config wrappers and existing window behavior intact.
- Replace the manual-append pagination example with an on-demand, fixed-server-page drain example. Explain full predicate/order translation, offsets, peek-ahead, unlimited requests, and authoritative endpoint exhaustion.
- Clarify that `initialPageParam` labels result pages; it neither sets a server cursor nor skips remote rows. Explain why explicit refetch can remove manually appended eager rows even with infinite stale time.

**Migration:** delete `getNextPageParam`. For remote loading, use an on-demand collection and make `queryFn` fulfill the exact requested subset. A capped endpoint response is not a fulfilled request merely because its promise resolved.

**Scope/trade-off:** minor bumps for the three pre-1.0 framework packages. No core or Query DB runtime changes, new cursor registry, coverage algebra, or `InfiniteQueryObserver` bridge. Opaque cursor handling remains adapter-local and may require fetching earlier pages again. Retaining a warning-only no-op option was rejected because it continues to suggest unsupported behavior.

## Verification

The callback rejection test was RED before the guard in each framework: query construction ran instead of the migration error. Each now passes, and type tests prove the option is no longer advertised.

The new Query DB/React matrix crosses 4 server page sizes × 3 UI page sizes × 3 row counts (including empty results). It uses the real QueryClient, QueryObserver, collection, compiler/window controller, and React hook. Every page compares full visible IDs and `hasNextPage` against an independent array slice. Additional controls cover eager transport, prefix retention, page labels, invalid capped responses, and explicit-refetch ownership.

The previous tests used fully materialized collections or supplied the callback without asserting its invocation. The new suite crosses that missing transport/framework boundary. It models static numeric ascending ID/rank queries, not every endpoint's ordering or concurrent-write protocol. The deliberately capped-provider control documents a protocol violation; it does not bless truncated responses as successful pagination.

Full runtime suites and their type gates pass:

- React: 201 runtime tests, 19 type cases.
- Vue: 89 runtime tests, 7 type cases.
- Svelte: 97 runtime tests, 4 type cases.
- Query DB: 195 runtime tests and package type checking.
- Focused ESLint and `git diff --check` pass.

```sh
pnpm --filter @tanstack/electric-db-collection build # declarations for Query DB's existing E2E types
pnpm --filter @tanstack/react-db exec vitest run --coverage.enabled=false --maxWorkers=2
pnpm --filter @tanstack/vue-db exec vitest run --coverage.enabled=false --maxWorkers=2
pnpm --filter @tanstack/svelte-db exec vitest run --coverage.enabled=false --maxWorkers=2
pnpm --filter @tanstack/query-db-collection exec vitest run --coverage.enabled=false --maxWorkers=2
```

Review the three hook entry guards/types first, then the server-pagination fixture and React matrix. The Query Collection guide and React overview carry migration guidance; `todo.md` preserves the investigation and design decision.

## Summary by CodeRabbit

- **Bug Fixes**
  - `useLiveInfiniteQuery` now clearly rejects the unsupported `getNextPageParam` option across React, Vue, and Svelte instead of silently ignoring it.
  - Server-pagination behavior now correctly handles on-demand loading, fixed-size server pages, cursors, ordering, and capped responses.

- **Documentation**
  - Updated pagination guidance, including the role of `initialPageParam` and requirements for on-demand query collections.
  - Added guidance for endpoints that return fixed-size pages.

- **Tests**
  - Added coverage for pagination boundaries, rejected options, server-page draining, ordering, and refetch behavior.

### DIFF
--- a/.changeset/reject-ignored-page-callback.md
+++ b/.changeset/reject-ignored-page-callback.md
@@ -1,0 +1,7 @@
+---
+'@tanstack/react-db': minor
+'@tanstack/vue-db': minor
+'@tanstack/svelte-db': minor
+---
+
+Remove the ignored `getNextPageParam` option from `useLiveInfiniteQuery` and reject it with a clear error when passed at runtime. Delete this callback from your config; for server pagination, use an on-demand Query Collection whose `queryFn` fulfills `meta.loadSubsetOptions`. Document fixed-server-page loading and clarify that `initialPageParam` labels result pages rather than setting a server cursor.
--- a/docs/collections/query-collection.md
+++ b/docs/collections/query-collection.md
@@ -777,9 +777,11 @@ request. This example assumes a zero-based page API with a fixed size of 50.
 The endpoint must apply the supplied filters and sorts **before** pagination,
 keep a consistent ordered result while its pages are read, and return
 `nextPage: null` only when it has authoritatively exhausted that result.
-The example `api.listPosts` translates the full `where` expression (including
-cursor predicates) and `orderBy` options into the endpoint's syntax, and rejects
-unsupported expressions. See [QueryFn and Predicate Push-Down](#queryfn-and-predicate-push-down)
+This example uses offset-based pagination. `api.listPosts` translates the full
+`where` expression and `orderBy` options into the endpoint's syntax, and rejects
+unsupported expressions. The separate `cursor` hints are deliberately unused;
+cursor-based adapters must handle those hints alongside `where`, not treat them
+as already included in it. See [QueryFn and Predicate Push-Down](#queryfn-and-predicate-push-down)
 for translation helpers. Do not drop predicates or filter after paginating:
 either changes the requested window.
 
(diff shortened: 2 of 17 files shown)


tannerlinsley (approved):
Reviewed the callback rejection across React, Vue, and Svelte and the server pagination guidance. This makes the previously ignored option fail clearly and points users to the supported adapter flow. The focused pagination tests passed locally across all three frameworks, and CI is green. Looks good to me.

PR #1810, opened 2026-09-11
fix(db): reclaim unused sync without aborting preloads

### LINKED ISSUE #1178: react-db's useLiveQuery seems to leave an extra subscriber in StrictMode

KyleAMathews:
Collections that start syncing before anything subscribes now release their source subscriptions after the GC delay. This fixes query graphs retained by discarded React renders while keeping pending `preload()` calls alive until they settle.

Builds on @edzis's work in #1744 and the [reproduction in #1178](https://github.com/TanStack/db/issues/1178#issuecomment-5326592699), preserving the original commits.

### Root cause and approach

GC previously started only when the last subscriber left. A live query constructed during rendering starts sync immediately, but a suspended or throwing render may never commit and subscribe. With no subscriber departure, that query retained its source subscriptions indefinitely.

Arm GC when sync starts without an owner. A subscriber or pending preload owns the collection; automatic cleanup requires neither to exist. Starting a preload cancels both queued GC and pending idle cleanup, and settlement starts a fresh unused retention period. Preloading already-ready data also renews that period, including when returning a cached preload promise. Explicit `cleanup()` can still abort a preload.

The initial delay is `max(gcTime, 50ms)`. This grace period reduces cleanup between render and commit; it does not guarantee React will commit within 50ms. A later subscriber restarts a reclaimed query. Once the last subscriber leaves, the configured `gcTime` applies without the floor. `gcTime: 0` continues to disable automatic GC.

The shared cleanup queue clears its timer when its last task is canceled and calls `unref()` on Node timer handles so background GC does not keep a finished process alive. Browser timer handles remain supported.

Cleanup also advances the observable state revision. A detached observer that misses the entire cleanup/restart cycle therefore discards its old snapshot even when the reload is empty and returns to the same ready status.

This keeps ownership and reclamation in the collection lifecycle rather than adding framework-specific disposal. It does not change adapter defaults or promise to retain a query for an arbitrarily delayed commit.

### Tests and verification

Lifecycle tests use fake-clock boundaries, source subscription counts, and projection-call assertions. The changed fixtures release their collections. Electric refresh tests verify cancellation of the specific refresh timer instead of counting unrelated shared timers.

Coverage includes discarded renders and late subscribers, orphan reclamation and restart, disabled GC, slow/shared preloads, repeated warm preloads and cached-promise reuse, preload failure and explicit cleanup, pending idle cleanup, Node process exit, and detached wholesale/granular snapshots across an unobserved cleanup and empty reload. Red/green controls reproduced the pending/warm preload, detached-observer, and Node regressions, disabled orphan GC, omitted final-task timer retirement, omitted Electric refresh timeout cancellation, and enabled erroneous child-facade GC while its root remained subscribed; the corresponding regressions failed and passed after restoration.

- DB suite: 5,145 tests across 176 files passed.
- React suite: 183 tests across 14 files passed.
- Electric suite passed; typechecking passed for all three suites.
- React and Electric builds, including dependencies, passed.
- Changed-file lint and formatting checks passed; existing Electric `no-shadow` warnings remain.

```sh
pnpm --filter @tanstack/react-db... --filter @tanstack/electric-db-collection... --workspace-concurrency=2 build
pnpm --filter @tanstack/db exec vitest run --pool-options.threads.maxThreads=2 --maxWorkers=2 --coverage.enabled=false
pnpm --filter @tanstack/react-db exec vitest run --pool-options.threads.maxThreads=2 --maxWorkers=2 --coverage.enabled=false
pnpm --filter @tanstack/electric-db-collection exec vitest run --pool-options.threads.maxThreads=2 --maxWorkers=2 --coverage.enabled=false
```

### Review map

- `collection/sync.ts` and `collection/lifecycle.ts`: initial GC scheduling and preload ownership through all cleanup stages.
- `collection/changes.ts`: cleanup revision invalidation for detached snapshots.
- `collection/cleanup-queue.ts`: last-task cancellation and Node process lifetime.
- DB, React, and Electric tests: lifecycle laws, adapter behavior, and precise resource assertions.
- Public `gcTime` docs, skills, and changeset: retention behavior and unchanged GC opt-out.

---
Fixes #1178

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved automatic cleanup of unused collections and live queries.
  - Added a minimum 50 ms grace period for collections that begin syncing without subscribers.
  - Pending preloads remain active until completion, and preloading ready data refreshes retention.
  - Automatic cleanup can be disabled with non-positive or non-finite `gcTime` values.
  - Background cleanup no longer prevents Node.js processes from exiting.
  - Detached observers now receive fresh snapshots after collections are reloaded.

- **Documentation**
  - Clarified garbage-collection timing, preload retention, and configuration behavior.
  - Updated guidance on when synchronization begins during rendering.

- **Tests**
  - Expanded coverage for cleanup, preloading, process exit, refresh lifecycle, and interrupted renders.


tannerlinsley (approved):
Reviewed unused sync collection, pending preload retention, timer cleanup, and the abandoned render coverage. The lifecycle guards preserve pending preloads while allowing unused sync to be collected. The selected core suites passed locally, 176 tests, along with all 4 uncommitted render tests. CI is green. Looks good to me.

Issue #1820, opened 2026-09-14
test: close remaining oracle contract and native execution gaps


KyleAMathews:
## Scope

Follow-up to the bounded oracle repair in #1808 / #1816. Preserve current contracts and useful tests. This is not a mandate to convert all ordinary tests or to add a universal model. Offline persistence/execution policy is owned by #1659.

### Contract decisions

- [ ] **Effect self-disposal.** External disposal waits for in-flight handlers. Synchronous disposal inside a handler is allowed. A handler returning independent pending work and a handler awaiting its own disposal are different histories: waiting for the former can deadlock the latter. Choose the public rule before applying the candidate wait. Preserve abort, source release, shared disposal outcome and no post-disposal publications. Existing disposal behavior remains unchanged in #1816.
- [ ] **Proxy native mutation and symbol writes.** Decide Date/typed-array mutator and end-to-end symbol-key support versus a clear rejection. Existing standard live Map/Set iteration and supported detachment must remain. Compare actual stored rows, not just a helper result, and do not erase native edits through cloning. Audit IDs A05 R8/R10.

### Real provider and host evidence

- [ ] **Eight progressive cells:** initial explicit snapshot/atomic transition, snapshot loading/swap, multiple snapshot predicates, incremental after swap, predicates during/after snapshot, snapshot dedupe, txid before/after swap, and cleanup/restart during snapshot. Electric's ready hook is after application and cannot prove the pre-publication phase. Use a conforming installed-SDK source gate before application and the original write's actual transaction identity. TrailBase supplies no progressive capability; an early-return pass is not evidence. Preserve existing partial witnesses and make unsupported capabilities explicit.
- [ ] **24 Electric move registrations:** reconcile exact case execution and content/cleanup results from the service CI artifacts. The current PR's successful E2E job is evidence for the job, not proof of every callback-phase claim. Existing finite helper probes are not service framing evidence.
- [ ] **Native host matrix:** distinguish Node SQLite/PowerSync SDK, browser shims, Capacitor/Tauri app builds and actual device/host runs. The 113-law persisted manifest intentionally excludes progressive/move suites. Expo's native-mode skipped branch and unresolved app-specific type/build/runtime cells need explicit owners and receipts; do not infer them from shared registration tests.

### Optional mechanics, after correctness boundaries

- [ ] Separate costly guarded-replay subprocess checks into a named campaign while keeping a real default execution smoke and false-target controls. Do not reintroduce progress starvation; yield only between completed histories/tests.
[...]

Issue #1808, opened 2026-09-11
test: spruce up oracle contracts, metamorphic laws, and async coverage


KyleAMathews:
## Goal

Strengthen our existing oracle suites without growing production code or building a universal testing framework. Prefer several small independent checks over increasingly elaborate reference implementations.

## Proposed follow-up PR

- [ ] Add metamorphic laws where the contract supports them: incremental versus fresh settled query results; backend page-size invariance for completed UI windows; idempotent authoritative replay; independent-key commutation; pending optimistic snapshot stability under unrelated sync/recomputation.
- [ ] Standardize public-observation checks: direct reads versus event-reconstructed rows and downstream results at publication boundaries, no torn transactions, no obsolete-session effects, balanced acquisition/release, and redundant-publication budgets where guaranteed.
- [ ] Generate meaningful async boundaries, not just operation order. Include `begin → truncate → mutation settles → commit`, startup/release reentrancy, replay overlap, cleanup, and late settlement. Reuse controlled promises and existing fast-check scheduler support.
- [ ] Test the checker with deliberate faults: dropped publication, stale completion, incorrect sibling release, and forbidden rebasing of optimistic snapshots. Each important law should reject an appropriate broken subject.
- [ ] Measure executed transitions and meaningful combinations, not just generated commands or test counts. Preserve fixed seeds, random campaigns, shrinking, and replayable minimal histories. Consider fast-check commands only where their preconditions and shrink tracking simplify an existing driver.
- [ ] Record each suite's intended contract, observation boundary, independent model inputs, supported history grammar, and explicit exclusions. Do not infer new product guarantees from a reference model's convenient behavior.

## Constraints and traps

- Optimistic mutations own whole-row snapshots. Do not merge newer synced fields into a pending mutation merely because a patch-based reference model expects it.
- Metamorphic equivalences need explicit preconditions. Fresh-query comparison checks settled results; it cannot replace continuous-observer checks for partial publication or lifecycle corruption.
- Control scheduling where possible. Permit multiple outcomes only when the actual contract allows them, with whole-outcome and causal constraints—not arbitrary per-field old/new choices. No general linearizability checker is proposed.
- Reference engines are useful only on a genuinely compatible semantic subset. A passing comparison can share a bug, and a disagreement is not proof that production is wrong.
- Keep valuable existing regression tests. Fold minimal counterexamples into a shared oracle driver when practical, rather than deleting coverage to improve test counts.
[...]

Issue #1659, opened 2026-07-08
RFC: Hardening the SQLite persistence / Electric sync stack


KyleAMathews:
# RFC: Hardening the SQLite persistence / Electric sync stack

**Status:** Draft · updated 2026-07-13 (cross-RFC ownership clarification)
**Scope:** Bug fixes and hardening only. No new public API surface except where
explicitly flagged as an open decision. The larger architectural reframings
that surfaced during this investigation (replica manifest, readiness state
machine, storage broker) are recorded as future direction in the appendix, not
proposed here.

---

## 1. Background

Fifteen open issues cluster around `persistedCollectionOptions` +
`electricCollectionOptions` + the browser/mobile SQLite persistence adapters:
#82, #865, #1415, #1416, #1443, #1453, #1456, #1478, #1486, #1487, #1498,
#1499, #1560, #1567, #1589.

These are not fifteen independent bugs. Every serious one reduces to a small
set of broken invariants, and the worst share a single shape: **the
persistence layer and the sync adapter each own state the other one wipes.**
The terminal user-visible state is always the same — a collection that is
`ready`, empty, and silent — which is why these reports looked random and
data-specific when they are actually deterministic.

Every mechanism below was ground-truthed with a failing test before being
accepted — see `review-claims.test.ts` in
[`db-sqlite-persistence-core/tests`](https://github.com/TanStack/db/blob/explore-persistence-electric-sqlite/packages/db-sqlite-persistence-core/tests/review-claims.test.ts),
[`browser-db-sqlite-persistence/tests`](https://github.com/TanStack/db/blob/explore-persistence-electric-sqlite/packages/browser-db-sqlite-persistence/tests/review-claims.test.ts),
and
[`react-native-db-sqlite-persistence/tests`](https://github.com/TanStack/db/blob/explore-persistence-electric-sqlite/packages/react-native-db-sqlite-persistence/tests/review-claims.test.ts)
on the
[`explore-persistence-electric-sqlite`](https://github.com/TanStack/db/tree/explore-persistence-electric-sqlite)
branch.

## 2. Verified defects

| #   | Defect                                                                                                                                                                                                                                                                                                                                      | Issues                          | Verification                                                                                               |
| --- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------- | ---------------------------------------------------------------------------------------------------------- |
[...]

PR #1822, opened 2026-09-14
fix: preserve accepted snapshots and fence stale replay reads

KyleAMathews:
## Summary

Fix three boundary failures found after #1816 merged: stale offline reads can replay already acknowledged work, top-K replacements can lose sparse-array/RegExp changes, and truncate can discard an accepted insert hidden by a later optimistic delete.

Extends the existing oracles first, then makes narrow runtime repairs: **23 net runtime lines**, no new public API, and patch changesets for core, IVM, and offline transactions.

## Why and how

- **Replay:** scheduler dedupe forgets an ID after durable deletion, but an older asynchronous outbox scan can still return it. A per-executor acknowledgment revision causes overlapping scans to reread before admission. Tests hold actual storage reads across successful deletion, with concurrent scans and unfinished peers that must still execute. No permanent completed-ID cache or exactly-once guarantee across independent owners.
- **Top-K:** enumerable keys alone miss sparse-array length and RegExp source, flags, and position. Compare those fields and guard the optional global `File` constructor. Existing value relations now check actual retained graph output, equal controls, both delta orders, and hosts without File. No structural hashing is added.
- **Retention:** truncate captured visible upserts, so an optimistic delete hid the accepted snapshot from retention. Preserve the existing accepted snapshot until ordinary sync retires it. The oracle crosses truncate before/during/after deletion, explicit rollback/provider rejection, and later sync/key reuse. No whole-row model change or synced-field rebasing.

The coverage map records these domains. Failed durable deletion and other offline policy questions remain with #1659; broader oracle/native-host gaps remain with #1820.

## Verification

Base: `150bde99ceb45b764f1b125753d7d31ef0cc733f` (origin/main after #1816).

Before fixes, the expanded top-K suite had **13 failing / 93 passing** tests; the new replay family had **2 failing** tests; retention had **2 failing / 4 passing** timing cells. After fixes, their complete suites pass: top-K **106**, leadership **24**, retention **38**. The original retention seed/path also passes unchanged.

Built IVM and core, then ran package suites with Node 24.5.0, installed Vitest 3.2.4, package configs, and two workers:

| Package | Reported passing tests |
| --- | ---: |
| Core | 6,155 |
| IVM | 549 |
| Offline transactions | 140 |
| Query DB | 454 |
| TrailBase | 67 |
| Electric | 820 |
| PowerSync | 136 |
| SQLite persistence core | 176 |

Standalone core/IVM/offline typechecks, enabled package Vitest typechecks, changed-file lint, and formatting pass. Query DB/Electric report additional collected type entries; the table counts passing tests only.

Reproduce from each package with `node ../../node_modules/vitest/vitest.mjs run --coverage.enabled=false --maxWorkers=2`, after building IVM and core. Standalone types: `node node_modules/typescript/bin/tsc --noEmit -p packages/<package>/tsconfig.json`.

PowerSync includes native SQLite/SDK tests. This is **not** a new service-backed Electric E2E run or mobile-device certification. CI remains a separate gate.

Related: #1657 (kept open for merged combined verification), #1808, #1816.

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved eligible accepted local inserts through truncation and subsequent edits, while correctly retiring inserts created after the snapshot.
  - Improved ordered-query replacements for sparse arrays and regular expressions, including regex state and environments without a global `File`.
  - Prevented transactions removed during replay from being scheduled or executed again.

- **Tests**
  - Added coverage for retention, ordered-query replacement, leadership replay, and durable acknowledgment scenarios.

- **Documentation**
  - Expanded contribution guidance with additional post-merge review areas.


tannerlinsley:
The remaining clear/read race is fixed in #1836. Each successful deletion is now fenced immediately, including while another deletion is pending or fails. Both new regression cases failed before the fix; all 148 offline transaction tests, types, lint, and the package build pass. The follow-up is running CI.

PR #1824, opened 2026-09-15
feat(query-db): add cached opaque cursor pagination

### LINKED ISSUE #863: Add collection-level pagination metadata support to avoid wasteful peek-ahead pattern
## Problem

Currently, `useLiveInfiniteQuery` uses a "peek ahead" pattern to determine if there's a next page:

```typescript
// Request one extra item
const expectedLimit = loadedPageCount * pageSize + 1 // +1 for peek ahead

// Check if we got the extra item
const hasMore = dataArray.length > totalItemsRequested
```

**This works well for:**
- **Electric collections**: Push-based sync, no control over what's available (though we could also modify this to avoid the extra fetch)
- **LiveQueryCollection**: Reactive local computation, no external API

**But it's wasteful for QueryCollection** where users control the `queryFn` and their API already returns pagination metadata:

```typescript
queryFn: async (ctx) => {
  const response = await fetch(url)
  const json = await response.json()
  
  // API already tells us hasNextPage!
  return json.items  // { items: [...], hasNextPage: true, totalCount: 100 }
}
```

The extra item fetch wastes:
- Network bandwidth
- Database query costs
- API rate limits

## Proposed Solution

Add collection-level pagination metadata storage:

```typescript
// Collections expose pagination metadata API
interface Collection {
  setPaginationMeta(sessionId: string, meta: PaginationMeta): void
  getPaginationMeta(sessionId: string): PaginationMeta | undefined
}

type PaginationMeta = {
  hasNextPage?: boolean
  hasPrevPage?: boolean
  nextCursor?: string
  prevCursor?: string
  totalCount?: number
}
```

### Session ID from Query Predicates

Session ID should be a deterministic hash of:
- Collection ID
- WHERE clauses
- ORDER BY clauses
- **Excluding**: limit, offset

This enables:
- Same query = same session = reuse pagination state
- Different query = different session = isolated pagination
- Multiple concurrent infinite queries on same collection don't interfere

Implementation using existing `serializePredicates` pattern:
```typescript
const sessionId = hashKey([
  collectionId,
  JSON.stringify({ where: opts.where, orderBy: opts.orderBy })
])
```

### Usage in QueryCollection

```typescript
queryFn: async (ctx) => {
  const response = await fetch(url)
  const json = await response.json()
  
  // Store pagination metadata
  ctx.collection.setPaginationMeta(ctx.sessionId, {
    hasNextPage: json.hasNextPage,
    totalCount: json.totalCount,
    nextCursor: json.nextCursor  // For future cursor support
  })
  
  return json.items
}
```

### Updated useLiveInfiniteQuery

```typescript
// Check for pagination metadata first
const meta = queryResult.collection.getPaginationMeta(sessionId)
const hasNextPage = meta?.hasNextPage ?? fallbackToPeekAhead()

// Adjust window: no +1 if metadata exists
const limit = meta ? loadedPageCount * pageSize : loadedPageCount * pageSize + 1
```

## Broader Context: Cursor-Based Pagination

This also lays groundwork for cursor-based pagination (from related discussion):

### Cursor Storage Concept
- Store cursor per infinite query session
- QueryFn retrieves stored cursor and passes to API
- If page under-fills, keep fetching until limit met
- Requires consistent ORDER BY matching cursor's order

### Metadata Columns Approach
To support multiple concurrent queries with different ORDER BY on same collection:
- Session ID isolates pagination sets
- Each session tracks its own cursor/position
- Prevents "page mixing" when different orderings are used

### Primary Key Tie-breaker
- Using PK as final ORDER BY element provides deterministic ordering
- Enables DB-native cursor positioning
- Could be enforced when cursor mode is enabled

### Cursor + Offset Hybrid
Potentially support starting with offset then switching to cursor:
```typescript
// Jump to page 5 via offset
// Then infinite scroll from there via cursor
useLiveInfiniteQuery(query, { 
  pageSize: 20, 
  initialOffset: 100  // Start at item 100
  // Then use cursor from API
})
```

## Benefits

- ✅ **QueryCollection**: No wasteful +1 fetch when API provides hasNextPage
- ✅ **Backward compatible**: Collections without metadata fall back to peek ahead
- ✅ **Foundation for cursors**: Enables future cursor-based pagination
- ✅ **Session isolation**: Multiple concurrent infinite queries work correctly
- ✅ **Extensible**: Can store other metadata (totalCount, sync progress, etc.)

## Open Questions

1. **Storage location**: Collection property vs separate registry?
2. **API surface**: Methods on collection vs separate pagination manager?
3. **Cursor implementation**: Should we tackle cursors now or separate issue?
4. **Order validation**: Should we detect/prevent ORDER BY changes within session?
5. **Cursor + offset**: Should both be supported simultaneously or exclusive modes?

## Related Use Cases

Beyond hasNextPage, this could also support:
- Total count for "X of Y" UI
- Sync progress tracking
- Stream positions/snapshots
- Transaction IDs for Electric collections
- Custom pagination metadata from APIs

KyleAMathews:
## 🎯 Changes

Add `createCursorPager` to Query DB Collection so opaque-cursor endpoints can fulfill DB's offset/limit requests. Fresh page growth reuses cached pages and fetches only the missing suffix.

This fully closes #863 with the chosen design: cached cursor loading over Query's cache, retaining N+1. The broader collection-metadata/no-peek proposal is intentionally not pursued, not deferred work needed to finish the issue.

### Approach and contracts

- **Use Query's cache, not another registry.** Query owns page freshness, retries, invalidation and GC. Each source/filter/order has a dedicated key that excludes the requested window. New pager instances share cached pages and acquisitions; only a retained instance shares its local read queue.
- **Keep row and page namespaces separate.** Use sibling keys such as `['posts', 'rows']` and `['posts', 'pages', scope]`. For a forced refresh, **cancel the shared resource prefix, then invalidate it**. Invalidation alone can join an old append and mark stale data fresh; `utils.refetch()` alone can reuse fresh pages.
- **Return complete windows.** Only `nextCursor: null` means exhaustion. Short or empty nonterminal pages continue loading; malformed or repeated continuations fail before cache publication. Preserve the full raw page prefix despite inherited `maxPages` or `select` defaults.
- **Separate reader cancellation from transport cancellation.** Reader abort rejects promptly without cancelling shared transport. Explicit Query cancellation rejects acquisition waiters with `AbortError`, not Query's internal control error that can strand an enclosing row query. Silent cancelling refetches move waiters to the replacement. Browser and server acquisition phases use the same resolved retry policy.
- **Protect other cache formats during manual row writes.** The raw-array writer leaves existing non-array records untouched, including freshness and invalidation state. Wrapped-response writes and empty row-cache seeding still work. This is a defensive guard, not support for arbitrary mixed cache schemas under one prefix.

### Trade-offs and non-goals

Stale reads refresh the entire loaded prefix, even for a shallow window. Query's freshness clock includes successful page growth; inactive page-cache GC can run while outer collection rows remain visible. Applications translate filters/order and supply a stable, totally ordered backend sequence—TTL does not create snapshot consistency.

Keep N+1. No metadata API, new D2 graph, total counts, previous-page API or offset-to-cursor jump. No-peek experiments remain test-only evidence. Core and framework hooks are unchanged.

### Verification

The independent reference filters, sorts and slices a complete relation. Generated histories exercise real Query caches and QueryCollection/window-controller paths, including ties, defaults, expiry, cancellation, replacement, recovery and manual writes. Review-driven oracle extensions reproduced bugs before fixes; a no-op cache-write mutant also fails because preserving page contents alone still clears invalidation. Refresh checks cover both new pagers and reads queued on a retained pager.

Receipts for `cbc987a5b`, based on `main` at `3b991173f` (Query Core 5.90.20):

- **374 package tests / 14 files pass**, including the default random-seed lane.
- **120 cursor tests / seven files pass at 100×**, seed 863: **174,000 generated histories**. This includes 18 retained no-peek experiment tests, not shipped metadata behavior.
- Package types, build and formatting pass. ESLint has no errors and two pre-existing `no-shadow` warnings in unchanged code.

Actual commands from `packages/query-db-collection`, with core and db-ivm built:

```sh
../../node_modules/.bin/vitest run --typecheck.enabled=false --maxWorkers=2
TANSTACK_DB_ORACLE_RUNS_MULTIPLIER=100 TANSTACK_DB_ORACLE_SEED=863 \
  ../../node_modules/.bin/vitest run tests/cursor-pagination*.test.ts \
  --typecheck.enabled=false --maxWorkers=1 --testTimeout=60000
../../node_modules/.bin/tsc --noEmit -p tsconfig.json
../../node_modules/.bin/vite build
```

Identical browser-ESM/esbuild/ES2020 diagnostic imports at this head: `queryCollectionOptions` alone is **46,655 minified / 15,000 gzip bytes**; adding the pager is **52,079 / 17,017** (+5,424 / +2,017). These are opt-in import measurements, not universal application bundle sizes.

`cursor-pagination.ts` and `index.ts` implement/export the helper; `query.ts` contains the small raw-cache guard. The guide explains integration. Oracle suites, fixtures and replay registration carry the evidence. The [suite README](https://github.com/TanStack/db/blob/cbc987a5bad08ec119816c2b6c6f4d776f004d9e/packages/query-db-collection/tests/cursor-pagination/README.md) and [loss audit](https://github.com/TanStack/db/blob/cbc987a5bad08ec119816c2b6c6f4d776f004d9e/packages/query-db-collection/tests/cursor-pagination/LOSS-AUDIT.md) preserve red/green receipts and limits. The tests do not establish arbitrary endpoint consistency or framework rendering behavior.

## ✅ Checklist

- [x] Tested locally with the package's underlying Vitest runner, using the commands above instead of the `pnpm test` launcher; types checked separately.
- [x] Added oracle coverage and verified red/green behavior for the fixes.
- [x] Updated integration docs and the changeset.

## 🚀 Release Impact

- [x] Affects published code; `.changeset/add-cached-cursor-pagination.md` covers `@tanstack/query-db-collection` as a **patch**, per maintainer decision.
- [ ] Docs/CI/dev-only (no release).

---
Closes #863


tannerlinsley (approved):
Reviewed the current head and accept the documented cursor API tradeoffs: sequential windows, full loaded-prefix refresh, separate page-cache keys, and reader versus transport cancellation. All 120 cursor tests and standalone types passed locally; the combined cursor/cache-ownership run also passed 168 tests. CI is green. Approved.

PR #1825, opened 2026-09-15
fix(solid-db): preserve live result identity during reconciliation

### LINKED ISSUE #1524: solid-db rendering issues when not using key named `id`

KyleAMathews:
`useLiveQuery` now keeps one stable Solid row store per live Collection result key. This fixes custom-key rows corrupting one another when an update reorders them, while also keeping distinct union results separate when they expose the same public `$key`.

## Root cause

Solid's `reconcile` defaults to matching rows by `id`, which does not work for collections keyed by another field such as `_id`. Matching by the row's public `$key` fixes that reproduction but is not a complete identity rule: live-query results can preserve an upstream `$key` while the Collection assigns a different result key, and multiple `unionAll` branches can therefore publish distinct results with the same `$key`.

The adapter was asking a public row field to carry an identity that belongs to the live Collection. Under a collision, Solid could retain or mutate the wrong mapper and DOM node even after the core Collection contained the correct survivor.

## Approach

- Maintain an adapter-private map from each live Collection entry key to one stable Solid row store.
- Reconcile updates into that row store without interpreting a public field as its identity.
- Rebuild the ordered result array from those stable row objects, following Collection entry order.
- Retire stores for deleted result keys and clear them when the query becomes disabled.
- Strengthen the original custom-key reorder regression with mapper/DOM-node identity checks.
- Add a public `unionAll` oracle where two distinct result keys deliberately share the same upstream `$key`.

## Key invariants

- Exactly one stable Solid row object exists per current live Collection result key.
- Public `$key` values are neither overwritten nor required to be unique in a derived result.
- Result order and cardinality continue to match the live Collection.
- Reorders preserve each logical row's mapper and DOM-node ownership.
- Deleted result identities are retired, and rendered roots still clean up normally.

## Non-goals

- No change to core live-query key or virtual-property semantics.
- No new public row field, wrapper, symbol, or API contract.
- No changes to other framework adapters.

## Trade-offs

This keeps a small adapter-private map and one Solid store per live row. That adds linear bookkeeping, but avoids leaking internal result keys into public values and keeps updates fine-grained. Using public `$key` directly is simpler, but is invalid for derived results where public and result identity differ.

## Verification

Commands below were run from `packages/solid-db` unless noted otherwise.

### RED on the pre-fix implementation

```bash
../../node_modules/.bin/vitest run tests/useLiveQuery.test.tsx --config vite.config.ts --configLoader runner --maxWorkers=2 --coverage.enabled=false --typecheck.enabled=false --reporter=verbose -t 'keeps union rows with colliding public keys tied to their live result identities'
```

Result: `1 failed | 41 skipped`. Setup proved two distinct live result keys with the same public `$key`, and the core Collection plus hook both reached one surviving Right row. The final identity checkpoint retained the deleted Left row's mapper/node: `mapper-1` and `retainedOwnNode: false`, instead of `mapper-2` and `true`.

### Targeted hostile mutant

Temporarily restored the stale implementation's choice:

```ts
reconcile(Array.from(currentCollection.values()), { key: `$key` })(previous).filter(Boolean)
```

Running the same command produced the same identity failure: `1 failed | 41 skipped`, with `mapper-1` / `false` instead of `mapper-2` / `true`. The mutant was then removed.

### GREEN

```bash
../../node_modules/.bin/vitest run tests/useLiveQuery.test.tsx --config vite.config.ts --configLoader runner --maxWorkers=2 --coverage.enabled=false --typecheck.enabled=false --reporter=dot -t 'keeps (custom-key rows distinct|union rows with colliding public keys)'
```

Result: `2 passed | 40 skipped`.

```bash
../../node_modules/.bin/vitest run --config vite.config.ts --configLoader runner --maxWorkers=2 --coverage.enabled=false --typecheck.enabled=false --reporter=dot
```

Result: `72 passed` across 2 files.

```bash
../../node_modules/.bin/vitest run tests/conformance.test.tsx --config vite.config.ts --configLoader runner --maxWorkers=2 --coverage.enabled=false --typecheck.enabled=false --reporter=dot
```

Result: `30 passed`.

From the repository root:

```bash
pnpm exec tsc -p packages/solid-db/tsconfig.json --noEmit --rootDir .
pnpm --filter @tanstack/solid-db build
pnpm --filter @tanstack/solid-db lint
./node_modules/.bin/prettier --experimental-cli --check packages/solid-db/src/useLiveQuery.ts packages/solid-db/tests/useLiveQuery.test.tsx
git diff --check
git diff HEAD^ HEAD --check
```

Result: all commands exited 0. TypeScript produced no diagnostics; the Solid package built with declarations; ESLint and Prettier passed; both diff checks were clean.

### Harness limitations

- The default coverage-enabled run could not start because the managed sandbox denied deleting a pre-existing `packages/solid-db/coverage/index.html`; the complete runtime suite was rerun with coverage output disabled, so this PR makes no coverage-percentage claim.
- Vitest's package-root built-in typecheck reports 137 pre-existing cross-package `rootDir` errors. The repository-rooted standalone TypeScript command above passed cleanly.

## Files changed

- `.changeset/fix-solid-custom-key-reconciliation.md` — records the Solid adapter patch.
- `packages/solid-db/src/useLiveQuery.ts` — keys stable private row stores by live Collection result identity.
- `packages/solid-db/tests/useLiveQuery.test.tsx` — covers custom-key reorder identity and colliding public keys across union results.

Reported by @fehmer in #1524. This supersedes #1598 from @Leonabcd123, whose earlier `$key`-based fix and regression established the initial direction; this replacement is implemented against current `main` and covers the additional public-key collision case.

---
Fixes #1524

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Solid live queries so rows retain their identity when custom-key records change order.
  - Fixed cases where multiple query results share the same public key, keeping each row associated with the correct data.
  - Deleting one of several similarly keyed rows now removes only the intended row.
  - Switching between collections now correctly refreshes rows with overlapping keys, preventing stale row identity from carrying over.


tannerlinsley (approved):
Reviewed the current head. Private result-key identity, reorders, union key collisions, and collection switching look good. All 73 Solid tests and the standalone type check passed locally. CI is green. Approved.

PR #1826, opened 2026-09-15
fix(query-db-collection): isolate query cache lifecycle ownership

KyleAMathews:
On-demand Query Collections now keep cache ownership and post-write readiness correct across shared QueryClients, co-owners, deferred cleanup, request reuse, error recovery, cache replacement, cancellation, and custom query hashes. Manual writes remain immediately visible, while only an authoritative provider fetch that started after the write may replace them.

## Root cause

The original direct-write fix coupled authority to this collection's wrapped `queryFn` and recursively refetched until that wrapper reported completion. That excluded successful foreign/shared fetches, treated cancelled raw promises as authoritative, retained stale Query identity after cache replacement, and could retry forever after errors. Its write path also skipped unobserved prefetched or hydrated sibling scopes and scanned the whole QueryClient.

## Approach

- Observe Query Cache fetch and success actions and assign a shared fetch-start epoch.
- Require a successful fetch whose start epoch is later than the manual write, preserving the pre-write request-reuse boundary.
- Share authority and cache-owner identity across Query Collections which own the same Query while keeping unrelated collections isolated.
- Recheck the observer's current Query after settlement so cache removal/replacement cannot strand authority on a retired object.
- Bound post-write refetching: one normal attempt and at most one follow-up for a reused or reverted request; errors terminate the attempt.
- Track relevant cache entries as they are added and removed, then operate directly on those Query objects.
- Remove unobserved matching sibling scopes so prefetched, hydrated, and `setQueryData` entries cannot retain ghost rows.
- Lazily materialize the normalized collection snapshot so on-demand writes do not allocate data they never use.

## Key invariants

- A fetch started before a manual write cannot establish post-write authority.
- Foreign/shared successes count; cancellation, reversion, and errors do not.
- Query replacement cannot create a refetch loop or readiness hang.
- One collection does not refetch or remove a Query owned by another collection.
- Unobserved matching scopes are evicted after a manual write.
- Overlapping writes remain bounded.

## RED/GREEN evidence

The external-review evaluation reproduced the failures with controlled probes before implementation. Permanent oracle regressions now cover:

- a failed post-write refetch stopping at exactly one attempt;
- overlapping writes producing exactly three starts and one abort;
- successful foreign fetches satisfying readiness;
- cache clear/replacement completing with exactly one replacement fetch;
- cancellation/reversion requiring another authoritative fetch;
- unobserved sibling eviction;
- zero whole-cache scans and zero normalized-snapshot materializations on the on-demand path.

Existing cross-collection isolation, disabled-observer, deferred cleanup, pre-write request-reuse, custom-hash, reset, and error-recovery regressions remain green.

Final verification:

```bash
pnpm --filter @tanstack/query-db-collection test:oracles -- --pool-options.threads.maxThreads=2
pnpm --filter @tanstack/query-db-collection test -- --pool-options.threads.maxThreads=2
pnpm --filter @tanstack/query-db-collection lint
git diff --check
```

All commands pass with no type errors.

Closes #1497.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved direct-write cache handling for on-demand collections by revalidating active queries and removing inactive or disabled entries.
  - Isolated cache ownership across collections to prevent unrelated queries from being modified or removed.
  - Improved handling of write-triggered refetches, errors, cancellations, custom query hashes, and remounts.

- **Documentation**
  - Clarified direct-write behavior in eager and on-demand sync modes, including refetch controls and query function guidance.


tannerlinsley:
The review reproduced one cache-tracking gap: a sibling entry prefetched after collection construction but before sync starts survives a manual delete. I am adding a focused startup/restart regression and fixing tracking initialization on this branch, then will rerun the ownership checks before approval.

tannerlinsley (approved):
Reviewed the original changes and the cache-start follow-up. Delayed startup and restart regressions now pass, along with all 283 Query Collection tests and standalone types. The fix rebuilds tracking once per sync start and preserves the no-scan manual-write path. Approved; waiting for the new CI run before merging.

Issue #471, opened 2025-08-29
useLiveQuery returning stale results when querying local storage collection


craigsmitham:
Changes to a local storage collection are successfully being persisted to local storage and `useLiveQuery` will return the updated data, but then revert back to the pre-modified data. Refreshing the window/tab then reflects the expected updated state. If I switch to a local only collection the issue no longer occurs. This doesn't happen everywhere I'm using `useLiveQuery` with the same local storage collection, so there is something unique about the way I'm using it combined with local storage. I'm not sure if I'll be able to make a repro case but curious if there is any idea of what might be going on ...

samwillis:
Hey @craigsmitham, thanks for the report. I've tried to reproduce it but not be able to so far. Could you check if you can still trigger it and is so show me the code that does? A reproduction would be ideal.

sudoskys:
Are you using electricSQL?

If so, I think it might be something like this?

https://github.com/TanStack/db/discussions/282

Changes are "loaded" from the HTTP cache

KyleAMathews:
Closing as stale because maintainers could not reproduce this and no reproduction or requested code followed in nearly a year. Please reopen or file a fresh issue with current versions and a minimal reproduction if the stale-result behavior remains.

PR #1840, opened 2026-09-16
fix(query-db-collection): publish mutation refetches

### LINKED ISSUE #478: 🐛 Bug: ⁠queryCollection stops re-rendering after rapid inserts with delayed ⁠onInsert

KyleAMathews:
Fixes Query Collection writes from idle state and guarantees that the newest authoritative cache result reaches the source Collection and downstream live views without a stale or torn intermediate publication. Mutations rejected by locally decidable validation remain inert, while valid state-dependent mutations hydrate synchronously before their row checks.

## Root cause

Core mutation startup did not distinguish locally decidable rejection from validation that needs synchronized Collection state. Query Collection direct-write utilities also needed a per-Collection way to request idle startup.

Separately, Query results were queued behind an older application that could be parked on a deferred commit or persisted-row scan. A focus refetch or mutation refetch could therefore publish a stale snapshot, settle `loadSubset` too early, or lose ownership rollback when the older generation was cancelled. Mutation-specific replacement did not cover focus refetches and could lose an authoritative server delete.

The first general supersession repair exposed a narrower phase bug: if a newer result arrived after core publication had begun, restoring the older application's provisional ownership could rewind the owner maps while already-published rows remained visible. The same boundary exists while a persisted wrapper is still awaiting durable completion.

## Approach

- Validate handlers, schemas, keys, empty key sets, and the update callback before starting an idle Collection.
- Start accepted insert/update/delete operations at the narrow mutation boundary; update/delete hydrate synchronously before checking state-dependent keys.
- Bind only Query Collection's existing `write*` utilities to the internal post-construction idle-start callback. Read-only utilities remain side-effect free.
- Supersede an older result application immediately when any newer successful cache result arrives, rather than special-casing mutation handlers.
- Forward `loadSubset` settlement to the newest application and restore ownership only for the cancelled generation.
- Fence ownership rollback at core's publication point of no return, including reentrant and persisted-wrapper paths, so an outer publication cannot rewind ownership or replace the settlement/error slot of a newer application.

## Key invariants

- Mutations rejected by locally decidable validation do not start sync or change lifecycle state.
- Valid idle update/delete operations synchronously hydrate before missing-row checks.
- Arbitrary utility reads do not start Query Collection sync.
- Every source and derived-view notification represents a complete semantic snapshot.
- The newest cache result supersedes an older application regardless of whether it came from mutation, focus, or direct refetch.
- `loadSubset` waiters follow the newest application instead of rejecting on supersession.
- A cancelled generation rolls back only its own provisional ownership; a reentrant newer generation remains authoritative.
- Once core publication begins, later supersession or listener failure cannot roll ownership behind visible rows; before publication, failed work still releases provisional ownership.
- Cleaned-up direct writes and late handlers retain base behavior and do not restart the Collection.
- Per-sync cache ownership and post-write authority from #1826 remain intact.

## Non-goals and unsupported boundaries

- No public API, new package export, dependency, queued-write protocol, or compatibility branch is added. The existing exported `@internal` sync-factory callback declaration gains its post-construction idle-start callback parameter; this emitted internal type change is included in package measurements rather than described as zero API delta.
- Direct `collection.utils.write*()` calls during internal `_deferSyncStart()` render/materialization coordination remain unsupported and fail explicitly with `SyncNotInitializedError`. The adapter has not entered its sync function, so no manual-write context exists. Starting immediately could expose a partially materialized graph; queuing would add ordering, replay, error, and cleanup semantics outside this fix. Framework commit/effect resumes startup normally.
- The internal sync-factory callback is post-construction-only; construction-time invocation remains unsupported rather than adding queued startup state.
- The inert-rejection guarantee covers locally decidable validation on idle Collections. State-dependent update/delete and hydrated-duplicate checks intentionally run after synchronous startup; existing cleaned-up mutation restart and error-precedence behavior are otherwise unchanged.
- #471 is not claimed because its missing use-pattern precondition was not reproduced.

## Review findings

The supplied external review contained eight independent rows. ER-01, ER-03, ER-05, ER-06, ER-07, and ER-08 are fixed here. ER-02's claimed retained-overlay collapse was refuted on the exact controlled persisted-scan path, although duplicate scanning remains a non-contractual performance idea. ER-04's claimed duplicate authoritative application was refuted with stable result-object identity and application/staging counters.

The review source did not identify its author, so no identity is inferred. Its raw preface also mentioned one refuted and two dropped candidates without supplying their claims, paths, or proposed fixes; that evidence gap remains explicit rather than inventing findings or credit.

Adversarial review additionally found synchronous reentrancy and settlement-slot hazards in the first general supersession repair. Both have permanent behavior-named oracles and hostile-mutant receipts.

An anonymous background review then found a high-severity orphaned-row regression in that repair: supersession after core publication could restore ownership even though the rows had already landed, and the persisted-GC/cold-revalidation path had the same exposure while durable commit was pending. Exact base/current probes confirmed both BGR-01 and BGR-02. The repair now retains the exact core transaction phase locally and permits rollback only before `applicationStarted`; post-publication listener errors, readiness generation, and pre-publication commit failure have independent permanent controls. The review also supplied four clean negative checks, all confirmed. Its author was not identified, so no identity is inferred.

The final CodeRabbit review found two further issues. An idle collection can already contain `initialData`, so duplicate inserts now reject before startup while retaining a second post-start check for keys discovered by hydration. An already-attached `loadSubset` waiter now observes failure of the current application; this reuses the existing failure record rather than adding the suggested per-application bookkeeping object.

The exact-head CodeRabbit rescan produced no actionable comments. Its residual risk summary said invalid direct writes can start idle sync. Controlled probes confirm the side effect but refute a blanket inert-write rule: update/delete may need startup to hydrate a valid target, and schema normalization may create insert/upsert keys. Prior art defines automatic startup for write-method entry but no invalid-idle contract. Adding two-phase preflight or a special untyped `writeBatch` rule would expand behavior and shipped machinery, so this PR leaves that design question outside scope.

## Core oracle ownership

The core mutation-startup law now has a dedicated registered owner rather than living as 441 examples in the conventional lifecycle suite. Existing cleanup/restart, subscription lifecycle, state-retention, optimistic-transaction, and publication owners were inspected; none coherently owns public idle mutation admission. The new 19-case owner has its own review card, finite ready/throwing adapter model, path and observation contract, hostile mutants, and explicit exclusions. The conventional suite is byte-identical to the refreshed base, the new owner is reached by `@tanstack/db`'s `test:oracles` campaign, and the coverage map names its exact domain. Query Collection's additional lifecycle laws were received by its existing ownership owner; that package's existing `test:oracles` registration remains unchanged.

## Shipped weight

Exact refreshed base `2841fde0fb4ef1383553df9545a2288bd1540fe2` to candidate `5ca8f919fc79c0acaeab87031aa0c9f317684d5d`. Raw values sum emitted production `.js`/`.cjs`; compression sums each file using `/usr/bin/gzip -n -c` and Node zlib Brotli under the same Vite 7.3.2 toolchain.

| Build | Output | Base raw | Head raw | Raw delta | Gzip delta | Brotli delta |
|---|---|---:|---:|---:|---:|---:|
| Normal | core ESM | 780,858 B | 781,353 B | +495 B | +97 B | +102 B |
| Normal | core CJS | 803,797 B | 804,306 B | +509 B | +94 B | +60 B |
| Normal | Query ESM | 72,292 B | 73,247 B | +955 B | +214 B | +173 B |
| Normal | Query CJS | 73,902 B | 74,827 B | +925 B | +209 B | +180 B |
| Minified | core ESM | 559,232 B | 559,631 B | +399 B | +107 B | +83 B |
| Minified | core CJS | 372,779 B | 373,114 B | +335 B | +77 B | +63 B |
| Minified | Query ESM | 42,488 B | 42,881 B | +393 B | +142 B | +140 B |
| Minified | Query CJS | 28,191 B | 28,479 B | +288 B | +137 B | +144 B |

Npm tarballs include shipped source, source maps, declarations, and package metadata as declared by each package:

| Build | Package | Packed | Unpacked |
|---|---|---:|---:|
| Normal | `@tanstack/db` | 1,550,545 → 1,551,195 B (+650 B) | 7,375,972 → 7,379,983 B (+4,011 B) |
| Normal | `@tanstack/query-db-collection` | 129,533 → 131,004 B (+1,471 B) | 626,140 → 633,723 B (+7,583 B) |
| Minified | `@tanstack/db` | 1,563,826 → 1,564,520 B (+694 B) | 6,763,971 → 6,767,820 B (+3,849 B) |
| Minified | `@tanstack/query-db-collection` | 130,485 → 131,973 B (+1,488 B) | 556,500 → 562,892 B (+6,392 B) |

The final deletion pass localized the private transaction phase reference, saving 23/23 B normal and 29/28 B minified Query ESM/CJS raw, and reused the existing core state alias, saving 10/10 B normal and 18/18 B minified core ESM/CJS raw. Earlier passes removed mutation-specific replacement, collapsed result tracking to one controller map, removed redundant rollback clears, derived write-helper names from the in-scope utility object, and reused the existing failure map. A factored duplicate assertion had worse compressed output and a per-call closure. Closure-only phase tracking, after-receipt tracking, baseline reconstruction, unconditional readiness, and deleting either duplicate checkpoint fail hostile oracles or add more work/state. The remaining startup hook and phase/settlement machinery are the smallest clear supported-path repair found.

The zero-growth target is not met. Positive production deltas remain in both packages, and the positive `@tanstack/db` delta is an explicit merge hold pending user approval. Tests and documentation are not netted against shipped production growth; package metadata and all tarball contents are reported above.

## Verification

Local receipts on the candidate:

- Query ownership oracle: 72/72 through explicit exact-worktree core/IVM source aliases
- Registered Query Collection oracle campaign: 8 files, 213 tests
- Core mutation-startup owner plus conventional lifecycle collateral: 73/73
- Full registered `@tanstack/db` oracle campaign: 38 files, 2,084 tests
- All Query Collection runtime tests: 14 files, 425 tests
- Core standalone TypeScript: green
- Normal/minified builds and declarations for both packages: green
- Changed-file formatting and `git diff --check`: green
- Changed-file ESLint: zero errors and five warnings (three existing `no-shadow`; two test-only `require-await`). The earlier `while (true)` error was removed by a behavior-equivalent `for (;;)` deletion pass.
- Query standalone TypeScript reaches the intended source and is blocked only by absent optional E2E dependencies: `@tanstack/electric-db-collection`, `pg`, and `@standard-schema/spec`

Hostile mutants killed: chained stale publication, missing pre-publication ownership rollback, unconditional post-publication rollback, transaction capture after durable commit, missing catch-side rollback, unconditional catch-side rollback, missing waiter forwarding, eager rejected-mutation startup, too-late update/delete startup, pre-start duplicate omission, post-start hydrated-duplicate omission, over-broad duplicate lookup, repeated startup, swapped update/delete dispatch, mutation application before startup failure, unconditional cleaned-up restart, missing reentrant point-of-no-return, missing newest-settlement guard, missing active-waiter failure propagation, unconditional readiness, and unfenced controller cleanup. The historical core product fails 9 of the dedicated owner's 19 cases.

The published branch was merged normally with current `origin/main` at `2841fde0` rather than rewriting PR history. Merged #1824, #1826, #1831, #1832, #1833, #1834, #1835, #1842, and #1847 are ancestors of the integrated base and are not duplicated in this branch. Both #1847's cold-join owner and this PR's mutation-startup owner remain registered.

## Files changed

- `packages/db/src/collection/index.ts`: makes the internal startup callback idle-only and preserves the construction boundary.
- `packages/db/src/collection/mutations.ts`: starts accepted mutations after local validation and before state-dependent checks.
- `packages/db/tests/collection-mutation-startup-oracle.test.ts`: independently owns the idle mutation admission law across local rejection, synchronous hydration, duplicate timing, accepted dispatch, and startup failure.
- `packages/db/package.json`: registers that owner in `test:oracles`.
- `docs/contributing/oracle-coverage.md`: records the owner's domain and explicit exclusions. The conventional lifecycle suite is restored byte-for-byte to the base after its 441 transferred lines were deleted.
- `packages/query-db-collection/src/query.ts`: implements general result supersession, waiter forwarding, core-phase-fenced ownership rollback, reentrancy fencing, and cleaned-up compatibility.
- `packages/query-db-collection/tests/ownership-lifecycle.oracle.test.ts`: owns delete publication, focus/mutation supersession, waiter settlement, post-/pre-publication rollback, readiness generation, cleanup, and publication integrity.
- `.changeset/fix-query-collection-lifecycle.md`: patch releases for both affected packages.

## Provenance and credit

| Source | Contributor(s) | What was reused or learned |
|---|---|---|
| #478 | Trey Hoover (`@treyhoover`) | Delayed rapid-insert reproduction and the cache-to-Collection/live-view publication symptom; the durable defect is publication, not `orderBy`. |
| [PR #918](https://github.com/TanStack/db/pull/918), commits [`192dd2c4`](https://github.com/TanStack/db/commit/192dd2c4dd0e18cd58e53cede07b4659fe935ace), [`b3057d6f`](https://github.com/TanStack/db/commit/b3057d6f363cd97e3d0532fc5c90b1daaa699523), [`0c76796b`](https://github.com/TanStack/db/commit/0c76796b9f5389280cd77006febc727a631e326a), and [`13c73147`](https://github.com/TanStack/db/commit/13c731477c61e4628aed3f0e41e5939d64bfa704) | Kyle Mathews (`@KyleAMathews`), Claude, `@mwalkersigma`, `@flybayer`, `autofix-ci[bot]` | Established the idle mutation/direct-write request and workarounds. The commits supplied, respectively, the rejected always-start-on-demand alternative, narrow direct-write binding, regular-mutation startup, and rejected all-`utils` getter. Direction and counterexamples were reused; no code was copied. |
| [#918 review: utility access](https://github.com/TanStack/db/pull/918#discussion_r2585824906), [regular mutations](https://github.com/TanStack/db/pull/918#discussion_r2585826877), [collateral behavior](https://github.com/TanStack/db/pull/918#discussion_r2585848046), and [`utils.status` counterexample](https://github.com/TanStack/db/pull/918#discussion_r2585851711) | Sam Willis (`@samwillis`), with [Kyle Mathews's path clarification](https://github.com/TanStack/db/pull/918#discussion_r2585840173) | Established that regular mutations must be covered while ordinary/read-only utility access remains inert; this PR keeps startup at mutation and named write-method boundaries. |
| [PR #929](https://github.com/TanStack/db/pull/929), original [`73237481`](https://github.com/TanStack/db/commit/73237481e136b2bc8d6658bdefc694c1eb053db3), merged as [`5f474f1e`](https://github.com/TanStack/db/commit/5f474f1eabd57e144ba05b0f33d848f7efc8fb07) | Kyle Mathews (`@KyleAMathews`), Claude; approved by Sam Willis (`@samwillis`) | Established same-batch duplicate-key rejection. The mutation-startup owner carries that guard as a pre-start admission cell; no code was copied. |
| [PR #560](https://github.com/TanStack/db/pull/560), merged as [`ac6250a8`](https://github.com/TanStack/db/commit/ac6250a879e95718e8d911732c10fb3388569f0f) | Sam Willis (`@samwillis`); reviewed by `@kevin-dp` | Introduced the Collection lifecycle/mutation-manager boundary extended by this repair and its focused owner. Kevin reviewed the manager decomposition; this does not attribute his unadopted mixin suggestion to the repair. |
| [PR #1840 CodeRabbit review: pre-start duplicate](https://github.com/TanStack/db/pull/1840#discussion_r4031807800) and [active-waiter failure](https://github.com/TanStack/db/pull/1840#discussion_r4031807808) | CodeRabbit (`coderabbitai[bot]`) | Distinguished keys visible before startup from keys synchronously hydrated during startup, and found that an attached waiter could miss the current application's failure. The chosen fixes retain both duplicate checks and reuse the existing failure record rather than adding the suggested per-application object. |
| PR #1840 background lifecycle review | Unidentified external reviewer | Found the mid-publication ownership rewind and named the persisted-GC family. Exact probes supplied the durable-wrapper precondition; the resulting core-phase fence covers both paths. Four accompanying clean checks were independently confirmed. The source did not identify the reviewer, so no name is inferred. |
| #629 | Sam Willis (`@samwillis`; `81007b5`, `94310c0`, `983dd7d`) | Earlier #478 reproduction plus stuck-batching and duplicate-echo fixes that this work preserves. |
| #631 | Kyle Mathews (`@KyleAMathews`; `56b870b`), approved by Sam Willis (`@samwillis`) | Preserved optimistic responsiveness during long/offline commits while retaining #629's teardown and duplicate-echo handling. |
| #1824 | Kyle Mathews (`@KyleAMathews`; `179d003`), Tanner Linsley (`@tannerlinsley`), `autofix-ci[bot]`, CodeRabbit | Merged pagination/cursor-acquisition work composed here without duplication. |
| #1826 | Kyle Mathews (`@KyleAMathews`; `fdcb078`), Tanner Linsley (`@tannerlinsley`), CodeRabbit | Merged per-query cache ownership and post-write authority machinery extended here without reverting per-sync tracking. |

---

Closes #478

Supersedes #918

## Summary by CodeRabbit

- **Bug Fixes**
  - Collections now validate mutations before starting synchronization, preventing rejected inserts, updates, and deletes from changing lifecycle state.
  - Duplicate keys are detected more reliably, including conflicts in initial or hydrated data.
  - Query collection refetches now publish authoritative results without exposing stale intermediate snapshots.
  - Overlapping refetches and asynchronous writes maintain consistent ordering and ownership, with errors propagated correctly.
  - Writes made during deferred startup are handled safely, and cleaned-up collections suppress further write or refetch activity.


KyleAMathews:
@coderabbitai review

PR #1841, opened 2026-09-16
test(query-db-collection): failing regression for orderBy + limit single request

ivasilov:
## 🎯 Changes

Adds a failing regression test to `packages/query-db-collection/tests/query.test.ts` for the case where a `.from({ todo }).orderBy(todo.id, "asc").limit(10)` query on an on-demand collection should issue a single bounded request. The query currently fires a second, unbounded full-table load (an extra `queryFn` call with empty `loadSubsetOptions`), so the test asserts `queryFn` is called exactly once and fails on the current code, documenting the over-fetch bug.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

- **Tests**
  - Added coverage confirming that on-demand ordered and limited queries make one bounded data request.
  - Verified ascending identifier ordering, a limit of 10 items, and no additional unbounded request.


KyleAMathews:
Thanks for digging into this and adding a clear reproducer! The second request is expected behavior rather than an extra full-table load. ELI5: after asking for the first 10 ordered rows, TanStack DB makes one small follow-up request asking “are there any more rows tied with the last one?” That prevents rows with the same sort value from being accidentally skipped. This test’s mock handles `limit` and `offset`, but ignores the follow-up request’s `where` filter, so it returns all 30 rows and makes the safety check look unbounded. The existing query and window-controller tests already model this tie-boundary request explicitly. Since asserting exactly one call would disable that correctness behavior, I’m closing this PR. Appreciate the careful report and test case.

ivasilov:
Reposting from Discord:

Ty for the explanation, I saw the behavior when using a cursor but I didn't knew it applied it to LIMIT calls also. 

I'd still say there's a regression because the second call is completely unbounded and from the explanation you gave, it should at least have a `WHERE={column to order by}=={column value of the last row}`, right?

I updated my [PR](https://github.com/TanStack/db/compare/main...ivasilov:db:ivasilov/query-collection-single-request?expand=1#diff-976a99cd346ccf0c2fb9e947f53056e71a8325e57f1befe3ec87364bedda1fd9R1717), you can see the second call is unbounded.

KyleAMathews:
You're right: the second request in your updated reproduction is genuinely unbounded. My previous explanation described the tie-boundary path, but that is not the path taken under the default configuration here.

The relevant distinction is collation. Query Collections default to locale string ordering, while a cursor/boundary request can only be expressed safely when the configured ordering is lexical. We cannot assume that the backend's ordering matches JavaScript's locale comparison. Once the loader sees that mismatch, it has to load the full source and compute the correct top 10 locally; adding a `WHERE orderedColumn == lastValue` request would not repair a potentially incorrect initial provider-ordered page.

So I still think this PR should remain closed: the full-source fallback is intentional correctness behavior, not a missing boundary filter. There is a narrower optimization opportunity here, though: detect that the locale fallback is inevitable before issuing the provisional bounded request, so this case makes only the full-source request instead of bounded-then-full. Thanks for pushing on the explanation and providing the stronger reproduction.

ivasilov:
> You're right: the second request in your updated reproduction is genuinely unbounded. My previous explanation described the tie-boundary path, but that is not the path taken under the default configuration here.
> 
> The relevant distinction is collation. Query Collections default to locale string ordering, while a cursor/boundary request can only be expressed safely when the configured ordering is lexical. We cannot assume that the backend's ordering matches JavaScript's locale comparison. Once the loader sees that mismatch, it has to load the full source and compute the correct top 10 locally; adding a `WHERE orderedColumn == lastValue` request would not repair a potentially incorrect initial provider-ordered page.
> 
> So I still think this PR should remain closed: the full-source fallback is intentional correctness behavior, not a missing boundary filter. There is a narrower optimization opportunity here, though: detect that the locale fallback is inevitable before issuing the provisional bounded request, so this case makes only the full-source request instead of bounded-then-full. Thanks for pushing on the explanation and providing the stronger reproduction.

That sounds like a good approach, I agree. The collation handling has to be replicated on both sides.

There's a small edge case there, if the results are less than LIMIT, there's no need for additional request, right?

String collation would've been my next question. Is it possible to change the `stringCollation` option in `queryCollectionOptions`  to accept default collation per field? Something like:
```js
...
  schema: {
    id: z.string(),
    name: z.string,,
    price: z.number()
  },
  defaultStringCollation: {
    id: { stringSort: 'lexical' },
    name: { stringSort: 'locale', locale: 'en-gb' },
  },
...
```
The main usecase I have with this is the primary keys are usually UUIDs which can be sorted lexically, but it's easy to forget the collation parameter in `.orderBy`.

Let me know if you want me contributing in any other way.

KyleAMathews:
Yes, those are both reasonable ideas, but they are separate feature-level contract/design changes rather than fixes for this PR. If you would like to pursue them, please feel free to open two feature requests: one for letting a query provider report that an exact filtered request has reached authoritative source exhaustion, so core can safely avoid a redundant full-source request; and another for field-specific default string collation, such as lexical IDs alongside locale-sorted names. Keeping them separate will let us work through the API and correctness implications independently. We will keep this PR closed. Thanks for digging into both edges.

Issue #1753, opened 2026-08-19
BrowserCollectionCoordinator: per-collection schemaVersions cause reset loops, and sync-ingested writes bypass leadership


EmilEinarsen:
Findings from an adoption attempt of `BrowserCollectionCoordinator` (`@tanstack/browser-db-sqlite-persistence` 0.2.12, `@tanstack/db-sqlite-persistence-core` 0.2.12, `@tanstack/db` 0.7.2, Electric collections). Verified empirically with two pages in one Chromium context (shared storage partition) plus source reading. Happy to share probe scripts.

### 1. The coordinator holds one adapter, so per-collection `schemaVersion`s destroy each other's tables

`createBrowserWASQLitePersistence` creates one adapter per distinct `(schemaMismatchPolicy, schemaVersion)` and calls `coordinator.setAdapter(...)` for each — the coordinator keeps whichever was created last. The schema check runs in `ensureCollectionReadyInternal`, which every adapter method reaches, and `ensureCollectionState` fires `acquireLeadership` → `getStreamPosition` for every subscribed collection through that single adapter. With per-collection versions (we compute one per Electric shape) the wrong-version adapter touches every other collection: under `sync-present-reset` we observed `collection_reset_epoch` climb 0→5 during one run, a 9,475-row table wiped to 0, every reload re-syncing cold, and the registry's `schema_version` flipping between two collections' hashes.

A second failure appears even with matching versions: two adapter instances race `ensureCollectionReadyInternal` for the same collection on a fresh registry (`UNIQUE constraint failed: collection_registry.tombstone_table_name`, then `no such table: t_…` when the loser tries to reset a table the winner has not created yet). Leadership is never acquired for those collections.

One global `schemaVersion` (all adapters collapse to one instance) avoids all of this, at the cost that any schema change resets the whole database. **Suggestion:** key leader-side adapter access by collection — e.g. `requireAdapter(collectionId)` with a resolver passed to the constructor — instead of a single `setAdapter` slot.

### 2. Sync-ingested transactions bypass leadership entirely

[...]

Issue #1498, opened 2026-04-28
Console flooded with "Function object could not be cloned" with multiple persisted tabs


c-kirkeby:
- [x] I've validated the bug against the latest version of DB packages

**Describe the bug**
When using an `on-demand` query collection with a persisted sqlite collection in the browser, if you open two tabs it will flood the client's console with the message (in the tab that was opened the latest):

```
Failed to ensure remote subset: DOMException: Function object could not be cloned.
    sendRPCOnce @tanstack_browser-db-sqlite-persistence.js:2723
    sendRPCOnce @tanstack_browser-db-sqlite-persistence.js:2704
    sendRPC @tanstack_browser-db-sqlite-persistence.js:2696
    requestEnsureRemoteSubset @tanstack_browser-db-sqlite-persistence.js:2517
    flushPendingRemoteSubsetEnsures @tanstack_browser-db-sqlite-persistence.js:776
    queueRemoteSubsetEnsure @tanstack_browser-db-sqlite-persistence.js:759
    hydrateSubsetUnsafe @tanstack_browser-db-sqlite-persistence.js:475
    loadSubset @tanstack_browser-db-sqlite-persistence.js:338
    taskPromise @tanstack_browser-db-sqlite-persistence.js:141
    promise callback*run @tanstack_browser-db-sqlite-persistence.js:141
    loadSubset @tanstack_browser-db-sqlite-persistence.js:338
    loadSubset @tanstack_browser-db-sqlite-persistence.js:1169
    loadSubset @tanstack_react-db.js:5083
    requestSnapshot @tanstack_react-db.js:4305
    subscribeChanges @tanstack_react-db.js:4537
    subscribeChanges @tanstack_react-db.js:6929
    subscribeToMatchingChanges @tanstack_react-db.js:11034
    subscribeToChanges @tanstack_react-db.js:11006
    subscribe @tanstack_react-db.js:10982
    loaders @tanstack_react-db.js:11633
    subscribeToAllCollections @tanstack_react-db.js:11621
    syncFn @tanstack_react-db.js:11415
    startSync @tanstack_react-db.js:4869
    CollectionImpl @tanstack_react-db.js:6624
    createCollection @tanstack_react-db.js:6545
    bridgeToCreateCollection @tanstack_react-db.js:12019
    createLiveQueryCollection @tanstack_react-db.js:12015
    useLiveQuery @tanstack_react-db.js:13189
    RouteComponent index.tsx:76
[...]

PR #1845, opened 2026-09-17
fix(sqlite): enforce crash-only persistence coordination

KyleAMathews:
Enforce crash-only, per-collection persistence coordination across the shared SQLite core, Browser, and Electron runtimes. Custom coordinators now fail during configuration if they cannot route complete committed transactions, and multiprocess writes, subset leases, transport values, and durability failures have explicit lossless contracts.

## Root cause

The persistence boundary had grown several partial paths independently: rich source commits could fall back to row-only mutation routing, remote subset requests exposed live values to structured-clone transport, ownership was not represented as an exact lease, and response-loss retries did not require a known stable leader route. Some asynchronous owner failures also escaped the lifecycle channel or were misclassified as conflicts. Together those paths could acknowledge incomplete work, route work to the wrong collection owner, duplicate an indeterminate mutation, or leave the only durability failure unobservable.

## Approach

- Make `requestApplyCommittedTx` a required coordinator invariant and validate untyped coordinators once during collection configuration.
- Route complete `PersistedTx` values through the elected adapter for the exact collection in Browser, Electron, and single-process operation.
- Export a structured-clone-safe remote-subset wire model and recursively reject unsupported nested values with `RemoteSubsetWireValueError` and an exact value path.
- Model remote subset ownership with acquisition IDs and exact release, reject duplicate owner registration, replay live leases during takeover, and preserve failure-safe cleanup.
- Replay a mutating RPC only when the original non-null leader id and term are still current. Otherwise reject with `IndeterminateCommitError` before sending work to another or unknown leader.
- Surface storage rejection as `PersistedCollectionDurabilityError` / `PERSISTENCE_ERROR`, preserve safe cause metadata, and move the collection through its existing error lifecycle after publication.
- Add fixed and generated oracle coverage for routing, wire admission, acknowledgement, takeover, response loss, durability, and cleanup schedules.

## Key invariants

- Every adapter-bound operation and sync-ingested write uses the one supported owner for its collection.
- A committed source transaction is transported in full; there is no row-only fallback or direct-writer escape hatch.
- Success is acknowledged only after the required owner work has completed.
- Mutation replay is same-known-leader and same-term only; an unknown or changed route is an indeterminate outcome that requires reconciliation.
- Remote subset release targets the exact acquisition, and ownership failure is reported once through the lifecycle channel without retry or an unhandled rejection.
- Publication still precedes durability settlement; a later persistence failure rejects the receipt with the named durability error and fail-stops the collection.

## Non-goals

- No codec, coercion, lossy normalization, or protocol-version negotiation for unsupported wire values.
- No durable cross-leader exactly-once guarantee or automatic retry after an indeterminate commit.
- No compatibility fallback for partial third-party coordinators; this is a pre-1.0 contract correction.
- Deterministic Browser/Electron harnesses do not claim real Web Locks, OPFS, native multiprocess Electron, or live Electric-service coverage.

## Trade-offs

This deliberately narrows and strengthens the public coordinator and wire contracts. Existing custom coordinators must implement complete committed-transaction routing, and values outside the documented structured-clone-safe domain now fail at admission instead of being partially transported. The added implementation and oracle weight buys explicit failure boundaries, exact ownership, and replay behavior that can be verified without relying on runtime-specific cloning accidents.

## Verification

```bash
pnpm --filter @tanstack/db-sqlite-persistence-core test -- --maxWorkers=2
pnpm --filter @tanstack/browser-db-sqlite-persistence test -- --maxWorkers=2
pnpm --filter @tanstack/electron-db-sqlite-persistence test -- --maxWorkers=2
pnpm --filter @tanstack/electric-db-collection test -- --maxWorkers=2
pnpm --filter @tanstack/db-sqlite-persistence-core build
pnpm --filter @tanstack/browser-db-sqlite-persistence build
pnpm --filter @tanstack/electron-db-sqlite-persistence build
```

Final audited results: Core 108/108, Browser 140/140, Electron 64/64, Electric 463/463. The coordinator oracle also passes 20/20 across 12 runs at seed `165902`, replay path `0:2:2:2`; direct typechecks, formatting, lint error gate, and cleanup checks pass.

## Files changed

- `packages/db-sqlite-persistence-core`: required coordinator contract, crash-only validation, wire model, named errors, single-process routing, and focused runtime/type coverage.
- `packages/browser-db-sqlite-persistence`: per-collection elected routing, lease/replay lifecycle, structured-clone admission, public exports/docs, and Browser oracle coverage.
- `packages/electron-db-sqlite-persistence`: Browser-parity routing and ownership over Electron transport, durability classification, public exports/docs, and IPC coverage.
- `packages/electric-db-collection/tests`: recovery control proving persistence failures cannot be silently swallowed.
- `docs/contributing/oracle-coverage.md` and the bundled persistence skill: ownership and coverage accounting for the strengthened boundary.

---

Part of #1659. Addresses the coordinator ownership and lifecycle evidence in #1498 and #1753.

## Summary by CodeRabbit

* **New Features**
  * Added coordinated persistence ownership for Browser and Electron collections, including multi-context transaction routing and replay handling.
  * Added remote subset leasing, release, and ownership management across Browser and Electron environments.
  * Added strict validation and transport support for remote subset requests.
  * Added named errors for durability failures, duplicate ownership, unsupported request values, and indeterminate commits.
  * Effect-free transactions no longer publish events or consume coordination sequence numbers.
* **Documentation**
  * Expanded public API and coordination guidance for Browser, Electron, and SQLite persistence.


KyleAMathews:
Lossless external-review evaluation at 3665806c: 7 atomic items = 2 fixed-now (unknown future RPC responses; lexical wire-shape violation) + 1 design-decision (bound passive heartbeat route state without losing pre-subscription mutation and lease-replay evidence) + 2 refuted as defects (unknown initial mutating route is intentionally indeterminate; stale heartbeat suppression is unobservable to persisted subscribers and now explicitly tested) + 1 deferred to RFC #1659 (shared Browser/Electron coordinator extraction) + 1 already-covered safe invariant (deterministic collection table naming makes the registration race safe). Same-path RED/GREEN evidence covers both host coordinators and the shared wire projector. Verification: core full 114/114 and focused 71/71; Browser coordinator plus composed oracle 126/126; Electron coordinator 47/47; all three TypeScript checks and ESM/CJS/declaration builds pass; ESLint 0 errors. The passive-state change was deliberately not guessed: simply ignoring unobserved heartbeats fixed the cardinality witness but broke six existing route/takeover laws, so its eviction or explicit-interest contract remains a real #1659 design choice.

KyleAMathews:
Second external-review evaluation at `e064fc5b`: 7 atomic items = 3 fixed-now + 1 refuted + 1 deferred + 2 duplicates of the prior review.

Fixed with same-path RED/GREEN evidence:

- process-local SingleProcess/Browser/Electron owners now retain the exact `signal` and `subscription` references for load and unload, while follower transport still strips both;
- remote-demand routing is recomputed after hydration/receipt waits, so a follower that becomes an ownerless leader does not dispatch through a stale owner route;
- `limit`/`offset` now require nonnegative safe integers, with exact-path coverage for NaN, infinities, negatives, fractions, and unsafe integers.

The unknown-route indeterminate claim duplicates the already-refuted crash-only finding. Cross-request `CONFLICT` branches remain because an adversarial same-ID/different-operation probe reaches them. Full-buffer cloning for offset views is a real cost but remains deferred: current wire semantics preserve backing-buffer identity, aliases, and byte offsets, so copying only the window would be a behavior change. Browser/Electron coordinator extraction remains the existing RFC #1659 follow-up.

Verification: core 233/233; Browser 248 passed (5 intentionally skipped); Electron 118/118; all package typechecks; core/Browser/Electron ESM/CJS/declaration builds; Prettier and diff checks. ESLint has 0 errors and 8 pre-existing warnings. Full loss audit: `7 = 3 fixed + 1 refuted + 1 deferred + 2 duplicates`; no evidence gap.

KyleAMathews:
@coderabbitai review

PR #1848, opened 2026-09-18
fix(react-native): decode op-sqlite results losslessly

### LINKED ISSUE #1499: OpSQLiteDriver: executeAsync result format not handled by extractRowsFromStatementResult, causing silent data loss on SELECT queries

**Describe the bug**

`OpSQLiteDriver` in `@tanstack/react-native-db-sqlite-persistence` silently returns empty arrays for SELECT queries when op-sqlite's `executeAsync` method is used. This causes the persistence layer to think collections don't exist in `collection_registry` on app restart, leading to UNIQUE constraint violations.

**Root cause**

`resolveExecuteMethod` picks the first available method from `[executeAsync, execute, executeRaw, execAsync]`. On op-sqlite v14, `executeAsync` is available and gets selected.

The problem is that op-sqlite's `executeAsync` returns a different result format than `execute`:

- **`execute`** returns: `{ rows: Array<Record<string, Scalar>> }` (object rows)
- **`executeAsync`** returns: `{ rowsAffected: number, rawRows: unknown[][], columnNames: string[] }` (raw columnar format)

`extractRowsFromStatementResult` only handles the `{ rows }` and `{ resultRows }` shapes. When it receives an `executeAsync` result:

1. `toRowArray(value.rows)` -> `value.rows` is undefined -> returns null
2. `toRowArray(value.resultRows)` -> `value.resultRows` is undefined -> returns null
3. `hasWriteResultMarker(value)` -> `"rowsAffected" in value` -> **true** -> returns `[]`

The SELECT result is silently treated as a write result with zero rows, even though the data is present in `rawRows`.

**Impact**

This causes a cascade of failures on app restart when the database already has data:

1. `ensureCollectionReadyInternal` SELECTs from `collection_registry` -> gets `[]` instead of the existing row
2. Code takes the INSERT branch -> fails with `UNIQUE constraint failed: collection_registry.tombstone_table_name`
3. `getStreamPosition` rejects -> the persistence runtime's `ensureStartupMetadataLoaded` fails
4. The Electric sync function is never called -> the collection stays in `loading` state forever with no data

The `ALTER TABLE ADD COLUMN` errors in `ensureInitialized` are a separate but related symptom — `executeAsync` is used for those DDL statements too, and the existing error handling (`isDuplicateColumnAddError`) works but still logs errors.

**To Reproduce**

1. Create a collection with `persistedCollectionOptions` and `createReactNativeSQLitePersistence` using op-sqlite v14+
2. Let it sync data from an Electric shape
3. Kill the app (full process kill, not hot reload)
4. Relaunch the app
5. The persistence layer crashes on startup — the collection never reaches `ready` status

**Expected behavior**

`extractRowsFromStatementResult` should handle the `{ rawRows, columnNames }` format returned by `executeAsync`, converting it into the expected `Array<Record<string, unknown>>` shape. Alternatively, `resolveExecuteMethod` should prefer `execute` over `executeAsync`.

- Device: iOS Simulator & physical Android device
- OS: iOS 18, Android 14
- Version: N/A (React Native app, not browser)

**Additional context**

Package versions:
- `@tanstack/db-sqlite-persistence-core`: 0.1.9
- `@tanstack/react-native-db-sqlite-persistence`: 0.1.9
- `@op-engineering/op-sqlite`: 14.1.4
- React Native (Expo)

**Current workaround:**

Remove `executeAsync` from the database handle before passing it to `createReactNativeSQLitePersistence`, forcing the driver to fall back to `execute`:

```ts
const database = open({ name: 'my-db.sqlite', location: 'default' });
delete (database as any).executeAsync;

const persistence = createReactNativeSQLitePersistence({ database });
```

**Suggested fix:**

Either:

1. Add `rawRows` + `columnNames` handling to `extractRowsFromStatementResult`:
```ts
function extractRowsFromStatementResult(value) {
  // Handle op-sqlite executeAsync format: { rawRows, columnNames }
  if (Array.isArray(value.rawRows) && Array.isArray(value.columnNames)) {
    return value.rawRows.map((row) =>
      Object.fromEntries(value.columnNames.map((col, i) => [col, row[i]]))
    );
  }
  // ... existing logic
}
```

2. Or change `resolveExecuteMethod` to prefer `execute` over `executeAsync`, since `execute` returns the `{ rows }` format the driver already handles.

KyleAMathews:
This fixes silent SELECT data loss in the React Native SQLite adapter by decoding op-sqlite v14 columnar results into exact object rows. Malformed or unknown result envelopes now fail with a configuration error, while persisted rows, registry identity, and stream position remain intact across restart.

### Root cause

`OpSQLiteDriver` understood direct row arrays and `rows`-style wrappers, but op-sqlite v14 can return `executeAsync` results as `{ rawRows, columnNames, rowsAffected }`. The old extraction path did not reconstruct those rows and could normalize an unrecognized SELECT result to an empty result, making real data appear to disappear.

### Approach

- Decode columnar rows by zipping each `rawRows` entry with `columnNames`, preserving column order, aliases, row order, values, and multiplicity.
- Keep direct row arrays, supported row-list carriers, write-result envelopes, and exactly one supported `results` wrapper distinct at the outer result boundary.
- Reject missing or conflicting carriers, duplicate columns, row-width mismatches, unsupported wrapper depth (including cycles), and all other malformed or unknown envelopes with `InvalidPersistedCollectionConfigError`.
- Exercise writes followed by exact SELECTs through the shared SQLite driver contract, then close and reopen a pre-populated registry using an ordinary-row writer and a columnar reader.

### Key invariants

- Unknown SELECT shapes never become an empty result.
- SQL aliases such as `rows`, `rawRows`, and `rowsAffected` remain ordinary row data.
- A restart reuses the single registry mapping and preserves the exact durable stream position and rows.
- Oracle cleanup preserves the primary failure and reports cleanup failures separately.

### Non-goals and trade-offs

- The decoder intentionally supports one documented result wrapper and rejects deeper or ambiguous envelopes instead of guessing or normalizing them.
- The better-sqlite3 fixture establishes adapter behavior with real SQLite semantics, but does not claim native iOS/Android bridge coverage; versioned native-device receipts remain a separate verification cell.
- This PR does not integrate or depend on #1837. That PR was considered as evidence only and is outside this branch's ancestry and scope.

## Verification

```bash
pnpm --filter @tanstack/react-native-db-sqlite-persistence test -- --maxWorkers=2
pnpm --filter @tanstack/db-sqlite-persistence-core test -- --maxWorkers=2
TANSTACK_DB_OP_SQLITE_ORACLE_SEED=165903 TANSTACK_DB_OP_SQLITE_ORACLE_PATH=0:0:0 TANSTACK_DB_OP_SQLITE_ORACLE_RUNS=1 pnpm exec vitest --run packages/react-native-db-sqlite-persistence/tests/op-sqlite-driver.test.ts --maxWorkers=2 -t "preserves generated legal SQL aliases"
pnpm --filter @tanstack/react-native-db-sqlite-persistence build
pnpm --filter @tanstack/db-sqlite-persistence-core build
```

Verified locally with 112/112 React Native persistence tests and 96/96 SQLite core tests passing, plus the exact seed/path replay, both package builds and typechecks, all shared SQLite-driver consumers, formatting, lint (zero errors), diff, and cleanup checks.

### Files changed

- `packages/react-native-db-sqlite-persistence/src/op-sqlite-driver.ts` — strictly classifies and decodes supported op-sqlite result envelopes.
- `packages/react-native-db-sqlite-persistence/tests/helpers/op-sqlite-test-db.ts` — adds a v14-style `executeAsync` columnar fixture.
- `packages/react-native-db-sqlite-persistence/tests/op-sqlite-driver.test.ts` — adds fixed witnesses, generated alias histories, malformed-envelope controls, replay, and cleanup assertions.
- `packages/react-native-db-sqlite-persistence/tests/react-native-persistence.test.ts` — verifies pre-populated registry reuse and exact row/stream restoration after close and reopen.
- `packages/db-sqlite-persistence-core/tests/contracts/sqlite-driver-contract.ts` — adds exact write-then-read and reserved-alias laws across shared drivers.
- `.changeset/fix-op-sqlite-result-decoding.md` — records the React Native persistence patch.

---

Fixes #1499

Part of #1659

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SQLite result handling to accurately preserve rows, column order, aliases, write-result fields, and insert IDs.
  * Added validation for malformed, ambiguous, unsupported, or unknown result formats.
  * Preserved persisted data and stream positions across database close-and-reopen cycles.
  * Improved error reporting when both an operation and its cleanup fail.


KyleAMathews:
@coderabbitai review

KyleAMathews:
@coderabbitai review

KyleAMathews:
@coderabbitai review

KyleAMathews:
@coderabbitai review

Issue #1858, opened 2026-09-20
useLiveSuspenseQuery never releases over an on-demand collection whose loadSubset returns a Promise


MAST1999:
- [x] I've validated the bug against the latest version of DB packages (`@tanstack/db@0.9.2`, `@tanstack/react-db@0.4.1`)

**Describe the bug**

`useLiveSuspenseQuery` over an `on-demand` collection never leaves its Suspense fallback whenever `loadSubset` returns a `Promise` — which is every real asynchronous data source.

`LoadSubsetFn` is declared `(options) => true | Promise<void>`. If an implementation answers with the synchronous `true`, the boundary releases immediately. If it answers with a promise — even one that resolves on the next tick — the component re-suspends without limit.

**To Reproduce**

Runnable repro, real Chromium via `@vitest/browser`: https://gist.github.com/MAST1999/c2261d268a68610f18c5a6c45525b468

```bash
npm install && npx vitest run
```

Plain `@tanstack/db`. **No persistence, no Electric, no `orderBy`, no `limit`.** The only variable between passing and failing is whether `loadSubset` returns `true` or a `Promise`; both deliver the same two rows.

```
✓ a synchronous loadSubset releases the boundary                      54ms
× an async loadSubset resolving after   0ms   → text="loading" renders=23525
× an async loadSubset resolving after  50ms   → text="loading" renders=416
× an async loadSubset resolving after 500ms   → text="loading" renders=56
× an explicit queryKey does not change the outcome
× StrictMode is not the cause
```

Two things a reviewer would reasonably suspect are ruled out in the repro itself: passing an explicit stable `queryKey` does not help, and it reproduces with `StrictMode` removed.

The render count *falling* as the delay grows (23525 → 416 → 56) is the signature of a re-suspend cycle rather than a busy loop — each retry waits for the promise, so a slower promise fits fewer retries into the same five-second budget.

**Expected behavior**

An async `loadSubset` should suspend once and release when it resolves.

**Likely mechanism**

`useLiveSuspenseQuery` resets `promiseRef`/`hasBeenReadyRef` whenever `collectionRef.current !== result.collection` and throws `observer.preload()`. A component that suspends *before it has ever mounted* does not keep its hook state, so each retry constructs a fresh live query, which is `loading`, which throws a fresh preload promise, which resolves, which retries. The synchronous-`true` path never suspends at all, which is why it is the only passing case.

[...]

PR #1859, opened 2026-09-20
fix(db-sqlite-persistence-core): answer a repeat subset acquisition synchronously

### LINKED ISSUE #1856: Persisted collections: the sync wrapper's loadSubset is always async, so a live query over a loaded on-demand collection is never synchronously ready
- [x] I've validated the bug against the latest version of DB packages (`@tanstack/db@0.9.2`, `@tanstack/react-db@0.4.1`, `@tanstack/db-sqlite-persistence-core@0.2.23`)

**Describe the bug**

`createWrappedSyncConfig` in `db-sqlite-persistence-core` returns its `loadSubset` as an `async` function, so it is always a `Promise` — even when everything it awaits is already settled and the wrapped sync answered with a synchronous `true`.

That matters because `LoadSubsetFn` is declared as a *sometimes-synchronous* contract in `packages/db/src/types.ts`:

```ts
export type LoadSubsetFn = (options: LoadSubsetOptions) => true | Promise<void>
```

and `CollectionSyncManager.loadSubset` branches on it (`packages/db/src/collection/sync.ts:853`):

```ts
if (result instanceof Promise) { this.trackLoadPromise(result); ... }
return true
```

An `async` implementation can never reach the `true` branch. So for **any** persisted collection, a live query over an already-loaded subset is never synchronously `ready`.

Under `useLiveSuspenseQuery` that is fatal rather than merely slow: a component that suspends before it mounts loses its refs, so each retry constructs a *fresh* live query, which is `loading`, which throws a fresh promise, which resolves, which retries. The boundary never releases.

**To Reproduce**

Full runnable repro (node + real-browser): https://gist.github.com/MAST1999/62526dceb542323d009d036e3bb8ed7e

```bash
npm install
npm run node      # the cause  — no React, no browser
npm run browser   # the symptom — real Chromium via @vitest/browser
```

There is **no SQLite in the repro**. It uses an in-memory `PersistenceAdapter` of the same shape as this repo's own `createRecordingAdapter` in `packages/db-sqlite-persistence-core/tests/persisted.test.ts`, because the defect is in the sync wrapper that `persistedCollectionOptions` puts in front of *any* adapter. The only difference between the two cases is whether the options went through `persistedCollectionOptions`.

The mock sync returns a bare `true` from `loadSubset` for a subset it has already loaded, which is exactly what the declared contract is for:

```js
loadSubset: (options) => {
  const key = JSON.stringify(options.where ?? null)
  if (loaded.has(key)) return true   // already here — nothing to await
  loaded.add(key)
  begin(); for (const row of ROWS) write({ type: `insert`, value: row }); commit()
  return true
}
```

`node-repro.mjs`:

```
no persistence : ready (2 rows loaded)
persisted      : loading (2 rows loaded)
```

`suspense.test.jsx`, in Chromium:

```
✓ an unpersisted on-demand collection settles under Suspense    54ms
× the same collection with persistence never leaves its fallback
  → the boundary never released; the component re-suspended 13778 times
```

**Expected behavior**

A persisted `on-demand` collection should honour the same `true | Promise<void>` contract as an unpersisted one: when the wrapped sync answers `true` and there is nothing left to await, the wrapper should return `true` too, and the live query should be `ready` at construction.

**Cause**

`packages/db-sqlite-persistence-core/src/persisted.ts`, `createWrappedSyncConfig` (line 2654 on `main` @ dffb17f):

```ts
loadSubset: async (options: LoadSubsetOptions) => {
  const acquisition = { forwarded: false }
  acquisitions.set(options, acquisition)
  await fullStartPromise
  const resolvedSourceResult = await sourceResultPromise
  ...
  return runtime.loadSubset(options, (loadOptions) => { ... })
}
```

Note also that `runtime.loadSubset` (line 1043) re-runs `hydrateSubsetUnsafe` against the store on *every* call, including a repeat call for a subset already in `activeSubsets` — so the redundant work is real, not just the lost `true`.

`async` syntax structurally cannot express `T | Promise<T>` (microsoft/TypeScript#33595), so conforming to `LoadSubsetFn` requires hand-writing the branch.

**Prior art in this repo**

`@tanstack/electric-db-collection` writes its raw fetcher as `async` too, but does not hand it back directly — it wraps it (`packages/electric-db-collection/src/electric.ts:860`):

```ts
return new DeduplicatedLoadSubset({ loadSubset })
```

and `DeduplicatedLoadSubset.loadSubset` returns the literal `true` for an already-completed demand key. `grep -rn "DeduplicatedLoadSubset" packages/db-sqlite-persistence-core/src/` returns nothing — the persistence package skips that layer entirely.

**Additional context**

This is distinct from #1855 (`orderBy` + `limit` chaining subset requests), though the two share a symptom. This one needs no `orderBy` and no `limit` — just persistence.

Happy to open a PR.

MAST1999:
## 🎯 Changes

Fixes #1856.

`createWrappedSyncConfig` returned its `loadSubset` as an `async` function, so it was always a `Promise` — even when startup had long since settled, the rows were already in the collection, and the wrapped sync answered with the synchronous `true` that `LoadSubsetFn` is declared to allow:

```ts
export type LoadSubsetFn = (options: LoadSubsetOptions) => true | Promise<void>
```

`CollectionSyncManager.loadSubset` branches on that (`packages/db/src/collection/sync.ts`), so an `async` implementation can never reach the `true` branch and a persisted `on-demand` collection is never synchronously `ready`. Under `useLiveSuspenseQuery` the result is not slowness but a hang: a component that suspends before it mounts does not keep its hook state, so every retry builds a fresh live query, which is `loading`, which throws a fresh preload promise, which resolves, which retries. Measured in real Chromium: **13 778 re-suspends in 3 s, never leaving the fallback**.

`createLoopbackSyncConfig` had the same defect (`loadSubset: (options) => runtime.loadSubset(options)`, where `runtime.loadSubset` is `async`), so local-only persisted collections were affected too.

## Approach and review guide

The fix is a synchronous fast path for a **repeat acquisition of an already-hydrated subset**. It is deliberately narrow: anything that could mean the rows are not there falls back to the existing asynchronous path.

**`PersistedCollectionRuntime` — new state.** The important distinction is that `getSubsetKey(options)` is *not* a demand key: it is a per-options-object `request:N` from a `WeakMap`, so `activeSubsets` is object-keyed and N live queries with identical predicates produce N entries for one demand. Releasing one acquisition must not end another's lease. "Are these rows already here?" is a different question, asked per *demand*, so it gets its own index:

- `demandCoverage: Map<DemandKey, number>` — how many live acquisitions cover each demand.
- `hydratedDemands: Set<DemandKey>` — which demands are in the collection.
- `startupSettled` — set at the end of `startInternal`, cleared in `advanceLifecycle()`.
- `sourceTruncateGeneration` — bumped by the source's `truncate`.

**`tryAcquireHydratedSubset(options)`** returns `true` only when startup has settled, no hydration is in flight, and some acquisition still holds a hydrated demand. On a hit it goes through the *same* `registerSubsetAcquisition` the slow path uses, plus `queueRemoteSubsetEnsure`, so the new acquisition owns a lease of its own and every invalidation path keeps treating its rows as wanted.

`registerSubsetAcquisition` is idempotent per options object, because `activeSubsets` is: its key is a per-object `request:N`, so acquiring the same object twice replaces one entry rather than adding a second. Coverage therefore counts entries, not calls — see the first of the two defects below for what happens when it does not.

**Three guards, each ruling out a way the rows could be absent:**

- `!isHydratingNow()` — `truncateAndReloadUnsafe` empties the collection and *then* awaits the reload. That window is exactly `hydratingGeneration !== null`, because `reloadActiveSubsetsUnsafe` sets it before its first `await`. Without this the fast path would answer `true` over an empty collection.
- `sourceTruncateGeneration` — a generation rather than a `clear()`, because a truncate arriving mid-hydration is *buffered* and applied later, so a naive clear would be immediately undone by the in-flight hydration marking its demand hydrated again. `runtime.loadSubset` captures it alongside `lifecycleGeneration` and marks `hydratedDemands` only if neither moved.
- `demandCoverage` refcounting — `releaseDemand` forgets the demand only when the count reaches zero, so releasing one of two sibling acquisitions leaves the other's rows intact, and releasing the last one forces the next request to re-read.

**Synchronous upstream throw.** The old `async` wrapper turned a synchronous throw from the source's `loadSubset` into a rejection. On the fast path it would escape synchronously, so the same upstream failure would surface differently depending on whether the rows happened to be cached. The fast path converts it back with `Promise.reject(error)`.

**Not in scope:** two *concurrent* acquisitions of the same demand still read the store once each. Sharing those needs its own promise-sharing and cancellation story; the changeset says so.

## Why the tests missed this

Nothing covered two acquisitions of the same demand against a source that defines `unloadSubset`. `keeps sibling requests owned after one release` looks like it does, but its source is `{ loadSubset: () => true }` with no `unloadSubset`, so it cannot observe an upstream lease being dropped.

**Two further holes were found by an adversarial pass over this branch, after the fix was written.** Both were real, and both had the same shape — a repeat acquisition answered `true` over rows that were no longer in the collection:

- **Acquiring the same options object twice.** `activeSubsets` is keyed per object, so the second `set` replaced one entry rather than adding a second — but the refcount was incremented anyway. One `unloadSubset` then emptied `activeSubsets` while coverage still read 1, leaving the demand marked hydrated with nothing covering it. A later full reload rebuilds from `activeSubsets`, the rows drop out, and the next acquisition of that demand is answered from a belief that is no longer true. Coverage now counts *entries*, not calls (`registerSubsetAcquisition`).
- **A source truncate landing during a reload.** `reloadActiveSubsetsUnsafe` re-marks its subsets hydrated when it finishes, which overwrote the invalidation a truncate had just performed. It now captures `sourceTruncateGeneration` on entry and only re-marks if it has not moved — the same guard `loadSubset` uses.

Both are covered by `tests/subset-fast-path.test.ts`.

Five more tests added:

- a repeat acquisition returns literal `true` and does not touch the store;
- a sibling acquisition keeps its lease when the first is released, and `unloadSubset` fires exactly once with the right options object;
- releasing the last acquisition forces a re-read;
- a source `truncate()` forces a re-read;
- the local-only (loopback) path is synchronous on a repeat.

Three of those five fail on unpatched `main` with `expected Promise{…} to be true`. The other two pass on `main` by construction — `main` always re-reads — and exist to stop the fast path over-caching.

One existing test changed: `does not release or acquire an upstream lease cancelled during hydration` used `{ limit: 1 }` for *both* acquisitions and blocks on the second reaching the adapter, so with the fast path it hung. Its `second` is now `{ limit: 2 }` — a different demand, so it still reaches the store and can still be blocked mid-hydration. Every assertion is unchanged (`leases` 1 → 1 → 0, `loads === 1`), and the invariant under test is still exercised.

## Verification

```
pnpm --filter @tanstack/db-sqlite-persistence-core test   → 14 files, 190 tests pass, no type errors
pnpm --filter @tanstack/node-db-sqlite-persistence test   →  8 files,  52 tests pass, no type errors
pnpm --filter @tanstack/db-sqlite-persistence-core lint   →  0 errors (7 warnings, identical to main)
```

Against the standalone reproduction from #1856, using a build of this branch:

```
before   no persistence : ready (2 rows loaded)
         persisted      : loading (2 rows loaded)

after    no persistence : ready (2 rows loaded)
         persisted      : ready (2 rows loaded)
```

and in real Chromium (`@vitest/browser` + playwright), the case that previously never left its fallback:

```
before   × the same collection with persistence never leaves its fallback
           → re-suspended 13778 times in 3000ms
after    ✓ the same collection with persistence never leaves its fallback   354ms
```

## Note on scope

This makes an *already-loaded* subset synchronously ready. A genuinely cold load still returns a promise, and I've filed #1858 separately: `useLiveSuspenseQuery` appears to re-suspend without limit over **any** on-demand collection whose `loadSubset` returns a promise, with no persistence involved. If that turns out to be the deeper issue, this PR is still worth having — it removes a redundant store read per acquisition and restores the declared `true | Promise<void>` contract — but it is not the whole story on its own.

## ✅ Checklist

- [x] I have tested this code locally with `pnpm test` (for the affected packages, above)

## 🚀 Release Impact

- [x] I have generated a changeset (`@tanstack/db-sqlite-persistence-core`: patch)

## Summary by CodeRabbit

- **Bug Fixes**
  - Repeat acquisitions of already-loaded subsets now complete synchronously.
  - Live queries over hydrated subsets are ready immediately instead of re-suspending.
  - Subsets are correctly reloaded after releases, source truncation, reloads, or before startup completes.
  - Independent acquisitions retain separate leases, preventing one release from invalidating another.
  - Local-only collections also benefit from synchronous repeat acquisitions.
  - Failed or cancelled loading no longer leaves subsets incorrectly marked as ready.


MAST1999:
Pushed `e91bded` — an adversarial pass over the first commit found two more ways the fast path could be wrong, both confirmed by running code, both now fixed and covered.

**1. A released demand could be marked hydrated after the fact.** `loadSubset` recorded its demand on completion having checked only `lifecycleGeneration` and `sourceTruncateGeneration`. If every acquisition of that demand was released *while the read was in flight*, the completion recorded it with nobody holding it — and only a release-to-zero clears an entry, so nothing could ever remove it. Now also gated on coverage being above zero.

**2. Registration is synchronous; the read is not.** `ApplyMutex.run` always defers, so there was a window where a demand was covered, `isHydratingNow()` was false, and no row had been read. Demands with a read in flight are now tracked and excluded. This is what makes the invariant in the docstring true rather than nearly true — I had written that "some acquisition must still hold this demand" was sufficient, and an acquisition that has registered but not yet read holds it while preserving nothing.

**3. An aborted fast-path acquisition was ensured remotely for ever.** The slow path deletes a cancelled demand from `pendingRemoteSubsetEnsures` on `AbortError` and re-queues any other failure. The fast path returned the upstream result raw, so none of that ran — while `tryAcquireHydratedSubset` had already called `queueRemoteSubsetEnsure`. An aborted acquisition was then retried against the coordinator indefinitely. `handles remote ensure after abort without resurrecting cancelled demand` already asserts this invariant for the slow path; it survived only because a first-ever load is never a fast path. The handling is now extracted and both paths use it.

`tests/subset-fast-path.test.ts` covers all of it. Four of the five cases fail on the previous commit, including a slow-path control that passes, which is what pins the divergence in 3 to the fast path specifically.

The fifth asserts the released-mid-read invariant directly rather than the wrong answer it prevents: reaching that answer needs the apply mutex held by *non-hydrating* work, and no public API on a persisted collection exposes a way to do that. Worth saying plainly — it is the one case here without a failing-first test.

Also noted while auditing: `hydrateBaseline`'s `{}` shares the `undefined` demand key with a genuinely unconstrained subset request. It is safe, but for a different reason than every other demand — that acquisition is never released and its entry lives as long as the collection, so every reload re-reads it. There is now a comment at the site saying so.

Verification re-run on `e91bded`:

```
pnpm --filter @tanstack/db-sqlite-persistence-core test   → 14 files, 196 tests pass, no type errors
pnpm --filter @tanstack/node-db-sqlite-persistence test   →  8 files,  52 tests pass, no type errors
pnpm --filter @tanstack/db-sqlite-persistence-core lint   →  0 errors, 7 warnings (identical to main)
```

and the reproduction from #1856, against a build of this branch, still goes from `loading` to `ready` in node and from 13 778 re-suspends to a 353 ms pass in Chromium.

MAST1999:
@coderabbitai thanks — all three were real. Fixed in `13fc08f`, and I verified each against the code rather than taking the finding at face value.

**Clear `hydratedDemands` before a truncate reload** — correct, and the reasoning generalises past the loopback case you named. `truncateAndReloadUnsafe` empties the collection and then refills it, and the refill can fail; it also never goes through the source's `truncate`, so `noteSourceTruncate` was not running for it at all. Cleared at the top of `truncateAndReloadUnsafe` as you suggested. There is now a failing-first test: a `collection:reset` from another node whose reload read rejects leaves the collection empty, and the next acquisition must re-read rather than be answered `true`.

**Do not replace an existing forwarded acquisition** — correct. `UnloadSubsetFn` is documented as releasing "the exact acquisition created for `options`", and the wrapper was replacing that record on every call, so two forwards against one release left a lease held for ever. The record is now reused, and a second forward for an object that already holds a lease is skipped. Test added exactly as you asked — repeated acquisition of one options object keeps upstream load and unload counts balanced; it fails on the previous commit.

**Tag `loadsInFlight` with the lifecycle generation** — correct, and the sharpest of the three. `advanceLifecycle` clears the map but a load from before it still runs its `finally`, so the late `leaveLoad` could decrement or delete a count belonging to the *current* generation and let the fast path answer while a current load was still settling. Entries now carry the generation that created them, `leaveLoad` takes the generation its load started in, and a mismatch is ignored. The in-flight check in `tryAcquireHydratedSubset` also compares the generation rather than just testing for presence.

That last one has **no failing-first test**, and I would rather say so than imply otherwise: provoking it needs a lifecycle advance during an in-flight read, and `advanceLifecycle` only runs from `setSyncControls` and `cleanup`, neither of which a test can interleave into that window through the public API.

```
pnpm --filter @tanstack/db-sqlite-persistence-core test   → 14 files, 200 tests pass, no type errors
pnpm --filter @tanstack/node-db-sqlite-persistence test   →  8 files,  52 tests pass, no type errors
pnpm --filter @tanstack/db-sqlite-persistence-core lint   →  0 errors, 7 warnings (identical to main)
```

The reproduction from #1856 still goes from `loading` to `ready` in node and from ~13 800 re-suspends to a 305 ms pass in Chromium.

KyleAMathews:
Hey checkout #1860 when you have a chance

MAST1999:
Closing in favour of #1860, which subsumes this fix.

We tested #1860's pkg-pr.new builds against the reporting monorepo (smartmix — the #1856/#1858 reporter): the persisted on-demand suites, the Electric-ordered on-demand pages, and the persisted-wrapper invariant suites all pass under it (41 + 7 + 116 + 69 + 91 tests across the five affected packages, zero regressions against released 0.9.2/0.4.1).

#1860's `persisted.ts` change implements the same synchronous fast path this PR contributes — a repeat subset acquisition settling with `true` instead of a fresh promise — and goes further than it did here: exact hydrated-demand keys, invalidation on release/truncate/reload/cleanup/failed-hydration, plus the ordered-loader normalization and the Suspense retry-loop fixes that the hang in #1858 also needed. Prefer the broader fix landing as one reviewed change.

Issue #1589, opened 2026-06-12
Persisted Electric collection ends up permanently empty: diverging schemaVersions wipe rows, schema reset leaves the resume point behind


eikowagenknecht:
**Packages:** `@tanstack/browser-db-sqlite-persistence` 0.2.0, `@tanstack/db-sqlite-persistence-core` 0.2.0, `@tanstack/electric-db-collection` 0.3.6, `@tanstack/db` 0.6.8

We hit a case where a persisted Electric collection comes up ready with 0 rows on every page load, with nothing logged, and never recovers. It took a while to track down because it looked random. There seem to be two separate bugs that line up to cause it.

## Diverging schemaVersions wipe rows through a shared coordinator adapter slot

`createBrowserWASQLitePersistence` caches one adapter per `(schemaMismatchPolicy, schemaVersion)` and calls `coordinator.setAdapter(adapter)` on every `resolvePersistenceForCollection` (see `browser-persistence.js`). `BrowserCollectionCoordinator` only has a single adapter slot, so whichever collection resolved last owns it.

If you have two persisted collections with different `schemaVersion`s, leader-side RPCs (`pullSince`, `applyLocalMutations`, `ensureRemoteSubset`) for collection A can end up running through collection B's adapter. `SQLiteCorePersistenceAdapter.ensureCollectionReady` then compares A's registry `schema_version` against the other adapter's version, sees a mismatch, and under `sync-present-reset` deletes all of A's rows and rewrites the registry version. With both collections live the wipes flip-flop depending on RPC timing, which is why it looked random for us (the difference between F5 and a hard reload was pure timing).

## A schema-mismatch reset doesn't clear `collection_metadata`, so Electric's resume point survives the wipe

The reset deletes rows, tombstones and `applied_tx`, but leaves `collection_metadata` intact. `@tanstack/electric-db-collection` keeps its resume point there (`electric:resume`, offset + handle, keyed by shape identity = url + params). So on the next load the collection restores 0 rows, reads the stale resume point, resumes the stream "up to date" past all the data, and marks itself ready. Nothing triggers a refetch, and the session even rewrites the resume record, so it stays empty across reloads.

## Reproduction

```ts
const persistence = createBrowserWASQLitePersistence({ database, coordinator });

const a = createCollection(persistedCollectionOptions({
  ...electricCollectionOptions({ id: "a", shapeOptions: { url: shapeA }, getKey: r => r.id }),
  persistence, schemaVersion: 2,
}));
[...]

Issue #1443, opened 2026-04-02
BrowserCollectionCoordinator + Electric adapter: collections never reach ready state


jwaltz:
# BrowserCollectionCoordinator + Electric adapter: collections never reach "ready" state

## Package & Version

- `@tanstack/browser-db-sqlite-persistence@0.1.5`
- `@tanstack/db@0.6.1`
- `@tanstack/electric-db-collection@0.2.43`
- `@electric-sql/client@1.5.14`

## Environment

- Chrome 136 (latest stable)
- Vite / TanStack Start
- Single browser tab (no multi-tab involved)

## Description

When `BrowserCollectionCoordinator` is passed to `createBrowserWASQLitePersistence` and the persistence wraps **Electric** collections (via `electricCollectionOptions`), collections never reach the "ready" state. Electric shape HTTP requests go out and return 200 with data (confirmed via Network tab), but `useLiveQuery` subscribers return empty results.

Removing the coordinator (falling back to the default `SingleProcessCoordinator`) immediately fixes the issue — Electric data flows through persistence and renders correctly.

### Isolation Testing

To isolate whether the issue is the coordinator generally or specific to Electric, I tested with `queryCollectionOptions` (TanStack Query adapter) + `BrowserCollectionCoordinator`:

| Config | Collections | Result |
|--------|:-----------:|--------|
| `queryCollectionOptions` + no coordinator | 1 | ✅ Works |
| `queryCollectionOptions` + `BrowserCollectionCoordinator` | 1 | ✅ Works |
| `queryCollectionOptions` + `BrowserCollectionCoordinator` | 21 | ✅ Works |
| `electricCollectionOptions` + no coordinator | 21 | ✅ Works |
| `electricCollectionOptions` + `BrowserCollectionCoordinator` | 21 | ❌ No data |

The coordinator works correctly with the query adapter at any scale. The issue is specific to the Electric adapter + coordinator combination.

### Test Coverage Note

The official example (`examples/react/offline-transactions/src/db/persisted-todos.ts`) uses `BrowserCollectionCoordinator` only with local-only persistence — no sync adapter. The e2e test (`browser-single-tab-persisted-collection.e2e.test.ts`) also uses no coordinator. The Electric + coordinator combination may be an untested code path.

## Reproduction (Real App)

```ts
const database = await openBrowserWASQLiteOPFSDatabase({
  databaseName: "myapp.sqlite",
});

const coordinator = new BrowserCollectionCoordinator({ dbName: "myapp" });
[...]

PR #1860, opened 2026-09-20
fix: restore warm on-demand query readiness

### LINKED ISSUE #1855: Live query with orderBy + limit over an on-demand collection is never ready at construction, so useLiveSuspenseQuery retries for ever

## Describe the bug

A live query that combines `orderBy` with `limit` over a `syncMode: "on-demand"` collection is **never `ready` at construction**, even when the source collection is `ready` and every subset the query needs is already loaded and answered synchronously by the adapter.

That alone is a performance wart. Under `useLiveSuspenseQuery` it is fatal: a component that suspends before it mounts loses its refs, so every retry runs `useLiveQuery` with `collectionRef.current === null` and builds a **brand-new live query collection**. If that fresh collection is `loading`, the hook throws a new `preload()` promise, which resolves, which re-renders, which builds another collection — for ever.

Removing `.limit()` — changing nothing else — makes the retry `ready` at construction and everything renders.

In a real app (React 19, `@tanstack/electric-db-collection`) this pegs the main thread: I measured ~190 renders of a single hook on one navigation, a permanently visible Suspense fallback, and **no network traffic after the first few requests** — the server had already answered, the rows were in memory, and `collection.status` was `ready` with `collection.size === 4`.

Related but not the same: #1418 (closed/fixed) was `useLiveSuspenseQuery` + on-demand stuck after a *dependency change*. This reproduces on 0.9.0 and 0.9.2 with no dependency change at all — the trigger is `orderBy` + `limit`.

## The cause, as far as I traced it

With `orderBy` + `limit`, `OrderedSourceLoader` needs **several sequential subset requests**, each issued only from the previous one's `complete()`:

1. `{ orderBy, limit: offset + limit }` — the ordered prefix
2. `{}` — full source (the `canExpressCursorOrder` / boundary fallback)
3. `{ orderBy, limit: 46, offset: 4, cursor }` — the cursor page

Because they are chained through promise callbacks, the query cannot reach `ready` synchronously **even when all three are cache hits that the adapter answers with a synchronous `true`**. Without `limit` there is exactly one subset request, it hits synchronously, and the query is `ready` at construction.

So the "already loaded ⇒ synchronously ready" fast path that makes `useLiveSuspenseQuery` terminate exists only for the single-request shape.

## To Reproduce

Standalone, `@tanstack/db` only — no React, no bundler, no Electric, no persistence. Save as `faithful.mjs` in a dir with `{"type":"module"}` and `@tanstack/db` installed, then `node faithful.mjs`.

```js
import { createCollection, createLiveQueryCollection, BTreeIndex } from "@tanstack/db";

const TABLE = [
  { id: "a", batched_date: new Date("2026-09-03T10:00:00Z") },
  { id: "b", batched_date: new Date("2026-09-03T09:00:00Z") },
  { id: "c", batched_date: new Date("2026-09-02T10:00:00Z") },
  { id: "d", batched_date: new Date("2026-09-01T10:00:00Z") },
];

const build = (withLimit) => {
  const served = new Set();
  const issued = [];
    id: `t-${withLimit}`,
    getKey: (r) => r.id,
    syncMode: "on-demand",
    autoIndex: "eager",
    defaultIndexType: BTreeIndex,
    sync: {
        markReady();
        return {
          loadSubset: (options) => {
            const key = JSON.stringify({
              limit: options.limit ?? null,
              offset: options.offset ?? null,
              ordered: Boolean(options.orderBy),
              cursor: Boolean(options.cursor),
            });
            issued.push(key);
            // Already loaded -> answer synchronously, as a real adapter does.
            if (served.has(key)) return true;
            served.add(key);
            return new Promise((res) => setTimeout(() => {
              begin();
              for (const row of TABLE) write({ type: "insert", value: row });
              res(commit());
            }, 5));
          },
          unloadSubset: () => {},
        };
      },
    },
  });
  const query = (q) => {
    const o = q.from({ t: collection })
      .orderBy(({ t }) => t.batched_date, { direction: "desc", nulls: "last" });
    return withLimit ? o.limit(50) : o;
  };
  return { collection, query, issued, served };
};

const wait = (ms) => new Promise((r) => setTimeout(r, ms));

for (const withLimit of [false, true]) {
  const { collection, query, issued, served } = build(withLimit);

  const first = createLiveQueryCollection({ startSync: true, query });
  first.subscribeChanges(() => {});
  await first.preload();
  await wait(300);
  const afterFirst = issued.length;

  console.log(`\n=== ${withLimit ? "WITH .limit(50)" : "WITHOUT limit"} — table has ${TABLE.length} rows ===`);
  console.log(`  source: status=${collection.status} size=${collection.size}`);
  console.log(`  first live query: status=${first.status} rows=${first.size}`);
  console.log(`  distinct subsets loaded: ${served.size}`);

  // Every useLiveSuspenseQuery retry builds a fresh live query, because the
  // suspended component's refs were discarded. The source is fully loaded.
  const statuses = [];
  for (let i = 0; i < 5; i++) {
    const retry = createLiveQueryCollection({ startSync: true, query });
    retry.subscribeChanges(() => {});
    statuses.push(retry.status);
    await wait(60);
  }
  console.log(`  retry live query status at construction: ${statuses.join(", ")}`);
  console.log(`  subsets requested by the 5 retries: ${issued.length - afterFirst}`);
}
process.exit(0);
```

### Output (identical on 0.9.0 and 0.9.2)

```
=== WITHOUT limit — table has 4 rows ===
  source: status=ready size=4
  first live query: status=ready rows=4
  distinct subsets loaded: 1
  retry live query status at construction: ready, ready, ready, ready, ready
  subsets requested by the 5 retries: 5

=== WITH .limit(50) — table has 4 rows ===
  source: status=ready size=4
  first live query: status=ready rows=4
  distinct subsets loaded: 3
  retry live query status at construction: loading, loading, loading, loading, loading
  subsets requested by the 5 retries: 15
```

Every one of those 15 retry requests is a cache hit answered with a synchronous `true`. The query still reports `loading`.

Note the table holds 4 rows against a limit of 50, which is what puts the loader on the "need more data" path (`dataNeeded()` stays at `limit - size`). A source with more rows than the limit may behave differently; I did not test that.

## Expected behavior

A live query over an on-demand collection whose every required subset is already loaded should be `ready` at construction, regardless of `limit` — so that `useLiveSuspenseQuery`'s retry can terminate.

Failing that, `useLiveSuspenseQuery` shouldn't depend on synchronous readiness for termination, since it cannot hold refs across a pre-mount suspend.

## Versions

`@tanstack/db` 0.9.0 and 0.9.2 (same result), `@tanstack/react-db` 0.3.8, `@tanstack/electric-db-collection` 0.4.8, React 19.2, Node 24.

The standalone repro above uses `@tanstack/db` alone; the React/Electric versions are for the app where it was first hit.

### LINKED ISSUE #1856: Persisted collections: the sync wrapper's loadSubset is always async, so a live query over a loaded on-demand collection is never synchronously ready

### LINKED ISSUE #1857: on-demand: every loadSubset forces an extra same-offset chunk request, tripping the Electric client's fast-retry-loop guard
- [x] I've validated the bug against the latest version of DB packages (`@tanstack/db@0.9.2`, `@tanstack/electric-db-collection@0.4.10`, `@electric-sql/client@1.5.28`)

**Describe the bug**

Every `loadSubset` on an `on-demand` Electric collection issues an **extra, non-subset chunk request at the same offset**, because each one calls `stream.forceDisconnectAndRefresh()` before `requestSnapshot`.

Against a log that is not changing, those extra requests are byte-identical and repeat the same offset — which is precisely the pattern `@electric-sql/client` treats as a broken cache or proxy. Ten ordinary sequential subset loads are enough to trip its fast-retry-loop guard, which clears the client's caches, resets the stream, and tells the developer to go and check their proxy configuration.

The proxy is fine. The duplicate requests come from the adapter.

**To Reproduce**

Runnable repro against a **real** Postgres + Electric (the pair this repo's own e2e suite expects), so nothing fakes the wire: https://gist.github.com/MAST1999/630589ae5beb96ce5c9222a5aad003b7

```bash
docker compose up -d
npm install
node repro.mjs
```

It creates ten live queries with distinct `where` predicates, each fully awaited before the next is created. No concurrency, nothing superseded, nothing racing, and the table never changes.

```
distinct subsets requested   : 10
non-live shape requests      : 20
  snapshots (POST)           : 0
  chunk refetches (GET)      : 20
distinct offsets requested   : 4
requests per subset          : 2.00
subset requests              : 10
```

The requests alternate — ten are the subset requests themselves, and ten carry **no `subset__` parameters at all** and repeat the same `offset=0_inf`:

```
GET ?log=changes_only&offset=now&subset__params=...&subset__where="n" >= $1&table=items
GET ?handle=...&log=changes_only&offset=0_inf&table=items
GET ?handle=...&log=changes_only&offset=0_inf&subset__params=...&subset__where=...
GET ?handle=...&log=changes_only&offset=0_inf&table=items
```

and the client's guard fires:

```
[Electric] Detected fast retry loop (5 requests in 500ms at the same offset).
Clearing client-side caches and resetting stream to recover. If this persists,
check that your proxy includes all query parameters (especially 'handle' and
'offset') in its cache key, and that required Electric headers are forwarded to
the client.
```

**Expected behavior**

Loading ten distinct subsets should cost ten subset requests. A refresh that is
only needed when the stream is actually behind should not be issued
unconditionally per subset, and ordinary usage should not trip a guard whose
message tells the developer to debug infrastructure that is working correctly.

**Scope — what this repro does and does not show**

It shows the request amplification (2.00 non-live requests per subset) and the spurious guard trip, both on the latest published versions.

It does **not** show a permanently dead stream — the client recovers in this repro. I've seen a stuck `paused` stream downstream of this in an application, but I could not reduce that to a minimal case, so I'm deliberately not claiming it here.

**Additional context**

`@tanstack/db@0.9.0` removed the subset-algebra helpers (`isLoadSubsetRequestSubsumedBy` and friends) in favour of `DeduplicatedLoadSubset` sharing only exact demand identities, so a narrower predicate is no longer recognised as already covered. That makes distinct-but-overlapping predicates more common in practice, and each one now carries this extra refresh.

Related but distinct: #1855 and #1856.

### LINKED ISSUE #1858: useLiveSuspenseQuery never releases over an on-demand collection whose loadSubset returns a Promise

KyleAMathews:
Restores synchronous readiness for warm on-demand queries and prevents React Suspense retries, persisted wrappers, and Electric subset snapshots from restarting work that has already been satisfied. Users no longer see permanently suspended ordered queries or redundant same-offset Electric requests.

## Root cause

The regressions came from lifecycle boundaries that treated settled work as still pending:

- the ordered loader normalized a synchronous `loadSubset` result through a promise, so a warm query could not become ready during construction;
- an uncommitted Suspense render recreated its live collection on each React retry;
- the persistence wrapper always re-entered asynchronous hydration, even when the exact demand was already hydrated and retained;
- Electric forced a normal refresh before asking the SDK for a subset snapshot, although `requestSnapshot` already owns that transition.

Review also found two narrower ownership errors: an ordered cleanup error could replace a newer nested request, and an explicitly keyed Suspense query rebuilt its query IR on every committed render.

The reported offset behavior is separate: an ordered offset window must load the prefix through `offset + limit` so earlier rows can shift the public window correctly. Electric receives that prefix request while the live query publishes only the requested slice.

## Approach

- Let literal synchronous subset completion settle inside the current ordered-loader call chain. D2 remains the authority for later demand; resident rows are not treated as proof of provider coverage.
- Keep one retry-stable React collection per query and source identity. Suspense-created collections use the existing `gcTime` lifecycle with a five-second default grace period, which covers nearby pre-commit retries without adding a second ownership system. Explicit `gcTime` still wins.
- Preserve a nested ordered request if a prior acquisition's cleanup reports an error, and avoid rebuilding an explicitly keyed query after it has committed.
- Track exact hydrated demands in the persisted wrapper. Use a synchronous fast path only after startup and replay have settled, and invalidate it on release, truncate, reload, cleanup, or failed hydration.
- Remove Electric's pre-snapshot disconnect/refresh. The installed SDK's snapshot request performs the only transition.
- Extend the stateful oracles across synchronous and asynchronous settlement, rejection, abandonment, reversed settlement, peer ownership, reload, cleanup, loopback mode, repeated Electric demands, and public offset publication.
- Correct `@tanstack/react-db`'s stale peer range to React 18. Its source has used React's built-in `useSyncExternalStore` since July 2025, and its Suspense hook has required React 18 since introduction. The never-imported shim dependency is removed.

## Key invariants

- A synchronously satisfied warm demand is ready before construction returns.
- Suspense retries with the same query identity reuse one live collection during the bounded GC grace period; failed and cleaned-up collections leave the registry.
- A cleanup failure cannot erase a newer nested ordered request.
- Persisted readiness belongs to an exact retained demand and never crosses a lifecycle reset.
- One Electric subset demand causes one SDK snapshot request and no ordinary refresh.
- Offset queries acquire the required prefix but expose only their requested window.
- Disabled Suspense queries fail with the documented API error before collection lifecycle setup.

## Non-goals

- This does not infer provider coverage from rows already resident in a collection.
- It does not promise permanent collection identity for an abandoned Suspense render; the default retry cache is bounded to five seconds.
- It does not change cursor compilation or Electric's snapshot wire protocol.
- It does not add support for mutable `LoadSubsetOptions`; acquisition options remain immutable lease identities.
- It does not broaden opaque client Suspense queries into the explicit-key requirement used by SSR streaming.

## Trade-offs

The React adapter keeps a small query-identity registry entry for the lifetime of its live collection. Suspense-created collections default to a five-second `gcTime`, trading brief retention for retry stability; callers can override it. The persistence wrapper keeps an exact set of hydrated demand identities. Both use existing lifecycle boundaries rather than treating resident rows as provider coverage.

## Verification

- `pnpm --filter @tanstack/db test` — 6,268/6,269 passed in the concurrent full run; the sole 5-second lifecycle-property timeout passed alone, 99/99.
- Focused ordered-loader cleanup regression — passed with no type errors.
- `pnpm --filter @tanstack/react-db test` — 294/294 passed with no type errors.
- Changed-file lint — no errors; one pre-existing async/no-await warning.
- `pnpm --filter @tanstack/react-db build` — passed.
- `pnpm --filter @tanstack/db-sqlite-persistence-core test` — 196 passed with no type errors, including release-during-read, registered-before-read, and fast-path abort boundaries.
- `pnpm --filter @tanstack/electric-db-collection test` — 863/863 passed.
- Builds passed for all affected packages.

## ✅ Checklist

- [x] Bug regressions are red/green covered, including hostile Suspense retry and cleanup.
- [x] A changeset documents the published package changes.
- [x] Affected package tests, types, lint, and builds pass.
- [x] The React peer range and installation docs match the runtime contract.

## 🚀 Release Impact

This publishes patch releases for the four affected packages via `.changeset/fix-on-demand-readiness.md`. It also corrects `@tanstack/react-db`'s declared React peer minimum from 16.8 to 18 and removes the unused `use-sync-external-store` dependency.

## Files changed

- `packages/db`: synchronous ordered settlement plus expanded ordered lifecycle and pagination oracles.
- `packages/react-db`: retry-stable Suspense collection reuse, React 18 metadata, and uncommitted-render lifecycle coverage.
- `packages/db-sqlite-persistence-core`: exact hydrated-demand readiness and state-machine coverage.
- `packages/electric-db-collection`: single-owner snapshot transition and SDK/public-query delivery coverage.

---

Closes #1855
Closes #1856
Closes #1857
Closes #1858
Closes #1859

## Summary by CodeRabbit

- **Bug Fixes**
  - Warm on-demand queries now become ready synchronously when requested data is already available.
  - Persisted subset data remains available across reloads and lifecycle transitions, with stale or failed reads retrying correctly.
  - Suspense queries recover more reliably after retries and abandoned renders.
  - Electric-backed queries avoid unnecessary stream refreshes for up-to-date snapshots.
  - Live query windows handle asynchronous loading, resets, and retries more consistently.

- **Documentation**
  - React 18 or later is now required for TanStack DB and React integrations.


MAST1999:
## Verified against a production monorepo: both reported symptoms fixed

We tested this PR's pkg.pr.new builds against the reporting monorepo (the #1855/#1856/#1858 reporter, ~53 workspace packages, Electric sync, persisted collections, React 19.3).

### The ordered + limited on-demand hang (#1855/#1858) — **fixed**

Our tickets list reads an `on-demand` collection through `useLiveSuspenseQuery` with `orderBy` + `limit(51)` + `offset(n)` pushed down as a subset — the exact shape that hung permanently on released `0.9.2`/`0.4.1` (measured: ~190 hook renders on one navigation before we switched to `useLiveQuery`; the Suspense boundary never released because each retry built a fresh live query that was `loading` again).

Under this PR's builds, we switched back to `useLiveSuspenseQuery` and it renders in the real browser:

- **Tickets list**: 45 rows across 10 columns, insight strip (7 plant groupings with production volume / cost / CO₂ aggregates), filter toolbar (6 filter controls + window selector), pagination. The Suspense boundary releases cleanly.
- **Ticket detail page** (on-demand, single-id read through the Suspense hook): full detail renders — status, plant, mix, address, timestamps.

### The persisted sync fast path (#1856) — **fixed**

Our PR #1859 (closed in favour of this one) contributed the same synchronous `loadSubset` fast path this PR implements more thoroughly (exact hydrated-demand keys with invalidation on release/truncate/reload/cleanup/failed-hydration).

### Not fixed by this PR (for the record)

- **#1589(b)**: a `schemaVersion` bump still strands the `electric:resume` cursor in `collection_metadata` — the collection restores nothing, resumes past all its data, and reports `ready` with zero rows. Our reproduction test still passes under this PR because the bug still reproduces. Our workaround (deleting the store file on version change) remains necessary.
- **#1443**: `BrowserCollectionCoordinator` still makes Electric collections never reach `ready` when passed to `persist()`. Our workaround (omitting it) remains necessary.

Both are pre-existing and outside this PR's scope — noted for maintainers' awareness, not as blockers.

### Environment

- `@tanstack/db@bb75989`, `@tanstack/react-db@bb75989`, `@tanstack/electric-db-collection@bb75989`, `@tanstack/db-sqlite-persistence-core@bb75989`, `@tanstack/db-ivm@bb75989` (pkg.pr.new, commit-pinned)
- pnpm 12.5.1, React 19.3.0, Vite 8.3.0, real Electric 1.8.1 shape proxy, real Chromium via Playwright

